PR #55, opened 2015-04-08
Redistribute restarts

johnomotani:
Added a redistribute(npes, path, nxpe, output, informat, outformat)
function to boutdata.restart. This collects the data from a set of
old restart files and creates a set of new ones for arbitrary (valid)
npes and nxpe.

### DIFF
--- a/tools/pylib/boutdata/restart.py
+++ b/tools/pylib/boutdata/restart.py
@@ -6,7 +6,9 @@ except ImportError:
     print "ERROR: restart module needs DataFile"
     raise
 
+import numpy
 from numpy import mean
+from math import sqrt
 
 try:
     import os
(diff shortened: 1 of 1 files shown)


JensMadsen:
Hi John 

Does this mean that I am able to 

1) run a code with 4 processors
2) restart the code with 16 processors? 

Could you add an example line in the python code:-D

JensMadsen:
Works  like a charm. Thanks. Just what I needed :-D

JensMadsen:
Hi John 

I can make the routine work if npeold< npenew? I I redistribute to a smaller number of processors It seems not to work?

johnomotani:
> I can make the routine work if npeold< npenew? I I redistribute to a 
> smaller number of processors It seems not to work?
>

It ought to work, as long as there is a valid split of the grid over 
npenew processors. I only tested very quickly, but going from 4->2 
seemed to work. Does it give an error for you, or just produce wrong output?

JensMadsen:
It just gives wrong output. I provide you with some more details in a minute:-D

JensMadsen:
The fields are just set to constant values.

PR #65, opened 2015-05-08
Python3 compatible

loeiten:
Merged brey's pull request.

All the major python routines are now py2 and py3 compatible (mostly done by using futurize (http://python-future.org/index.html)). However, there may still be some corner cases in lesser used python procedures not covered (these should be easy fixable once eventually spotted). These eventual corner cases will not yet be py3 compatible, but should be py2 compatible.

For the procedures to be py2 compatible, the package 'future' must be installed (it usually is). If you find 'future' is hard to install, or if you would like to easily install py3 yourself (on your laptop or a server), the package manager anaconda is recommendable.
See:
https://store.continuum.io/cshop/anaconda/
http://conda.pydata.org/docs/intro.html
for more info.


johnomotani:
I have tried to remove the dependency on past.old_deriv so that we can at least run on vanilla python3 installs without installing (a newer version of) future. This involved inserting a lot of '/' and '//' by hand, so I may have broken something, but at least the tests all pass when running them with python3.

Are we standardising on python3 now? Maintaining backward compatible python3 code seems to be recommended over maintaining forward compatible python2 code. At the moment test_suite launches each runtest script from a shell, and so they use the system default python (i.e. python2) regardless of which python was used to run test_suite: should we change this (maybe check if we are using python3 and use ```status,out = shell("python3 ./runtest")``` if we are)?

erikgrinaker:
You can use sys.executable to get the full path to the Python interpreter that is executing the current script. This is less brittle than making assumptions about how Python is installed (i.e. that python3 is a valid executable and in the user's $PATH).

bendudson:
Ideally things would work with python 2 or 3, but eventually we have to
move everything to python 3. I agree that backwards compatible python 3
code is probably the way to go for now, at least for core functionality
stuff.

I made the test suite a set of shell scripts, so that tests could be
written in many ways, rather than forcing them to all be python. If a test
script needs python 3, then it can specify that in its shebang.

loeiten:
Hmm...
It looks like the installation through anaconda runs the runtest scripts from python3 even though they are called from the shell (as this is set to an environmental path through .bashrc).

Regarding the dependencies on old_div. I agree that it looks much better with '/' and '//', but other that that I thought it would be quite straight forward to install/upgrade future with easy_install or something similar if one does not want to use conda? And if it is easy to install/upgrade future, maybe it is ok to stick with old_div?

johnomotani:
"And if it is easy to install/upgrade future..."

The issue I have is that it creates yet another barrier for new users 
installing the code. BOUT++'s other dependencies should be available 
either through the system package manager or modules. python-future is 
apparently not (universally) available yet, so one needs some python 
package-manager as well to install it. Since test_suite really ought to 
work, a working pylib is an essential part of BOUT++ and the minimum set 
of things needed for a working BOUT++ installation should be as easy as 
possible to acquire: is the cost of making python-future necessary worth 
the gains in python2/python3 compatibility?

bendudson:
I think John's right that at least the basics should work, and the rest
should fail gracefully. I have made some changes which worked on my laptop
(without future installed), so will push later. Many of the "old_div" calls
are unnecessary, since the result was then cast to an integer anyway.

int(old_div(a, b))

should be the same as

int(a / b)

regardless of how "/" behaves. The exception is when a or b might be
negative, but the uses here are for processor number and other things which
are always positive.

loeiten:
Ok, that makes good sense.

bendudson:
I've pushed commit 7cf9068c64152ed which seems to work on my laptop with Python 2.7.8 . Previously failed, complaining about old_div and other things not being defined. The test suite seems to run now, for me at least.

PR #73, opened 2015-07-07
Adding HDF5 support

bendudson:
John's HDF5 interface seems to work, and has updated Python collect and DataFile routines to work with either HDF5 or NetCDF. The only parts of the test-io test which fail are when reading single numbers: the NetCDF routines return a 1-element NumPy array, whereas the HDF5 routines return a number (e.g. int32). 

Currently no documentation


johnomotani:
For me the NetCDF DataFile routines return non-arrays when reading single numbers, the same as the HDF5 ones. If there is a need to enforce consistency, one could wrap the variables in numpy.array() in the return statements, e.g. ```return numpy.array(data)``` instead of ```return data```, etc.

I have added some documentation to the user manual. Developer documentation would take more time to write... hopefully the comments and code will be enough.

I noticed that the ```dump_format``` option is in the root section at the moment: it would probably make more sense to move it to ```[output]```.

bendudson:
Nice, thanks John!

PR #78, opened 2015-07-08
Fixed tridagMatrix calls

### LINKED ISSUE #71: Possible bug in call to tridagMatrix in serial_tri and cycl_laplace
The call in ```serial_tri.cxx``` goes like this

```c++
    tridagMatrix(avec, bvec, cvec, bk1d, jy, 
		 iz == 0, // Called "int kz" in the function. Used to check if DC.
		 iz*2.0*PI/mesh->zlength, // kwave
		 global_flags, inner_boundary_flags, outer_boundary_flags,
		 &A, &C, &D);
```
this calls to the ```invert_laplace.cxx``` function
```c++
void Laplacian::tridagMatrix(dcomplex *avec, dcomplex *bvec, dcomplex *cvec,
                             dcomplex *bk, int jy, int kz, BoutReal kwave,
                             int global_flags, int inner_boundary_flags, int outer_boundary_flags,
                             const Field2D *a, const Field2D *ccoef,
                             const Field2D *d,
                             bool includeguards) {
```
I think there is a problem here as ```iz == 0``` (```iz``` is the looping index for the mode numbers) will be casted to ```int kz```. As ```iz == 0``` will evaluate to ```1``` when ```iz``` is ```0```, the result will be that ```kz = 1``` in the ```invert_laplace.cxx``` function.
 ```kz``` will now (as far as I can understand) have a false value, and this value is used to check whether we are dealing with DC or AC components, and it is used if the ```INVERT_IN_CYLINDER``` flag is being used.

The same call is being made in ```cycl_laplace``` if ```dst``` is ```True```.

I thought it would be a good idea to post this as an issue, rather than to just "fix" it, in case I am wrong.

loeiten:
This should fix #71


loeiten:
Tested, and can confirm that this closes #71

PR #87, opened 2015-08-28
Visboutit

peternaylor:
3D visualisation using VISiT in Python by Peter Naylor. Examples and manual for cylindrical and toroidal coordinates (both poloidal planes and toroidal planes)


dschwoerer:
This would need some cleaning (purging the binaries).
It seems to add some quite useful functionality. 
I haven't tested it yet though ...

bendudson:
Yes, I think this definitely is useful but needs tidying. I think Peter (H) started having a look at it a while ago, but maybe never got around to it. Peter Naylor is a bit busy at the moment, but maybe when he's finished his dissertation...

bendudson:
Merged into a branch `peternaylor-Visboutit` so it's not lost. The manual could be merged into the main manual, and the visualisation code is quite self-contained.

d7919:
As this is standalone should this in fact go into a top level repository rather than the main BOUT++ one?

bendudson:
Yes I think that would make sense. There are some other projects which are related to BOUT++ but don't need to go in the main repository such as BOUTgui https://github.com/joe1510/BOUTgui

ZedThree:
Pulled out into stand alone repo: https://github.com/boutproject/VisBoutIt

PR #88, opened 2015-08-28
Cleanup whitespace

ZedThree:
Whitespace commit of doom. Remove all trailing whitespace, indent
everything consistently, destroy all tabs

### DIFF
--- a/src/bout++.cxx
+++ b/src/bout++.cxx
@@ -7,7 +7,7 @@
  * Copyright 2010 B.D.Dudson, S.Farley, M.V.Umansky, X.Q.Xu
  *
  * Contact Ben Dudson, bd512@york.ac.uk
- * 
+ *
  * This file is part of BOUT++.
  *
  * BOUT++ is free software: you can redistribute it and/or modify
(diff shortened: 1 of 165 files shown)


ZedThree:
This was generated by a short lisp program that takes <5 minutes to run, so it could be run on every commit.

dschwoerer:
Re-indention makes git blame reports the user who did the re-indention, instead of the user who actually changed it. Therefore I think this commit should only merged, if this indention style is the one which is preferred, and also will be enforced in the future.

loeiten:
I really like the idea :smiley:.
Looks like the lisp program is also editing on the spaces between a variable and ```=```, as in line 403 in src/bout++.cxx. Could it be that it would make the code less readable certain places?

ZedThree:
@dschwoerer This is true, but there are various tools that do recursive blame, letting you see the previous changes. Commits with only whitespace changes are generally not great for this reason, but a single massive commit like this is preferable perhaps.

@loeiten That particular line is the result of previous alignment of the `=`s being done with a mix of tabs and spaces, so it would only look correct if your tabs were set the same as the original author's. This commit replaces all tabs with 4 spaces, so there's likely several places where it does look worse. If you spot any more, let me know and I'll realign everything with spaces.

PR #86, opened 2015-08-27
Travis CI

ZedThree:
Basic configuration for Travis CI (Continuous Integration). Creates a fresh virtual machine and attempts to configure, compile and test BOUT++ for each commit or pull request, and displays the status of the latest attempt (building/failing) on the front page of the github repo.

PR #106, opened 2015-10-02
Fix tests

dschwoerer:
Some tests failed if run on Fionn.
These changes fixed them, and I think they should not change the overall result.


dschwoerer:
Already fixed in pull-request #86.

PR #112, opened 2015-10-16
Clean up init

### LINKED ISSUE #68: Clean up __init__.py
I suggest to clean up `__init__.py`.
As things are implemented now the whole `boututils` and `boutdata` is loaded (even if one is only interested in using let's say `collect.py`).

A fix could be to let `__init__.py` only contain the information on what to do when typing
```python
from boututils import all
```
, and let other dependencies be specified within the individual python routines.

loeiten:
Suggestion of new ```__init__.py```.

Currently these files loads all modules of the folders, even though only one module is requested. This can cause an annoying waiting time on certain clusters when running scripts.

This fixes #68


bendudson:
Wow, this is quite an epic commit! Thank you very much. It appears to pass tests (Thanks Travis CI!). Anyone object to this change?

ZedThree:
I broadly support this, I just have a couple of suggestions. As most of the package is one function per file, you now have to do `from boututils.function import function`. This looks like it would be a little annoying for functions which are often used together. Could the overall package structure be tidied up a bit as part of this PR? For example, moving `shell`, `launch` and `getmpirun` into a directory called, e.g., `run`, so that we can do `from boututils.run import shell, launch, getmpirun`. Maybe the same for the plotting functions.

Also, as `boutdata` imports `collect`, I'm not sure there's a need to change `from boutdata import collect` to `from boutdata.collect import collect`, [here](https://github.com/boutproject/BOUT-dev/commit/72bb4790eb8de1b8d66e893e805d258fe24efae8#diff-fddfaa163db53d3d4e62d783e76c0392L8) for example.

ZedThree:
These files all seem to be related and could be moved to a submodule: analyse_equil_2.py, efit_analyzer.py, moment_xyzt.py, red_gedsk.py, surface_average.py, volume_intergral.py.
Coincidentally, they all use Bunch, which I suggest could be replaced with either plain dicts or namedtuples, thus removing an otherwise unneeded dependency.

loeiten:
I agree with @ZedThree.

So...
* If all agree, I could merge ```shell```, ```launch``` and ```getmpirun``` into one file.
* Indeed, the ```boutdata.collect import collect``` in unnecessary as it is already loaded in ```__init__.py```. No problem to revert it back to normal syntax.
* If it is straight forward, I could replace ```bunch``` with normal dicts.
* As for the plotting modules and the modules using ```bunch```, I do not too much experience with them, but I could combine them as well.

loeiten:
```shell```, ```launch``` and ```getmpirun``` has now been merged to ```run_wrapper.py```.

If we are to merge the plotting routines, I suggest to merge those which are using mayavi into one file (```anim```, ```polplotslice``` and ```View3D```) and those who works out of the box (```showdata``` and ```plotdata```) into another. This, however, will probably take some time as several of the current files are written in a way such that they use some globals and up to several functions.

Also, merging the files using ```bunch``` is not that straight forward. Usually, the bunch-objects are returned from functions, which makes it a bit time consuming to track down.

Thus, I suggest that (if that is what we would like):
* The authors of the plotting routines rewrites them in such a way that they can be merged into one file (a good excuse to remove out-commented stuff).
* The author of the files using ```bunch``` rewrites them in such a way that they can be used out of the box with standard python tools.

PR #115, opened 2015-10-29
Laplacian solvers

bendudson:
Adds LaplaceXY, which solves the n=0 component including both X and Y derivatives

Adds LaplaceXZ, which solves a similar equation to Laplacian, but formulated in a different way and implemented in a (slightly) more efficient way.


bendudson:
Travis-CI fails because this branch doesn't include the Travis hooks. The test suite does still pass.

loeiten:
Merged master into this pull-request (that means that the newest IMEX-BDF2 from master was chosen over changes made previously in this branch).

This branch builds fine on my system with petsc installed. Not sure why travis fails (travis is not set with petsc options?)

However, there is a problem:
Running ```test-laplacexz``` throws range fault. The reason for this can be found in ```src/invert/laplacexz/impls/petsc/laplacexz-petsc.cxx:232:```, the outer loop is starting at ```z=0```, whereas ```A(x,y,z-1)``` is requested.

d7919:
Yes the travis build doesn't include PETSc (as far as I'm aware) so if there are any locations not protected by directives then the compilation will fail. Looking at the travis build log it looks like this is indeed what is happening as it's trying to compile the laplacexz-petsc source file and makes reference to routines which will involve petsc calls.

ZedThree:
Should just be a matter of adding

    sudo apt-get install petsc petsc-dev

to `.travis.yml` to get PETSc installed for Travis.

bendudson:
Thanks for catching these Michael! Good thing we have pull requests for things like this... I have added #ifdef BOUT_HAS_PETSC  so should compile, and fixed the Z index out of range bug (I think)

Issue #121, opened 2015-11-25
Adding traceback to BoutException


loeiten:
Would it make sense to add some sort of back-trace to `BoutException`?
The advantage is of course the potential to shorten debugging time, while the drawbacks could be difficulties to make this portable/that it may confuse users not familiar with the stack trace.

Examples of implementation using `execinfo.h` is given in the answers below
http://stackoverflow.com/questions/77005/how-to-generate-a-stacktrace-when-my-gcc-c-app-crashes

d7919:
I've not looked into it but doesn't the `msg_stack` variable contain essentially this information, so it would simply be a matter of printing the entire list?

bendudson:
This might indeed be quite nice. I have some concern about how to make sure it compiles with non-gnu compilers, and maybe on things like AIX, *BSD, OSX etc. One possibility might be to have configure try compiling a test case (as is done now to determine the floating point type for CVODE), and only include this functionality if the test works.

msg_stack provides something like this, but it relies on the code developer pushing and popping messages. The advantage is that the messages can (in principle) be more informative, but it does have some overhead. The entire msg_stack list should be printed in the event of a crash, see line 517 of bout++.cxx

  output.write("\n****** SEGMENTATION FAULT CAUGHT ******\n\n");
#ifdef CHECK
  /// Print out the message stack to help debugging
  msg_stack.dump();
#else
...

d7919:
One other issue to keep in mind could be the demangling of names, though there is some discussion of this in a later answer on the same thread.

Presumably one option to encourage the developer to populate the `msg_stack` would be to provide a macro along the lines of the below, this also has the benefit of having no overhead when doing a production run, being portable and allows more fine scale traces if required (i.e. can have multiple messages added within a single routine).

~~~~
#ifdef CHECK
#define TRACE(name,idname) int idname = msg_stack.push("In "+str(name)+ " in "+__FILE__+" at "+to_string(__LINE__))
#define ENDTRACE(idname) msg_stack.pop(idname)
#else
#define TRACE(name)
#define ENDTRACE
#endif
~~~~

Of course this still requires the developer to add the `TRACE(name)` and `ENDTRACE` lines manually. It may be possible to remove the requirement on the presence of `ENDTRACE` by making `msg_stack.push` return something which `pop`s the message when it goes out of scope (i.e. leaves the routine).

bendudson:
[...]

PR #124, opened 2015-12-08
basic backtrace in exception

dschwoerer:
A possible implementation of a backtrace, see issue #121 
Before it can be pulled, I need to add a check at configure time, whether this is supported on the environment - but only if this backtrace is of interest ...

the backtrace example produces for me something like:
Initialising physics module

```
[bt] Execution path:
[bt] #1 ./backtrace() [0x420d1c]
BoutException::BoutException(char const*, ...) at ??:?
[bt] #2 ./backtrace() [0x40f423]
physics_init at .../examples/backtrace/backtrace.cxx:11
[bt] #3 /lib64/libc.so.6(__libc_start_main+0xf0) [0x7f5deca24fe0]
?? ??:0
[bt] #4 ./backtrace() [0x410965]
_start at ??:?
Failed to initialise physics. Aborting
```

Note the line number, which I think can be very useful, if the physics_init / physics_run functions are longer. Additionally, if no/not many loops in the $TARGET.cxx are, the performance penalty should be negligible, as only this file is compiled with -g to allow the line number to be printed.


dschwoerer:
I updated configure.ac:
 * use enable if appropriate (track, check,debug, etc.)
 * add option to enable FPEs
 * add option to call abort in exceptions

The point of abort in exceptions is, if run inside e.g. gdb, you get to take over where the exception is thrown, check the memory etc. I found this to be quite handy, but of course can also be done by setting a breakpoint in the exception.

Right now, if an exception is thrown (with ABORT disabled) this is done quietly, i.e. no backtrace is immediately printed. I think that is consistent with the old behaviour. I am not sure this is the way it should be, as if I throw an exception in the phisics_init function, what() will not get called/printed and I need to add some effort to get the error message (which exists, but is not printed).

Are there any exception which can be properly catched? I.e. where the run can and is supposed to continue?

ZedThree:
I'm reasonably certain all the exceptions are fatal, and the run should not continue. I think in a few cases it is possible for the run to continue, but an exception is thrown because it indicates some missing/bad configuration in the input file, so you probably don't *want* it to continue. I guess you might want to do `try {} catch {}` in your physics model though.

Also, for gdb at least, you can do `catch throw` to let gdb automatically catch exceptions (i.e. no need to set a breakpoint inside the exception). This will catch it before the stack is unwound, so you do get a backtrace.

dschwoerer:
Thanks, putting `catch throw` in .gdbinit is probably a much better way to do it :+1: 
I didn't get the error message in the physics_init because I didn't throw the exception :flushed:

bout_error calls now BoutException - I think it doesn't make sense to maintain the code twice ...

dschwoerer:
Should I close this?

bendudson:
Having a backtrace and an option to catch SIGFPE seem like useful things to have

PR #125, opened 2015-12-08
typo in configure.ac

dschwoerer:
I think it should be sundials ...


d7919:
I think this has already been fixed in master. See 9e35fc3ca20d6b9e02ddd10f5165e48a1908b5dc or pr #98

loeiten:
Indeed, it has been fixed in ```configure.ac```. I looks, however, that I forgot to type ```autoconfig configure.ac``` after, so there is still a typo in ```configure``` file in master.

dschwoerer:
I thought I had pulled before ...
But as long as its fixed ...

PR #126, opened 2015-12-08
throw exception if given boundray is not found

dschwoerer:
I think if the given boundary does not exist, this should cause some serious action, and not only be printed.
It can cause quite serious errors, if the wrong boundary is set, and might not be noted otherwise ...


loeiten:
This would be nice indeed.

PR #130, opened 2015-12-14
Unique field group

### LINKED ISSUE #129: Repeated references in FieldGroup leads to incorrect behaviour when TwistShift=True
The twist shift condition is applied during the communicate stage. If the communicated FieldGroup has multiple references to the same field then the twist shift modification can be applied multiple times.

Commit  bbb20bf demonstrates a proposed simple fix to this which eliminates repeat references from FieldGroup.

d7919:
Fixes #129 and adds a small test case

### DIFF
--- a/examples/test-fieldgroupComm/cyclone_68x32.nc
+++ b/examples/test-fieldgroupComm/cyclone_68x32.nc
@@ -1,0 +1,1 @@
+../test-gyro/cyclone_68x32.nc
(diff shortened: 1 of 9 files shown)


bendudson:
Merged this fix for now, but I think a better solution might be to prevent FieldGroup from ever containing duplicate references. This could be checked once when fields are added, rather than every time communicate is called.

d7919:
Yes I agree, making sure FieldGroup can never contain multiple references is probably the better approach. I think the only reason I didn't do this in the fix is because it would (I think) require changing all of the FieldGroup::add definitions in include/bout/fieldgroup.hxx and all of the constructors in src/field/fieldgroup.cxx, but this isn't really a *good* reason for not doing it.

PR #144, opened 2016-01-12
Configure optimize

dschwoerer:
added --with-optimize to the code, allowing to use the compiler optimization without having to change the make file / specifying the flags manually.
Currently the flags are intended for gcc, but are checked whether the compiler supports them.


dschwoerer:
Are there any objections to this change?

bendudson:
Apparently no objections, so merged. Thanks David! Have a happy Easter.

PR #146, opened 2016-01-13
Added target check to makefile

dschwoerer:
just run

```
make && make check -j 8
```

to run the whole test_suite in parallel.

The output is not as nice as examples/test_suite but its faster and easier ...

also `make check` fails to build the library reliably.

Additional make doc builds the manual

### DIFF
--- a/configure
+++ b/configure
@@ -657,6 +657,7 @@ EXEEXT
 ac_ct_CXX
 CPPFLAGS
 CXX
+ARFLAGS
 RANLIB
 SET_MAKE
 LN_S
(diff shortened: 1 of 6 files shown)


dschwoerer:
Should I split the ar change into another pull request?

bendudson:
Has this PR been replaced/superseded by other changes? Is it still needed?

dschwoerer:
The issue with `ar` still exists. (If `ar` has been build with default for deterministic mode, the `u` flag results in a warning)

Further, I would like to be able to call `make check -j <n>` - but I guess the conflicts are by now so bad, it would need to be redone.
I would be willing to do that, but only if there is interest (and it will get merged this time)

bendudson:
Can you re-submit the ar changes as a new PR? The faster checks looks a bit more involved, and can probably wait. I think that could be part of a later push to add tests:

a) Checks which run really fast, and could be used frequently as code is being developed
b) Middling checks which can be run in Travis in reasonable time
c) Thorough "extreme vetting" which could be done very occasionally e.g. for releases

bendudson:
I think this PR is now obsolete

PR #147, opened 2016-01-14
Travis matrix

dschwoerer:
Test travis


dschwoerer:
This would be helpful to automatically check the OpenMP branch ...

bendudson:
Which bits of this are still useful / necessary?

dschwoerer:
Some parts have been fixed independently in the last year ...

The main idea, to also test with clang, is, as far as I can see not yet done in the 4.0 RC (openmpi is install but mpich flags are set)

PR #163, opened 2016-02-09
Inner outer mesh iterators

nick-walkden:
Four new boundary iterators have been added to boutmesh.cxx to allow iteration over either inner or outer boundaries in double null configurations.

The iterators have been tested and function as desired, but the test has not been included in this branch as it seemed too minor to include. I can commit this if needed.

If all are happy this is ready for merging.


bendudson:
I've modified the BoutMesh code slightly to simplify it a little, and also keep the old versions of the Y iterators to guarantee backwards compatibility. I think there must be a better way to handle these different boundaries, but this will do for now.

dschwoerer:
Merged as #486
can probably be deleted?

PR #140, opened 2016-01-11
Configure

dschwoerer:
Should fix issue #103

Issue #103, opened 2015-10-02
./configure can no longer find libfftw3.a


loeiten:
When ```./configure``` BOUT++, the configure can no longer find libfftw3-dev as installed as described in the manual. 

The output of configure reads

```
Searching for FFTW library
 -> Path not given. Searching for library...
 -> Found fftw-wisdom executable
checking for /usr/include/fftw3.h... yes
checking for /usr/lib/libfftw3.a... no
checking for /usr/lib64/libfftw3.a... no
ERROR: FFTW not found. Required by BOUT++
```

I guess this comes from commit 2bad94c04346d1c7a41431011988b6eecbb342dc, where search paths of ```libfftw3.a``` has been changed. Should these be changed back?

dschwoerer:
In earlier versions, a missing libfftw3.a was ignored.
This was inconsistent with the other library, where both the .a and .h was required.

dschwoerer:
For fedora-like distribution, the packages fftw-devel and fftw-static.
I think there are two options, ether check for the shared objects, or update the documentation.

loeiten:
As pointed out by Dr. Alf Köhn, this is still a bug on some (maybe several) distributions.

Installing ```libfftw-dev``` with ```sudo``` puts some of the files into the ```/usr/lib/x86_64-linux-gnu/``` folder, which ```./configure``` currently does not search for.

If ```fftw``` is installed both with ```sudo``` and locally, it looks like the search starts in the ```/usr/``` directory and throws a message when the lib-files are not found there, without searching in ```$HOME/local/```.

I suggest to append ```/usr/lib/x86_64-linux-gnu/``` to the searchable paths.

ZedThree:
This is the case on Ubuntu 14.04, at least. I think this is also a more general concern than just FFTW.

One option would be to use pkg-config to check the location of a package. Something like:

    PKG_CHECK_MODULES([fftw3], [fftw3 >= 3.3.3])

should set `FFTW3_LIBS` and `FFTW3_CFLAGS` to their correct values. However, I'm not sure how they interact with setting the paths manually. Maybe a check to see if a path is given to the `--with-package` option, and if so, use that path, if not, use pkg-config.

Alternatively, we could do some parsing of `/etc/ld.so.conf`, but I feel that might be a little gross.

Another option would be to enable setting `--with-package-lib` and `--with-package-inc` as separate arguments.

[...]

PR #166, opened 2016-02-23
D Schwoerer's configure improvements

bendudson:
Fixed PETSc configure problems, caused errors on York's "snake" machine.

Now compiles on my machine, York servers, and Archer. Could everyone please try this branch on their machines?


loeiten:
The configure now works on both my laptop and on the cluster, and fixes the problems mentioned in #140.

Issue #103 is still a problem though. In some distro packages, ```libfftw``` is located in ```/usr/lib/x86_64-linux-gnu```, and as far as I see this is not a search path. The error message I get when building with

```
./configure --with-checks=3 --with-track --with-debug --with-petsc=$HOME/petsc-3.4.5 --with-sundials
```

is

```
Searching for FFTW library
 -> Path not given. Searching for library...
 -> Found fftw-wisdom executable
checking for /usr/include/fftw3.h... yes
checking for /usr/lib/libfftw3.a... no
checking for /usr/lib64/libfftw3.a... no
checking for /usr/lib/libfftw3.so... no
checking for /usr/lib64/libfftw3.so... no
ERROR: FFTW not found. Required by BOUT++
```

dschwoerer:
What os/version are you using?

I couldn't find any, that causes this behaviour.

I found a bug, I only looked for $prefix/x86_64-linux-gnu , not $prefix/lib/x86_64-linux-gnu.

Still, if it is installed with apt-get (or similar), it should be in the default searchpath, and this should not be needed.

Secondly, I am puzzled, as configure should not look for these files. It should try to link with them, i.e. only the linker should look for them.

Finaly, the error message in the configure.ac is:
AC_MSG_ERROR('FFTW3 is needed by BOUT++') - not 'FFTW not found. Required by BOUT++'. - I can't find this error string in the configure - are you using the most recent version?
If so, could you provide me e.g. the config.log?
What distro / version are you using?

I am really confused and would like help to fix this issue.

loeiten:
So, I'm using ```Linux Mint 17 'Qiana' MATE 64-bit```.

I tried to build ```BOUT++``` with the package
```libfftw3-bin version:3.3.3-7ubuntu3``` (I also tried installing ```libfftw3-dev``` with the same version, but it gives the same error).

The FFTW3 package was installed through the software manager (equivalent to ```sudo apt-get install``` AFAIK).

The error message above was a copy paste from the mentioned issue, sorry about that. The current error message is:

```
Searching for FFTW library
 -> Found fftw-wisdom executable
checking for g++... g++
checking whether the C++ compiler works... yes
checking for C++ compiler default output file name... a.out
checking for suffix of executables... 
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C++ compiler... yes
checking whether g++ accepts -g... yes
checking for fftw3.h... yes
checking for libfftw3... no
configure: error: "FFTW3 is needed by BOUT++"
```

The [config.log](https://github.com/boutproject/BOUT-dev/files/150159/config.log.txt) (renamed to ```config.log.txt``` to make it easier to share with github).

Hope this helps :)

loeiten:
Oh dear....I managed to pull the wrong version (by using git's ```command line instruction``` ```Step 1```). I've pulled the correct version, and it builds both on my laptop and on the cluster. 

My complaints above are to be ignored. Sorry for the inconvenience.

PR #167, opened 2016-02-26
Clean up

loeiten:
Small clean up in `\manual` and matlab-files in `\tools`.

I suggest that the `tools` directory should be cleaned up as well (both because a clean code in general is nice, and because it is easy to get lost in what is going on in that folder). However, I do not know if there is any dependencies there, or any particular reason for the folder structure there (e.g. why `.pro` files are scattered around and not in the `idllib`).


bendudson:
I think the criteria for a routine being in idllib (or pylib, matlablib etc.) should be that it's generic enough to be useful in at least two different tools/codes. Many of the .pro IDL codes under tools/ are specific tools for doing things like generating grid files, and not generic library routines which could be imported or used in another context. I'm willing to revise this criterion though if people think differently (peer pressure...)

loeiten:
That makes good sense :).

Maybe they don't need to be moved, but maybe just cataloged in a way so it is clear what the purpose of each folder/file is?

At the moment I have questions like:

* What the relationship between ```numlib```, ```plasmalib```, ```pdb2idl``` and the other IDL files? 
* Why are there two ```line_tracing``` folders?
* What would I have to do in order to run ```cyl_and_helimak_grids```
* Should ```hypnotoad``` be mentioned in the ```user_manual```

loeiten:
Closing this, as a real clean-up is needed.

PR #168, opened 2016-03-14
PETSc 3.6

bendudson:
Small fixes so BOUT++ configures and compiles with version 3.6. Tested on Archer with 3.6.1


bendudson:
Pull requested too soon. Some parts must not have recompiled, now still failing.

bendudson:
Now compiles with PETSc-3.6.1

PR #172, opened 2016-03-31
make travis output more verbose if configure fails

dschwoerer:
Just trying to understand why the build fails ...


dschwoerer:
got it right in the mail, and failed implementing it ...

PR #182, opened 2016-04-14
easy 1D Data Access

dschwoerer:
I am looking for an efficient way to loop over all elements of a field(3d).
So far i used (`f` is a `Field3D` and `nxyz` is `ngx*ngy*ngz` of `mesh`):
```
BoutReal * d=f[0][0];
for (int i=0;i<nxyz;++i)
  d[i]=0;
```
In the next branch I could use
```
BoutReal * d=f(0,0);
for (int i=0;i<nxyz;++i)
  d[i]=0;
```
but that disables all checks.

Would you be in favour to add:
```
for (int i=0;i<nxyz;++i)
  f(i)=0;
```
or do you think something in the way of
```
for (int i=0;i<nxyz;++i)
  f.data_1d(i)=0;
```
would be more intuitive to use?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,13 @@
 language: cpp
 compiler: gcc
+matrix:
+  include:
+  - env: CONFIGURE_OPTIONS='--enable-openmp'
+  allow_failures:
+  - env: CONFIGURE_OPTIONS='' MPICH_CC=clang MPICH_CXX=clang++
+    compiler: clang
+  - env: CONFIGURE_OPTIONS='--enable-openmp' MPICH_CC=clang MPICH_CXX=clang++
+    compiler: clang
 before_install:
   - sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y
   - sudo apt-get update -qq
--- a/.travis_script.sh
+++ b/.travis_script.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-./configure
+./configure $CONFIGURE_OPTIONS
 conf=$?
 if test $conf -gt 0
 then
--- a/configure
+++ b/configure
@@ -649,6 +649,7 @@ LTLIBOBJS
 PACT
 SLEPC_VARS
 PETSC
+OPENMP_CXXFLAGS
 LIBOBJS
 EGREP
 GREP
@@ -734,6 +735,8 @@ with_pvode
 with_hypre
 with_mumps
 with_arkode
+with_openmp
+enable_openmp
 '
       ac_precious_vars='build_alias
 host_alias
@@ -1355,6 +1358,12 @@ if test -n "$ac_init_help"; then
    esac
   cat <<\_ACEOF
 
+Optional Features:
+  --disable-option-checking  ignore unrecognized --enable/--with options
+  --disable-FEATURE       do not include FEATURE (same as --enable-FEATURE=no)
+  --enable-FEATURE[=ARG]  include FEATURE [ARG=yes]
+  --disable-openmp        do not use OpenMP
+
 Optional Packages:
   --with-PACKAGE[=ARG]    use PACKAGE [ARG=yes]
   --without-PACKAGE       do not use PACKAGE (same as --with-PACKAGE=no)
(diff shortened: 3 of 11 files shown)


bendudson:
Is there a reason not to use an iterator?

for(auto& i : f)
  f[i] = 0;

In simple tests I think this is as quick as a single c loop, or close enough to not worry about the difference (in the next branch). It also has the advantage of optional checking, openmp (thanks! I guess a single c loop has this too), stencils (i.xp(), i.xm() etc.) and c++11 loop goodness.

My preference would be to make the iterator as efficient as possible, since I think it's the safest and most convenient method. If it really can't give the performance needed (and I think we'll be doing very well if this matters) then maybe making the underlying Array object a public data member would be ok. This would follow the c++ convention used for vectors, and a syntax like the last of your options. I'm a bit reluctant to allow public access to the internal data structures though unless it's really really needed.


-------- Original Message --------

dschwoerer:
I had in memory that the iterator was slower than plain c single for loop.

Unfortunately the performace/iterator is broken, so I didn't test that ...

bendudson:
In the master branch the iterator definitely is slower, but when testing the next branch a while ago (6 months? Maybe more) I think the difference was marginal. More testing always good. I'll have a look at the performance test tomorrow.


-------- Original Message --------

dschwoerer:
Just checked with test-dataiterator/speed:
Without optimization

    DataIterator       : 1.379897
    triple for loop    : 0.373417
    single for loop    : 0.166649

With optimization:

    DataIterator       : 0.920417
    triple for loop    : 0.167611
    single for loop    : 0.166200

It's nice to see that the triple loop does get optimized, but I should check why the fielditerator is so much slower, I also remember that the difference was much smaller ...

dschwoerer:
If the internal data shouldn't be exposed - should the function `BoutReal * Field3D::opoerator(int,int)` then be removed or maybe rather be declared depricated?

bendudson:
I've cleaned up the performance/iterator test, which was broken by changes in next (eg no Field3D::getData function). The timing is for a loop over the domain, adding two field3Ds together.

With gcc 4.8.0, flags "-std=c++11 -O" (no checking)

C loop                     : 0.043142
----- (x,y,z) indexing ----
Nested loops               : 0.147492
MeshIterator               : 0.129266
DataIterator (begin/end)   : 0.181778
DataIterator (begin/done)  : 0.085671
C++11 range-based for      : 0.090283
------ [i] indexing -------
C++11 Range-based for      : 0.090225
DataIterator (done)        : 0.085563

With gcc 4.8.0, flags "-std=c++11 -O3" (no checking)

C loop                     : 0.04229
----- (x,y,z) indexing ----
Nested loops               : 0.043516
MeshIterator               : 0.089536
DataIterator (begin/end)   : 0.221265
DataIterator (begin/done)  : 0.046353
C++11 range-based for      : 0.0517
------ [i] indexing -------
C++11 Range-based for      : 0.051503
DataIterator (done)        : 0.044618

bendudson:
As a test I tried making the "data" member of Field3D public, allowing operations like:

for(auto& val : result.data)
   val = 0.0;

or 

for(int i=0;i<result.data.size();++i)
   result.data[i] = a.data[i] + b.data[i];

With -O3 optimisation
C loop                     : 0.042405
C++11 Range-based for      : 0.051688
data member access         : 0.054621

and -O optimisation:
C loop                     : 0.043126
C++11 Range-based for      : 0.089832
data member access         : 0.064003

Looks like the 1D access to the underlying Array class is easier to optimise (in some sense) than the DataIterator method, but with sufficient optimisation the compiler can make them comparable speeds. 

If people feel that a 1D array access method is useful, then I'd feel happier about adding this form of 1D access than a operator(int index) 1D access to Fiel3D, if only because it has a size() method and iterators.

ZedThree:
Personally, I feel like the gains made by using the `data` member are too marginal to make it worth exposing it. Exposing it makes the code more fragile, as we may change the details of how the data is stored (this has already happened once in the last year).

Also, I think using the `DataIterator` results in much cleaner code.

dschwoerer:
Seems only my non-OpenMP code with DataIterator is extremly slow:

    DataIterator       : 0.906596
    DataIterator(done) : 0.122796
    triple for loop    : 0.131485
    single for loop    : 0.119244

I will try to fix this again ...

You can always get a pointer to the data, nothing prevents you from using
```BoutReal *data = &field(0,0,0);```
I would rather give a function to get direct access and declare it as deprecated - nobody should be surprised if it changes ...

PR #186, opened 2016-04-18
Fast non-OpenMP DataIterator

dschwoerer:
As you might have noted the DataIterator begin/end is currently extremly slow:

```
DataIterator (begin/end)   : 0.0747581
DataIterator (begin/done)  : 0.0227261
```

This fixes this issue, but unfortunately in an unsatisfying way.
C++17 would allow end and begin to return different types, therefore a dummy type could be returned, which would allow to safely overload !=, but C++11 doesn't.
This is a temporary workaround*.

*https://i.imgur.com/M5wl14r.png (i.e. we should look for a better fix)


bendudson:
Thanks for the the handy lookup table, like it :). Agree this is a mildly awful hack (I've seen and perpetrated far worse!), but maybe acceptable. Is this the main cause of the slow performance? @ZedThree pointed out to me that the performance/iterator test was inefficient:

for(DataIterator i = begin(result); i != end(result); ++i) 

creates a new "end" object every iteration, whereas the C++11 range for loop uses a form close to:

for(DataIterator i = begin(result), rend=end(result); i != rend; ++i) 

which then gives performance closer to the begin/done form.

dschwoerer:
Unfortunatly I didn't understand why the performance dropes.
The end only gets generated once.

Maybe the compiler isn't any more sure what the loop does and is therefore not able to optimize it. I only tested with gcc-5.3.1, maybe other compilers perform better ...

dschwoerer:
Intel [1] without checks and `-fast` results in:
```
C loop                     : 0.0236002
----- (x,y,z) indexing ----
Nested loops               : 0.0226514
MeshIterator               : 0.0336627
DataIterator (begin/end)   : 0.0389248
DataIterator (begin/done)  : 0.036247
C++11 range-based for      : 0.0501092
------ [i] indexing -------
C++11 Range-based for      : 0.0495127
DataIterator (done)        : 0.0363222
```
Which is a bit better than `gcc-5.3.1` in regard to the end comparison, but otherwise not as good.
[1] mpiicpc -v
mpiicpc for the Intel(R) MPI Library 5.0 Update 3 for Linux*
Copyright(C) 2003-2015, Intel Corporation.  All rights reserved.
icpc version 15.0.3 (gcc version 5.1.0 compatibility)

dschwoerer:
Any objections to this?
The most recent version isn't as bad as the initial one, as the dataiterator has an extra flag, so the != can check whether the other one is the end - and only in that case do the fast comparison, and otherwise a real comparison ...

bendudson:
This seems like a nice solution, thanks David!

bendudson:
Please update base branch to v4.0.0-rc

PR #187, opened 2016-04-18
Fix array.cxx

dschwoerer:


bendudson:
Why is this fix needed for the Intel compiler? Does it not follow the C++11 standard?

dschwoerer:
Hi, as far as I could find, Intel claims that the Intel compiler (version 15 I think) does support C++11.
I don't know whether this is a bug in the compiler, but this works also with gcc, and I personally think it's also easier to read.

ZedThree:
Partial support from Intel C++ 13.0, fully supported from 14.0 according to [this list][1]
(see "General initializer lists"). If it doesn't work, it's a compiler bug.

This is fine in this case, as it's just empty, but in general, the two aren't equivalent, and `T foo {...}` should be preferred. See [this SO answer][2] for reasoning, but basically `T foo {...}` doesn't allow implicit casting, but `T foo = {...}` does.

[1]: https://software.intel.com/en-us/articles/c0x-features-supported-by-intel-c-compiler 
[2]: http://stackoverflow.com/questions/18222926/why-is-list-initialization-using-curly-braces-better-than-the-alternatives

bendudson:
Thanks Peter! Agree since empty here it's probably ok, but something to watch out for I guess.

dschwoerer:
In case anybodies interrested:
Intels compiler, as well as the cray compiler fail to compile this initialiser list - it seems to be an issue when initializer lists in combination with templates are used:
```
template<typename T>
struct Array {
  static bool used;
};

template<>
bool Array<double>::used {false};


int main(){
  return Array<double>::used;;
}
```
I don't know where to submit these bugs ...

PR #188, opened 2016-04-19
add FieldPerp::operator(int,int)

dschwoerer:
In order to make the master branch closer to next.
I would like to be able to compile the modules both with the current master, as well as with next, therefore I think we should add all operator(int,int[,int]) to the fields etc.

I think we should also declare the operator[int] as deprecated, but I wanted to ask first what you think about this.

If we want to push for the next branch, I think we should make the transition as smooth as possible ...

### DIFF
--- a/include/fieldperp.hxx
+++ b/include/fieldperp.hxx
@@ -57,6 +57,8 @@ class FieldPerp : public Field {
   // operators
 
   BoutReal* operator[](int jx) const;
+  BoutReal & operator()(int jx, int jz);
+  const BoutReal & operator()(int jx, int jz) const;
 
   FieldPerp & operator=(const FieldPerp &rhs);
   FieldPerp & operator=(const BoutReal rhs);
--- a/src/field/fieldperp.cxx
+++ b/src/field/fieldperp.cxx
@@ -110,6 +110,36 @@ BoutReal* FieldPerp::operator[](int jx) const {
   return data[jx];
 }
 
+const BoutReal & FieldPerp::operator()(int jx, int jz) const {
+#if CHECK > 2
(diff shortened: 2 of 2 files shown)


loeiten:
Nice :)

Are there any reasons not to merge master and next at the current time?
When is an eventual merge planned?

bendudson:
Thanks! Yes, easing the transition is a good plan. The main reason for not merging yet is that next changes/breaks a bunch of crucial things. In particular:

1. when using shifted X derivatives (the default for tokamak simulations) fields are now in X-Z orthogonal coordinates rather than field-aligned coordinates, and all the shiftZ stuff has been removed and hidden away so that FCI can be used transparently. The outputs will be in a different coordinate system, and this change needs to be thoroughly tested. Probably MMS'd... sigh :(

2. Smaller changes like the metric components have been moved. mesh->g11 becomes mesh->coordinates()->g11 etc.

I think the next step is for people to test their codes in "next" branch, find and squash the bugs, then merge can happen

loeiten:
Ok, that makes sense :). Just some questions before changing to `next`

1. Is the `next` branch supposed to be as stable as the `master`, or is it meant as something more experimental?
2. Are there any policy of merging from `master` to `next`?

bendudson:
1. The next branch is supposed to pass the test suite like master, but is at the moment more experimental than the master branch. I think boundary conditions in y, and possibly the twist-shift condition may well be broken or have issues which are not caught by the test suite. 

2. Not particularly, though I'd like to keep next following any fixes or improvements in master. I last merged in ebd406c6 and don't think anything major has changed in master since then. Hopefully the merge will be less difficult next time.

PR #190, opened 2016-04-22
added operator(DataIterator) to field3d

dschwoerer:


bendudson:
Not sure I see why this is needed, as duplicates operator[]. Is there a readability or functionality improvement?

dschwoerer:
My mistake, operator() only exists for int,int,int - I thought it existed in more cases (and then it should exist imho in all), but the way it is, is probably better

Issue #19, opened 2015-01-12
Remove ngz-1


bendudson:
The mesh in Z currently uses ngz-1 points, though ngz length arrays are allocated. ngx, ngy and ngz are also not particularly descriptive names.

Propose replacing ngx, ngy, and ngz with variables such as LocalNx, LocalNy, LocalNz  (to go with GlobalNx, GlobalNy, GlobalNz). The extra point in Z can then be amputated.

bendudson:
This is fixed in the next branch. master branch not going to be fixed (probably)

PR #194, opened 2016-05-10
Renamed ngx,ngy,ngz to LocalN*

bendudson:
o Renamed local mesh variables to a more readable
  format, similar to GlobalN* variables.

o Extra point in Z removed, so LocalNz is the
  actual size of the domain, not ngz-1

o Increased BOUT_VERSION to 4, updated Python and IDL collect
  routines to handle lack of extra Z point in new outputs


dschwoerer:
I'm in favor of merging, as it should not break any code (name has changed -> its easy to detect where ngz needs adjustment)

If we wait to long, the branches might diverge ... and I am really looking forward to losing the extra point in z.

Solves issue #19

Issue #225, opened 2016-08-19
Making grid index routines public


loeiten:
Is there a reason for why the grid index routines (line 136-144 in `boutmesh.hxx`) are `private`?
If not, would it be OK if I made these `public`?

The same question applies for the data which are given in `BOUT.inp` or a grid file (for example `nx` and `ny`). I have no need (currently) to make these public, but I can see these can be handy to access from time to time.

bendudson:
The reason I've been reluctant to make these public is that it makes future changes to the mesh harder. For example nx and ny are currently constants, but in an irregularly shaped mesh might not be. The grid index routines in boutmesh.hxx really only handle the case where NXPE and NYPE are fixed, so for exampe NXPE doesn't depend on y position. 

I agree some kind of public mapping from local to global index would be useful, or finding which processor contains point (x,y,z), but I'd prefer them to be more general if possible.

loeiten:
I see, that makes sense.

A possible solution could be to throw a `NotImplemented` on these methods in mesh-classes where `NXPE` and `NYPE` are not fixed. Eventually they could deprecate and go back to being `private` if a general method is made.

loeiten:
Closing this, see discussion in #226 for details.

PR #226, opened 2016-08-19
Made the grid index routines public

loeiten:
- Made grid index routines public
- Added getters for `MXG`, `MYG`, `nx` and `ny`
- Added virtual functions in mesh (no warnigns are thrown if not implemeted as
  `BoutException` is not available)
- Fixed bug in `IS_MYPROC`

This fixes issue #225

### DIFF
--- a/include/bout/mesh.hxx
+++ b/include/bout/mesh.hxx
@@ -188,7 +188,23 @@ class Mesh {
   /// Local ranges of data (inclusive), excluding guard cells
   int xstart, xend, ystart, yend;
 
-  // These used for differential operators 
+  /// Getters for grid data
+  virtual int getnx() const {}; /// Return nx from the grid
+  virtual int getny() const {}; /// Return ny from the grid
+  virtual int getMXG() const{}; /// Return MXG from the grid
+  virtual int getMYG() const{}; /// Return MYG from the grid
+
+  /// Processor number, local <-> global translation
+  virtual int PROC_NUM(int xind, int yind) const{}; // (PE_XIND, PE_YIND) -> MYPE
+  virtual bool IS_MYPROC(int xind, int yind) const{};
+  virtual int XLOCAL(int xglo) const{};
+  virtual int YGLOBAL(int yloc, int yproc) const{};
+  virtual int YLOCAL(int yglo) const{};
+  virtual int YLOCAL(int yglo, int yproc) const{};
+  virtual int YPROC(int yind) const{};
+  virtual int XPROC(int xind) const{};
+
+  // These used for differential operators
   Field2D dx, dy;      // Read in grid.cpp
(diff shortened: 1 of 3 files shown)


bendudson:
The bug fix in IS_MYPROC is important (and a little worrying...). I'm not sure about making some of the other things public unless there's a really good reason, as they expose details of how the mesh is currently organised which may change in future

loeiten:
Maybe this pull is not necessary/could be rewritten, as it is a "quick and dirty" fix of a feature i needed.

I used `IS_MYPROC` a while back when I wanted to do a surface integral of my domain. This requires to compute `vDotDS` for the different surfaces. As I wanted to compute `vDotDS` only on the surfaces where it was needed, I found the `IS_MYPROC` routine handy.

The bug was found when MESing the integral.

bendudson:
Is this still necessary, or has it been replaced? I see some of these public routines in Jens' latest PR  #479

dschwoerer:
I am not sure about the IS_MYPROC.

Shouldnt it be rather:
```
bool BoutMesh::IS_MYPROC(int xind, int yind)
{
  return (((xind-MXG) / MXSUB) == PE_XIND) && (((yind-MYG) / MYSUB) == PE_YIND);
}
```
however, this may depend, on whether the global coordinates include the guard cells ...
As having guard cells in the global indexes is anyway dodgy (will return wrong results for guard cells, which may be as well be on this processor, and also for the first and last cells, the old code may be better, but never the less non intuitive.

I cannot understand why we should divide by `MXSUB+1`, if I read correctly `MXSUB` is the number of cells on this node, and not something like the last index ...

loeiten:
@bendudson 

I see that 
`BoutReal getNx() const {return nx;}`
`BoutReal getNy() const {return ny;}`
is introduced in #479 (would it not be better to change this to `int`?), however, it does not include
`getMXG()`
`getMYG()`
and the processor information functions.

I'm still in favor of having such functions accessible for the user, but the best option is maybe to close this PR, and edit #225 to a enhance request.

@dschwoerer 

It's been a while since I looked at this, but I remember that I got the correct result with my version.
Looks like your version would be right, assuming that `xind` and `yind` starts counting from `0` and is including the ghost points.


In any case `IS_MYPROC` is **never** used, so should be removed if we decide that the function should not be publicly accessible.

dschwoerer:
Concerning `IS_MYPROC` - I guess the function should be removed, as it is quite hard to tell what it should do. 
I think if you include the boundaries, you may, for only a few processors in that direction, closer to the expected result, than with the old implementation ...

Concerning the getters for nx and ny:
I added the getters for nx in #485 to the fields.
The reason is, that I otherwise first need to ask the field for its mesh, and then query the mesh, in case it has one, otherwise query the global mesh.

Instead of the getMXG, i always used xstart. Is there a good reason to use getMXG instead?

loeiten:
Good point, I'll close this request, and make one which removes `IS_MYPROC`.

Issue #251, opened 2016-09-23
Parallel derivative order /=2 incorrect on next for field3d


d7919:
With the new `yup`/`ydown` treatment of the parallel transform the `stencil.pp` and `stencil.mm` values are now set to 0 in `field3d::setYStencil` -- this means that schemes for the derivatives that require these (e.g. `C4`) no longer give the correct answer.

Note this does not apply to `field2d` which does not use `yup` or `ydown` when populating stencils.

bendudson:
To fix this we either have to have an option to store two planes in each y direction (5 fields rather than 3, compared to 1 in master), or give up on 4th-order operators.

JensMadsen:
I vote for making the machinery flexible for higher order schemes. In principle the number of stored planes should always equal MGY?

dschwoerer:
I'd consider it very bad coding style to break stuff silently.

If .pp isn't set, set it to nan("");
If somebody is using it, he will know very fast it is wrong. Without that people will spend weeks or months debugging!

d7919:
Yes I think it would be useful to ensure that this is easily noticed -- Perhaps we can add a check for the derivative order and abort if not supported?

dschwoerer:
Seems to me rather complicated.

Set it to nans.

Merge #124 (it adds support for enabling SIGFPE) and you will get the immediate feedback if the nan is used. True it might not be obvious where it came from, but at least you don't need to know what stancil is what order ...

dschwoerer:
Somewhat worrying: interchange-instability uses forth order in parallel direction, yet it didn't notice this error ...

We really need more (and more precise) tests ...

dschwoerer:
Another idea: Does the FCI scheeme need to be in the standard mesh?
As the derivative options are part of the mesh, using an FCI mesh, then FCI could drop the support for higher order, without breaking the rest of bout. Also it would still allow the non FCI code to be optimized ...

However, that would require set?Stencil not to be part of field3D, but rather of the mesh. But then if the Mesh knows what is where connected, and the field has only the data, it might make sense ...

d7919:
Whilst this is perhaps obvious I thought I'd also point out that this issue doesn't just impact the fourth order accurate stencils, it also impacts:
* The `d4dy4` operator that may sometimes be used in artificial dissipation etc.
* The `U2` upwind operator (`VDDX_U2`)
* The `W3` upwind operator (`VDDX_WENO3`/`VDDX_CWENO3`)
[...]

PR #253, opened 2016-09-23
Make sure people notice this bug ...

dschwoerer:
helps with #251 
Obviously more beneficial if SIGFPE is enabled, but even without better ...


dschwoerer:
I really think this should be merged.
It is bad if an simulation failes with an error, but it is far worse if the simulation produces a wrong results, and this is a well known issue!

bendudson:
Agreed, should be merged. Should this pull request be updated to merge into v4.0.0-rc?

dschwoerer:
Updated base to RC

PR #254, opened 2016-09-23
New test

dschwoerer:
* checks if solver can solve for field2d
* checks if min/max work as expected
* checks if staggered grids works

### DIFF
--- a/examples/test-minmax/data/BOUT.inp
+++ b/examples/test-minmax/data/BOUT.inp
@@ -1,0 +1,29 @@
+MZ=5
+
+nout=1
+timestep=1
+
+periodicY=true
+
+[MESH]
+nx=68
+ny=68
+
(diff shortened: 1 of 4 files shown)


ZedThree:
Given that this is testing three different things, could they be split out into different tests?

dschwoerer:
I don't really want to, as I check with min/max whether the staggered grid works
I don't like throwing errors, so in order to exit smoothly, I just solve for the fields with zero time derivative.
That's why the min/max test, also fails if something other is broken.

dschwoerer:
Should I split this in 3 tests?

I don't really see the point, but if that is required, I could have a go (and basically triple all the code)

bendudson:
I think as long as the test prints a useful error message which says which
part failed then it's ok

PR #272, opened 2016-09-29
include -fno-inline in debug mode

dschwoerer:


bendudson:
This seems useful, please update to v4.0.0-rc

bendudson:
If used with compilers other than gcc, which presumably may not understand -fno-inline, will this just result in a warning or the compilation failing?

d7919:
The flag is also recognised by the intel compilers (at least v14) but causes an error with the cray compiler. I'm not sure how configure handles this though.

bendudson:
Tricky... supporting Cray compiler would be good: Archer, NERSC machines widely used for BOUT++ runs. Is it possible to try compiling a small snippet with these debugging flags, falling back on just "-g" or similar if the compilation fails?

dschwoerer:
The flag is added to a list, which is then checked with the compiler, and in that case removed.
If you happen to know, what flags other compiler use, I can add them ...

bendudson:
Awesome! Thanks David. Hadn't realised it was that clever. 

How do I merge manually then update this PR without pushing directly to v4.0.0-rc? I'd prefer it to go through Travis before going into v4.0.0-rc

d7919:
For the cray compiler I believe `-hipa1` should reasonably well match `-fno-inline`, it disables all inlining except from that which is due to a directive. Replacing 1 with 0 disables all inlining.

dschwoerer:
I updated the PR, added the `-hipa1` flag and merged into the RC.

dschwoerer:
I needed `aclocal.m4` (generated by `aclocal`) for `autoconf` to be successful - is that only me, or should I add `aclocal.m4`?

ZedThree:
Try running `autoreconf -fvi` instead. There shouldn't be any need for
`aclocal.m4` to be in the repo.

It's possible you might have to install an autotools package if it can't
find some macros.

dschwoerer:
```
$ autoreconf -fvi
autoreconf: Entering directory `.'
autoreconf: configure.ac: not using Gettext
autoreconf: running: aclocal --force 
autoreconf: configure.ac: tracing
autoreconf: configure.ac: not using Libtool
autoreconf: running: /usr/bin/autoconf --force
autoreconf: configure.ac: not using Autoheader
autoreconf: configure.ac: not using Automake
autoreconf: Leaving directory `.'
```
That calls aclocal and generates `aclocal.m4` as well ...
Do you mind if I add `aclocal.m4` than to `.gitignore`?

ZedThree:
Good idea, yes please!

PR #290, opened 2016-10-10
Adding two new boundary modifiers

d7919:
toFieldAligned : Convert field fromFieldAligned and then apply boundary
fromFieldAligned : Convert field toFieldAligned and then apply boundary

Naming may not be the best choice?

### DIFF
--- a/include/boundary_standard.hxx
+++ b/include/boundary_standard.hxx
@@ -341,7 +341,9 @@ public:
   BoundaryOp* cloneMod(BoundaryOp *op, const list<string> &args);
   
   void apply(Field2D &f);
+  void apply(Field2D &f, BoutReal t);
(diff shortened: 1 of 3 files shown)


include/boundary_standard.hxx
```diff
@@ -333,4 +333,38 @@ private:
   int width;
 };
 
+/// Convert toFieldAligned
```
ZedThree: Could you expand on the documentation here a little bit? These modifiers are designed to take a *function* specified in non-field-aligned/field-aligned space and apply it to a *field* specified in the other space.

d7919: I've just added a bit more description, hopefully enough to make it clear what is happening.

Issue #286, opened 2016-10-08
vsprintf should be replaced by vsnprintf


dschwoerer:
The inputs may be longer than expected ... nasty stuff could happen (however probably not remotly exploitable)

ZedThree:
This could possibly just be replaced with a C++ function.

dschwoerer:
I don't think this is possible, without changing the interface.

But setting the current interface as deprecated, an offering a C++ interface would probably be nice ...

bendudson:
The "output" class already has a C++ interface, though probably not as nice
as it could be:

    output << "Hello World" << endl;

is an alternative to

    output.write("Hello World\n");

For now just replacing vsprintf with vsnprintf seems like the simplest
solution

ZedThree:
Fixed in #291

dschwoerer:
Please reopen.

The output should of `vsnprintf` should be checked, making sure the write was as expected - and ether allocating a large enough buffer, or throwing an exception.

bendudson:
I think this was fixed in pull request #319

PR #291, opened 2016-10-10
Replace vsprintf with vsnprintf

bendudson:
Safe version, which guards against buffer overruns
by specifying the length of the buffer into which text
is being printed.

This was used in both Output and BoutException classes
to provide C printf-style output.

This fixes issue #286

### DIFF
--- a/src/sys/boutexception.cxx
+++ b/src/sys/boutexception.cxx
@@ -13,12 +13,10 @@ void BoutParallelThrowRhsFail(int &status, const char* message) {
   if (allstatus) throw BoutRhsFail(message);
 }
 
-BoutException::~BoutException() throw()
-{
+BoutException::~BoutException() throw() {
 }
 
-BoutException::BoutException(const char* s, ...)
-{
+BoutException::BoutException(const char* s, ...) {
   va_list ap;  // List of arguments
 
   if(s == (const char*) NULL)
@@ -32,8 +30,7 @@ BoutException::BoutException(const char* s, ...)
   message.assign(buffer);
 }
 
-const char* BoutException::what() const throw()
-{
+const char* BoutException::what() const throw() {
   #ifdef CHECK
     /// Print out the message stack to help debugging
     msg_stack.dump();
@@ -43,7 +40,7 @@ const char* BoutException::what() const throw()
   return message.c_str();
 }
 
-BoutRhsFail::BoutRhsFail(const char* s, ...)  : BoutException::BoutException(s){
+BoutRhsFail::BoutRhsFail(const char* s, ...)  : BoutException::BoutException(s) {
   va_list ap;  // List of arguments
 
   if(s == (const char*) NULL)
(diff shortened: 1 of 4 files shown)


src/sys/boutexception.cxx
```diff
@@ -13,27 +13,24 @@ void BoutParallelThrowRhsFail(int &status, const char* message) {
...
+BoutException::BoutException(const char* s, ...) {
   va_list ap;  // List of arguments
 
   if(s == (const char*) NULL)
     return;
   
   char buffer[1024];
   va_start(ap, s);
-    vsprintf(buffer, s, ap);
+    vsnprintf(buffer, 1024, s, ap);
```
ZedThree: Can this be a constant too? `BoutException::BUFFER_LEN`?

dschwoerer:
The return value of `vsnprintf` should be checked, and if smaller than the buffer size, a sufficiently large buffer should be allocated - or an exception should be thrown (easier)

bendudson:
Throwing an exception inside BoutException could end badly. In output I don't think throwing an exception is the best way. These outputs will be mainly used for printing information on a run as it goes, so terminating just because an output couldn't be printed is probably not what the user would want. Possibly allocating more memory, but 1024 characters should be enough for anyone... 

One question is how fixes like this should be applied to next. Should there be a second branch from next, which would presumably then conflict in any merge, or wait and merge master into next later?

dschwoerer:
In output, this would only happen if the path of the file, you are trying to open would be longer than whatever.
In that case, ignoring is terrible, as it might overwrite anything. Maybe something important ...

In exception it shouldn't be to much work to reallocate memory ...

ZedThree:
With a bit of luck, next will be ready to be merged into master soon-ish. Then, hopefully the development branch and master won't be too far apart, so hotfixes like this can be applied to both at the same time, though that would have to be done from the command line (or with a second PR, I guess?)

ZedThree:
Good point about output filenames.

Linux has a maximum path length of 4096 characters. Although I'm not sure we want to be allocating that much all the time.

bendudson:
ok, this is used in three places:

1) In Output as a printf() convenience. This doesn't open any files, and only writes text to stdout and the BOUT.log files. Here allocating memory would be nice, but not essential. Throwing an exception is not desirable.

2) In BoutException to format error messages. I think allocating or reallocating memory here is not advisable, because the exception may have been thrown due to memory corruption and so allocs/frees may cause further exceptions. I think the only option here is to ignore anything not printed.

3) In DataFile, where it is used to create file names. This still uses vsprintf, not vsnprintf. I agree there is a potential issue with this, and either it needs to allocate more memory or throw an exception.

The correct fix in "next" would be to use variadic templates for the functions, rather than C vararg.

dschwoerer:
About 1):
```
  int open(const char *fname, ...); ///< Open an output log file                                                                             
```
seems to me like it would be opening a file, not writing to it.
On the other hand, I am not sure this is often used ...

about `Output::write` I think an error here could be ignored ... ( I missed earlier that vs(n)printf is used here as well)

about 2):
We are anyway allocating memory for the buffer, and in general a whole exception is constructed.
I am not sure an extra delete/new call would make a difference.

about 3)
This was the main reason I opened this issue :)

PR #292, opened 2016-10-10
fixing the issue with vsprintf

dschwoerer:
I changed it, so that if the buffer is to small, a larger one will be allocated.
I wrote a macro in the utils.hxx (I hope this is the right place) to deal with checking whether the call was successful, and if not reallocating.

I think this is the most clean way to deal with it.

I am in favour of `printf()` statements over the more C++ like `<<`, as they allow for short and well defined expression, concerning numbers.

Fixes issue #291

### DIFF
--- a/src/field/field.cxx
+++ b/src/field/field.cxx
@@ -46,15 +46,12 @@ Field::Field() {
 void Field::error(const char *s, ...) const {
   int buf_len=512;
   char * err_buffer=new char[buf_len];
-  va_list ap;  // List of arguments
 
   if(s == (const char*) NULL) {
     output.write("Unspecified error in field\n");
   }else {
   
-    va_start(ap, s);
     myvsnprintf(err_buffer,buf_len, s, ap);
-    va_end(ap);
 
 #ifdef TRACK
       output.write("Error in '%s': %s", name.c_str(), err_buffer);
--- a/src/field/field.cxx
+++ b/src/field/field.cxx
@@ -31,6 +31,7 @@
 #include <output.hxx>
 #include <msg_stack.hxx>
 #include <boutexception.hxx>
+#include <utils.hxx>
 
 Field::Field() {
 #ifdef CHECK
@@ -40,10 +41,11 @@ Field::Field() {
 
 /////////////////// PROTECTED ////////////////////
 
-char err_buffer[512];
 
 // Report an error occurring
 void Field::error(const char *s, ...) const {
+  int buf_len=512;
+  char * err_buffer=new char[buf_len];
   va_list ap;  // List of arguments
 
   if(s == (const char*) NULL) {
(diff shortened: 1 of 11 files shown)


bendudson:
Thanks David! This seems like a nice fix if it works. Some suggestions:

1. Please document the myvsnprintf macro (DOxygen comments preferred), since it modifies its inputs and makes assumptions about how the buffer was allocated (that it can be deleted)

2. Perhaps put a limit on the size of the buffer (e.g. 4096), and throw an exception or ignore if too long

3. Rename BUFFER_LEN to buffer_len since not constant anymore

src/fileio/datafile.cxx
```diff
@@ -41,13 +41,13 @@
 #include <boutexception.hxx>
 #include <output.hxx>
 #include <boutcomm.hxx>
-
+#include <utils.hxx>
 #include "formatfactory.hxx"
 
 Datafile::Datafile(Options *opt) : parallel(false), flush(true), guards(true), floats(false), openclose(true), enabled(true), file(NULL) {
+  filename=new char[filenamelen];
```
bendudson: Is filenamelen set to a value before this line runs?

src/field/field.cxx
```diff
@@ -62,5 +64,6 @@ void Field::error(const char *s, ...) const {
   }
   
   throw BoutException("Error in field: %s", err_buffer);
+  delete[] err_buffer;
```
dschwoerer: Is there a way to free this buffer?

src/sys/boutexception.cxx
```diff
@@ -41,29 +45,7 @@ const char* BoutException::what() const throw() {
...
-  char buffer[BoutException::BUFFER_LEN];
-  va_start(ap, s);
-    vsnprintf(buffer, BoutException::BUFFER_LEN, s, ap);
-  va_end(ap);
-  
-  message.assign(buffer);
 }
 
 BoutIterationFail::BoutIterationFail(const char* s, ...) : BoutException::BoutException(s) {
-  va_list ap;  // List of arguments
```
dschwoerer: BoutIterationFail doesn't do anything ...
Well, currentl it does what the BoutException did once more - I don't think that is beneficial, thats why I removed the whole code.

Furthermore I couldn't find any documentation on va_list having any effect - by now I hide all the va_list/ap handling in the myvsnprint() macro ...

dschwoerer:
Ok, I added the documentation.

I am not sure that limiting the buffer size is required.
Even 4000000 is only 4 MB - no computer used for bout should have any problems with this amount of memory. Further more, 4 MB of actual text is basically impossible to fill ...
I couldn't think of any way to generate that much text into a single buffer - at least not without causing other, more severe problems.
Nevertheless, such a limit could be implemented in the macro, that would just be an additional line.

`buffer_len` is now the current length, whereas `BUFFER_LEN` is the default length.

This time I run the test_suite, so it should work :)

bendudson:
Replaced by pull request #319

PR #306, opened 2016-10-12
Ensure a new line after exception

d7919:
Most exception messages don't include a newline -- this can make the backtrace etc. a little messy so forcing a new line after the message.


d7919:
Downside of this PR is that it adds an extra new line to those messages that already end with "\n". 

Alternative is to adjust a large number of exception messages in lots of files and ensure that any new messages end with "\n".

PR #316, opened 2016-10-17
Fix ddz of vector

### LINKED ISSUE #123: Vector3D DDZ not calculated correctly
`Vector3D DDZ` is declared in [derivs.hxx](../blob/master/include/derivs.hxx) (currently) line 57-59, however, it's implementation in [derivs.cxx](../blob/master/src/sys/derivs.cxx) (currently) line 1970-1996 only takes into account the derivatives of the component, and not the derivatives of the basis vectors (see D'Haeseeler eqation (2.6.17) and (2.6.32) for details).

I suggest to replace this with the correct definitions. The correct definitions are handy when defining terms like `brackets(const Field3d &f, const Vector3D &a, ...)`.

loeiten:
This closes #123.

Should be used with caution, as it has not been MMSed.


bendudson:
Can the base please be changed from master to v4.0.0-rc? Thanks Michael

loeiten:
Oh, I intended to do the merge, but it got a bit late last night. Sorry about that.
And thanks for doing it :)

PR #321, opened 2016-10-18
User requested exit

d7919:
Allows a user to request that the run exits at the end of the next monitor call by creating a specific file. 

This can be quite useful alternative to just killing the job as it ensures things exit cleanly with no risk (or at least less risk) of causing problems in the output files etc.


ZedThree:
Looks good to me. The only thing that I'm not sure about is whether it's better to have this feature off by default or on by default. This does seem like something I would want all the time, and it might be annoying if I'd forgotten to turn it on. In fact, I might even go so far as to say that actually you probably don't *ever* want to turn it off.

Anyone with strong opinions either way?

d7919:
You might not want it on if you're running on a slow file system with a high monitor frequency, but I'd be happy if the default was to be on.

dschwoerer:
I am not sure why a file is chosen instead of doing an MPI communication.

I guess there are no guarantees that all processes will try to read the file at the same time, nor that all filesystems will be synced at the same time.

I am wondering whether a global variable + a MPI call after calling all monitors is not the more reliable way.

d7919:
The reason for checking for a file is that the user can generate this during the run externally from the code (i.e. I start running a simulation and then decide that I've it's going to take too long to do all the requested steps so I want to stop the run. I can either kill the job -- which risks "corrupting" the output, or I have some external way of signalling to the code that I want it to finish).

I agree that it's probably safer to have one process do the read and then broadcast the result.
This is presumably true for the abort triggered by the `wall_time` check as well (currently if the processes are not in sync then each process will end up checking how much time is left at a different point in time -- some may decide they've run out whilst others don't).

I would propose that this can be fixed (maybe in a separate PR) through simply adding a status variable to record the result of all checks, remove "early" returns, broadcast the status and then return it. This could however be costly for large scale jobs. Alternatively this logic (i.e. the broadcast/reduction) could be moved to a higher level so that if you write a custom monitor then you don't need to worry about it.

dschwoerer:
It shoudn't be hard to trap e.g. signal USR1 and let that set the flag.

I don't think that because other code may be broken, we should introduce new broken code.

Again, it could be disabled, if that extra communication is harmful. Also we could try an asynchronous communication - overlapping with the dump call would be straight forward and I guess in most cases should hide the latency.

d7919:
I don't have much experience with sending and trapping signals etc., would this work when running on a HPC system (for example) where you don't have access to the nodes on which the code is running? From my quick search it looks like the ability to forward signals is mpi implementation dependent and relies on you having the ability to send the signal to the running `mpirun` process I think.

d7919:
To fix the general issue we would probably want code similar to `BoutParallelThrowRhsFail` in `solver::call_monitors` instead of directly throwing a `BoutException`.

dschwoerer:
I guess most HPC system should support a mechanism to signal the processes, with PBS, which is used on archer, the command is qsig.

Signal trapping is already done in BOUT++, if -DSIGHANDLE is given, SIGSEGV/SIGFPE is caught.
So adding SIGUSR1 would be easy. (i.e. I would be willing to implement it)

I think it should be affordable to have an extra communication in the output monitor. Then you would have to signal only one thread (and avoid race conditions)

Note that BoutParallelThrowRhsFail always does communications, so we shouldn't use it to often ...

d7919:
Excellent, I wasn't aware of qsig and similar, thanks. I'm happy to look at implementing this as well if you have other priorities, but probably won't get chance until Friday.

d7919:
Just a brief reminder for later; for slurm a signal can be sent with `scancel -s <SIG>`.

dschwoerer:
I have started, so the signal handler is done (it's just one line).

What I am think it would be probably the most save way to modify nout in the solver (or probably easier in the call_monitors function.)

Otherwise, if I have a monitor that writes data, but does so without open/closing the file every time, it would fail to do so.

Further, if looking towards several output monitor frequencies, when should we actually exit?
The next time the iteration number is a multiple of the least common multiple of all monitors frequency?

src/bout++.cxx
```diff
@@ -430,7 +430,14 @@ int bout_monitor(Solver *solver, BoutReal t, int iter, int NOUT) {
   // Data used for timing
   static bool first_time = true;
   static BoutReal wall_limit, mpi_start_time; // Keep track of remaining wall time
-
+  
+  static bool stopCheck;       // Check for file, exit if exists?
+  static string stopCheckName; // File checked, whose existence triggers a stop
```
ZedThree: I forgot our policy on stdlib stuff (if we have one...) -- are we preferring new variables use `std::`?

src/bout++.cxx
```diff
@@ -430,7 +430,14 @@ int bout_monitor(Solver *solver, BoutReal t, int iter, int NOUT) {
...
+  
+#ifdef CHECK
+  int msg_point = msg_stack.push("bout_monitor(%e, %d, %d)", t, iter, NOUT);
```
ZedThree: Can be replaced with `TRACE("bout_monitor(%e, %d, %d)", t, iter, NOUT);`

src/bout++.cxx
```diff
@@ -458,31 +465,42 @@ int bout_monitor(Solver *solver, BoutReal t, int iter, int NOUT) {
...
     OPTION(options, wall_limit, -1.0); // Wall time limit. By default, no limit
-    wall_limit *= 60.0*60.0;  // Convert from hours to seconds
+    wall_limit *= 60.0 * 60.0;         // Convert from hours to seconds
+
+    OPTION(options, stopCheck, false);
+    if (stopCheck) {
+      // Get name of file whose existence triggers a stop
+      OPTION(options, stopCheckName, "BOUT.stop");
+      // Now add data directory to start of name to ensure we look in a run specific location
+      string data_dir;
```
ZedThree: `std::string` here as well, possibly?

src/bout++.cxx
```diff
@@ -512,6 +530,23 @@ int bout_monitor(Solver *solver, BoutReal t, int iter, int NOUT) {
...
 
+  // Check if the user has created the stop file and if so trigger an exit
+  if (stopCheck) {
+    std::ifstream f(stopCheckName);
+    if (f.good()) {
+      output << "\n" << "File " << stopCheckName
+             << " exists -- triggering exit." << endl;
+#ifdef CHECK
+      msg_stack.pop(msg_point);
+#endif
```
ZedThree: These three lines can be removed if `TRACE` is used above

src/bout++.cxx
```diff
@@ -512,6 +530,23 @@ int bout_monitor(Solver *solver, BoutReal t, int iter, int NOUT) {
...
+      return 1;
+    }
+  }
+
+#ifdef CHECK
+  msg_stack.pop(msg_point);
+#endif
```
ZedThree: These three lines can be removed if `TRACE` is used above

src/bout++.cxx
d7919: Actually the TRACE macro has already been added for this function so I'm just going to remove this `msg_stack` line.

src/bout++.cxx
d7919: Not sure this is consistently handled in this file anyway, but I've updated the two strings that are related to this PR

PR #334, opened 2016-10-20
Optimise next shift z

d7919:
Use a cached complex phase for the shiftedmetric calculations. Can lead to significant performance improvements in testing (~2x speed up for problems with a lot of shifting)


d7919:
Note we might want to look into doing a similar thing for the twistShift parallel boundary that's currently done inside `mesh::wait` (part of `communicate`).

Issue #322, opened 2016-10-18
Reducing compiler warnings


d7919:
As part of recent efforts to test the next branch I've started compiling with as many compiler warnings enabled as possible (`-Wall -Wextra -Wpedantic`). This generates `14365` lines of extra output (up to 5 lines per warning). Note this is with 

```
> g++ --version
g++ (SUSE Linux) 4.8.3 20140627 [gcc-4_8-branch revision 212064]
```

Extracting the unique warnings and counting (done using `make 2>ERRS ; grep "\[\-W" ERRS | awk -F"[" '{print $2}' | awk -F"]" '{print $1}' | sort | uniq -c | sort -nr | awk '{print "|\t" $1 "\t|" $2 "\t|"}'`) I find:

| Count | Warning |
| --- | --- |
| 3667 | unused parameter |
| 510 | reorder |
| 359 | pedantic |
| 205 | ignored qualifiers |
| 58 | missing field initializers |
| 29 | return type |
| 22 | unknown pragmas |
| 19 | unused variable |
| 12 | sign compare |
| 10 | unused but set variable |
| 6 | vla |
| 6 | deprecated declarations |
| 5 | unused value |
| 5 | switch |
| 3 | init self |
| 3 | format= |
| 1 | delete non virtual dtor |

In other words we're dominated by a few types of warning. Further I can note that some of these warnings are repeated in multiple places for the same line of source code (e.g. the `reorder` warning appears wherever we construct certain object instances so these 510 warnings can probably be fixed with just a few lines changed).

I would be keen to try to eliminate as many of these warnings as possible. One reason for this is that some of these warnings indicate code that can lead to undefined behaviour. I suspect I won't be able to fix everything simply but I may try to pick some of the low hanging fruit for now. I propose using this issue to keep track of efforts to remove/improve these warnings.

I expect a lot of the fixes will involve relatively minor changes (for example e185e38d148d6a440e00c12e0b04e49645273038 fixes the "delete non virtual dtor" warning which could potentially have led to undefined behaviour) but may touch lots of files.

d7919:
The `format=` warning results form three lines in `domain.cxx` of the form:

~~~
fprintf(stderr, "%x: (%d, %d) %d [%d, %d] [%d, %d]",
                this, nx, ny, s, bndry->getMin(s), bndry->getMax(s), b->getMin(s), b->getMax(s));
~~~

[...]

PR #337, opened 2016-10-24
Add macro to silence unused-parameter warning

ZedThree:
This is a little bit horrendous as it touches so many places.

Adds the macro defined in [this SO question](http://stackoverflow.com/questions/7090998/portable-unused-parameter-macro-used-on-function-signature-for-c-and-c) to a new file, `unused.hxx`. This wraps unused parameters in function signatures to indicate to the compiler that we know that the parameter is not used. Importantly, it also renames the parameter (prepending `unused_`) so that if it is accidentally used, we'll know about it immediately.

This commit reduces the number of `unused-parameter` warnings from ~3500 to 52 (see #322). The remaining warnings are either suspicious (i.e. possibly a bug!), should be dealt with differently (e.g. function should be appropriately overloaded or even removed entirely), or I just don't know whether it is correct or not (e.g. all the solver callback functions).

An alternative to this approach would be to `/* comment */` out parameter names in function signatures. This will also silence errors, but maybe doesn't make it quite so obvious as to *why* we've done so.

### DIFF
--- a/include/boundary_op.hxx
+++ b/include/boundary_op.hxx
@@ -10,6 +10,7 @@ class BoundaryModifier;
 #include "field3d.hxx"
 #include "vector2d.hxx"
 #include "vector3d.hxx"
+#include "unused.hxx"
 
 #include <cmath>
 #include <string>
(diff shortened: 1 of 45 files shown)


bendudson:
I think this is the best solution to this problem, since it makes the intentions of the code author clear, and catches cases where inconsistencies could arise. Thanks Peter!

PR #341, opened 2016-10-27
More warning fixes

ZedThree:
Fix lots more warnings

### DIFF
--- a/include/cyclic_reduction.hxx
+++ b/include/cyclic_reduction.hxx
@@ -131,7 +131,7 @@ public:
     int ns = Nsys / nprocs; // Number of systems to assign to all processors
     int nsextra = Nsys % nprocs;  // Number of processors with 1 extra 
     
-    MPI_Request req[nprocs];
+    MPI_Request* req = new MPI_Request[nprocs];
 
     if(myns > 0) {
       // Post receives from all other processors
@@ -343,6 +343,7 @@ public:
     ///////////////////////////////////////
     // Solve local equations
     back_solve(Nsys, N, coefs, x1, xn, x);
+    delete req;
   }
   
 private:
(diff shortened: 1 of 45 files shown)


include/cyclic_reduction.hxx
```diff
@@ -343,6 +343,7 @@ public:
     ///////////////////////////////////////
     // Solve local equations
     back_solve(Nsys, N, coefs, x1, xn, x);
+    delete req;
```
dschwoerer: This causes undefined behaviour! Use delete[]

src/field/globalfield.cxx
```diff
@@ -164,6 +164,7 @@ void GlobalField2D::gather(const Field2D &f) {
         }
       }while(pe != MPI_UNDEFINED);
     }
+    delete req;
```
dschwoerer: use `delete[]`

src/field/globalfield.cxx
```diff
@@ -329,6 +330,7 @@ void GlobalField3D::gather(const Field3D &f) {
...
```
dschwoerer: delete[]

src/mesh/data/gridfromoptions.cxx
```diff
@@ -91,45 +91,45 @@ bool GridFromOptions::get(Mesh *m, vector<int> &var, const string &name, int len
...
+
   // Parse, and evaluate with x,y,z,t = 0
   FieldGenerator* gen = FieldFactory::get()->parse(expr, options);
 
   var.resize(len);
-  
+
   switch(dir) {
   case GridDataSource::X: {
-    for(int x=0;x<var.size();x++){
```
dschwoerer: I assume the compiler warnung was comparison of signed and unsigned int?
Wouldn't it make more sense to use an unsigned int for `x` and `len`?

include/cyclic_reduction.hxx
ZedThree: Good catch, thanks!

PR #343, opened 2016-10-27
signal based user requested exit

dschwoerer:
Alternative to #321 here implemented using signals.

One change is that signal.h is always needed ...


d7919:
Looks good, just a small question. Currently if the monitors signal an abort we throw an exception, which is immediately caught allowing us to then write a restart, before returning 1 to allow the rest of the code to finish cleanly. In other words I think a monitor abort behaves more like a clean shutdown than an error event, so it may be more consistent to treat a user requested abort in the same way?

This would then presumably also allow us to treat the abort status uniformly across the user requested and the monitors, i.e. we can accumulate the return status and user requested status into a single variable and then communicate this in order to ensure that if any monitor on any processor has signalled an abort we do the abort?

If we end up caring about the cost of the `mpi_allgather` we could look at something like [this](http://stackoverflow.com/a/4961000)

dschwoerer:
I don't think that would be that easy.

I think we should first check the status, so the monitors can now it is there last call. If any of the monitors singals to exit, we would have to call all again before that one and tell them that it is there last call, so they can shut down cleanly. As they would be called twice for the same time, this could make things even worse.
So the save solution would be, continue for another timestep. But then it would be easier, if the monitor would just set the user_reqeuest_exit variable, and we don't need any extra code to handle this.

d7919:
As the gather is a blocking call presumably putting this after all the monitor calls would ensure that we have called them all the monitors on every processor the same number of times (assuming we move the throw+catch after this comm). Something like

~~~~
bool status=user_requested_exit;
for each monitor:
    status = status or monitor->call();

allgather(status);
if(status){
    save restart
    tidy up monitors
}
return status;
~~~~

Where `tidy up monitors` can be whatever needs to be done at the last timestep, such as calling the monitors again with the information this is the last step if needed. Given that #346 is talking about a rewrite of the monitor code I'd suggest it might be nice to add a `monitor` class with members like `init`, `call` and `finish`, where `finish` should do any finalisation of the monitor (e.g. cleanly closing files) and then in the future this could be done as part of the clean bout shutdown rather than here.

This is just a quick idea, I've probably missed something that would prevent this from working.

dschwoerer:
Now with script, to stop bout.
It requires the path, and sends a signal to the process.
This is not suitable for HPC systems, as every one has its own signalling system, and it is quite easy for the user to use them instead.

@d7919 probably a better idea. I cannot see anything that would prevent this from working, probably worth having a go if this and #364 are merged.

d7919:
As this is an existing PR should it the target be v4.1-rc rather than next so that it actually gets into the new release rather than having to wait?

src/bout++.cxx
```diff
@@ -75,18 +75,18 @@ using std::list;
...
 #ifdef BOUT_FPE
 #include <fenv.h>
 #endif
 
+
 #include <output.hxx>
 
 BoutReal simtime;
 int iteration;
+int user_requested_exit=0;
```
ZedThree: Elsewhere this is a bool -- is there a reason it's an int here?

src/bout++.cxx
```diff
@@ -221,7 +222,16 @@ int BoutInitialise(int &argc, char **&argv) {
...
+  {
+    const size_t len=512;
+    char buf[len];
+    snprintf(buf,len,"%s/.BOUT.pid.%d",data_dir, MYPE);
+    FILE * fp = fopen(buf, "w");
+    if (fp == nullptr)
+      return 1;
+    fprintf(fp,"%d\n",getpid());
+    fclose(fp);
+  }
```
ZedThree: Could this be rewritten as idiomatic C++ rather than C-style?
Something like

```cpp
std::string filename;
std::stringstream(filename) >> data_dir >> "/.BOUT.pid." >> MYPE;
std::ofstream pid_file;
pid_file.open(filename, std::ios::out);
if (pid_file.is_open()) {
    pid_file << getpid() << "\n";
    pid_file.close()
}
```

dschwoerer:
Now with documentation and test 👍

PR #351, opened 2016-10-28
More warning fixes

dschwoerer:
fixes the unknown pragma warning and also signifcantly reduces the number of warnings, if CHECK is disabled ...

### DIFF
--- a/include/field2d.hxx
+++ b/include/field2d.hxx
@@ -27,8 +27,6 @@
 class Field2D;
 
 #pragma once
-#ifndef __FIELD2D_H__
-#define __FIELD2D_H__
 
 class Mesh;
 #include "field.hxx"
@@ -36,6 +34,7 @@ class Mesh;
 class Field3D; //#include "field3d.hxx"
 #include "fieldperp.hxx"
 #include "stencils.hxx"
+#include "unused.hxx"
 
 #include "bout/dataiterator.hxx"
 
@@ -275,7 +274,7 @@ Field2D pow(BoutReal lhs, const Field2D &rhs);
 #ifdef CHECK
 void checkData(const Field2D &f);
 #else
-inline void checkData(const Field2D &f) {}
+inline void checkData(const Field2D &UNUSED(f)) {}
 #endif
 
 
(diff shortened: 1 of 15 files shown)


ZedThree:
The second commit is good, but I'm less sure about the first one. I'm not terribly keen on replacing `#pragma`s with macros. The point about pragmas is that the compiler is supposed to ignore them if it doesn't understand. I think I'd much rather use `-Wno-unknown-pragmas` to disable the warning, even if this means not catching typos in pragmas. The warning could just be enabled when adding new pragmas.

Also, I'm not sure about getting rid of the header guards. I don't think there's any harm in keeping them and there's very little benefit in getting rid of them.

dschwoerer:
About replacing pragmas with macros - I don't think it is extremely bad, but of course it can cause confusion. Therefore if you think that it is not worth to remove the warnings, I am fine with that. I think this solution is still better than having `#ifdef _OPENMP` always in front ...

About using `#pragma once` instead of the `#ifdef` guards: I don't see any advantage in keeping them, but agree that the gain is small, but it results in a cleaner code. I wouldn't remove them in all files in one go, but if I am anyway changing a file, I think it is worth removing them ...

d7919:
One thing to note is that (I think) the pgi compiler doesn't currently support `#pragma once` so keeping the header guards will make it easy for someone to still compile if they really want to use pgi.

dschwoerer:
Thanks, I wasn't aware PGI doesn't support this.
I will use `#ifdef` guards in the future ...
https://en.wikipedia.org/wiki/Pragma_once#Portability

ZedThree:
Yeah, it's annoying. To be honest, I think `pragma once` looks cleaner and
has some advantages, but I think the most portable solution is having both.

d7919:
Yes it's a pain, hopefully they'll change this in the future!

bendudson:
I think this PR has been split, with guards being kept, but warnings about unused arguments being suppressed using UNUSED. Is this PR still needed?

bendudson:
I think this PR is now obsolete

PR #358, opened 2016-11-01
Various fixes and improvements to configure

ZedThree:
- Better ways to check for NetCDF and HDF5
- Better check for MPI compiler, now fails if not found
- Check compiler supports C++11; fail if not
- Compile PVODE using same flags as BOUT++
- Move various macros out into separate files


ZedThree:
Oh god NetCDF + Travis is painful. In order to use python 3 properly, we need miniconda. This brings in its own version of NetCDF, which doesn't have the C++ API, and is linked against a bunch of other libraries (including HDF**4**!). That version also has a bug which results in `nc-config --libs` being empty, making it difficult to find out what we need to link against to get NetCDF to work.

So we end up needing to make sure we still use the system NetCDF, and either compile without HDF5 or, again, use the system version. Unfortunately, on precise this uses a deprecated API, so we have to use trusty, which also means changing mpich to openmpi, so as to avoid installing Fortran.

loeiten:
Sound like some painful hours. If it is of any comfort, you can have my thank-you points :).

Would it not be possible to have two `NetCDF` versions though? One built through `miniconda` used in the `collect` routines, and another one "built by hand"?

ZedThree:
Thanks!

Yep, that's what happens. The pain is making sure BOUT++ sees and *uses* the "right" NetCDF and not the miniconda one. These changes let you specify the *exact* version you want like this:

     ./configure --with-netcdf=/path/to/nc-config # or --with-netcdf=/path/to/ncxx4-config

This ensures that it uses the "right" version... assuming that that what `nc-config--libs` reports is correct, which with the miniconda one, it isn't!

PR #363, opened 2016-11-03
Non-orthogonal mesh

ZedThree:
Created on behalf of @JarrodLeddy 

I've made the PR against master to avoid bringing master's changes into next before the Great Merge. It will need fixing up after, but not very much.

It looks like this is currently failing in `drift-instability`.


ZedThree:
Merged in release candidate

PR #319, opened 2016-10-18
Dschwoerer master vsnprintf fix

bendudson:
Merged changes to datafile.cxx

Replaces pull request #292

PR #365, opened 2016-11-03
Next fix vsnprintf

dschwoerer:
Inlike the PR  #319 (the PR for the master branch) this should also work for the derived exceptions.


src/fileio/datafile.cxx
```diff
@@ -41,13 +41,14 @@
 #include <boutexception.hxx>
 #include <output.hxx>
 #include <boutcomm.hxx>
-
+#include <utils.hxx>
 #include "formatfactory.hxx"
 
-Datafile::Datafile(Options *opt) : parallel(false), flush(true), guards(true), floats(false), openclose(true), enabled(true), shiftOutput(false), file(NULL) {
+Datafile::Datafile(Options *opt) : parallel(false), flush(true), guards(true), floats(false), openclose(true), enabled(true), file(NULL) {
```
ZedThree: This removes `shiftOutput(false)` from the initializer list. I assume this is accidental?

include/utils.hxx
```diff
@@ -103,4 +103,29 @@ string trim(const string &, const string &c=" \t\r");
...
 string trimRight(const string &, const string &c=" \t\r");
 string trimComments(const string &, const string &c="#;");
+
+/// the myvsnprintf macro:
+/// The first argument is an char * buffer of length len.
+/// It needs to have been allocated with new[], as it may be
+/// reallocated.
+/// len: the length of said buffer. May be changed, mussn't be const.
+/// fmt: the const char * descriping the format.
+/// va: A dummy argument. will
```
ZedThree: Is the dummy argument needed? It looks confusing when the macro is called.

src/field/field.cxx
```diff
@@ -62,5 +61,6 @@ void Field::error(const char *s, ...) const {
   }
   
   throw BoutException("Error in field: %s", err_buffer);
+  delete[] err_buffer;
```
ZedThree: This will never be hit, and presumably leads to a memory leak if the exception isn't caught?
I'm not really sure a) if this is a problem and b) if so, how to fix it.

ZedThree:
Maybe in future it would be possible to replace all this C stuff with nice C++ idioms? [This library][1] looks like a really nice replacement for the `printf` family. It could possibly reduce a lot of the complexity about formatting messages.

[1]: http://fmtlib.net/latest/index.html

src/field/field.cxx
dschwoerer: This leads to a memory leak, no matter if caught or not. (not caught however doesn't really matter)
b) use std::string

dschwoerer:
Does "this library" (fmt?) support traditional formatting like `%-10.3g`?
This is the main reason why I don't like the C++ way - it is to bloated for number printing ...

ZedThree:
From what I can tell, it supports not only traditional `printf` style printing like that, but also python style:

     fmt::format("The answer is {}", 42);

dschwoerer:
Only the python format style - python also supports c-printf like statements ...( as the one above, left centered, at least 10 chars, and 3 digits after the point)

ZedThree:
Yep, look at `fmt::printf`, `fmt::fprintf` and `fmt::sprintf`:

     fmt::printf("The answer is %d\n", 42);

Also, see the [SO answer here][1]. Looks like we can do all this in pure C++, but still use `printf` format strings. Everyone's happy! :)

I'll have a bash at it tomorrow.

[1]: http://stackoverflow.com/a/26221725/2043465

src/fileio/datafile.cxx
dschwoerer: yes, thanks

ZedThree:
Is this still needed for `v4.0.0-RC`? It looks quite similar to what is in there now - I think there was also a PR against master?

Also, is it possible that this functionality could be done using C++11 features instead of macros, e.g. `std::snprintf`? See http://stackoverflow.com/a/26221725/2043465

dschwoerer:
Using variadic templates requires that all this is defined in header files ...

I cannot see that this is already fixed in v4 RC:
https://github.com/boutproject/BOUT-dev/blob/v4.0.0-RC/src/sys/boutexception.cxx#L80

bendudson:
This would be better than what's there now, but using type safe templates
might be better

bendudson:
To avoid best becoming the enemy of better, I vote we merge this for now, and then look at changing to a C++ solution later. 
Can you (David) please change the base to v4.0.0-rc?

dschwoerer:
base changed + conflicts resolved:
should be ready to merge ...

PR #366, opened 2016-11-08
Co- and contravariant metric tensors simultaneously read from grid file.

### LINKED ISSUE #338: Why throw exception when both gij and g_ij are given?
I calculate the metric tensor using lambdified sympy expressions. I have symbolic expressions for gij and g_ij and is therefore able to calculate both tensor with high precision. 

When I provide e.g gij and let BOUT++ calculate the inverse g_ij i get much higher errors than if I calculate it using lambdified sympy expressions. 

My problem is that if I provide gij and g_ij BOUT boutmesh throws an exception line 565. I suggest to remove this exception and perhaps testing $$g^{ij} \* g_{ij} - I < \epsilon$$ instead?

If you agree I of course do the fix :-D

JensMadsen:
This fixes #338

Consistentcy is checked by calculating gij*g_ij - I on all points in the X-Y
direction. The L2 norm of the the \infty matrix norm is then calculated. This check

	modified:   include/bout/mesh.hxx
	modified:   src/mesh/impls/bout/boutmesh.cxx
	modified:   src/mesh/mesh.cxx

### DIFF
--- a/include/bout/mesh.hxx
+++ b/include/bout/mesh.hxx
@@ -231,6 +231,7 @@ class Mesh {
   int calcCovariant(); ///< Inverts contravatiant metric to get covariant
   int calcContravariant(); ///< Invert covariant metric to get contravariant
   int jacobian(); // Calculate J and Bxy
+  BoutReal gijXg_ijMinusI(BoutReal epsilon = 10E-10); //<Returns inf norm of \| g_ij gij - I \|_inf = \epsilon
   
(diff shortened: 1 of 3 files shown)


JensMadsen:
In the case that the matrix product between gij and g_ij is not consistent the code no longer throws an exception. Only is the error printed to the screen. 

Can it be merged now? I need it:-D

PR #370, opened 2016-11-15
Make PETSc work in next

ZedThree:
- Update the various PETSc interfaces for next
- Some small fixes in configure for PETSc and NetCDF

The PETSc detection stuff in configure is probably quite fragile, given how complicated PETSc is, and also how they keep moving stuff around, e.g. `PETSC_CONFIGDIR` changes locations between 3.3 and 3.4.


ZedThree:
Merged into release candidate

PR #372, opened 2016-11-15
Fix potential bug with writing scalars in NetCDF4

ZedThree:
Only affects the NetCDF C++ API v4.2.1, fixed in v4.3.0

We don't actually need to treat scalars differently here, as getDimVec
returns an empty vector for scalars, which is the correct thing to do
anyway.


ZedThree:
Merged in release candidate

PR #373, opened 2016-11-16
small improvement for collect

dschwoerer:
* Not sure why it was limited to 0 in some cases.
* added support for getting the last element with `-1` (e.g. `tind=-1` to get the final state of the simulation)


loeiten:
I agree with @d7919.
Once at it, would it be an idea to let `collect` accept `slices`-objects for `xind`, `tind` etc.?

dschwoerer:
If anybody wants to implement slice objects, that would be a much nicer solution :)

bendudson:
Can the base of this merge be updated to v4.0.0-rc please?

PR #380, opened 2016-11-22
Adding Coordinates::gijXg_ijMinusI() which calculates the L2 vector n…

JensMadsen:
…orm of the Linf matrix norm of gij*g_ij - I.

This function can be used to test the consistency of metric the co- and contravariant tensors

 Changes to be committed:
	modified:   include/bout/coordinates.hxx
	modified:   src/mesh/coordinates.cxx

### DIFF
--- a/include/bout/coordinates.hxx
+++ b/include/bout/coordinates.hxx
@@ -85,6 +85,7 @@ public:
   int calcCovariant(); ///< Inverts contravatiant metric to get covariant
   int calcContravariant(); ///< Invert covariant metric to get contravariant
   int jacobian(); // Calculate J and Bxy
+  BoutReal gijXg_ijMinusI(); //<Returns L2 norm of \| gij*g_ij - I \|_inf
 
   // Operators
 
--- a/src/mesh/coordinates.cxx
+++ b/src/mesh/coordinates.cxx
@@ -505,6 +505,43 @@ int Coordinates::jacobian() {
   return 0;
 }
 
+/** This function calculates the L2 vector norm of the Linf matrix norm of gij*g_ij - I.
+ * 	This function can be used to test the consistency of metric the co- and contravariant tensors
+ **/
+BoutReal Coordinates::gijXg_ijMinusI(){
+	BoutReal globalError = 0.;
+	BoutReal localError = 0.;
+	///2d array of field3d pointers. Using array because vector of const references aor const pointers is not allowed
+	Field2D * const gij[3][3] = {{&g11,&g12,&g13}, {&g12,&g22,&g23}, {&g13,&g23,&g33}};
+	Field2D * const g_ij[3][3]= {{&g_11,&g_12,&g_13}, {&g_12,&g_22,&g_23}, {&g_13,&g_23,&g_33}};
+
+	BoutReal res[3][3] = {{0.,0.,0.},{0.,0.,0.},{0.,0.,0.}};
+	const BoutReal I[3][3] = {{1.,0.,0.},{0.,1.,0.},{0.,1.,0.}};
+
+	for(int x=mesh->xstart;x<mesh->xend;x++)for(int y=mesh->ystart;y<mesh->yend;y++)
+	{//do matrix multiplication for each point in space
+		for(int i = 0; i<3;i++)
+		{
+			for(int j = 0; j<3;j++)
+			{
+				for(int k = 0; k<3;k++){
+
+					res[i][j] += (*gij[i][k])(x,y) * (*g_ij[k][j])(x,y);
+				}
+				res[i][j] -=  I[i][j];
+
+			}///L2 norm of matrix inf norm
+			localError += mesh->dx(x,y)*mesh->dy(x,y)*(*std::max_element(res[i],res[i]+3));
+		}
+
+		for(int i = 0; i<3;i++)	for(int j = 0; j<3;j++){
+			res[i][j] = 0.;
+		}
+	}
+	MPI_Allreduce(&localError,&globalError,1,MPI_DOUBLE,MPI_SUM,MPI_COMM_WORLD);
+	return globalError;
+}
+
 /*******************************************************************************
  * Operators
  * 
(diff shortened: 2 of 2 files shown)


include/bout/coordinates.hxx
```diff
@@ -85,6 +85,7 @@ public:
   int calcCovariant(); ///< Inverts contravatiant metric to get covariant
   int calcContravariant(); ///< Invert covariant metric to get contravariant
   int jacobian(); // Calculate J and Bxy
+  BoutReal gijXg_ijMinusI(); //<Returns L2 norm of \| gij*g_ij - I \|_inf
```
dschwoerer: 2 norm, or inf norm?

src/mesh/coordinates.cxx
```diff
@@ -505,6 +505,43 @@ int Coordinates::jacobian() {
   return 0;
 }
 
+/** This function calculates the L2 vector norm of the Linf matrix norm of gij*g_ij - I.
+ * 	This function can be used to test the consistency of metric the co- and contravariant tensors
+ **/
+BoutReal Coordinates::gijXg_ijMinusI(){
+	BoutReal globalError = 0.;
+	BoutReal localError = 0.;
+	///2d array of field3d pointers. Using array because vector of const references aor const pointers is not allowed
```
dschwoerer: typo: aor - should be or?

src/mesh/coordinates.cxx
```diff
@@ -505,6 +505,43 @@ int Coordinates::jacobian() {
...
+	{//do matrix multiplication for each point in space
+		for(int i = 0; i<3;i++)
+		{
+			for(int j = 0; j<3;j++)
+			{
+				for(int k = 0; k<3;k++){
+
+					res[i][j] += (*gij[i][k])(x,y) * (*g_ij[k][j])(x,y);
+				}
+				res[i][j] -=  I[i][j];
```
dschwoerer: I am not sure wheter I miss something - but shouldn't there be an abs?

JensMadsen: yes. I am an idiot. Sorry. I did it in a hurry. 

I cannot test my revised version though because I cannot run any program due to problems with netcdf....

src/mesh/coordinates.cxx
JensMadsen: yes I have corrected this

src/mesh/coordinates.cxx
JensMadsen: Anyway. Such things should be done using BLAS/LAPACK....

ZedThree:
While `gijXg_ijMinusI` is a nice and descriptive name, it's a bit hard to read. Do you think something like `checkMetricConsistency` would be better?

For now, I'm just going by your previous PR and if both tensors are given, it calls `gijXg_ijMinusI` and prints the norm, but in future maybe it should take an `epsilon` argument (from `[metric]` section in input? defaults to `1e-5`?), and throw an exception if the norm is larger than epsilon.

JensMadsen:
@ZedThree. I personally prefer that a function do one thing and one thing only:-D Actually I dislike that my function do multiplication and two norms... I would prefer that the exception is not thrown in the routine. That would also exclude using it in e.g. the init in the testphase

I do not have a strong opinion about the name and I am not sure about what the epsilon should be?

dschwoerer:
I think the epsilon should be multiplied by the area(s) dx*dy.
The tensor itself should be order unity. If you do not devide by (integrated) dx*dy the check depends on the units used - who now what length scale people use (a run would start if you use km as length units, but fail with mm as length units ...)

JensMadsen:
In the original flawed PR the flawed L2 norm is correct and has a factor dx*dy

dschwoerer:
No, I meant the other way:
`\int  \| gij*g_ij -I\|_inf^2 dA / \int dA`
`= (\sum_{x,y} dx*dy \sqrt{(\| gij*g_ij -I\|_inf)^2 }) / (\sum_{x,y}dx*dy)` 
This way you get a number or order unity - in the other case you would get something of dimension `m^2` - which depends on the length units used.

src/mesh/coordinates.cxx
```diff
@@ -505,6 +505,43 @@ int Coordinates::jacobian() {
...
+	BoutReal globalError = 0.;
+	BoutReal localError = 0.;
+	///2d array of field3d pointers. Using array because vector of const references aor const pointers is not allowed
+	Field2D * const gij[3][3] = {{&g11,&g12,&g13}, {&g12,&g22,&g23}, {&g13,&g23,&g33}};
+	Field2D * const g_ij[3][3]= {{&g_11,&g_12,&g_13}, {&g_12,&g_22,&g_23}, {&g_13,&g_23,&g_33}};
+
+	BoutReal res[3][3] = {{0.,0.,0.},{0.,0.,0.},{0.,0.,0.}};
+	const BoutReal I[3][3] = {{1.,0.,0.},{0.,1.,0.},{0.,1.,0.}};
+
+	for(int x=mesh->xstart;x<mesh->xend;x++)for(int y=mesh->ystart;y<mesh->yend;y++)
```
dschwoerer: Isn't `mesh->xend` still within the grid?
I think this is where we deviate from the C++ end definition ...

JensMadsen: Ok I agree: 

Just to clarify. Assume equidistant grid (dx and dy const)
`\int \| gij*g_ij -I\|_inf^2 dA / \int dA
= (\sum_{x,y} dx*dy (\| gij*g_ij -I\|_inf)^2) / (\sum_{x,y}dx*dy)
= (\sum_{x,y}  (\| gij*g_ij -I\|_inf)^2) / (nx*ny)
`

ZedThree:
I'm leaving this until 4.1, if that's alright. The changes I've just pushed should be able to handle reading both co and contravariant tensors, and won't recalculate the other one if you supply both.

There's no code to choose to supply just the covariant and calculate the contravariant from that though. Pull requests welcome, but it'll have to wait till 4.1 I'm afraid.

dschwoerer:
@JensMadsen did you fix the issues? If so, could you push your changes?

Further, you should probably change this PR to go into next. This can be changed at the very top of this page, after clicking on the `Edit` button.

JensMadsen:
No i have not looked at it since. Sorry. 

Changed base to next

dschwoerer:
Should be closed as the repo is gone.

Is this still needed and should be reopened from a new branch?

ZedThree:
Something to do in the Coordinates refactoring

PR #381, opened 2016-11-22
Added Div_Par(f, metric) to mms.py. Calculates the parallel divergence.

JensMadsen:
modified:   mms.py


JensMadsen:
There is a bug in there..... Sorry. Missing second optional argument in Grad_par

PR #392, opened 2016-11-29
Update the package info, add a fancier version number

ZedThree:
Version number uses git describe to get the latest tag, appending
number of commits since that tag and the current commit hash


configure
```diff
@@ -1,8 +1,8 @@
 #! /bin/sh
 # Guess values for system-dependent variables and create Makefiles.
-# Generated by GNU Autoconf 2.69 for BOUT++ BOUT_VERSION.
+# Generated by GNU Autoconf 2.69 for BOUT++ v3.0-852-g2cd825c.
 #
-# Report bugs to <bd512@york.ac.uk>.
+# Report bugs to <benjamin.dudson@york.ac.uk>.
```
dschwoerer: Shouldn't bugs be reported to the github page?

bendudson: I think detailed instructions for reporting bugs should go on the website (along with many other things)

bendudson:
The improvement to configure.ac looks good, but should the string "v3.0-852-g2cd825c" be hard-coded into the configure script? I guess this was created when autoconf ran, but this should be generated when configure runs I think.

ZedThree:
Er, good point! It probably needs more quoting. Or less.

As a side point, `BOUT_VERSION` is currently hard-coded in `bout/bout.hxx`. It should probably use the variable set in configure.

bendudson:
touche, a good point to my point. Yes, BOUT_VERSION should I guess be passed at compile time rather than hard coded in the header

PR #409, opened 2016-12-06
Coding style

dschwoerer:
Resolves Issue 402


bendudson:
Thanks David! The test seems like a nice solution for now, but could be replaced with a more complete enforcement of style later.

dschwoerer:
The tests are mainly what I used to find these issues ...
They are not intended to be complete.

PR #410, opened 2016-12-07
add test-minmax to test suite

dschwoerer:
Add the minmax test to the test suite.

### DIFF
--- a/examples/test-minmax/README.md
+++ b/examples/test-minmax/README.md
@@ -1,0 +1,7 @@
+Min/Max test
+============
+
+Tests the min and max functions for fields.
+
+The min and max functions of fields are tested. Additional to the
+Field3D, it includes a case with staggered grids and Field2D.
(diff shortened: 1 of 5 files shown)


ZedThree:
Looks good. As you've been looking at this, please could you also add a README.md that briefly explains what the test does? :)

dschwoerer:
Is there a specific format the file should be in?

Especially if we want to generate a file for the examples folder, a common format would be handy to extract the short description ...

d7919:
I agree, having a standard readme template/style will help encourage documentation and make it easier to generate summaries etc.

Perhaps we could use markdown (with a widely supported set of extensions, e.g. pandoc) to use metadata to provide a summary that can easily be extracted?

ZedThree:
@dschwoerer Only really basic:

```
name-of-test
============

Description of test
```

We could put a YAML style header with metadata in, that pandoc understands, but then this renders badly on github ([see here](https://github.com/aaronwolen/pandoc-letter/blob/master/example/letter.md) - view the raw text to see what the header looks like). Not a showstopper, but something to consider.

Easiest thing is just maybe like this for now:

```
name-of-test
============

Brief description of test to be extracted. Might
be multiple lines.

Longer description of test. Basically, like a doxygen 
comment or git commit, with first line(s) the brief, 
everything else is longer.
```

PR #413, opened 2016-12-08
Short laplace3d dc fix

bshanahan:
Fixes (I think) a small bug in laplace3d.hxx which tried to take the DC component of a Field3D using the old method.


bshanahan:
Also just removed calls for mesh->ShiftXderivs in a couple files, as this doesn't exist anymore.

ZedThree:
Oh bugger, I didn't mean to commit `{nonorth,orig}_boundary_standard.cxx`. They should be removed completely.

PR #415, opened 2016-12-12
Fix several warnings in v4

ZedThree:


src/mesh/impls/bout/boutmesh.cxx
```diff
@@ -2014,57 +2014,58 @@ void BoutMesh::clear_handles() {
  *                   Communication utilities
  ****************************************************************/
 
-int BoutMesh::pack_data(const vector<FieldData*> &var_list, int xge, int xlt, int yge, int ylt, BoutReal *buffer) {
-  int jx, jy, jz;
-  int len = 0;
-  std::vector<FieldData*>::iterator it;
+int BoutMesh::pack_data(const vector<FieldData *> &var_list, int xge, int xlt, int yge,
```
dschwoerer: why not wrap at 80 chars?

src/mesh/impls/bout/boutmesh.cxx
```diff
@@ -2014,57 +2014,58 @@ void BoutMesh::clear_handles() {
...
-  return(len);
+  return (len);
 }
 
-int BoutMesh::unpack_data(const vector<FieldData*> &var_list, int xge, int xlt, int yge, int ylt, BoutReal *buffer)
-{
-  int jx, jy, jz;
-  int len = 0;
-  std::vector<FieldData*>::iterator it;
+int BoutMesh::unpack_data(const vector<FieldData *> &var_list, int xge, int xlt, int yge,
```
dschwoerer: Why write data to const fields?

bendudson: Here I think it's the vector which is const, not the fields. Fields can't be added/removed from the vector, but they can be modified.

dschwoerer: Oh, you are right. I was thinking why we need the dynamic cast, but that is due to field data being possibly of another type ...
Sorry for the confusion :-/

ZedThree:
> getData(int,int,int,ptr) is still used in ./src/sys/diagnos.cxx.

It's actually commented out there :) Actually, I think probably the whole of diagnos can be deprecated.

The manual needs some serious updating, I think :(

PR #417, opened 2016-12-13
Remove unsupported quiltmesh implementation

ZedThree:
Pruning dead code


dschwoerer:
The PR also helps also with the documentation mile stone 👍

dschwoerer:
Trying to use the quilt mesh (or any other non implemented mesh) results now in an error, instead of being silently ignored.
It doesn't seem to be an issue anywhere in the test_suite. If any code should try to use quilt, I think a clear error message is better ...

PR #432, opened 2017-01-12
Small fix for configure NetCDF

bendudson:
Typo LDFALGS -> LDFLAGS meant that checks for NetCDF failed.
Also added include flags to CXXFLAGS

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -369,7 +369,6 @@ AS_IF([test "x$with_netcdf" != "xno"],
         AC_MSG_CHECKING([if we can compile NetCDF with C++])
         # Note netcdf_c++ needed
         NCLIB=`$NCCONF --libs | sed s/-lnetcdf/-lnetcdf_c++\ -lnetcdf/`
-        LDFLAGS="$LDFLAGS -Wl,-rpath,`nc-config --prefix`/lib"
 
         save_LIBS=$LIBS
         save_LDFLAGS=$LDFLAGS
--- a/configure.ac
+++ b/configure.ac
@@ -19,7 +19,9 @@
 #
 #####################################################################
 #
-# Process this file with autoconf to produce a configure script.
+# Process this file with autoreconf to produce a configure script.
+#
+#     $ autoreconf -if 
 #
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -367,12 +369,15 @@ AS_IF([test "x$with_netcdf" != "xno"],
         AC_MSG_CHECKING([if we can compile NetCDF with C++])
         # Note netcdf_c++ needed
         NCLIB=`$NCCONF --libs | sed s/-lnetcdf/-lnetcdf_c++\ -lnetcdf/`
+        LDFLAGS="$LDFLAGS -Wl,-rpath,`nc-config --prefix`/lib"
```
ZedThree: Unfortunately, I think this needs to be a bit more complicated, as it might be `lib64` on some systems. I think NetCDF 4.4.1.1 now has `nc-config --libdir` which tells you the correct path. Otherwise, we need to parse `nc-config --libs` and extract the path on `-L`.

bendudson: I'm  not sure how critical that line is, as the linker flags should be included in NCLIB. I included it here because it was in the master branch

ZedThree:
What system and netcdf version was this failing on, btw?

bendudson:
It was failing on snake, using the old module system. I think it should fail anywhere that netcdf is not in the standard paths, since LDFLAGS was not being set correctly.

configure.ac
bendudson: I have now removed this line, as it doesn't seem to be needed

ZedThree:
I think [L391](https://github.com/boutproject/BOUT-dev/blob/5bee9b7914ea68ca743eb3e34803cd0e00560e4d/configure.ac#L391) should now be;

    CXXFLAGS="$save_CXXFLAGS -DNCDF"

[`$EXTRA_INCS` includes `$NCINC`](https://github.com/boutproject/BOUT-dev/blob/5bee9b7914ea68ca743eb3e34803cd0e00560e4d/configure.ac#L361), so CXXFLAGS doesn't need to include it again

PR #436, opened 2017-01-19

bendudson:
  are 20 and 5.


dschwoerer:
Shouldn't only bug fixes go into master?
I would assume that should be merged into next?

bendudson:
Yes, I suppose you're right. There is a separate pull request for merging into v4.0.0-RC, so happy to drop this one. Only reason is that Hermes is not yet on v4, but I probably should eat my own dog food...

PR #437, opened 2017-01-19

bendudson:
are 20 and 5.

### DIFF
--- a/src/solver/impls/imex-bdf2/imex-bdf2.cxx
+++ b/src/solver/impls/imex-bdf2/imex-bdf2.cxx
@@ -183,8 +183,8 @@ int IMEXBDF2::init(bool restarting, int nout, BoutReal tstep) {
 
   rhs = new BoutReal[nlocal];
 
-  OPTION(options, adaptive, false); //Do we try to estimate the error?
-  OPTION(options, nadapt, 1); //How often do we check the error
+  OPTION(options, adaptive, true); //Do we try to estimate the error?
+  OPTION(options, nadapt, 4); //How often do we check the error
   OPTION(options, dtMinFatal, 1.0e-10);
   OPTION(options, dtMax, out_timestep);
   OPTION(options, dtMin, dtMinFatal);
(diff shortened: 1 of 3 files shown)


src/solver/impls/imex-bdf2/imex-bdf2.cxx
```diff
@@ -788,18 +806,16 @@ int IMEXBDF2::run() {
...
+        BoutReal dtNoLimit = dtNext; // What dt would have been without artificial limit
 	if(cumulativeTime+dtNext > out_timestep){
 	  artificalLimit = true;
 	  dtNext = out_timestep - cumulativeTime;
 	}
 
-	// output << "Attempting internal step "<<counter<<" (attempt "<<adaptCounter<<")"<<endl;
-	// output << "Using dt = "<<dtNext<<endl;
+	//output << "At t=" << cumulativeTime << " attempting internal step "<<counter<<" (attempt "<<adaptCounter<<")"<<endl;
+	//output << "Using dt = "<<dtNext<<endl;
```
ZedThree: Can these two lines be deleted? Or uncommented and wrapped in a conditional of some nature?

src/solver/impls/imex-bdf2/imex-bdf2.cxx
```diff
@@ -1170,9 +1216,11 @@ PetscErrorCode IMEXBDF2::solve_implicit(BoutReal curtime, BoutReal gamma) {
     KSPConvergedReason kreason;
     KSPGetConvergedReason(ksp,&kreason);
     if(kreason<0){
-      output<<"KSP Failed to converge with reason "<<kreason<<endl;
+      //output<<"KSP Failed to converge with reason "<<kreason<<endl;
+      linear_fails++;
     }else{
-      output<<"KSP Succeeded with reason "<<kreason<<endl;
+      nonlinear_fails++;
+      //output<<"KSP Succeeded with reason "<<kreason<<endl;
```
ZedThree: Same for these commented lines

ZedThree:
This changes default values/behaviour, isn't a new feature, and I'm guessing @bendudson has thoroughly tested it, so it should go in 4.0.0, rather than 4.1.

src/solver/impls/imex-bdf2/imex-bdf2.cxx
bendudson: Useful for debugging internal timestep behavior. Perhaps a "verbose" switch?

bendudson:
I was quite keen for this to be in 4.0, because I think it's a useful feature. On refection though I would be happy for this to go into 4.1 with more testing. I have tested it, but not particularly thoroughly, and it might be good for other people to try it out on their problems first.

src/solver/impls/imex-bdf2/imex-bdf2.cxx
ZedThree: Verbose switch seems reasonable. I don't think there's a VERBOSE #define?

ZedThree:
Ok, let's get in 4.1 then

bendudson:
New pull request #476 replaces this one, merges with next

Issue #394, opened 2016-11-30
Fix tools to have the correct number of z points in version 4


loeiten:
I stumbled upon one of the bugs in `restart.py`, but I guess the problem are occurring elsewhere as well.

ZedThree:
I think this is mostly fixed in #440, but that leaves all the other interfaces broken (IDL, Mathematica, Matlab, Octave). I don't think any of them are used nearly as much as the python routines, so I'm not too worried. A nice easy task for someone who knows those languages though!

bendudson:
Agreed, I think this would be nice but is not essential. IDL is probably the next-most used, and should be fixed at some point... more urgent things to do though

dschwoerer:
Would it make sense to break them in an obvious way, so people know they need to be fixed, and don't silently get wrong results?

ZedThree:
That's a good idea, but it might be just as much effort to fix them properly.

Also, I think only idllib has this functionality anyway.

bendudson:
Yes, agreed. The collect routines for IDL, matlab, mathematica could just throw an error saying "This is broken, fix before using"

d7919:
Is this going to make v4.2 or will it be pushed back?

ZedThree:
#1312 should at least bring people to here, where they will hopefully read this nice message asking them to please fix it for us? :)

I'm removing the milestone for now, but leaving the issue open. I don't think these tools are used much, else we would've had more complaints!

ZedThree:
Non-python tools are now unsupported. Let's remove them entirely or shift them to other repos with big "Here be dragons" messages

bendudson:
Agreed, the IDL, matlab, octave tools should be separate repos

bshanahan:
My only objection would be the case for tokamak grid generation.

Is the python version of hypnotoad fully functioning?

bendudson:
Sadly no, the python replacement is not a full replacement. Hypnotoad could be a separate repo though.

PR #440, opened 2017-01-20
Boutdata/restart fix new nz

JensMadsen:
fixing expand and resize to comply with new nz convention


tools/pylib/boutdata/restart.py
```diff
@@ -159,15 +159,16 @@ def resize3DField(var, data, coordsAndSizesTuple, method, mute):
     newNx, newNy, newNz = coordsAndSizesTuple
 
     if not(mute):
-        print("    Resizing "+var)
+        print("    Resizing "+var + ' to (nx,nt,nz) = ({},{},{})'.format(newNx,newNy,newNz) )
```
loeiten: nx,ny,nz

tools/pylib/boutdata/restart.py
```diff
@@ -176,8 +177,9 @@ def resize3DField(var, data, coordsAndSizesTuple, method, mute):
                 newData[xInd, yInd, zInd] = gridInterpolator([x, y, z])
 
     return var, newData
-
-
+    #return var
```
loeiten: Reomve?

tools/pylib/boutdata/restart.py
```diff
@@ -42,8 +42,8 @@ def split(nxpe, nype, path="data", output="./", informat="nc", outformat=None):
     if outformat is None:
         outformat = informat
 
-    mxg = 2
-    myg = 2
+    mxg = mxg
+    myg = myg
```
ZedThree: You can remove both of these lines

tools/pylib/boutdata/restart.py
```diff
@@ -33,7 +33,7 @@
 except ImportError:
     raise ImportError("ERROR: os, sys or glob modules not available")
 
-def split(nxpe, nype, path="data", output="./", informat="nc", outformat=None):
+def split(nxpe, nype, path="data", output="./", informat="nc", outformat=None, mxg = 2, myg = 2):
```
ZedThree: No spaces around `=` in `def` lines:

     mxg=2, myg=2):

tools/pylib/boutdata/restart.py
JensMadsen: I do not understand. Remove what?

loeiten: `#return var`

ZedThree: You can just delete the `#return var` rather than commenting it out :)

tools/pylib/boutdata/restart.py
```diff
@@ -341,7 +329,7 @@ def is_pow2(x):
 
 
 
-def expand(newz, path="data", output="./", informat="nc", outformat=None):
+def resizeZ(newNz, path="data", output="./", informat="nc", outformat=None):
```
ZedThree: `resizeZ` is probably a better name than `expand`, but I'm worried this might break existing scripts. Anybody have strong feelings about this renaming?

ZedThree:
Might be a stupid question, but is this a complete/partial fix for #394?

loeiten:
I guess it is only partial, as this only fixes the python part of the problem, and it seems like the it is still a problem in the IDL routines (https://github.com/JensMadsen/BOUT-dev/blob/bfdb3fd074ce89c9ece6594758aef9ebfd51cc18/tools/idllib/is_pow2.pro).

tools/pylib/boutdata/restart.py
bendudson: I suspect this is not used in an automatic script anywhere, and the name is much better

PR #444, opened 2017-01-27
Fixes python restart.resizeY to handle Field2D, scalars

bendudson:
Previously only resized Field3Ds, and ignored Field2D or other fields


bendudson:
Oh very well... not in v3

PR #450, opened 2017-01-31
Sphinx/readthedocs documentation

ZedThree:
This changes the user's and developer's manuals into reStructuredText, and uses sphinx to build them into a website. This will allow us to then use readthedocs to automatically build and host the documentation. To build this locally, you will need to install breathe from pip.

While I've updated various out of dates where I've found them, it could still use a lot of work!

Importantly, I've added a paragraph to CONTRIBUTING.md about the expectation of documentation in pull requests. Basically, if you change/add anything, please keep the doxygen comments up to date.


ZedThree:
Ok, so after all that time getting doxygen + breathe + sphinx working, it turns out that sphinx uses 1.6 GB of RAM when building on readthedocs, which is more than their 1 GB limit. So, no doxygen docs on readthedocs just yet :(

Issue #367, opened 2016-11-11
Datafile copy constructor


dschwoerer:
If I copy a datafile, I lose the file name.

Therefore, I cannot write a function to return a datafile, I can only return a pointer.
Is there a good reason to do this?

This costed me some time so track it down why my data file didn't work ...

ZedThree:
That sounds like a bug to be honest. There should almost certainly be a copy-constructor, as well as a copy-assignment operator. There should probably also be a move-constructor and a move-assignment operator. See [rule of three/five][1].


[1]: http://en.cppreference.com/w/cpp/language/rule_of_three

dschwoerer:
There is a copy operator, but that doesn't keep the file name. Not having the copy constructor (i.e. being private) would have been easier to track down ...

I would have fixed it, but I thought it may be intended:
```
 // Same added variables, but the file not the same
```
[datafile.cxx:69](https://github.com/boutproject/BOUT-dev/blob/master/src/fileio/datafile.cxx#L69)

dschwoerer:
@bendudson could you comment? Git blames you :grin:

dschwoerer:
Should be fixed by #487 

The reason for the unexpected behaviour of the copy constructor is the definition of `bool Datfile::write(format,...) const` in https://github.com/boutproject/BOUT-dev/blob/next/src/fileio/datafile.cxx#L535

ZedThree:
@dschwoerer Could you confirm this is fixed?

dschwoerer:
the file it self is not copied, which may produce still unexpected behaviour, but the file name is, so I think it should work as expected, in most cases.
Further there is a move constructor - I hope that is sufficient.

PR #452, opened 2017-02-01
copy file, if datafile gets copied

dschwoerer:
Not sure that this is what was intended ...
@bendudson never commented on #367, but I hope I am not the only one that thinks this is better ...


bendudson:
This is definitely better, but generally we don't really want multiple references to the file. The copy constructor exists for returning DataFile objects from functions, but perhaps what is really needed is a move constructor (or unique_ptr) so only one DataFile object has a pointer to the file.

bendudson:
I think this PR has been replaced by PR #487. Thanks David!

PR #456, opened 2017-02-02
clean datafile from mem leaks

### LINKED ISSUE #412: Memleaks in Datafile
It seems like there are some memleak in `Datafile`.
Running
```sh
valgrind --leak-check=yes ./conduction
```
in the conduction example yields
```
==26241== 
==26241== HEAP SUMMARY:
==26241==     in use at exit: 75,331 bytes in 17 blocks
==26241==   total heap usage: 59,299 allocs, 59,282 frees, 195,833,386 bytes allocated
==26241== 
==26241== 8 bytes in 1 blocks are definitely lost in loss record 9 of 17
==26241==    at 0x4C2E0EF: operator new(unsigned long) (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==26241==    by 0x476E38: Datafile::writeVar(double const&, char const*) (datafile.cxx:555)
==26241==    by 0x45F2E7: BoutInitialise(int&, char**&) (bout++.cxx:300)
==26241==    by 0x406BF7: main (boutmain.hxx:119)
==26241== 
==26241== 184 (152 direct, 32 indirect) bytes in 1 blocks are definitely lost in loss record 13 of 17
==26241==    at 0x4C2E0EF: operator new(unsigned long) (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==26241==    by 0x47D58F: FormatFactory::createDataFormat(char const*, bool) (formatfactory.cxx:95)
==26241==    by 0x478388: Datafile::openw(char const*, ...) (datafile.cxx:141)
==26241==    by 0x44D0D2: Solver::init(bool, int, double) (solver.cxx:675)
==26241==    by 0x4A59E2: PvodeSolver::init(bool, int, double) (pvode.cxx:84)
==26241==    by 0x44FF80: Solver::solve(int, double) (solver.cxx:507)
==26241==    by 0x45DDF1: bout_run(Solver*, int (*)(double)) (bout++.cxx:325)
==26241==    by 0x406D16: main (boutmain.hxx:150)
==26241== 
==26241== 512 bytes in 1 blocks are definitely lost in loss record 14 of 17
==26241==    at 0x4C2E80F: operator new[](unsigned long) (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==26241==    by 0x477627: Datafile::Datafile(Options*) (datafile.cxx:49)
==26241==    by 0x45F67C: __static_initialization_and_destruction_0 (globals.hxx:75)
==26241==    by 0x45F67C: _GLOBAL__sub_I_mesh (bout++.cxx:562)
==26241==    by 0x4DDA3C: __libc_csu_init (in /home/mmag/BOUT-dev-r4.0.0.0/examples/conduction/conduction)
==26241==    by 0x6D437BE: (below main) (libc-start.c:247)
==26241== 
==26241== 512 bytes in 1 blocks are definitely lost in loss record 15 of 17
==26241==    at 0x4C2E80F: operator new[](unsigned long) (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==26241==    by 0x477627: Datafile::Datafile(Options*) (datafile.cxx:49)
==26241==    by 0x44C17C: Solver::Solver(Options*) (solver.cxx:52)
==26241==    by 0x4A5474: PvodeSolver::PvodeSolver(Options*) (pvode.cxx:51)
==26241==    by 0x45102E: SolverFactory::createSolver(char const*&, Options*) (solverfactory.cxx:71)
==26241==    by 0x4517D8: SolverFactory::createSolver(Options*) (solverfactory.cxx:63)
==26241==    by 0x4487CB: Solver::create(Options*) (solver.cxx:853)
==26241==    by 0x406C11: main (boutmain.hxx:130)
==26241== 
==26241== LEAK SUMMARY:
==26241==    definitely lost: 1,184 bytes in 4 blocks
==26241==    indirectly lost: 32 bytes in 1 blocks
==26241==      possibly lost: 0 bytes in 0 blocks
==26241==    still reachable: 74,115 bytes in 12 blocks
==26241==         suppressed: 0 bytes in 0 blocks
==26241== Reachable blocks (those to which a pointer was found) are not shown.
==26241== To see them, rerun with: --leak-check=full --show-leak-kinds=all
==26241== 
==26241== For counts of detected and suppressed errors, rerun with: -v
==26241== ERROR SUMMARY: 4 errors from 4 contexts (suppressed: 0 from 0)
```

dschwoerer:
Should fix #412 
However, I don't think it works with #452.

Using a [shared_ptr](http://en.cppreference.com/w/cpp/memory/shared_ptr) should solve the issue ...

### DIFF
--- a/src/fileio/datafile.cxx
+++ b/src/fileio/datafile.cxx
@@ -88,12 +88,14 @@ Datafile& Datafile::operator=(const Datafile &rhs) {
   f3d_arr      = rhs.f3d_arr;
   v2d_arr      = rhs.v2d_arr;
   v3d_arr      = rhs.v3d_arr;
-  filenamelen=FILENAMELEN;
-  filename     = new char[filenamelen];
   return *this;
 }
 
 Datafile::~Datafile() {
+  if (file){
+    delete file;
+    file = NULL;
+  }
   delete[] filename;
 }
 
@@ -528,7 +530,7 @@ bool Datafile::write(const char *format, ...) const {
   if(format == (const char*) NULL)
     throw BoutException("Datafile::write: No argument given!");
 
-  int filenamelen=512;
+  int filenamelen=FILENAMELEN;
   char * filename=new char[filenamelen];
 
   bout_vsnprintf(filename, filenamelen, format);
@@ -539,6 +541,8 @@ bool Datafile::write(const char *format, ...) const {
   tmp.openw(filename);
   bool ret = tmp.write();
   tmp.close();
+
+  delete[] filename;
   
   return ret;
 }
--- a/src/fileio/datafile.cxx
+++ b/src/fileio/datafile.cxx
@@ -44,7 +44,7 @@
 #include <utils.hxx>
 #include "formatfactory.hxx"
 
-Datafile::Datafile(Options *opt) : parallel(false), flush(true), guards(true), floats(false), openclose(true), enabled(true), shiftOutput(false), file(NULL) {
+Datafile::Datafile(Options *opt) : parallel(false), flush(true), guards(true), floats(false), openclose(true), enabled(true), shiftOutput(false), file(nullptr) {
   filenamelen=FILENAMELEN;
   filename=new char[filenamelen];
   if(opt == NULL)
@@ -64,7 +64,7 @@ Datafile::Datafile(Options *opt) : parallel(false), flush(true), guards(true), f
 Datafile::Datafile(const Datafile &other) :
   parallel(other.parallel), flush(other.flush), guards(other.guards), 
   floats(other.floats), openclose(other.openclose), Lx(other.Lx), Ly(other.Ly), Lz(other.Lz), 
-  enabled(other.enabled), shiftOutput(other.shiftOutput), file(NULL), int_arr(other.int_arr), 
+  enabled(other.enabled), shiftOutput(other.shiftOutput), file(nullptr), int_arr(other.int_arr), 
   BoutReal_arr(other.BoutReal_arr), f2d_arr(other.f2d_arr), 
   f3d_arr(other.f3d_arr), v2d_arr(other.v2d_arr), v3d_arr(other.v3d_arr) {
   filenamelen=FILENAMELEN;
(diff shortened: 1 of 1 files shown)


src/fileio/datafile.cxx
```diff
@@ -88,12 +88,14 @@ Datafile& Datafile::operator=(const Datafile &rhs) {
...
   v3d_arr      = rhs.v3d_arr;
-  filenamelen=FILENAMELEN;
-  filename     = new char[filenamelen];
   return *this;
 }
 
 Datafile::~Datafile() {
+  if (file){
+    delete file;
+    file = NULL;
```
ZedThree: Could you switch this to `nullptr` rather than `NULL`?

src/fileio/datafile.cxx
```diff
@@ -539,6 +541,8 @@ bool Datafile::write(const char *format, ...) const {
   tmp.openw(filename);
   bool ret = tmp.write();
   tmp.close();
+
+  delete[] filename;
```
ZedThree: If you `delete[] filename` here, does the destructor need to do it as well? I think here is probably the better place.

dschwoerer: `filename` shadows `filename` of the class. Therefore the local version needs to be deleted here.
The local version is needed, as this is in a const method.

ZedThree: Ah, good point. In that case, should the destructor also check `filename` before deleting it?

src/fileio/datafile.cxx
dschwoerer: I could, but the code uses NULL every where else in this file.
Should this be changed everywhere?
Probably the whole code should be checked for this. Usage of NULL: 1282. Usage of nullptr: 63
Isn't the only difference the one is already casted to pointer type?

src/fileio/datafile.cxx
dschwoerer: `filename` is always allocated - maybe it would be cleaner not to do so - but that is a different issue. Shouldn't be needed ...

src/fileio/datafile.cxx
ZedThree: `nullptr` is preferred to `NULL`, because `NULL` can be implicitly cast to `int`, which can cause all sorts of headaches. `nullptr` is its own type, which is comparable with all pointer types.

We probably should change `NULL` to `nullptr` everywhere

Issue #278, opened 2016-10-01
DataIterator::offset()


dschwoerer:
`DataIterator::offset(dx,dy,dz)` returns invalid Indices, if `nz < -dz`, e.g. if there is only one z point and `.offset(0,0,-2)` is used.

For `dz=2` I have implemented 

```
  const Indices zpp() const { int zp=(z == zend ? zstart : z+1); return {x, y, zp == zend ? zstart :zp+1 }; }
  const Indices zmm() const { int zm=(z == zstart ? zend : z-1); return {x, y, zm == zstart ? zend : zm-1}; }
```

the more general solution would be:

```
    return {x+dx, y+dy, ((z+dz-zstart + 8*nz) % nz) + zstart};
```

hoping that nobody is using an larger offset than 8 while only using a single point in z direction ...

dschwoerer:
In more detail:
% of a negative number is negative (or zero)
therefore, if -dz is larger than z+nz, the argument for modulo is zero, and the result is outside of the domain. This may not be the case for most simulations, but it is still wrong, and I don't think we should keep wrong code, just because it is faster (which I am not sure)

bendudson:
Another possibility:

(( ( (z+dz-zstart) % nz) + nz) % nz) + zstart;

This should (I think) always be in the range zstart to (zstart+nz-1) inclusive. It does involve two modulus operations though. 

Or we could put in guard cells in Z...

dschwoerer:
| range | model | mean | stddev | min | max |
|------|--------|---------|-----------|--------|-----------|
| 0 1 | BEN | 0.166138 | 0.003665 | 0.157014 | 0.169787 |
| 0 1 | ORG | 0.102617 | 0.002643 | 0.095798 | 0.106146 |
| 0 1 | DAVE | 0.061518 | 0.018036 | 0.038645 | 0.089279 |
| 0 63 | BEN | 0.165118 | 0.001613 | 0.162381 | 0.168256 |
| 0 63 | ORG | 0.096228 | 0.000952 | 0.094580 | 0.097771 |
| 0 63 | DAVE | 0.087226 | 0.026324 | 0.043176 | 0.127765 |
| 1 126 | BEN | 0.169272 | 0.002777 | 0.164576 | 0.174476 |
| 1 126 | ORG | 0.098711 | 0.001328 | 0.095508 | 0.101161 |
| 1 126 | DAVE | 0.082887 | 0.024832 | 0.040657 | 0.119596 |

[...]

PR #457, opened 2017-02-02
fix the dataiterator z-offset for abitrary offsets

dschwoerer:
fix for #278 
As one of the labels is bug, and the other is enhance request, should I rebase the PR to the next branch?

### DIFF
--- a/include/bout/dataiterator.hxx
+++ b/include/bout/dataiterator.hxx
@@ -158,7 +158,6 @@ public:
    * Add an offset to the index for general stencils
    */
   const Indices offset(int dx, int dy, int dz) const {
-    int nz = zend-zstart+1;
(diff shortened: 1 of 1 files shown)


bendudson:
I think this is already fixed; the existing code

((z+dz-zstart + nz) % nz) + zstart

wraps the z index within the correct range for any value of dz. I haven't done timings, but I assume this is quicker than a bunch of conditionals and for loops.

dschwoerer:
As mentioned in the issue, this is not true.
The `%` is extremly slow, if I remember arround 80 cycles, so it might actually be faster, as the offset is small.
But this is not the point, the old implementation is broken, please read #278, and if needed I can elaborate.

Issue #396, opened 2016-12-01
Field3D Operation in shifted metrics


dschwoerer:
It seems that if `a*b` where `a` and `b` are at different grids (e.g. CELL_CENTRE and CELL_YLOW), this is currently ignored, and the results location is set to `a`'s location.

I would have assumed that it interpolates to one location.
I guess it may be best to throw an error if this happens, as the user is the one who should make sure both fields are at the same location ...
This would cause currently probably issues with Field2d, as they are not shiftable, and therefore e.g. in the metrics area (after calculating the derivative, multiplying with dy etc)

bendudson:
Agree that this should throw an exception rather than silently interpolating

-------- Original Message --------

dschwoerer:
This is fixed by #458 ...

Issue #83, opened 2015-08-04
configure file present


dschwoerer:
Is it intentional that the generated file 'configure' is present in git?

bendudson:
Yes, since not all users have autoconf installed (or know what it is).

PR #458, opened 2017-02-02
Fieldops

dschwoerer:
Improve the fieldops (make them faster + less code duplication)

I assumed that the bug fixes from the v4-RC branch should also be in next?
That doesn't seem to be the case, so should I rebase?


dschwoerer:
also fixes #396

bendudson:
This seems to have pulled in all the changes, so difficult to see what the new bit is. I guess a rebasing is needed. Thanks for all the good work David!

ZedThree:
I'll get v4-rc merged into next today, then you can rebase onto that, should make it all nicer. Likewise for your other PRs, I think. Thanks again for the good work!

dschwoerer:
Ok, I will update the PR after that  👍

ZedThree:
I've not had proper chance to go through this yet, but I have some general comments after a quick skim:

- Documentation please! :)
- Please delete lines rather than commenting them out
- You've removed `const` from the binary arithmetic operators -- this is a pretty big change that I'm not sure about. Please could you comment why?
- Your test uses the "ailoi" mesh, which I think is not generally available yet :)

Also, I'm undecided about whether the generated code should be under version control or not. I'm currently leaning towards "should be in". Maybe with a massive header at the top "THIS IS A MACHINE GENERATED FILE, DO NOT CHANGE BY HAND!" or something. Anyone with opinions on this?

dschwoerer:
I guess I should have cleaned up before pushing.

Sorry for no comments. Will fix asap.

I am not in favour of shipping auto generated code, as it increases the size of the git repo - but you don't necessarily agree with me (see #83)

ZedThree:
To be honest, I can't 100% convince myself either way. Not including the generated code then puts hard dependencies on both python and the generating script. Python is pretty much everywhere though, and the generator doesn't use any external dependencies -- though there is always the python 2/3 issue.
Not including it does also makes it harder to review changes to the generated code.

Including the generated code risks people trying to commit changes to it, though hopefully that would be caught at code review. As you say, it does also increase the size of the repo, but not by that much.

As I said, I'm currently leaning more towards including it, but I'm willing to be persuaded!

dschwoerer:
I would like to make it compatible with both python 2 and 3. Currently is it only python2.
However as most of the script is printing, is there a simple way to port it without adding additional dependencies?

Concerning the Inclusion of the cxx file:
Pro including it:
 * I have more contributions in therms of source lines
 * Easier to see changes (if the auto generated file is added)

Contra:
 * git wants to merge changes
 * it doesn't matter what is in the file, as it is overwritten by make

Concerning: 
> You've removed `const` from the binary arithmetic operators -- this is a pretty big change that I'm not sure about. Please could you comment why?
I removed the const for all returning functions, I think that's better:
https://stackoverflow.com/questions/8716330/purpose-of-returning-by-const-value

loeiten:
I thought BOUT was meant to be py3 compatible?

Anyways, automated conversion to py2/3 compatible code can be done with http://python-future.org/automatic_conversion.html.
It can in principle be done without introducing dependencies of the `future` package.

bendudson:
Thanks for this David. I'm not convinced though that changing from macro-generated code to python-generated code is a good thing: With the macro it was pretty clear what code was generated, but it's much harder (for me at least) to understand from the python code what C++ code will be generated. Perhaps documentation would help.

dschwoerer:
I have added the [generated file](https://github.com/boutproject/BOUT-dev/pull/458/commits/cf7f0da1c5e5d6cf2b176b320601c1f48c431eec). Not sure what documentation I should add. I already added some comments to [`gen_fieldops.py`](https://github.com/dschwoerer/BOUT-dev/blob/cf7f0da1c5e5d6cf2b176b320601c1f48c431eec/examples/test-fieldops/gen_fieldops.py) ...

dschwoerer:
I have added the [generated file](https://github.com/boutproject/BOUT-dev/pull/458/commits/cf7f0da1c5e5d6cf2b176b320601c1f48c431eec). Not sure what documentation I should add. I already added some comments to [`gen_fieldops.py`](https://github.com/dschwoerer/BOUT-dev/blob/cf7f0da1c5e5d6cf2b176b320601c1f48c431eec/examples/test-fieldops/gen_fieldops.py) ...

dschwoerer:
I tried to rewrite this in bash - but it was again around 300 lines of code - and also not much easier to read ...

While the macro version is easier to read, it involves a lot of code duplication.
This is reduced with this code - with the disadvantage of less readability.
But as some of the macros are missing some code (conserving the mesh, checking locations, ...) I think it might be worth considering this patch ...

src/field/gen_fieldops.py
```diff
@@ -185,10 +187,17 @@ def mymin(f1,f2):
             for d in dims:
                 print("  }")
             print("  }")
```
bendudson: I think this approach to balancing braces is a bit fragile, since changes to the generation code need be made in several places simultaneously. I have seen this done elsewhere using "with" blocks which emit open braces on entering, and closing braces on exit:

    with braces:
        some code

dschwoerer: It now has the disadvantage that the there is only one set of braces for the triple for loop - I hope that this is fine - as it is represented in python.

dschwoerer:
Note: this includes the #752 PR.

dschwoerer:
The merge was not working as expecting.
I had to format some files in next to not having to resolve everything by hand 👎
I will open a new PR for the formatting.
If someone has more success with a merge I'd be happy to replace the 6 merge commits with a cleaner version ...

ZedThree:
I'm going to give it a go now. Although, I do want to apply clang-format to the whole codebase, so we might keep #755 anyway...

I've also spotted a couple of nice things that can be pulled out into separate PRs -- I've cherry-picked them so you retain authorship.

PR #463, opened 2017-02-07
Fix for use with intel compilers without openmp.

d7919:
For some reason the references used to store *min and *max in the data
iterator were being set to the wrong memory locations when using intel
17 compilers. Using separate int vars (as done when running with openmp)
seems to fix this problem. Whilst this removes the nicely explicit
behaviour that without openmp *min should be == *start etc. it at least
means we get the correct behaviour with intel and there's a unified
declaration of these vars.

### DIFF
--- a/include/bout/dataiterator.hxx
+++ b/include/bout/dataiterator.hxx
@@ -222,20 +222,10 @@ private:
   DataIterator(); // Disable null constructor
 
   int xstart, ystart, zstart;
-
-#ifndef _OPENMP
-  int &xmin, &ymin, &zmin;
-#else
   int xmin, ymin, zmin;
-#endif
-
   int xend, yend, zend;
-
-#ifndef _OPENMP
-  int &xmax, &ymax, &zmax;
-#else
   int xmax, ymax, zmax;
-#endif
+
   const bool isEnd;
   /// Advance to the next index
   void next() {


dschwoerer:
It is a shame that icc doesn't handle this correctly.

I am not really in favour of this fix, as this increases the memory consumption unnecessarily, but also, and I think this is the more troubling part, leads to the fact that we have have a duplicate of a variable, that, if it should get changed needs to update the copy as well. This can lead to bugs, therefore I would like to consider other possibles before merging.

Maybe we should make ?min/?max const, and in the non-omp case ?start/?end as well?
another option would be macros, for xstart e.g. resolving in the OMP case to xstart, and otherwise to xmin ...

d7919:
I was surprised that intel couldn't handle this as well and as the issue was holding back the work of others I thought I'd share this work around for now and the PR acts as a slightly more urgent reminder that something needs fixing soon. 

Making things `const` that can be sounds like a good idea to me.  I'm not super familiar with these iterators so I could be quite wrong, but I think these vars are private members so can only be changed by the iterator themselves. As such adding `const` in non-openmp should be enough to protect against accidental changes, and the openmp version is unchanged (at risk of accidental changes?). 

I'm not sure we can be certain how this will impact memory usage as according to the standard the storage requirements for references is not defined (at least according to [here](http://stackoverflow.com/a/13964709)), but I'd have thought this is likely negligible compared to typical field data sizes or am I missing something (quite possible!).

d7919:
30b71d6 provides an alternative fix where the reference is replaced with `const`, I can close this PR and open one based off that.

ZedThree:
`const` version looks good to me

dschwoerer:
I guess the `const` version is even better then the reference version.
Therefore the variables can be stored in registers, and don't need to be in memory (to allow pointers)

d7919:
Closing as improved fix in #466 . Thanks for the suggestion!

dschwoerer:
Intels c++ is support isn't great.
#187 is another example of this ...

PR #468, opened 2017-02-10
Alt fix for use with intel compilers without openmp. (more const)

dschwoerer:
Alternative to #466 
In this case, also ?start and ?end are const.
This is an implementation detail, so if there should be ever a reason to change this, it shouldn't break anything.


bendudson:
Is it agreed that this is the right way to fix this problem?

PR #469, opened 2017-02-10
Fix rare double free with BoutException

ZedThree:
Strange double free occurred with some MPI libraries, running with certain number of processes.


dschwoerer:
Good to check we don't `delete[] nullptr;` 👍

PR #473, opened 2017-02-10
Travis matrix (new)

dschwoerer:
... trying to understand why clang builds are ignored.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,18 +2,16 @@ sudo: required
 dist: trusty
 language: cpp
 compiler: gcc
-python:
-  - "3.5"
 matrix:
   include:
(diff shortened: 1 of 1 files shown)


ZedThree:
I'm very much liking having it built with clang as well. Just curious, why the removal of the python version?

dschwoerer:
I checked it [here](https://lint.travis-ci.org/) and travis was complaining that the python version is irrelevant with this language.
I can reintroduce it, if you think it has any effect ...

ZedThree:
That looks like a useful tool! Ah, I guess it doesn't actually matter because we use anaconda to sort out all the python stuff, and we specify python 3.5 there.

PR #476, opened 2017-02-15

bendudson:
are 20 and 5.

o Added doxygen comments and a manual section to explain how to use the solver

### DIFF
--- a/examples/IMEX/drift-wave/README.md
+++ b/examples/IMEX/drift-wave/README.md
@@ -21,17 +21,34 @@ Running with CVODE, treating both convective and diffusive parts together:
 
     $ ./test-drift
 
-    1.000e+02        574              574       1.87e+00    57.2   29.8    4.7    0.3    8.0
-    2.000e+02        233              233       7.91e-01    57.2   29.5    4.9    0.6    7.7
-    3.000e+02        237              237       7.73e-01    57.1   29.6    4.7    0.7    7.8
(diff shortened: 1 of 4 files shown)


examples/IMEX/drift-wave/README.md
```diff
@@ -21,17 +21,34 @@ Running with CVODE, treating both convective and diffusive parts together:
...
+
+    $ ./test-drift solver:type=imexbdf2 solver:maxl=100
+
+    1.000e+02          2               64       2.54e-01    41.8   15.6    2.1    2.5   38.1
+    2.000e+02          2               57       2.24e-01    42.3   15.9    2.1    2.8   36.9
+    3.000e+02          2               53       2.08e-01    42.3   15.9    2.1    3.1   36.6
+   
+Without adaptive timestep:
+
+    $ ./test-drift solver:type=imexbdf2 solver:maxl=100 solver:adaptive=false
```
dschwoerer: `solver:adaptive=false` doesn't seem to have any influence?

bendudson: no; in that case because maxl is larger than the number of linear iterations the timestep is never cut by the adaptive timestepping. The accuracy criteria are met, and the timestep is the same as the output interval, so the adaptive timestepping has no effect. If maxl is smaller (default 20), or if the accuracy is tighter, then adaptive timestepping does make a difference

PR #479, opened 2017-02-17
Support of reading Field2D from gridfiles which have values at ghost points

JensMadsen:
Previous default behaviour was when reading a field2d: 
x-direction) read including ghostpoints and write including ghostpoints
y-direction) read excluding ghostpoints irrespective of whether these are in the field being read from the file. Furthermore, ghost point values were set to nearest neighbor

Now the ghostpoint values are read fi they are there and in that case ghostpoint values are not set to nearest neighbor

### DIFF
--- a/src/mesh/data/gridfromfile.cxx
+++ b/src/mesh/data/gridfromfile.cxx
@@ -166,27 +166,62 @@ bool GridFile::get(Mesh *m, Field2D &var,   const string &name, BoutReal def) {
   int ys = m->OffsetY;
 
   // Index offsets into destination
-  int xd = 0;
-  int yd = m->ystart;
+  int xd = -1;
+  int yd = -1;
 
-  // Number of points to read
-  int nx = m->LocalNx;
-  int ny = m->yend - m->ystart + 1;
-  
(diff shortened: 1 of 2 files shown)


src/mesh/data/gridfromfile.cxx
```diff
@@ -166,27 +166,62 @@ bool GridFile::get(Mesh *m, Field2D &var,   const string &name, BoutReal def) {
...
-
-  // Number of points to read
-  int nx = m->LocalNx;
-  int ny = m->yend - m->ystart + 1;
-  
-  for(int x=xs;x < xs+nx; x++) {
+  int xd = -1;
+  int yd = -1;
+
+  ///Ghost region width. This is error prone. Should check if really integer!!
```
bendudson: Maybe could use "/2" rather than "0.5*" so it uses integer arithmetic rather than floating point (where rounding may do strange things). You could also have a check before, something like:

ASSERT1((m->LocalNx - (m->xend - m->xstart + 1)) % 2 == 0);

to check that it's a multiple of 2

bendudson:
This seems like a nice and simple fix for now. Thanks Jens!

Issue #419, opened 2016-12-16
BoutMesh ignores non-available differentiation-options


dschwoerer:
* If I manage so set the derivative scheme to something non-available, this is (nearly silently) ignored:
```
Setting X differencing methods
	First       :  No match for 'awesomescheeme' ->  Second order central (C2)
```
* Expected outcome:
Exit with error, as the code cannot do what is expected.
The aiolos mesh prints in such a case the possible options and throws an exception.

dschwoerer:
This breaks existing code - so maybe a feature for 5.0?

bendudson:
I would be in favour of fixing this earlier, since silently ignoring an input is probably a bug

PR #481, opened 2017-02-22
throw if diffmethod is unknown

dschwoerer:
It is a bug fix, but it also could break code - should it go in master?


dschwoerer:
Fixes #419

dschwoerer:
In the past if `C1` was given, `C4` for was returned. Now an error is thrown. I think this is cleaner, than guessing ...

I kept that when no label is given, the first method is returned - although I am not sure that this is the best way to set the default value.

PR #488, opened 2017-03-08
rename LocalNx etc

dschwoerer:
The name is not according to coding guide (and I cannot remember the capital L in the beginning)

update your files with:
```
for d in x y z; do for f in ` find . -type f | grep x\$` ; do sed "s/LocalN$d/local_n$d/g" $f -i ; done;done
```


bendudson:
Agree this is probably the right thing to do. The names LocalN* were chosen to be consistent with GlobalN*, so whilst we're at it can we rename those variables too?

https://github.com/boutproject/BOUT-dev/blob/master/include/bout/mesh.hxx#L403

Probably also Offset* 

https://github.com/boutproject/BOUT-dev/blob/master/include/bout/mesh.hxx#L404

dschwoerer:
As this will not be implemented before bout 5.0 - closed for now ...

PR #489, opened 2017-03-09
remove using namespace std;

dschwoerer:
I guess we don't want to have `using namespace std;` in a header.

not sure we want to remove `using std::string;` and the like.

### DIFF
--- a/include/boutexception.hxx
+++ b/include/boutexception.hxx
@@ -8,7 +8,6 @@ class BoutException;
 #include <string>
 
 using std::string;
-//using namespace std;
 
 void BoutParallelThrowRhsFail(int &status, const char* message);
 
--- a/include/boutexception.hxx
+++ b/include/boutexception.hxx
@@ -7,11 +7,12 @@ class BoutException;
 #include <exception>
 #include <string>
 
-using namespace std;
+using std::string;
+//using namespace std;
 
 void BoutParallelThrowRhsFail(int &status, const char* message);
 
-class BoutException : public exception {
+class BoutException : public std::exception {
 public:
   BoutException(const char *, ...);
   BoutException(const std::string);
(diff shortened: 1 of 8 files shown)


ZedThree:
Looks good. I'm guessing removing `using std::string` touches a *lot* of lines?

include/boutexception.hxx
```diff
@@ -7,11 +7,12 @@ class BoutException;
 #include <exception>
 #include <string>
 
-using namespace std;
+using std::string;
+//using namespace std;
```
ZedThree: You can delete this line rather than just commenting it out

bendudson:
Agree with removing "using namespace std;". using std::string is probably
ok though, as it's specific to one class and it's quite obvious if someone
uses "string" in code what they mean.

dschwoerer:
That's why slow merging is annoying :-/

removing `using std::string;` would probably require around 177 changes in header files ...

There are still some more `using` statement in header files:
```
include/bout/solver.hxx:using std::string;
include/bout/rkscheme.hxx:using std::string;
include/bout/rkscheme.hxx:using std::setw;
include/options.hxx:using std::map;
include/options.hxx:using std::string;
include/boundary_op.hxx:using std::string;
include/boundary_op.hxx:using std::list;
include/boutexception.hxx:using std::string;
include/dataformat.hxx:using std::string;
include/dataformat.hxx:using std::vector;
include/output.hxx:using std::endl;
include/interpolation_factory.hxx:using std::string;
include/boundary_region.hxx:using std::string;
include/field_data.hxx:using std::string;
include/field_data.hxx:using std::vector;
include/field_data.hxx:using std::map;
include/utils.hxx:using std::abs;
include/utils.hxx:using std::swap;
include/boundary_factory.hxx:using std::string;
include/boundary_factory.hxx:using std::map;
```
I am not sure about abs, as we have our own definitions ...

Issue #493, opened 2017-03-21
MYG and MXG not read from gridfile


JensMadsen:
If MXG and/or MYG are defined in gridfile they should be read

PR #494, opened 2017-03-21
MXG and MYG read from grid files

bendudson:
Tries to read number of guard cells from the grid input
file (or [mesh] section if no file used). If not found,
reverts to top-level options as before.

Fixes issue #493

### DIFF
--- a/src/mesh/impls/bout/boutmesh.cxx
+++ b/src/mesh/impls/bout/boutmesh.cxx
@@ -127,8 +127,20 @@ int BoutMesh::load() {
   output << "\tGrid size: " << nx << " x " << ny << " x " << MZ << endl;
 
   // Get guard cell sizes
-  options->get("MXG", MXG, 2);
-  options->get("MYG", MYG, 2);
+  // Try to read from grid file first, then if not found
+  // get from options
+  if (Mesh::get(MXG, "MXG")) {
+    // Error code returned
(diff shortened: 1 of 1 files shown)


src/mesh/impls/bout/boutmesh.cxx
```diff
@@ -127,8 +127,20 @@ int BoutMesh::load() {
...
   // Get guard cell sizes
-  options->get("MXG", MXG, 2);
-  options->get("MYG", MYG, 2);
+  // Try to read from grid file first, then if not found
+  // get from options
+  if (Mesh::get(MXG, "MXG")) {
+    // Error code returned
+    options->get("MXG", MXG, 2);
+  }
+  ASSERT0(MXG >= 0);
```
dschwoerer: This is an extremely cheap check.
Shouldn't we always assert this, not just with CHECK given?

bendudson: ASSERT0 is always checked, since if CHECK is not given the level is set to zero:
https://github.com/boutproject/BOUT-dev/blob/master/include/bout/assert.hxx#L29

dschwoerer: Sorry, missed the declaration of `CHECKLEVEL` above the definition.
So this is only disabled if somebody declares `-DCHECK=-1` 👍

PR #495, opened 2017-03-27
Petsc slepsc bug fixes

loeiten:
Fix of missing declaration of `verbose` and a missing `std`, which gave errors when making `next` with `petsc` and `slepsc`.


loeiten:
Good point, I forgot to turn of the white space strip.
Will take better care next time :)

PR #500, opened 2017-03-27
Removing IS_MYPROC

loeiten:
This is not used anywhere, and is also wrong, as mentioned in #226.


dschwoerer:
strictly speaking this changes the interface of bout.
Shouldn't we deprecate it, and drop it in 5.0?

loeiten:
I guess this would be the proper way of doing it :).
However, it is a private class member which is wrong and not in use.
If anyone wants to use it, then strictly speaking, the changes will have to wait to 5.0 anyway.
Therefore, my opinion is that it is fine to merge at the current time.

dschwoerer:
I didn't remember that it is private ...

PR #502, opened 2017-04-03
C++11 snprintf

ZedThree:
Replaces C-style variadic functions with lovely C++ variadic templates, and provides a wrapper, `string_format`, around `snprintf` that replaces `bout_vsnprintf`. For instance:

in `include/datafile.hxx`:

```cpp
    bool openr(const char *filename, ...);
```

in `src/fileio/datafile.cxx`:

```cpp
    bool Datafile::openr(const char *format, ...) {
      if(format == (const char*) NULL)
        throw BoutException("Datafile::open: No argument given for opening file!");
      bout_vsnprintf(filename,filenamelen, format);
```

can be replaced with:

in `include/datafile.hxx`:

```cpp
    template <typename... Args>
    bool openr(const std::string &filename, Args... args) {
      return openr(string_format(filename, args...));
    }
    bool openr(const std::string &filename);
```

and `filename` can be used directly in `Datafile::openr`.

Notes:

- You can't pass a variable of type `std::string` to `string_format()` as one of the variadic arguments, as `snprintf` can't handle this:

```cpp
     std::string message = "this won't work";
     string_format("%s", message); // Error!
     string_format("%s", "this will work"); // Ok!
     std::string message2 = "this will also work";
     string_format("%s", message2.c_str()); // Ok!
```

- While `bout_vsnprintf` is now deprecated, I've not found a nice portable way of marking it as such.
- The string utilities are now split out of `utils.hxx` into `string_utils.hxx` -- this is to avoid a circular dependency with `boutexception.hxx`
- Almost completely removes need for `new char[]` anywhere. Might not be possible to remove `char *` yet though
- Added a brief test for the `TRACE` macro -- `test-msgstack`
- A more robust replacement for the `printf` family is the [fmt library](http://fmtlib.net/) -- we should consider including it (code is BSD licensed, so we could just bundle it)


dschwoerer:
I just wanted to note that the functions are not type safe - in that sense I cannot see an immediate advantage. Further as std::string cannot be constructed via move semantics from an char * pointer - this constructs seems like unncessary calls to malloc (I guess it shouldn't degrade performance in most cases, but haven't tested).
I think we should rather change the interfaces to allow `std::string` to be passed - while keeping the internals of the library fast (using `char*` as output buffer).

So may I ask, what are the advantages?

More general:
* You again mixed white space changes with code changes.
* I think the Simplify Field error function commit is useful - but could be a different PR.

ZedThree:
Good point that it's still not type safe -- we'd have to move to something like fmt for that. In fact, any type-safe printf can be a drop-in replacement now.

One immediate advantage though is reducing the risk of memory leaks and double frees. This method removes the need to keep a `new char[]` buffer about that needs to be freed using `delete[]`, and possibly reallocated to a larger size. This removes a bunch of code around keeping track of `char[]` buffers and their lifetimes. Less code == better code!

It should also give better error messages than the C-style variadic functions.

It's not an enormous improvement, but I think it is useful.

src/fileio/datafile.cxx
```diff
@@ -68,11 +66,8 @@ Datafile::Datafile(Datafile &&other) :
   enabled(other.enabled), shiftOutput(other.shiftOutput), file(other.file.release()), int_arr(other.int_arr),
   BoutReal_arr(other.BoutReal_arr), f2d_arr(other.f2d_arr),
   f3d_arr(other.f3d_arr), v2d_arr(other.v2d_arr), v3d_arr(other.v3d_arr) {
-  filenamelen=other.filenamelen;
-  filename=other.filename;
-  other.filenamelen=0;
-  other.filename=nullptr;
-  other.file = nullptr;
+  filename = other.filename;
```
dschwoerer: would be nice to use move semantics, similar to `file`

src/fileio/datafile.cxx
```diff
@@ -68,11 +66,8 @@ Datafile::Datafile(Datafile &&other) :
   enabled(other.enabled), shiftOutput(other.shiftOutput), file(other.file.release()), int_arr(other.int_arr),
   BoutReal_arr(other.BoutReal_arr), f2d_arr(other.f2d_arr),
   f3d_arr(other.f3d_arr), v2d_arr(other.v2d_arr), v3d_arr(other.v3d_arr) {
-  filenamelen=other.filenamelen;
-  filename=other.filename;
-  other.filenamelen=0;
-  other.filename=nullptr;
-  other.file = nullptr;
```
dschwoerer: I wasn't sure whether after a move the old pointer is set to `nullptr` is this guaranteed by the standard (and done by compilers)?

src/fileio/datafile.cxx
```diff
@@ -81,9 +76,6 @@ Datafile::Datafile(const Datafile &other) :
   enabled(other.enabled), shiftOutput(other.shiftOutput), file(nullptr), int_arr(other.int_arr),
   BoutReal_arr(other.BoutReal_arr), f2d_arr(other.f2d_arr),
   f3d_arr(other.f3d_arr), v2d_arr(other.v2d_arr), v3d_arr(other.v3d_arr) {
-  filenamelen=other.filenamelen;
-  filename=new char[filenamelen];
-  strncpy(filename,other.filename,filenamelen);
```
dschwoerer: Why don't you copy the filename?

src/fileio/datafile.cxx
```diff
@@ -105,29 +97,14 @@ Datafile& Datafile::operator=(Datafile &&rhs) {
...
-    filename=nullptr;
-    filenamelen=0;
-  }
-}
+bool Datafile::openr(const std::string& format) {
 
-bool Datafile::openr(const char *format, ...) {
-  if(format == (const char*) NULL) 
-    throw BoutException("Datafile::open: No argument given for opening file!");
+  filename = format;
```
dschwoerer: I would prefer to directly change the argument of the function to `filename`, instead of `format` ...

PR #507, opened 2017-04-11
Changes to PetscOptions in PETSc 3.7

bendudson:
Fixes compilation with PETSc 3.7, tested using 3.7.5

Extra first argument added from version 3.7:
https://www.mcs.anl.gov/petsc/documentation/changes/37.html

affects petsc time integration solver implementation.


ZedThree:
We might want to think about moving to a docker image on Travis that contains most/all of the third-party libraries we support, so we can check they all work.

Issue #483, opened 2017-02-27
Restart files should include MXSUB such that boutdata.collect can read them


JensMadsen:

dschwoerer:
fixed by #508

PR #508, opened 2017-04-11
Solver adds Mesh variables to restart files

bendudson:
By calling mesh->outputVars(restart), the mesh indices and
metric tensor components are added to the restart files.

This allows collect to read restart files, fixing issue #483

To avoid mesh variables being overwritten by a restart file read,
they must be added after the restart file is read. Whilst loading
the mesh from the restart file would be desirable (I think), this
would currently introduce inconsistencies and unexpected behaviour.
It would also require changes to code which modifies restart files,
since these would now have to also resize/modify metric components etc.

Since NXPE is now added twice, DataFile::add has been modified:
If a variable has already been added, it checks if it is the same
object by comparing pointers. Adding the same object twice with the
same name results in a warning message, but if the object is

### DIFF
--- a/include/datafile.hxx
+++ b/include/datafile.hxx
@@ -121,7 +121,12 @@ class Datafile {
   bool write_f2d(const string &name, Field2D *f, bool save_repeat);
   bool write_f3d(const string &name, Field3D *f, bool save_repeat);
 
-  bool varAdded(const string &name); // Check if a variable has already been added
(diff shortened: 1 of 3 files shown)


bendudson:
This PR will conflict with PR #535, so probably only one should be merged. If #535 is approved, then the mesh variables need to be added to restart in PhysicsModel

Issue #451, opened 2017-02-01
Increase verbosity of option reader


dschwoerer:
The problem:
-------------
I have in my `BOUT.inp`
```
[blob]
dx                    = perp_size
dz                    = perp_size
dz_n                  = dz
```
So that I can set easily the filament size of my simulations with the command line, by running
```
./filaments blob:perp_size=0.005
```
Although this is very convenient, it is completely hidden in the logs that the perp_size was set on the command line:
```
        Option blob:dz = perp_size (size_05/data_rd_n_1p000_t_1p20/BOUT.inp)
        Option blob:dz_n = 0.005 (size_05/data_rd_n_1p000_t_1p20/BOUT.inp)
```
Why this is an issue
---------------
In order to reproduce it is beneficial to know how a simulation was run.

Possible solutions
-------------
* It would be nice if the complete stack, how an option was read is printed.
 Depending on how this is done, it may however lead to extremly verbose logs. So a compromis may be to only write out the root-option only once. The root option in this case would be `blob:perp_size`

* Another option would to dump additional information to e.g. `BOUT.cfg.nc` - including BOUT++ version, the version of the physics model (if applicable), the command line options and the content of `BOUT.inp`

d7919:
Presumably we can dump the entire `argv`  passed into main? This way we would also capture the options passed through to `PETSc` etc.

d7919:
I should note that I'm strongly in favour of including the input file (and other provenance information like version/hash etc.) as a part of the dump file(s).

bendudson:
I think the origin of perp_size should also be printed. In addition to the line
    Option blob:dz = perp_size (size_05/data_rd_n_1p000_t_1p20/BOUT.inp)
there should also be a line
    Option blob:perp_size = 0.005 (Command line)
It might be printed earlier, as it's only printed once when perp_size is first resolved.

Another possibility would be to dump the Options tree to an output (BOUT.settings?) which could then be renamed to BOUT.inp to rerun the simulation

bendudson:
I think this is now fixed in next: The whole stack is printed to the log file, and the settings used are written to BOUT.settings

Issue #233, opened 2016-09-09
commandline options not recognized should halt program execution


JensMadsen:
For instance

giving nype at the commandline is not recognized, but no error.

bendudson:
This is a little difficult, because:
1. There is no way to know which options will be used until the program has
run. Options are put into a database (tree), then requested by various
parts of the code as it runs. Perhaps unused options could be checked after
the first iteration, when hopefully all options have been used?

2.Some options on the command line may be used by other libraries e.g.
PETSc. Also difficult to tell which are unused, so PETSc prints unused
options at the end of the run.

JensMadsen:
Ok. If it is difficult then perhaps a warning after BoutInitialise if options have not been used?

dschwoerer:
Duplicate of #35 

Can be closed

Issue #35, opened 2015-03-12
non-recognized command line option should raise a message or halt the program 


JensMadsen:
When running a program called "diffusion": 

./diffusion -ben=44

The program has no option named ben. A suggestion is to warn the user that he/she has used a commandline option not recognized by the program. Especially if one has misspelled an option it would be nice with a warning.

ZedThree:
This is solved in my help message branch. Might need some tidying up before
merging now.

JensMadsen:
Ok. Should I keep the issue (you will then delete it when merged) or should I remove it ?

Jens Madsen
PPFE, DTU Physics
Denmark
________________________________

ZedThree:
Keep it, I'll merge it and close the issue

ZedThree:
I think this might be a "won't fix". My branch only solves a subset of this problem - the "-d" and "-f" flags, which are not handled by the options parser, and adds a "-h/--help" flag. All the other command line options are added to a list and (possibly) consumed later, so it's not possible to know beforehand what options exist.

One workaround might be the following:
There is an [`Options::printUnused()`](https://github.com/boutproject/BOUT-dev/blob/cbe528207ba3230321b9482e7516f4d492cd460e/src/sys/options.cxx#L197) method which could be called (maybe in the `BOUTMAIN` macro?) and if any options are unused, pause the simulation for ~15-30 seconds with a message to the effect of "Warning, following options unrecognised. Hit Ctrl-C to abort run".

ZedThree:
Won't fix -- doesn't seem possible to do, short of requiring all possible options to be registered at load time. Even then, we'll still have the problem of e.g. mutually exclusive options.

PR #512, opened 2017-04-12
Output settings to file

bendudson:
The options used by BOUT++, including any overwritten by command-line arguments, are saved to a file BOUT.settings

This is done twice:
* Once during initialisation, so that the options are saved if the simulation stops/crashes
* Again at the end, so the comments include whether the option was used during the simulation.

The format used is the same as the BOUT.inp file, so in principle it should be possible to rename BOUT.settings to BOUT.inp to run the simulation again. It can also be parsed by ConfigParser and BoutOptionsFile (https://github.com/boutproject/BOUT-dev/blob/master/tools/pylib/boutdata/data.py#L137) for post-processing.

Note that this does not capture command-line arguments to libraries such as PETSc. That could be done by storing the command-line arguments in an option string, but it would need the user to manually copy and re-use those options.

This fixes issue #451, and is a partial fix for issue #35 and issue #233

### DIFF
--- a/include/options.hxx
+++ b/include/options.hxx
@@ -42,7 +42,6 @@ class Options;
 #include "bout_types.hxx"
 
 #include <map>
-using std::map;
 #include <string>
 using std::string;
 
@@ -143,14 +142,10 @@ public:
 
   /// Print the options which haven't been used
   void printUnused();
- private:
-  static Options *root; ///< Only instance of the root section
-  
-  Options *parent;
-  string sectionName; // section name (if any), for logging only
 
+  
   /*!
-   * Private class, used to store values, together with
+   * Class used to store values, together with
    * information about their origin and usage
    */
   struct OptionValue {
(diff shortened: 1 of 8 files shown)


dschwoerer:
I really like this 👍 
I have some suggestions for improvement in #515 
* I think leaving the implementation of OptionsParser::write purely virtual helps, as this will be noticed at compile time, otherwise the code will crash just after start.
* thought it is mayby nicer to only write a comment to the options that are not used, where they are from. I think this could help to see easier what has not been used ...

dschwoerer:
I really like this 👍 
I have some suggestions for improvement in #515 
* I think leaving the implementation of OptionsParser::write purely virtual helps, as this will be noticed at compile time, otherwise the code will crash just after start.
* thought it is mayby nicer to only write a comment to the options that are not used, where they are from. I think this could help to see easier what has not been used ...

PR #514, opened 2017-04-12
fix for datafile

dschwoerer:
Not sure why, but without I cannot compile ....

### DIFF
--- a/src/fileio/datafile.cxx
+++ b/src/fileio/datafile.cxx
@@ -43,6 +43,7 @@
 #include <boutcomm.hxx>
 #include <utils.hxx>
 #include <msg_stack.hxx>
+#include <cstring>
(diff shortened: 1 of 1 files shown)


d7919:
Odd, what where the issues without it? Was it complaining about not knowing `strncpy`? This is provided by  `cstring` (or `string.h`) -- note we include `string` in the headerfile *not* `string.h` (which can be hard to spot).

d7919:
Ah, https://github.com/dschwoerer/BOUT-dev/commit/962319d302f78166f34b824d0bc214da97c5ced1 seems to have clobbered the addition of `string.h` from https://github.com/dschwoerer/BOUT-dev/commit/507fa8ec4d84772f2ba9176ef83a0350e47e9054

PR #517, opened 2017-04-18
Important bugfix: Read dz from input

bendudson:
In the move to v4, the code which read "dz" was clobbered.
This broke blob2d example, and probably some others.

### DIFF
--- a/examples/blob2d/blob2d.cxx
+++ b/examples/blob2d/blob2d.cxx
@@ -128,7 +128,7 @@ protected:
     // Density Evolution
     /////////////////////////////////////////////////////////////////////////////
 
-    ddt(n) = -bracket(phi,n,BRACKET_ARAKAWA)    // ExB term
+    ddt(n) = -bracket(phi,n,BRACKET_SIMPLE)    // ExB term
       + 2*DDZ(n)*(rho_s/R_c)               // Curvature term
       + D_n*Delp2(n)
       ;                                         // Diffusion term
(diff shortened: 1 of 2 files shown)


examples/blob2d/blob2d.cxx
```diff
@@ -128,7 +128,7 @@ class Blob2D : public PhysicsModel {
     // Density Evolution
     /////////////////////////////////////////////////////////////////////////////
 
-    ddt(n) = -bracket(phi,n,BRACKET_ARAKAWA)    // ExB term
+    ddt(n) = -bracket(phi,n,BRACKET_SIMPLE)    // ExB term
```
loeiten: Why the change to `BRACKET_SIMPLE`? Would this not cause a blow up of terms?

examples/blob2d/blob2d.cxx
```diff
@@ -143,7 +143,7 @@ class Blob2D : public PhysicsModel {
     // Vorticity evolution
     /////////////////////////////////////////////////////////////////////////////
 
-    ddt(omega) = -bracket(phi,omega, BRACKET_ARAKAWA)                     // ExB term
+    ddt(omega) = -bracket(phi,omega, BRACKET_SIMPLE)                     // ExB term
```
loeiten: The same here.

examples/blob2d/blob2d.cxx
bendudson: Reverted to SIMPLE because that was what was used in v3.1. The ARAKAWA bracket has less dissipation, so results in more noisy solutions which seems less robust than SIMPLE. Since this is one of the easier examples to start with, I thought it better to have it as robust as possible.

PR #519, opened 2017-04-18
Supplements to bout_3to4.py

loeiten:
Added function which takes care of `ng@` -> `LocalN@` replacements, where `@` in {`x`,`y`,`z`}.

**NOTE:** `mesh->ng@` is appearing several places in the BOUT++ code. In these cases, an alias `ng@ = LocalN@` has been used. Thus, running the new `bout_3to4.py` could possibly break code.

Also, found that these replacements are missing, but they are not implemented in this commit as they need a bit of thinking:
* `a^b` -> `pow(a,b)`
* `Field3D.max(bool_or_NULL)` -> `max(Field3D, bool_or_NULL)`

### DIFF
--- a/bin/bout_3to4.py
+++ b/bin/bout_3to4.py
@@ -3,6 +3,7 @@
 import argparse
 import re
 import fileinput
+import sys
 
 nonmembers = {
     'DC': ['DC', 1],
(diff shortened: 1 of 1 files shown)


dschwoerer:
instead of renaming to LocalN@ - should this be changed to localN@? (see #488)


for pow and max, would it make sense to just add a comment before lines containing this, similar to
```
#error use pow(a,b) instead of a^b
```
I think it is better to let the user fix it, than trying to write a script that makes sure it doesn't break any code.

loeiten:
I agree with you that the lower case `l` would make more sense @dschwoerer.
However, `next` currently uses upper case `L`, so I think it would be best to change to lowercase `l`  once it is changed in `next`.

Yes, maybe a warning should be thrown when running the script, and append something (breakable) in the line before the problem for `^` and `.max(`

dschwoerer:
You are right, it first needs to be changed in next ...
Just in case that #488 will be accepted, one should be aware that this needs to adjusted in this script as well.

loeiten:
Come to think about it, maybe `ngz` -> `LocalNz` should throw an error as well as `ngz` used to include one extra point.

`ngz\s-\s1` -> `LocalNz` would be probably be safe, but there could be places where conditions on `ngz` is used, and the script would happily ignore this.

Issue #522, opened 2017-04-18
Communication of vectors gives segfault in BOUT++ 4


loeiten:
Communication of `Vector3D` gives segfault in BOUT++-4

To illustrate the bug, condiser the following MWE

`testVec.hxx`

```cpp
#ifndef __VECTEST_H__
#define __VECTEST_H__

#include <bout/physicsmodel.hxx>
#include <vecops.hxx>         // Gives the vec diff options

class VecTest : public PhysicsModel {
protected:
  int init(bool restarting);
  int rhs(BoutReal t);

public:
  Field3D n;
  Vector3D gradPerpN;
  string ownOpType;
};

#endif
```

`testVec.cxx`

```cpp
#include "vecTest.hxx"

int VecTest::init(bool restarting) {
  TRACE("Halt in VecTest::init");
    SOLVE_FOR(n);
  return 0;
}

int VecTest::rhs(BoutReal t) {
  TRACE("Halt in VecTest::rhs");
  mesh->communicate(n);
  gradPerpN = Grad_perp(n);
  mesh->communicate(gradPerpN);
  ddt(n) = 0;
  return 0;
}

BOUTMAIN(VecTest);
```

`BOUT.inp`

```md
# Root options
###############################################################################
nout = 1         # Number of output timesteps
timestep = 1e-10 # Time between outputs

dump_format="nc" # Write NetCDF format files

# Setting the z coordinate
ZMIN = 0.0
ZMAX = 1.0 # dz = 2*pi(ZMAX - ZMIN)/(MZ - 1)

# Number of guard cells
MXG = 1
MYG = 1
###############################################################################


# Mesh options
###############################################################################
[mesh]
# Number of points
nx = 16
ny = 16
nz = 16
# Puts the boundaries half a step outside the last gridpoints
symmetricGlobalY=true
symmetricGlobalX=true
# Points 0 <= x < ixseps1 are inside separatrix, ixseps1 <= x < ngx outside
# --------Non-periodic----------
ixseps1 = -1
ixseps2 = -1
# ------------------------------
###############################################################################


# Methods options
###############################################################################
# Methods used for the radial (x) derivative terms
[ddx]
first  = C2 # d/dx (f)
second = C2 # d^2/dx^2 (f)
upwind = U1 # u*d/dx (f)

# Methods used for the axial (y) derivative terms
[ddy]
first  = C2 # d/dy (f)
second = C2 # d^2/dy^2 (f)
[...]

PR #524, opened 2017-04-19
add basic test for vector communication

dschwoerer:
This is the test from #522 
I cannot reproduce the issue in next, only in master.


examples/test-vec/testVec.cxx
```diff
@@ -0,0 +1,32 @@
...
+
+class VecTest : public PhysicsModel {
+protected:
+  int init(bool restarting);
+  int rhs(BoutReal t);
+
+public:
+  Field3D n;
+  Vector3D gradPerpN;
+  string ownOpType;
```
loeiten: This line can be removed

PR #527, opened 2017-04-20
check data_dir exists

### LINKED ISSUE #526: Error message is not informative when specifying a non existing directory on command line -d
sub directory heste does not exist. The error message is 
```
yolen@yolen-HP-EliteBook-820-G3:~/numerics/BOUT-HESEL$ mpirun -np 4 ./hesel -d heste
Could not open log file 'heste/BOUT.log.0'
Could not open log file 'heste/BOUT.log.1'
Could not open log file 'heste/BOUT.log.2'
Could not open log file 'heste/BOUT.log.3'
```

dschwoerer:
further add test case
Resolves #526


ZedThree:
Also, it makes absolutely no difference, but for small tests like this, it's probably enough to do something like:

```cpp
#include <bout.hxx>

int main(int argc, char **argv) {
  return BoutInitialise(argc, argv);
}
```

Given how quickly this runs though, it's fine as is.

dschwoerer:
I like adding a test if I fix a bug, although this one is not very thorough (I didn't wanted to enforce a specific string)

ZedThree:
We really should be doing that more -- fix a bug -> add a test.
That way, we know it won't come back!

PR #528, opened 2017-04-20
cleanup test_suite

dschwoerer:
move tests to external text file

If somebody who knows python better has a one liner to read the file ...
I thought comments would be handy, otherwise it would be easier ...


ZedThree:
Something like

```py
tests = []
with open('test_suite_list') as testfile:
  for line in testfile.readlines():
    if not line.lstrip().startswith('#'):
      tests.append(line.strip())
```
The context manager (`with` statement) is important, it automatically closes the file, even if an exception is thrown.

ZedThree:
test-initial failed to build because it uses the python script to generate some of the code, i.e. it won't work through `test_suite_make`.

I think we need some machinery to distinguish levels of tests, but for now, you could just hardcode skipping it in `test_suite_make`.

dschwoerer:
I wasn't aware that `test-initial` also couldn't just call make ...
I had this option, but removed it, as I thought only `test-code-style` would use it ...
I am not sure the `;` is the best character to add in front, if the test should be skipped in the make step ...

ZedThree:
What about `!`? It has a sense that there's something important about that
line.

Could you add some nature of documentation as well please? :) Maybe a small
header to the list of tests or something

PR #531, opened 2017-04-25
make build faster

dschwoerer:
To get old behaviour run `make -C src`

### DIFF
--- a/examples/test_suite_make
+++ b/examples/test_suite_make
@@ -16,7 +16,6 @@ from boututils.run_wrapper import shell
 tests=[]
 with open('test_suite_list') as testfile:
     for test in testfile:
-        print(test)
         test=test.strip()
         if test.startswith("#"):
             pass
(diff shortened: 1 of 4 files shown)


dschwoerer:
Reading log helps fixing errors 👍

dschwoerer:
yes. It shouldn't be needed, ar should be doing that.
I didn't remove it in case some strange ar flags could break it in the future or on some platform ...

from the ar man page:
```
       s   Write an object-file index into the archive, or update an existing one, even if no other change
           is made to the archive.  You may use this modifier flag either with any operation, or alone.
           Running ar s on an archive is equivalent to running ranlib on it.

       S   Do not generate an archive symbol table.  This can speed up building a large library in several
           steps.  The resulting archive can not be used with the linker.  In order to build a symbol
           table, you must omit the S modifier on the last execution of ar, or you must run ranlib on the
           archive.
```

d7919:
Ah OK I see, thanks! I was going to mention that we probably don't need the
ranlib call because of this, but it's a good point that there could be an
odd platform. I can note that adding "S" to arflags and maintaining the
ranlib call is somewhat slower than the current approach.

PR #538, opened 2017-04-26
bout 4to5 script

dschwoerer:
replaces #488 

There is no point in changing it now.
Once we are starting to work on bout 5.0 we can run this script (and finish it before that)


dschwoerer:
Should I merge the script of #888 into this one?

It shouldn't hurt to have the script before it is needed - that way it can easily be run if it is applied and merge conflicts arise (i,e. there will be no need for the cherry-pick from #888 which could cause conflicts.

ZedThree:
That's a good idea. It would be good to develop a `bout_4_to_5` tool, similar to the `bout_3_to_4` one, for mechanical changes like this.

dschwoerer:
Should now be rather safe:
 * It is safely possible to run the script several times
 * Checks git status and warns if not clean
 * Check if bout is already version 5 - otherwise it runs only after asking

dschwoerer:
I have added clang-format in 4f40030 - that way we should get everything clang formatted.
I think it will be a good thing to do it within this push - as such a change will cause significant merge conflicts.
If this script is applied to all branches at a give time, there should be no issues - as the change can be applied independently.
This will allow to add a clang-format test to test-code-style, such that all code will always be nicely formatted in next.

However, I would not mind to remove this commit, as it might make it harder to detect / diagnose when something goes wrong.

dschwoerer:
Should clang-format be removed?
Due to #1017?

Besides that, this is ready for merging ...

dschwoerer:
Can we merge master into next, to fix this?

There are some conflicts, which I am not confident which version we want to keep ...

ZedThree:
Yes, 4.2.2 release coming soon

dschwoerer:
I think it is failing now because the PR is not coming from the boutproject repository.

This is ready to merge :+1:

dschwoerer:
Once more, ready for review / merge ...

bendudson:
Thanks @dschwoerer ! Some things will need watching and updating occasionally I guess, for example the list of header files in the script. This will be handy though when the time comes for v5.

dschwoerer:
Can we merge this?

bendudson:
yes, please do. Thanks!

ZedThree:
One of the tests is still failing:

```
Creating compat headerfiles
mv: cannot stat 'comm_group.hxx': No such file or directory
```

dschwoerer:
While I have updated the list of header file, now cython fails to compile on the updated codebase ... I will investigate ...

dschwoerer:
Now only the code coverage fails :+1: 

The list of header files is still static, but the travis script now checks that the list is up to date, and it is easier to update, as it is a separate file ...

dschwoerer:
Once more, conflicts resolved, and ready for review / merge ...

dschwoerer:
Can we merge this?

bendudson:
Yes, happy to. Does anyone understand why Travis is failing now? I tried restarting the failed tests, but this doesn't seem to have helped, and the log is not very enlightening.

dschwoerer:
Sorry, didn't notice that an actual test was failing.

It seems netcdf.hxx has been added - I have also updated the script to include a more useful error ...

dschwoerer:
Now the tests are passing again - except coverage.

Thanks for noting the issue @bendudson

dschwoerer:
Ready for merging (try no 7 or so)

dschwoerer:
Should this also go into 4.3?

bendudson:
Replaced by #1774

PR #548, opened 2017-05-03
cleanup of tools/pylib

dschwoerer:
* fix permissions
* fix stray #!
* fix whitespace

Is a bug fix, if accepted should also go to master


loeiten:
Why remove the shebang line?

dschwoerer:
They are just libraries, and not executables (they didn't provide a main)
From [SO](http://stackoverflow.com/a/19305076):

> The shebang line in any script determines the script's ability to be executed like an standalone executable without typing python beforehand in the terminal or when double clicking it in a file manager(when configured properly). It isn't necessary but generally put there so when someone sees the file opened in an editor, they immediately know what they're looking at. However, which shebang line you use IS important; Correct usage is:

loeiten:
Ahh, I see. That makes sense :).

Issue #551, opened 2017-05-03
Various test fails


dschwoerer:
I am trying to build BOUT++ on fedora rawhide.
There are several issues:

On `x86_64` only `drift-instability` fails:
```
Making resistive drift instability test
Running drift instability test, zeff =  2
Traceback (most recent call last):
  File "./runtest", line 111, in <module>
    lbNorm = lZeta*((grid['Bpxy'])[0,old_div(ny,2)] / (grid['Bxy'])[0,old_div(ny,2)])  # binormal coord range [cm]
IndexError: only integers, slices (`:`), ellipsis (`...`), numpy.newaxis (`None`) and integer or boolean arrays are valid indices
rm: cannot remove 'data_1/BOUT.dmp.*': No such file or directory
rm: cannot remove 'data_10/BOUT.dmp.*': No such file or directory
```

On `i686` in the `test-initial` the `mixmode` and `mixmode_seed` fail - probably a usage of float instead of double.

Why `arm` and some more exotic architectures fail: not sure.
But the binaries crash in all cases.

Full logs are available here:
https://koji.fedoraproject.org/koji/taskinfo?taskID=19383165

bendudson:
Could this be a problem with old_div? The runtest script tries to import it from past.utils, but if it's not defined it uses
    def old_div(a,b):
      return a / b
https://github.com/boutproject/BOUT-dev/blob/master/examples/drift-instability/runtest#L19

Perhaps this is returning a float rather than an integer? Replacing this with
    def old_div(a,b):
      return int(a / b)
should fix it if this is the problem

bendudson:
The errors on ppc (https://kojipkgs.fedoraproject.org//work/tasks/3167/19383167/build.log) are odd...

in test-field:

    Couldn't get integer from mesh:nx = '6'
    Mesh::get(ival) on line 59 of 'mesh.cxx'

which is reading this line: https://github.com/boutproject/BOUT-dev/blob/master/examples/test-field/data/BOUT.inp#L5

from here:
https://github.com/boutproject/BOUT-dev/blob/master/src/mesh/mesh.cxx#L58

through here:
https://github.com/boutproject/BOUT-dev/blob/master/src/mesh/data/gridfromoptions.cxx#L18

to the parser call here which returns null:
https://github.com/boutproject/BOUT-dev/blob/master/src/sys/options.cxx#L79

which checks a cache here:
[...]

PR #555, opened 2017-05-05
fix test-initial to allow more precise calculation

dschwoerer:
I think this is the last fix needed for #551

### DIFF
--- a/examples/test-initial/runtest
+++ b/examples/test-initial/runtest
@@ -209,8 +209,16 @@ for nproc in nprocs:
                 success_string = "PASS"
             else:
                 if ( var == "mixmode" or var == "mixmode_seed" ) and E2 < 1e-3:
-                    # This can happen due tue excess precision e.g. on X87 architecture
-                    success_string = "WARNING"
+                    import platform
+                    arch=platform.machine()
+                    if arch=='i686':
+                        # This can happen due tue excess precision e.g. on X87 architecture
+                        success_string = "WARNING"
+                    else:
+                        print("This should only happen in i686 with an x87 math architecture.")
+                        print("We detected however an %s architecture."%arch)
+                        success_string = "FAIL"
+                        success = False
                 else:
                     success_string = "FAIL"
                     success = False
--- a/examples/test-initial/runtest
+++ b/examples/test-initial/runtest
@@ -208,8 +208,12 @@ for nproc in nprocs:
             if E2 < tolerance:
                 success_string = "PASS"
             else:
-                success_string = "FAIL"
-                success = False
+                if ( var == "mixmode" or var == "mixmode_seed" ) and E2 < 1e-3:
+                    # This can happen due tue excess precision e.g. on X87 architecture
+                    success_string = "WARNING"
+                else:
+                    success_string = "FAIL"
+                    success = False
             print("\tChecking {var:<12}: l-2: {err:.4e} ... {success}".format(var=var, err=E2,
                                                                       success=success_string))
 


ZedThree:
To be honest, this change makes me a little uncomfortable, as I worry it might cause genuine bugs in BOUT++ to be missed. This is only a problem on one architecture which I'm not certain is used very much any more. I think this might be better solved by noting it as a "known bug" in the test README.md

dschwoerer:
I think this specific test is anyway not well designed.
For example, python does not guarantee a specific precision for floats. Therefore assuming that any python implementation has to use IEEE 754 seems not necessary.
What I think would be way better, is a test that makes sure that e.g. the BOUT PRNG actually produces uniform random numbers, and that the modes are on average initialized as they should. This however would mean a far more complicated, stochastic test.

The current test just makes sure that all the bugs from the C++ code where reproduced in python.
Sure, it is better then the old test, but this is imho not a test that convinces me that the code does what it is supposed to do ...

bendudson:
I think this is not going to be merged for now, since we're uneasy about making accuracy exceptions for particular tests. Perhaps just document that this failure is expected on some architectures

dschwoerer:
I would prefer getting the exception.
Having it documented is no use, if I want to run the check in the build process, as then I have to ignore the results (and thereby disable all checks) or not generate binaries for i686.

Further, I don't think the PRNG is decently tested ...

dschwoerer:
This is a histogram of the random output from using the python implementation of the bout's PRNG.
If we would care about this code - this should be fixed first.
bout_prng
This is the most basic test of the PRNG I could think of ... iterating a logistici map isn't a good idea if you want a good prng ...

So back to this PR:
The test is anyway not good, so allowing for it to not fail on higher precision machines shouldn't be our main issue ...

bendudson:
Thanks for the testing. Yes, this does suck as a PRNG. Cryptographic use not recommended, but I put it in because it's simple and (I thought) reproducible. Do we need a better PRNG? I would prefer not to use a standard library because I don't think there's a way to make sure it will be the same in 5 year's time on a different system.

bendudson:
Presumably a PRNG which uses integers and then at the end generates a float should be more reproducible and less sensitive to floating point errors...?

dschwoerer:
Something like this maybe?
Thats the bad one from C++ [1](http://www.cplusplus.com/reference/random/minstd_rand0/)
```
typedef unsigned int uint;
struct Ran{
  uint x;
  Ran(uint x_):x(x_){};
  uint unt(){
    x = x * 16807 % 2147483647;
    //printf("%d ",x);                                                                                                                                                            
    return x;
  }
  double flt(){
    return unt()*4.65661287524580e-10;
  }
};
```
It is better than what we have right know. It is not a good one - but should do the job.
The issue with C++ std <random> suite seems to be, that the way the random numbers are converted to double/normal/... is not well defined [2](https://stackoverflow.com/a/35449608/3384414)

I think we would also be fine with std::mt19937 and provide our own code to create uniform/normal[3](https://en.wikipedia.org/wiki/Box%E2%80%93Muller_transform#Implementation)/...[4](http://www.netlib.org/random/index.html)

d7919:
There seem to be several possible issues here, but as I understand it the primary issue is that `test-initial` fails on i686 and that this means automated build systems that include the checks don't like it and abort (or similar). 

Would a suitable workaround for now be to simply add some code into the specific `runtest` that just skips the `mixmode` test when on i686 (optionally with a warning)? That way we're not weakening the test for any other system but still allowing progress on i686. We can document the reason for the skip in the `runtest` and remove the skip should we address the somewhat separate issue (see #603 for example) of updating the `mixmode` modifier as several improvements are desirable but it is likely to break reproducibility.

PR #556, opened 2017-05-05
Next merge

dschwoerer:
make sure all bug fixes from muster end up in next


dschwoerer:
This is far more than I expected 😁 
I just wanted to make sure 7ac550f ends up in next 🐀

Seems one bug was fixed in two slightly different ways ... git added the change then twice.

Would it make sense to submit bug fixes only to master, and once they are accepted merge into next?
* no cherry-picking required
* master gets patched
* no divergence between next and master

bendudson:
Not sure about this pull request... seems to pull in a lot of changes. I agree though that a bugfix in master followed by a merge into next is probably cleaner than cherrypicking.

dschwoerer:
The changes are mainly:
* License
* bout runners
* documentation
* bug fix:
  * drift-instabilty
  * `fval.pp` -> `fval.mm`
  * `D2DX2_C4_stag` -> `D2DX2_C2_stag`

The bug fixes all seem to be merged correctly, the documentation update are changes from `ngx` to `localNx`.
I am not sure about bout runners, but cannot see why they shouldn't be updated in next, but I assume @loeiten is more qualified to comment on that.

loeiten:
Regarding `bout_runners`: There should be no problem to merge this into `next`.
I also agree that bugfix in master followed by a merge into next seems like a good idea :).

dschwoerer:
I have reviewed this PR, but as I opened it, I cannot approve it.
I still think this is should be merged, to avoid we need a great merge, as we had it before the 4.0 release.
Small changes are easier to verify.

dschwoerer:
This will not get easier...
I think we should try our best to prevent master and next from diverging.

Issue #540, opened 2017-04-27
FieldGroup adding const fields


bendudson:
Passing const fields to FieldGroup constructor results in segfault. 

    Field3D a(0.0);
    FieldGroup g(a);  // ok
    FieldGroup h(2*a); // compiles, but segfault

This then also affects mesh->communicate(). It's ok that this doesn't work; communicate shouldn't be able to modify const or temporary fields, but a more useful error message than "segfault" would be good. A stack trace produces:

...
#223592 FieldGroup<Field3D const> (this=0x7fffffffc6d0)
    at ../..//include/bout/fieldgroup.hxx:53
#223593 FieldGroup<Field3D const> (this=0x7fffffffc6a0)
    at ../..//include/bout/fieldgroup.hxx:53
#223594 FieldGroup<Field3D const> (this=0x7fffffffc670)
    at ../..//include/bout/fieldgroup.hxx:53
#223595 FieldGroup<Field3D const> (this=0x7fffffffc640)
    at ../..//include/bout/fieldgroup.hxx:53
#223596 FieldGroup<Field3D const> (this=0x7fffffff0250)
    at ../..//include/bout/fieldgroup.hxx:53
#223597 physics_run (t=<optimized out>) at elm_6f.cxx:1381

so it looks like the constructor is calling itself until it overruns the stack. Some way to catch const fields being added to FieldGroup maybe?

ZedThree:
We could use `std::is_const` to check for `const`ness and then `throw` an exception?

dschwoerer:
This does not fix the issue:
```
int i;
FieldGroup f(i);
```

We should add a template that requires at least 2 arguments, passes the first to a real implemntation, and calls with the remaining ones ether itself or implemntation for that:
```
template <typename T, typename... Args>
FieldGroup (T a, Args... args){
  return FieldGroup(a)+FieldGroup(args);
}
```

bendudson:
Another possibility is to add:

template<typename T>
void add(T &v) = delete;
  
which results in the error message:

error: use of deleted function ‘void FieldGroup::add(T&) [with T = const Field3D]’
     add(ts...);
     ^
../..//include/bout/fieldgroup.hxx:112:8: error: declared here
   void add(T &v) = delete

It does also mean changing void add(FieldData &f)  to void add(Field2D &f), so only Field2D, Field3D, Vector2D and Vector3D types can be added.

PR #557, opened 2017-05-05
FieldGroup catch incorrect types at compile time

bendudson:
If FieldGroup was constructed using a type which could not be added, such as an integer or const Field3D, then an infinite loop would result. This is because the call to add() inside the constructor would try to implicitly convert the type to FieldGroup, using the variadic template constructor.

This fixes issue #540 by making the template constructor explicit. I tried some other ways to fix this, but this seemed the most straightforward.

The infinite loop no longer occurs since the call to add() inside the constructor fails to compile. Instead an error message like this is printed:

    ../../include/bout/fieldgroup.hxx: In instantiation of ‘FieldGroup::FieldGroup(Ts& ...) [with Ts = {int}]’:
    test_fail.cxx:14:17:   required from here
    ../../include/bout/fieldgroup.hxx:57:36: error: no matching function for call to ‘FieldGroup::add(int&)’
        explicit FieldGroup(Ts&... ts) { add(ts...); }

A test case has also been added (examples/test-fieldgroup) which tests for this issue and a couple of other sanity checks.

### DIFF
--- a/examples/test-fieldgroup/runtest
+++ b/examples/test-fieldgroup/runtest
@@ -1,0 +1,50 @@
+#!/bin/bash
+#
+# This scipt is intended to be quite quiet when it succeeds,
+# printing only the test name and "Passed". 
+# If the test fails then the whole log is printed, hopefully
+# helping diagnose where the problem lies.
+#
+
+echo ""
+echo "Running FieldGroup test"
+
+# Ensure we clean old run, in case it succeeded last time
+make -f makefile_fail clean
+
+# This test should fail to compile
+make -f makefile_fail > run.log  2>&1
+
+if [ $? -eq 0 ]; then
+  # Compiled ok, something wrong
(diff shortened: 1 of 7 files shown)


examples/test-fieldgroup/runtest
```diff
@@ -0,0 +1,50 @@
...
+  echo "Fail. Log follows:"
+  echo ""
+  cat run.log
+  exit 1;
+fi
+
+# Second test should pass
+
+make clean
+make >> run.log 2>&1
```
dschwoerer: Not sure why the forced rebuild is needed
I have disabled the build in test_suite_build to prevent it from being build twice

d7919: I think this test is actually a test that the model can/can't compile in certain conditions, hence we have to make sure that `make` is going to try to rebuild the target when we run the test.

Issue #484, opened 2017-03-07
y initial conditions always at CELL_YLOW


JosephThomasParker:
Initial conditions for y seem to be evaluated at CELL_YLOW, even when a variable is located at CELL_CENTER.

I modified the test-staggered example ![here](https://github.com/JosephThomasParker/BOUT-dev/tree/staggered-bc/examples/test-staggered).  In this is test, density is located at the cell centre and has the initial condition `1+0.01*cos(y)`.  Velocity is located at YLOW and has ICs `1+0.01*sin(y)`.  

To plot the ICs actually used, do
`./runtest`
`cd test`
`python plot_staggered_ic.py`

Velocity (green) is correct, but the density (blue) should be offset by half a grid spacing, so that it would be 1.01 at the boundary and 1+0.01*cos(dy/2) at the first grid point. 

![figure_1](https://cloud.githubusercontent.com/assets/15211161/23663039/5145537c-0349-11e7-96ee-fd427acf1af8.png)

bendudson:
Ah, it might need to have mesh:symmetricGlobalY = true set, in order to define the Y coordinate properly. This is what is done in the MMS tests:
https://github.com/boutproject/BOUT-dev/blob/master/examples/MMS/wave-1d-y/data/BOUT.inp#L32

The default is currently false, since changing would "break" some tests
https://github.com/boutproject/BOUT-dev/blob/master/src/mesh/impls/bout/boutmesh.cxx#L60

This modifies how the global Y index is calculated here:
https://github.com/boutproject/BOUT-dev/blob/master/src/mesh/impls/bout/boutmesh.cxx#L2256

then if the cell location is YLOW the index is interpolated to the cell boundary:
https://github.com/boutproject/BOUT-dev/blob/master/src/field/field_factory.cxx#L133

whilst if it's CENTRE then the index at cell centre is used:
https://github.com/boutproject/BOUT-dev/blob/master/src/field/field_factory.cxx#L144

JosephThomasParker:
Thanks Ben!  With `mesh:symmetricGlobalY = true` it works as expected (though the lower BC for the staggered grid must be put in manually, as is the point of the example).

![figure_2](https://cloud.githubusercontent.com/assets/15211161/23666471/06c4e11e-0353-11e7-97a5-861a5f16f09b.png)

bendudson:
We probably shouldn't depend on a magic "make this work properly now" switch. 
[...]

PR #561, opened 2017-05-09
Changing mesh:symmetricGlobalY to true by default

bendudson:
This is the "correct" setting, which corresponds to setting y to 0 and 2pi on the boundaries in the input expressions.

Some test cases depend on the original setting, so now set symmetricGlobalY to false in the BOUT.inp settings.

This changes user inputs, but if anyone has been doing testing (particularly MMS), they will already have been setting symmetricGlobalY to true (or getting frustrated trying to work out why their test
doesn't converge properly).

This fixes issue #484 since the other part of the issue concerning FieldFactory has already been fixed

### DIFF
--- a/examples/test-fieldfactory/data/BOUT.inp
+++ b/examples/test-fieldfactory/data/BOUT.inp
@@ -13,6 +13,7 @@ dump_format = "nc"  # NetCDF format. Alternative is "pdb"
 
 [mesh]
 symmetricGlobalX = false
+symmetricGlobalY = false
 
 [output]
 floats=true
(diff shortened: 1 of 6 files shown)


dschwoerer:
Would it make sense to at least give a warning?
```
if (! options->isset("symmetricGlobalY")){
  output << "WARNING: The default of this option has changed in release 4.1.\nIf you want the old setting, you have to specify mes:symmetricGlobalY=false in BOUT.inp\n"
}

bendudson:
Yes, a warning is a good idea

PR #562, opened 2017-05-12
Modify the IDL_PATH suggested by configure

bendudson:
In order to work correctly, IDL_PATH must retain the <IDL_DEFAULT>
token. This is discussed here: https://www.harrisgeospatial.com/docs/prefs_directory.html

Issue found and fixed by Haruki Seto


dschwoerer:
shouldn't this go to master, and then be merged into next?

Issue #449, opened 2017-01-31
User defined flush frequency


d7919:
It can be helpful to be able to change the frequency with which the data gets flushed to file (e.g. if each output step only takes a very short time the data flushing to disk can be a significant percentage of the step, similarly if a step is very slow we might want to flush each step to ensure we have the up to date data available). 

I made a branch a while ago (next_Feature_UserDefinedFlushFreq) against the pre v4 next branch that implemented this. This will need some updating for v4 but might be fairly simple to get in.

d7919:
Closed in next by #568

PR #568, opened 2017-06-01
Next feature user defined flush freq

d7919:
Addresses #449 to add user control over the frequency of datafile flushing to disk.

This is for use with openclose=true. Rather than opening and closing the
datafile in every call to write, only do it every writeFlushFreq calls
to write. This defaults to 1 in order to give the standard
behaviour. Note this will not be useful for many runs but can help when
you would like high time resolution output (e.g. when I/O is taking a
significant amount of each output step).

Note you are likely to get jumps in the solver time every writeFlushFreq
outputs (it looks like this write time is actually ending up being
attributed to solver rather than I/O).


src/fileio/datafile.cxx
```diff
@@ -61,12 +62,17 @@ Datafile::Datafile(Options *opt) : parallel(false), flush(true), guards(true), f
   OPTION(opt, enabled, true);
   OPTION(opt, init_missing, false); // Initialise missing variables?
   OPTION(opt, shiftOutput, false); //Do we want to write 3D fields in shifted space?
+
+  writeFlushCounter = 0;
```
d7919: Good point!

src/fileio/datafile.cxx
```diff
@@ -61,12 +62,17 @@ Datafile::Datafile(Options *opt) : parallel(false), flush(true), guards(true), f
...
+  OPTION(opt, writeFlushFreq, 1); //How frequently do we flush the file
```
d7919: This was to match the style of the other options in this section, where the default is given again (e.g. `enabled`), but happy to change if preferred.

src/fileio/datafile.cxx
```diff
@@ -98,6 +105,8 @@ Datafile& Datafile::operator=(Datafile &&rhs) {
   enabled      = rhs.enabled;
   init_missing = rhs.init_missing;
   shiftOutput  = rhs.shiftOutput;
+  writeFlushCounter = rhs.writeFlushCounter;
```
d7919: Should I initialise to 0 here instead then?

include/datafile.hxx
```diff
@@ -86,6 +86,8 @@ class Datafile {
   bool enabled;  // Enable / Disable writing
   bool init_missing; // Initialise missing variables?
   bool shiftOutput; //Do we want to write out in shifted space?
+  int writeFlushCounter; //Counter used in determining when next openclose required
+  int writeFlushFreq; //How many write calls do we want between openclose
```
d7919: Yes that sounds sensible.

src/fileio/datafile.cxx
dschwoerer: Yes, I think that should make sure the file always gets opened.

I am not sure whether we should make sure the file gets closed in the destructor. It seems to me this is currently not the case for files that are not openclose - did I miss something?

src/fileio/datafile.cxx
dschwoerer: I don't like having defaults duplicated, as it makes it harder to change them, but keeping the current style is also a valid point.

dschwoerer:
I am still not quite sure the file should be closed, in the de-constructor, but I assume the file gets closed if it goes out of scope - at least that's the case for netcdf ...

PR #572, opened 2017-06-07
Small bug fix: Ensure msg_stack.push call inside directive

d7919:
Important to match the msg_stack.pop call to ensure messages added are
removed in the same setup. 

There was a mismatch in `smoothing : averageY` which meant `msg_stack.push` was being called for any `CHECK` value but `msg_stack.pop` was only being called if `CHECK` was defined.


d7919:
May be better fixed by replacing `msg_stack` usage with `TRACE` macro (as in b419cde459d9b47daf918248943d9fa5dd42868b)

d7919:
Should be fixed by #575

Issue #569, opened 2017-06-02
Consistency of CHECK usage


d7919:
In most locations in the code we check if the value of `CHECK` is greater than a certain value to decide if we do that relevant test, however in a few places we just have `#ifdef CHECK`. This may be slightly confusing for some as it means even if `CHECK` is set to 0 (or -1 etc.) then we will still have some checks. 

Of course whilst we'd hope the users are using the configure option `--enable-checks=no` to disable checks, it's still common for people to edit make.config manually, which could lead to people setting `-DCHECK=0`.

Should we consider ensuring that `CHECK` is always used in a comparison operation (and if not defined we give it a default value)?

d7919:
A related point to note is that using `--enable-checks=4`, `--enable-checks=-100` or  `--enable-checks=off` etc. in the configure command we will end up with `-DCHECK=2` in all cases. Is it possible to catch invalid values and abort with a helpful message? Could/should values >3 be capped to 3 (guessing someone setting to a large number wants most rigorous checking)?

bendudson:
I think I would prefer aborting with a helpful message over capping values. Does the compiler treat a missing definition as 0? Would replacing all ifdefs with if CHECK > 0 be sufficient?

d7919:
I think the C standard says that an undefined identifier will be replaced by a 0, so switching ifdef to CHECK>0 should be sufficient ([see here](https://stackoverflow.com/a/13149212))

I agree that aborting with a helpful message would probably be the best option.

d7919:
fbad805 should provide a "fix" for the main "issue".

Note a large number of these checks are actually associated with `msg_stack.push/pop` calls which can probably be replaced with the `TRACE` macro.

d7919:
Helped by #575 which removes a number of `#ifdef CHECK` blocks associated with `msg_stack` calls.

d7919:
Closed by #577

PR #574, opened 2017-06-08
Let trace do formatting

### LINKED ISSUE #573: Expanding the capability of `TRACE`
I'm currently looking at replacing the `msg_stack.push/pop` calls with use of the `TRACE` macro (see b419cde459d9b47daf918248943d9fa5dd42868b). One thing that `TRACE` can't currently do is to deal with formatting strings (i.e. can only pass a string with no expansions).

With a bit of magic (or gcc specific extension), it's possible to allow the `TRACE`  macro to take additional arguments that can be used for filling in values in the passed message. This is provided at ef912322a30429e21476fae79e91a45e7878d89f . This perhaps isn't the nicest code ever, but does provide the requested feature. 

If anyone has any thoughts/suggestions/alternatives I'd be keen to hear them!

d7919:
Allows the TRACE macro to take additional arguments used for formatting.

In other words `TRACE("Some string %d",2)` would lead to a `msg_stack`
entry "Some string 2 on line __LINE__ of '__FILE__'". This means `TRACE`
has the full capability of `msg_stack.push`.

Two original constructors kept currently but are not used anywhere in
the BOUT++ code base so should perhaps be removed. If kept for long term
then may want to consider reordering their arguments as well for consistency.

Closes #573

### DIFF
--- a/include/msg_stack.hxx
+++ b/include/msg_stack.hxx
@@ -30,6 +30,7 @@ class MsgStack;
 #define __MSG_STACK_H__
 
 #include <stdio.h>
+#include <stdarg.h>
 #include <string>
 
 /// The maximum length (in chars) of messages, not including terminating '0'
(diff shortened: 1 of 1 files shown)


d7919:
Also helps improve #569 as a lot of `#ifdef CHECK` uses are around `msg_stack` uses.

PR #577, opened 2017-06-12
Replacing all `#ifdef CHECK` lines with `#if CHECK > 0`

### LINKED ISSUE #569: Consistency of CHECK usage

d7919:
Fixes #569

Note this will change the behaviour of builds with `-DCHECK` and `-DCHECK=0` (the
first is treated as the second). This is now currently equivalent to no
checks.

May wish to treat some of these checks as essential, can now be explicit
with these by setting line to `#if CHECK >= 0` or equivalent.


dschwoerer:
I am not sure about the rst files in the documentation folder ...
Do they need to be updated as well?
They seem to have in part the same content as the tex files...

bendudson:
Yes please, I think the rst files are now the main documentation and are what goes on readthedocs. I think the tex files should probably be removed now, to avoid duplicating things.

d7919:
Thanks for updating the documentation!

PR #592, opened 2017-06-30
Laplacian maxmode fix

### LINKED ISSUE #589: Laplacian maxmode limits
Looking through some of the Laplacian solvers I noticed a few possible issues/inefficiencies.

* Shoot_laplace seems to use loop limits that are up to but not including `maxmode`, whilst other solvers are up to and including `maxmode`.
* The following impls solve for all mode numbers, not just those `<=maxmode`; serial_tri, serial_band

d7919:
Closes #589 

* Ensure serial Laplacian methods skip modes greater than specified `maxmode`.
* Disallow use of `type=shoot` for standard Laplacian by throwing in the constructor.

### DIFF
--- a/src/invert/laplace/impls/serial_band/serial_band.cxx
+++ b/src/invert/laplace/impls/serial_band/serial_band.cxx
@@ -48,9 +48,23 @@ LaplaceSerialBand::LaplaceSerialBand(Options *opt) : Laplacian(opt), Acoef(0.0),
   int ncz = mesh->LocalNz;
   bk = matrix<dcomplex>(mesh->LocalNx, ncz/2 + 1);
   bk1d = new dcomplex[mesh->LocalNx];
+
+  //Initialise bk to 0 as we only visit 0<= kz <= maxmode in solve
+  for(int kz=maxmode+1; kz < ncz/2 + 1; kz++){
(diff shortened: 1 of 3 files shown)


src/invert/laplace/impls/serial_band/serial_band.cxx
```diff
@@ -106,21 +106,19 @@ const FieldPerp LaplaceSerialBand::solve(const FieldPerp &b, const FieldPerp &x0
     xend = mesh->LocalNx-2;
   }
 
-  for(int iz=0;iz<=ncz/2;iz++) {
+  for(int iz=0;iz<=maxmode;iz++) {
```
bendudson: Is the value of the mode the result (xk1d?) set to zero anywhere for iz>maxmode?

d7919: That's a good point, it's not (not in `serial_tri` either) -- presumably something like

~~~~~
  for(int kz=maxmode+1; kz < ncz/2 + 1; kz++){
    for (int ix=0; ix<=ncx; ix++){
      xk[ix][kz] = 0.0;
    }
  }
~~~~~

after this loop would be an appropriate fix? 

Alternatively, we could do the initialisation at the point where we allocate the storage (i.e. in the constructor) such that this is only done once.

PR #595, opened 2017-07-04
fix typos in bout-config

dschwoerer:
`CONTRIBUTION.md` hints that the use of Fortran is not directly preferred (i.e. to be never used).
`bin/bout-config` still provides a fortran mpi compiler :D

### DIFF
--- a/bin/bout-config.in
+++ b/bin/bout-config.in
@@ -9,7 +9,6 @@
 cc="@MPICXX@"
 cxx="@MPICXX@"
 ld="@MPICXX@"
-fc="none"
 checks="@CHECK_LEVEL@"
 cflags="@CONFIG_CFLAGS@"
 libs="@CONFIG_LDFLAGS@"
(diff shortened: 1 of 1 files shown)


d7919:
Whilst it's requested that there's no Fortran in BOUT++ I think this doesn't mean a Fortran project can't use the BOUT++ library (with appropriate wrapping), so the Fortran compiler flag could perhaps be useful for people who are trying to do this -- although I may be wrong!

bendudson:
Sadly Fortran is sometimes needed, for example coupling to EIRENE. Whilst no Fortran should go in the core code, coupling to external libraries may be necessary.

dschwoerer:
I am quite sure nobody used this.
Otherwise `--fc` would not return an error that it is an unknown option.

Further, $fc is set to none. Rather sure, you cannot compile EIRENE with this 👍

If I implement bout's configure to look for a fortran compiler, I can add this again, also providing apropriate flags ...

PR #598, opened 2017-07-04
add fieldmesh and make derivatives default to fieldmesh

### LINKED ISSUE #596: Derivatives - why global mesh?
Should derivatives not default to there own mesh to take derivatives?

Currently:
https://github.com/boutproject/BOUT-dev/blob/master/src/sys/derivs.cxx#L87

I think something like:
```
const Field3D DDY(const Field3D &f, CELL_LOC outloc, DIFF_METHOD method) {
  return f.getMesh()->indexDDY(f,outloc, method) / f.getMesh()->coordinates()->dy;
}
```
and
```
Mesh * Field3D::getMesh(){
  if (fieldmesh){
    return fieldmesh;
  } else {
    return mesh;
  }
}
```
I hope this helps allowing several meshes 👍

dschwoerer:
fixes #596

### DIFF
--- a/include/field2d.hxx
+++ b/include/field2d.hxx
@@ -100,11 +100,11 @@ class Field2D : public Field, public FieldData {
    */
   int getNx() const override {return nx;};
   /*!
-   * Return the number of nx points
+   * Return the number of ny points
    */
   int getNy() const override {return ny;};
   /*!
-   * Return the number of nx points
+   * Return the number of nz points
    */
   int getNz() const override {return 1;};
 
--- a/include/field2d.hxx
+++ b/include/field2d.hxx
@@ -98,15 +98,15 @@ class Field2D : public Field, public FieldData {
   /*!
    * Return the number of nx points
    */
-  int getNx() const {return nx;};
+  int getNx() const override {return nx;};
   /*!
    * Return the number of nx points
    */
-  int getNy() const {return ny;};
+  int getNy() const override {return ny;};
   /*!
    * Return the number of nx points
    */
-  int getNz() const {return 1;};
+  int getNz() const override {return 1;};
 
   // Operators
 
--- a/include/field3d.hxx
+++ b/include/field3d.hxx
@@ -212,15 +212,15 @@ class Field3D : public Field, public FieldData {
   /*!
    * Return the number of nx points
    */
-  int getNx() const {return nx;};
+  int getNx() const override {return nx;};
   /*!
    * Return the number of nx points
    */
-  int getNy() const {return ny;};
+  int getNy() const override {return ny;};
   /*!
    * Return the number of nx points
    */
-  int getNz() const {return nz;};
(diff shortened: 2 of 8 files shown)


include/field2d.hxx
```diff
@@ -98,15 +98,15 @@ class Field2D : public Field, public FieldData {
...
    */
-  int getNx() const {return nx;};
+  int getNx() const override {return nx;};
   /*!
    * Return the number of nx points
    */
-  int getNy() const {return ny;};
+  int getNy() const override {return ny;};
   /*!
    * Return the number of nx points
```
d7919: Could you fix the comments whilst you're here?

include/field3d.hxx
```diff
@@ -212,15 +212,15 @@ class Field3D : public Field, public FieldData {
...
    */
-  int getNx() const {return nx;};
+  int getNx() const override {return nx;};
   /*!
    * Return the number of nx points
    */
-  int getNy() const {return ny;};
+  int getNy() const override {return ny;};
   /*!
    * Return the number of nx points
```
d7919: Could you fix the comments whilst you're here?

d7919:
Do some operators, like `vddx`, require some extra checking to make sure that both fields involved are on compatible meshes?

bendudson:
Yes, probably best to add ASSERT statements. Since it should have very little overhead, I think the checks should be done all the time (ASSERT0), though ASSERT1 may be ok.

d7919:
I guess we need to define what we mean by two compatible meshes -- do they have to be exactly the same mesh, just have the same number of points in all/some directions, be convertible with interpolation or similar. For now I guess we could just force it to be the same mesh as the "safest" bet.

bendudson:
I think ensuring that they are the exact same mesh is the best and safest thing. If (oh glorious day) we ever get to the point of having multiple meshes, it should probably be the physics model author who explicitly converts from one mesh to another.

dschwoerer:
Now all derivatives are checked in the mesh.
If another mesh is implemented, that removes this restriction, it can easily remove the check 👍
Also now, results should be created with the right mesh.

ZedThree:
I think a similar thing to the derivatives needs to be done in `F3D_OP_REAL` etc., as currently the result will use the global mesh. Basically, anywhere there is a temporary `Field3D` created to store a result, it needs to use the same mesh as the input arguments.

dschwoerer:
Yes, I didn't wanted to fix everything in BOUT in one go - but rather start somewhere ...

On the specific issue with F3D_OP_REAL: that would be easier to fix with PR #458 :)
#458 checks already whether both fields are at the same position ...

Besides that: I would prefer open an issue for every case you find where it is not fixed (if you don't want to fix it your self)

This PR shouldn't cause any issues that haven't been there before.

bendudson:
Fixing everything in BOUT(++) may be a bit too much to ask, and probably impossible.

Issue #600, opened 2017-07-05
BOUT overwrites input file with -f BOUT.settings


dschwoerer:
Bout generetes on exit the file BOUT.settings.
Nothing is preventing me from using this as new input file.
Is this intended behaviour that this is overwritten on exit?

bendudson:
Yes. It's written at the start and again at the end. The settings don't (or shouldn't) change, but the comments do: it's only known at the end whether a setting is used.  The write at the start is in case the run is stopped or crashes before the end

dschwoerer:
My question was rather whether we are fine to happily overwrite the __input__ file?

I think there should be some protection for the user - just in case the disc is full, or for some reason the code crashes while writing, so that the user does always have an input file.

And the options might very well change, if I supply other or new options, `BOUT.settings` will have changed.

bendudson:
Ah ok... I suppose someone could try to do that. I had assumed people would rename BOUT.settings -> BOUT.inp. We could prevent the user from setting the input file to BOUT.settings, but they could make a soft or hard link. 

Perhaps we could use a different name for the settings file written at the start and end, but I think it's difficult to prevent a user from potentially overwriting their input if they really try. Maybe just a warning in the manual?

dschwoerer:
fixed by #601

PR #601, opened 2017-07-06
make sure input file is not overwritten

dschwoerer:
Not sure this PR is welcome, but I think we shouldn't overwrite the input file.
This PR allows to use `BOUT.settings` as input file, by allowing to specify a settings file.

related to question #600


bendudson:
Yes, I think this is useful. It doesn't prevent someone linking BOUT.inp and BOUT.settings, but we can't catch everything

PR #605, opened 2017-07-10
Scorep wrapper

d7919:
Adds a configure flag to turn on `scorep` profiling support and some helper macros. This has two main effects:

1. Update the compiler command (`MPICXX`) to use `scorep`  with the options `--user` and `--nocompiler`. For example if initially `MPICXX=mpicxx` then when configured `--with-scorep` this will become `MPICXX=scorep --user --nocompiler mpicxx`. The option `--nocompiler` disables automatic instrumentation of every routine (found to cause too much overhead due to large number of calls to small routines). The `--user` option enables user defined instrumentation. By default there are no user defined regions so the profiled code will only typically record time in `MPI` routines.
2. Provides macros of the form `SCOREP0, SCOREP1` etc.  which can be placed at the start of routines the user wishes to instrument. The number signifies the "level", with `SCOREPLVL` defining which levels are to be included. This follows the same approach as the assert macros.

The variable `__func__` is part of the `c++11` standard and gives the current routine name. This is used within the macros to ensure that the profiler records an appropriate name for each instrumented region. However, some compilers provide a `__PRETTY_FUNCTION__` variable which provides the entire function signature. This is used if available instead of `__func__`  as it provides more information. A test for this support is a part of the configure stage.

Currently there are no uses of `SCOREPN` in any of the source code, and hence configuring with `--with-scorep` does not provide any timing information. It's possible that we may want to include some of these macros in the committed code at some point (and also possible that we don't). But this may lead to more clutter than desirable so we might want to think about having a `BOUT_FUNCTION_HEADER` macro that includes things like the `SCOREPN` macro, the `TRACE` macro and any other common items (alternatively we could imagine including the `SCOREPN` macro inside the `TRACE` macro).

### DIFF
--- a/include/bout/scorepwrapper.hxx
+++ b/include/bout/scorepwrapper.hxx
@@ -13,10 +13,9 @@
 //of the standard. The __func__ variable *is* a part of the c++11 standard so we'd like to fall back
 //to this if possible. However as these are variables/constants and not macros we can't just
 //check if __PRETTY_FUNCITON__ is defined or not. Instead we need to say if we support this
-//or not by defining __HAS_PRETTY_FUNCTION (to be implemented in configure)
-//#define __HAS_PRETTY_FUNCTION
+//or not by defining HAS_PRETTY_FUNCTION (to be implemented in configure)
 
-#ifdef __HAS_PRETTY_FUNCTION
+#ifdef HAS_PRETTY_FUNCTION
 #define __thefunc__ __PRETTY_FUNCTION__ 
 #else
 #define __thefunc__ __func__
--- a/include/bout/scorepwrapper.hxx
+++ b/include/bout/scorepwrapper.hxx
@@ -1,0 +1,59 @@
+#ifndef __BOUT_SCOREP_H__
+#define __BOUT_SCOREP_H__
+
+#ifdef BOUT_HAS_SCOREP
+#include <scorep/SCOREP_User.h>
+#endif
+
+#ifndef SCOREPLVL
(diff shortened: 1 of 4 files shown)


d7919:
Not sure if this is the sort of thing that we want to include with the main code, but it is quite helpful for those wishing to profile with `scorep` (/`scalasca`).

include/bout/scorepwrapper.hxx
```diff
@@ -0,0 +1,58 @@
...
+#define SCOREPLVL 0
+#endif
+
+//The __PRETTY_FUNCTION__ variable is defined by GCC (and some other families) but is not a part 
+//of the standard. The __func__ variable *is* a part of the c++11 standard so we'd like to fall back
+//to this if possible. However as these are variables/constants and not macros we can't just
+//check if __PRETTY_FUNCITON__ is defined or not. Instead we need to say if we support this
+//or not by defining HAS_PRETTY_FUNCTION (to be implemented in configure)
+
+#ifdef HAS_PRETTY_FUNCTION
```
bendudson: Can this just check #ifdef __PRETTY_FUNCTION__  rather than checking in configure?

ZedThree: Unfortunately not, because `__PRETTY_FUNCTION__` (and `__func__`) are actually variables and not preprocessor macros

d7919: Yes can confirm what @ZedThree says about this -- slightly painful but the configure check is relatively easy.

PR #606, opened 2017-07-11
Warnings squash

bendudson:
Three small(ish) changes:

1. Adding override to Field, Vector classes

2. Modifying Array so it's all in the header. This involves moving the static data stores into functions. Possible performance change, not yet tested.

3. Changing the default ParderivFactory function, so it reads the type option. This should only have good consequences...


bendudson:
Testing the performance impact of the changes to Array: I compared the examples/performance/arithmetic output between next and this branch. Best of 5 timings:

next branch:
Fields: 0.00703659
C loop: 0.00140747
Templates: 0.0035166
Range For: 0.00156046

this branch:
Fields: 0.00710322
C loop: 0.00138715
Templates: 0.00348272
Range For: 0.00156449

I don't think there's much in it...

PR #609, opened 2017-07-14
Field3d shift fix and const auto loops, Field2D/3D consistency

d7919:
A collection of small commits that provide more uniform treatment between 2D and 3D fields and provides one small fix to Field3D and one to Field2D.

1. Fix `Field3D` non-member `shiftZ` which had a potentially infinite loop.
2. `Field2D::applyBoundary` would either silently ignore attempts to apply a boundary to an unallocated field, or would print a warning and then just exit routine depending on CHECK. This now includes an ASSERT (as done for Field3D) to ensure we don't ignore such mistakes.
3. Make all range based loops in `Field3D` be const reference form.
4. A number of small changes to ensure that the same checks are being done in Field3D and Field2D where appropriate. In some places checks where only done on one field type and in other places they were done at different levels in different places. This PR tries to unify the treatment but as a consequence does change the checking behaviour slightly (i.e. for a fixed `CHECK` may get different number of exceptions now).
5. Update Field2D to use `Field::fieldmesh` as the relevant mesh, to improve consistency with `Field3D` (related to #598). Also update some uses of global `mesh` to now use `fieldmesh` in 2D and 3D.

### DIFF
--- a/src/field/field2d.cxx
+++ b/src/field/field2d.cxx
@@ -53,6 +53,12 @@ Field2D::Field2D(Mesh *msh) : Field(msh), deriv(nullptr) {
     nx = fieldmesh->LocalNx;
     ny = fieldmesh->LocalNy;
   }
+#if CHECK > 0
+  else {
+    nx=-1;
+    ny=-1;
+  }
+#endif
   
 #ifdef TRACK
   name = "<F2D>";
@@ -66,6 +72,12 @@ Field2D::Field2D(const Field2D& f) : Field(f.fieldmesh), // The mesh containing 
     nx = fieldmesh->LocalNx;
     ny = fieldmesh->LocalNy;
   }
+#if CHECK > 0
+  else {
+    nx=-1;
+    ny=-1;
+  }
+#endif
   
   boundaryIsSet = false;
   *this = f;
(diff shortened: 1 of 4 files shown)


src/field/field2d.cxx
```diff
@@ -45,33 +45,51 @@
...
+#if CHECK > 0
+  else {
+    nx=-1;
+    ny=-1;
+  }
+#endif
   
   boundaryIsSet = false;
-  *this = f;
+  *this = f; //This copies all the data -- unlike 3D which delays, using Copy on write+ref counting
```
bendudson: Doesn't Field2D use the same CoW system as Field3D? The operator= function below uses the same Array reference counting as Field3D

d7919: Yes, you're right -- I didn't mean to include that comment in the commit!

I think this line is probably not required however as we have `data(f.data)` in the initialiser list (as is done for `Field3D`, which doesn't include a line `*this = f`).

src/field/field2d.cxx
```diff
@@ -86,8 +86,7 @@ Field2D::Field2D(const Field2D& f) : Field(f.fieldmesh), // The mesh containing
 #endif
   
   boundaryIsSet = false;
-  *this = f; //This copies all the data -- unlike 3D which delays, using Copy on write+ref counting
-}
```
ZedThree: You've accidentally killed the `}` too

Issue #608, opened 2017-07-13
BOUT.log files too noisy


bendudson:
The output log files (BOUT.log.*) contain too much noise, which makes it hard to pick out important information or warnings. Most of this is options, but these are (mostly) now output to BOUT.settings.

One problem is how to handle default settings which are not set in BOUT.inp, and are not (currently) output to BOUT.settings. The value used could be put in BOUT.settings perhaps? i.e. if an option is requested from the options tree which has not been set, then it could be added to the options with a "default" label for the source (rather than "command line" or "BOUT.inp"). A problem could arise if the same option has different defaults in different places, but that would probably be a bug anyway.
An advantage is that the default values used are recorded in a machine-readable format (BOUT.settings), and can be reproduced even if the default is changed in the code.

d7919:
Agreed. The most helpful bit about the "header" in my usage is that I can diff the logs of two runs to find any difference in setup -- this should now be possible by diffing BOUT.settings (if the defaults are recorded).

Would it be possible to check when on the default path if that route has already been followed but with a different default (for example checking if the variable is in the option tree with a source of default and a different value -- I'm not familiar with the tree so don't know if it stores this information)?

This change would of course mean it wouldn't be so easy to compare to old log files, but I guess that's not a particularly important change.

bendudson:
This is addressed by PRs #615, #629 and #630

Issue #402, opened 2016-12-05
How to pass BoutReal


dschwoerer:
There are different ways to pass a BoutReal to a function `f`
```
f(const BoutReal a);
f(BoutReal a);
f(const BoutReal &a);
```
They are all the same from the point of the user (I am not sure about the reference in terms of performance).
However, in terms of function signatures, they aren't.

All are used:
Option 1:
https://github.com/boutproject/BOUT-dev/blob/v4.0.0-RC/include/field3d.hxx#L405
Option 2:
https://github.com/boutproject/BOUT-dev/blob/v4.0.0-RC/include/field3d.hxx#L512
Option 3:
https://github.com/boutproject/BOUT-dev/blob/v4.0.0-RC/include/field3d.hxx#L410

Which way should be preferred?
Or did I miss something, and there is a good reason to have the differences?

bendudson:
Sorry, this is what happens when a code is around for a while without any style guide (and a developer who can't make up their mind).

The consensus here:
http://stackoverflow.com/questions/270408/is-it-better-in-c-to-pass-by-value-or-pass-by-constant-reference

 seems to be that it doesn't matter very much, but that option 2 (pass by value) is preferred for small things like BoutReal (double).

I think option 1 is the same as option 2 (const redundant)


-------- Original Message --------

dschwoerer:
You are right, I didn't know that const is an implementation detail:
http://stackoverflow.com/a/3969745

dschwoerer:
Closed with #409

PR #615, opened 2017-07-15
Options logging

bendudson:
* Fixes a bug in Options::set methods
* Adds a test case  (test-options) which catches the original bug
* Adds default values to the Options tree if they are not set, and checks for consistency. All options used are now output to BOUT.settings, including default values not overridden by the inputs
* Turns off printing of options to output (stdout and BOUT.log.* files) by default.
* Adds a "-v" and "--verbose" switch to re-enable logging

This is a partial fix for issue #608

### DIFF
--- a/examples/test-options/README
+++ b/examples/test-options/README
@@ -1,0 +1,5 @@
+Options test
+============
+
+This is a short unit test of the Options class. It tests if the get and set
+methods work, and that the default values are handled correctly.
--- a/examples/test-options/data/BOUT.inp
+++ b/examples/test-options/data/BOUT.inp
@@ -1,0 +1,8 @@
+
+mxg = 0
+myg = 0
+
+[mesh]
+nx = 1
+ny = 1
+nz = 1
--- a/examples/test-options/makefile
+++ b/examples/test-options/makefile
@@ -1,0 +1,6 @@
+
+BOUT_TOP	= ../..
+
+SOURCEC		= test-options.cxx
+
+include $(BOUT_TOP)/make.config
--- a/examples/test-options/runtest
+++ b/examples/test-options/runtest
@@ -1,1 +1,14 @@
-make && ./test-options > log && echo "Success"
+make
+./test-options > log 
+
+# Check return code
+if [ $? -eq 0 ]; then
+  echo "Success"
+  exit 0
+else 
+  echo "Failed."
+  # print log so it's seen in Travis output
+  cat log
+  exit 1
+fi
+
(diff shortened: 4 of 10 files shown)


d7919:
The style test seems to trigger on any const by-value arguments so the change of `const int &var` to `const int var` still fails (instead of failing at [this line](https://github.com/boutproject/BOUT-dev/blob/master/examples/test-code-style/runtest#L8) it now fails at [this line](https://github.com/boutproject/BOUT-dev/blob/master/examples/test-code-style/runtest#L11)).

dschwoerer:
Agree, the test could removed. As mentioned in #402 it does not matter to the function signature whether a passed value is const or not. ([SO](http://stackoverflow.com/a/3969745))
The main motivation for this test was, to make it easier to overwrite functions. But as the second check is only a hint for humans reading the header - it is probably better removed.

dschwoerer:
This breaks the AiolosMesh if used in a multi mesh scenario.
The problem is, that the BoutMesh sets the default to "".
The AiolosMesh checks whether an options is set, as it tries a few options, and if none is found, uses the same default as the BoutMesh.
But as the BoutMesh does not set the default it is using, this breaks the AiolosMesh.
I am against merging, until the BoutMesh sets sane defaults.

bendudson:
Sorry, which default does BoutMesh set to "" which then cause a problem?
I didn't want to invest too much time in test-options, since Peter is close to merging in his unit testing stuff which would replace it. I can add a README and script though for now.

dschwoerer:
The issue is that upwinding derivatives have in the non staggered, and in the staggered case different function signatures. (Introduced by #303 )
This causes different functions to be called, which then sets different default options - which causes min-max to fail, as it enables staggered grids, but does not set derivatives ...

bendudson:
I have to apologise for the index_derivs code again... it is fractally broken, and showing its age, but seemed like a good idea at the time (~10 years ago). 
For now I guess the default options for staggered and non-staggered should be made consistent.

bendudson:
Should be working now... I've taken out the log switch in Options and "-v" command-line argument, so hopefully this PR won't conflict with PR #623

bendudson:
Does this PR still break AiolosMesh? If not, can we merge it and then merge PR #623 ?

dschwoerer:
It is still not working with the AiolosMesh, but that is rather an issue with the AiolosMesh.

However, I am slightly worried of chaning the default of operators - as some code might rely on getting `SPLIT` as default, instead of `U1`?

bendudson:
Thanks David. I'm not aware of any that use the flux methods, but agree it could be an issue. Perhaps a warning could be printed if the default value is used? 
I think a good thing to try and do longer-term would be to identify which operators are actually used, and which can be removed.

PR #621, opened 2017-07-17
Options::cleanCache()

### LINKED ISSUE #620: Options:set()
I have been trying to use the set feature of the options class.

I have a mesh, and try to set the number of z points.
`BOUT.inp` looks like this:
```
[meshz]
nz=n
n=4
```
If I use:
```
    boutcore.setOption("meshz:n","%d"%nz)
    boutcore.setOption("meshz:nz","n")
```
The first mesh created from this option is created with correct settings, however if I call this section again, and create a new mesh, it keeps the first setting.

What works, is setting directly `meshz:nz`:
```
    boutcore.setOption("meshz:nz","%d"%nz)
```

Note: I set all options as string.
While I have a work around, it would be great if I could write MMS code that was less depending on the direction, and I could just set n. Further, I also need to update `dz` by hand ...

dschwoerer:
Fixes #620

### DIFF
--- a/examples/MMS/GBS/gbs.cxx
+++ b/examples/MMS/GBS/gbs.cxx
@@ -272,7 +272,7 @@ int GBS::init(bool restarting) {
   output.write("g_12 = %e, g_23 = %e\n", coords->g_12(2,2), coords->g_23(2,2));
 
 
-  FieldGenerator *gen = FieldFactory::get()->parse("source", Options::getRoot()->getSection("ne"));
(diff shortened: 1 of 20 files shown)


bendudson:
At the moment this cache cleaning will lead to a memory leak; the parsed objects (FieldGenerators) are not currently deleted anywhere, since they're stored in the cache and could all be deleted at the end. Deleting them might cause a problem if there's still a pointer to them somewhere (e.g. in a Boundary object for MMS time-dependent boundaries). We could wrap all returned FieldGenerators in a shared pointer so they're deleted when not needed.

dschwoerer:
I have pushed changes, that should avoid the memory leak.

PR #622, opened 2017-07-19
Optimise pack_data/unpack_data

d7919:
Two approaches to speed up routines

1. Reorder loops to try to have better memory access (small impact)
2. Replace `dynamic_cast` with `static_cast` (large impact).

In testing on 32 cores for a `68x32x512` grid this has been observed to
reduce the time spent in `pack_data` and `unpack_data` by a factor of
between 4 and 5, leading to a reduction in the communication time by
between 15% and 30%.

### DIFF
--- a/src/mesh/impls/bout/boutmesh.cxx
+++ b/src/mesh/impls/bout/boutmesh.cxx
@@ -2016,28 +2016,29 @@ int BoutMesh::pack_data(const vector<FieldData *> &var_list, int xge, int xlt, i
 
   int len = 0;
 
-  for (int jx = xge; jx != xlt; jx++) {
-    /// Loop over variables
-    for (const auto &var : var_list) {
-      if (var->is3D()) {
-        // 3D variable
-        ASSERT2(dynamic_cast<Field3D*>(var)->isAllocated());
-        
-        for (int jy = yge; jy < ylt; jy++) {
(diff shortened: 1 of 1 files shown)


src/mesh/impls/bout/boutmesh.cxx
```diff
@@ -2016,29 +2016,30 @@ int BoutMesh::pack_data(const vector<FieldData *> &var_list, int xge, int xlt, i
...
-        }
+  /// Loop over variables
+  for (const auto &var : var_list) {
+    if (var->is3D()) {
+      // 3D variable
+      ASSERT2(static_cast<Field3D*>(var)->isAllocated());
+      for (int jx = xge; jx != xlt; jx++) {        
+	for (int jy = yge; jy < ylt; jy++) {
+	  for (int jz = 0; jz < LocalNz; jz++, len++) {
+	    buffer[len] = (*static_cast<Field3D*>(var))(jx, jy, jz);
```
d7919: I didn't test this but suspect it won't make much difference -- static_cast is supposed to not have much (any?) runtime penalty.

bendudson: Looks like static cast of a pointer should have negligible cost: https://stackoverflow.com/questions/6445841/c-static-cast-runtime-overhead

PR #623, opened 2017-07-19
Verbosity

dschwoerer:
An alternative approach to reduce the verbosity.
`output` is replaced with several output objects:

name| useage
------|---------
`output_debug` |For highly verbose output messages, that are normally not needed. Needs to be enabled with a compile switch
`output_info`| For infos like what options are used
`output_prog`|For infos about the current progress
`output_warn`|For warnings
`output_error`|For errors

The specific output streams are just a suggestion and can easily be changed.

With the option `-v` more streams are enabled, and with `-q` less are enabled.
Currently all are enabled, as I think that it makes sense to keep the old output for a while, until the `BOUT.setting` has been around for a while. (To allow simple comparison of old and new simulation)

### DIFF
--- a/include/options.hxx
+++ b/include/options.hxx
@@ -45,6 +45,7 @@ class Options;
 #include <string>
 using std::string;
 
+class ConditionalOutput;
 /// Class to represent hierarchy of options
 /*!
  * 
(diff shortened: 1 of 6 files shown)


bendudson:
I quite like this, and think it's an improvement. I would prefer to keep "output" ; can we make it a synonym for output_info?

Longer-term perhaps it would be worth using a logging library like https://github.com/gabime/spdlog

bendudson:
Replaced by PR #629

PR #627, opened 2017-07-21
Fixes mapping of profiles in gridgen

bendudson:
As pointed out by Dr Seto-san, the mapping from
EQDSK 1D profiles to 2D mesh profiles assumed that both
used the same convention for the boundary psi. This
commit should fix this issue, by mapping the
mesh psi onto the EQDSK normalisations.

Only fixed the IDL version (Hypnotoad)


d7919:
This should probably target v4.1-rc now as this is a bug fix so we would want it in the next release.

PR #629, opened 2017-07-25
Dschwoerer verbosity

bendudson:
Update to @dschwoerer 's PR #623 

* Merging, resolved small conflicts
* Removes debugging print statements
* Replaces output macro with a ConditionalOutput object, with the same level as output_prog 
* Adds some doxygen comments
* Formatting with clang-format


src/mesh/coordinates.cxx
```diff
@@ -414,27 +414,16 @@ int Coordinates::calcCovariant() {
...
-
-  if ((err = max(abs((g_13 * g13 + g_23 * g23 + g_33 * g33) - 1))) > maxerr) {
-    maxerr = err;
-  }
+  BoutReal maxerr;
+  maxerr= BOUTMAX(max(abs((g_11 * g11 + g_12 * g12 + g_13 * g13) - 1)),
+                  max(abs((g_12 * g12 + g_22 * g22 + g_23 * g23) - 1)),
+                  max(abs((g_13 * g13 + g_23 * g23 + g_33 * g33) - 1)));
 
   output_info.write("\tMaximum error in diagonal inversion is %e\n", maxerr);
```
dschwoerer: Shouldn't we do a global operation?
Otherwise this info can be misleading, if another processor has a larger error

include/output.hxx
```diff
@@ -50,64 +49,211 @@ using std::endl;
...
 /// To allow statements like "output.write(...)" or "output << ..."
-#define output (*Output::getInstance())
+/// Output for debugging
+#ifdef DEBUG_ENABLED
+extern Output output_debug;
+#else
+extern DummyOutput output_debug;
+#endif
+extern ConditionalOutput output_warn;  ///< warnings
+extern ConditionalOutput output_prog;  ///< progress
```
ZedThree: I think this should just be `output_progress`, `prog` is very ambiguous for the sake of saving 4 characters.

manual/sphinx/user_docs/bout_options.rst
```diff
@@ -63,6 +63,7 @@ Command-line switches are:
 =============   ============================================================
 -h, --help      Prints a help message and quits
 -v, --verbose   Outputs more messages to BOUT.log files
```
ZedThree: I'm in favour of adding both `--verbose` and `--quiet` arguments.

include/options.hxx
```diff
@@ -104,14 +104,14 @@ using std::string;
...
-
+  Options();
+  
   /// Constructor used to create non-root objects
   ///
   /// @param[in] p         Parent object
   /// @param[in[ secname   Name of the section, including path from the root
-  Options(Options *p, string secname) : parent(p), sectionName(secname) {};
-
+  Options(Options *p, string s) : parent(p), sectionName(s) {};
```
ZedThree: I think this renaming should go the other way: `parent` and `sectionName` instead of `p` and `s` :)

include/options.hxx
```diff
@@ -104,14 +104,14 @@ using std::string;
...
 public:
   /// Constructor. This is called to create the root object
-  Options() : parent(nullptr) {}
-
+  Options();
+  
   /// Constructor used to create non-root objects
   ///
   /// @param[in] p         Parent object
   /// @param[in[ secname   Name of the section, including path from the root
```
ZedThree: Typo: should be `[in]`

include/output.hxx
```diff
@@ -29,11 +29,10 @@ class Output;
 #ifndef __OUTPUT_H__
 #define __OUTPUT_H__
 
-#include <stdio.h>
 #include "multiostream.hxx"
 #include <iostream>
 #include <fstream>
-
+#include <boutexception.hxx>
```
ZedThree: Very minor, but `<>` should be reserved for system headers, with double-quotes for our headers

include/output.hxx
```diff
@@ -50,64 +49,211 @@ using std::endl;
...
+
+/// Layer on top of Output which passes through calls to write, print etc
+/// if it is enabled, but discards messages otherwise.
+/// This is used to provide different levels of output
+/// (info, prog, warn, error) which can be enabled and disabled at run time.
+///
+class ConditionalOutput : public Output {
+public:
+  /// @params[in] base    The Output object which will be written to if enabled
+  ConditionalOutput(Output *base_) : base(base_), enabled(true), base_is_cond(false) {};
```
ZedThree: `base_` could just be `base` -- it's not used in the constructor body, so no chance for confusion. It would also make it match the doxygen comment

manual/sphinx/user_docs/bout_options.rst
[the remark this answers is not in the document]
bendudson: Both --verbose and --quiet added. After looking at a few command-line argument parsers I wrote one which perhaps might be useful: https://github.com/bendudson/argopts

src/mesh/coordinates.cxx
bendudson: I've changed "Maximum error" to "Local maximum error" in the output

src/field/field_data.cxx
```diff
@@ -21,13 +21,13 @@ void FieldData::setBoundary(const string &name) {
   /// Get the boundary factory (singleton)
   BoundaryFactory *bfact = BoundaryFactory::getInstance();
   
-  output << "Setting boundary for variable " << name << endl;
+  output_prog << "Setting boundary for variable " << name << endl;
```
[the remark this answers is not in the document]
bendudson: yes, changed to info

src/mesh/index_derivs.cxx
```diff
@@ -929,7 +929,7 @@ void derivs_initialise(Options *options, bool StaggerGrids,
 void Mesh::derivs_init(Options* options) {
   TRACE("Initialising derivatives");
 
-  output.write("Setting X differencing methods\n");
+  output_prog.write("Setting X differencing methods\n");
```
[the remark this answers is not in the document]
bendudson: Also changed to info

src/solver/impls/cvode/cvode.cxx
```diff
@@ -328,12 +327,12 @@ int CvodeSolver::run() {
     /// Check if the run succeeded
     if(simtime < 0.0) {
       // Step failed
-      output.write("Timestep failed. Aborting\n");
+      output_error.write("Timestep failed. Aborting\n");
```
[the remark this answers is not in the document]
bendudson: Removed

include/output.hxx
bendudson: Done

include/output.hxx
bendudson: Done

include/output.hxx
bendudson: All output_prog changed to output_progress

include/options.hxx
bendudson: Done

include/options.hxx
bendudson: Done

src/mesh/impls/bout/boutmesh.cxx
```diff
@@ -1938,16 +1816,16 @@ BoutMesh::CommHandle* BoutMesh::get_handle(int xlen, int ylen) {
...
-      delete[] ch->umsg_sendbuff;
-      delete[] ch->umsg_recvbuff;
-      delete[] ch->dmsg_sendbuff;
-      delete[] ch->dmsg_recvbuff;
+  if (ch->ybufflen < ylen) {
+    if (ch->ybufflen > 0) {
+      delete[] ch -> umsg_sendbuff;
+      delete[] ch -> umsg_recvbuff;
+      delete[] ch -> dmsg_sendbuff;
+      delete[] ch -> dmsg_recvbuff;
```
ZedThree: Is this just clang-format being weird?

ZedThree:
The unit tests for `Options` might need updating now

bendudson:
I don't understand the failing job: https://travis-ci.org/boutproject/BOUT-dev/jobs/272514957

    sys/test_array.cxx:110: Failure
      Expected: a[4]
      Which is: 0
    To be equal to: 4
    [  FAILED  ] ArrayTest.RetrieveData (0 ms)

This seems to be only on one build, where CONFIGURE_OPTIONS='--enable-checks=no --enable-optimize=3 --disable-signal --disable-track --disable-backtrace'

The same options for a different merge with v4.1-RC ( https://github.com/boutproject/BOUT-dev/pull/674 ) passes: https://travis-ci.org/boutproject/BOUT-dev/jobs/272475092
but the files involved, include/bout/array.hxx and tests/unit/sys/test_array.cxx  are the same between the two branches. 

Suggestions anyone?

ZedThree:
Looks like 857f512 fixed that failing test? Did `Array::cleanup` change with the recent changes in `Array`?
I was a little bit worried that some of the tests relied on the store being cleaned, but after a quick look, I don't think that is the case.

bendudson:
The array cleanup() call at the end of each test was disabling the store for the next test, so a mystery is why the test passed previously, then started failing on that particular test. I think since the store was disabled, the array was being deleted, then re-allocated with new. Since the sizes of the arrays are the same, perhaps new just grabbed the deleted block of memory, so the contents were the same as if the store was working. I'm not sure how to really test whether new and delete are being called, or perhaps we could add something to Array to report the size of the store?

PR #630, opened 2017-07-27
Add color to output

bendudson:
Adds a shell script and command-line option ("-c") to color BOUT++ output

* bout-log-color is a shell script which uses sed to insert text color escape codes. This currently highlights:

  1. Any word "warning" or "error" (case insensitive) is turned red
  2. The word "default" after options is turned blue
  3. The phrase "Command line" after options is turned green

  The aim of this is to highlight important things to note
  i.e. when something might be wrong; when the code has set something
  for you, and for the few settings which you override on the command line

  If the BOUT-dev "bin" directory is in PATH, then existing log files can be printed in color using:

  $ cat BOUT.log.0 | bout-log-color

* A command-line option "-c" has been added, which pipes stdout through bout-log-color. This uses posix functions, so should be quite portable. This functionality can be disabled in configure with "--disable-color" if needed.

An issue is that if "-c" is used, but bout-log-color is not in PATH, then the code exits. The popen command succeeds, returning a valid pipe, even though the shell command failed.


d7919:
Could the full path to `bout-log-color` be defined at the configure stage so we can use the full path when opening the pipe? This might reduce issues with not having path set correctly, although still not avoiding all possible issues (e.g. with the script not working for other unlikely reasons).

configure.ac
```diff
@@ -261,6 +263,14 @@ else
   echo "Segmentation fault handling disabled"
 fi
 
+if test "$enable_signal" != "no"
```
dschwoerer: `enable_signal` should be `enable_color`

src/bout++.cxx
```diff
@@ -148,6 +148,9 @@ int BoutInitialise(int &argc, char **&argv) {
 	      "  -o <settings filename>\tSave used OPTIONS given to <options filename>\n"
 	      "  -v \t\tIncrease verbosity\n"
 	      "  -q \t\tDecrease verbosity\n"
+#ifdef LOGCOLOR
+              "  -c \t\tColor output using bout-log-color\n"
```
dschwoerer: provide long version as well? Both `-c` and `--color`?

PR #633, opened 2017-07-29
Replacing old stencils with DataIterator loops in index_derivs

bendudson:
Rather than having conditionals in the inner loop (inside calls to Field::set*Stencil), the applyXdiff, applyYdiff and applyZdiff functions now check for:

* Staggered vs non-staggered
* one or two guard cells
* yup/down fields

and have multiple loops, one for each combination of possibilities. 

This improves performance a little: For "-03" flags, timing "DDX() + DDY() + DDZ()"  in examples/performance/difops this code is about 10% faster than the old code. With the standard "-O -DCHECK=2) it's about 6% faster.

This should make parallelisation with OpenMP easier, since DataIterator already does most of the work. Hopefully it will also allow the old stencils code to be removed.


d7919:
It looks like `applyXdiff` for `Field2D` has not been updated but the 3d and 2d versions have for `y` (and 3d for `z`) -- is it possible to update the 2D `x` version as well?

src/mesh/index_derivs.cxx
```diff
@@ -1067,23 +1067,103 @@ const Field2D Mesh::applyXdiff(const Field2D &var, Mesh::deriv_func func, Mesh::
...
+      (loc != CELL_DEFAULT) && (loc != var.getLocation())) {
+    // Staggered differencing
 
-  bindex bx;
+    CELL_LOC location = var.getLocation();
+    
+    if (mesh->xstart > 1) {
+      // More than one guard cell, so set pp and mm values
+      // This allows higher-order methods to be used
+      for(const auto &i : result.region(RGN_NOX)) {
```
d7919: `RGN_NOX` is used here but in the `y` versions `RGN_NOBNDRY` is used. Is this intentional (i.e. do we intentionally want to set the derivatives in the y guard cells here but not the other way around in this case) and if so should there be a comment around here to say why?

src/mesh/index_derivs.cxx
```diff
@@ -1278,6 +1499,7 @@ const Field3D Mesh::applyYdiff(const Field3D &var, Mesh::deriv_func func, Mesh::
   result.bndry_xin = result.bndry_xout = result.bndry_yup = result.bndry_ydown = false;
 #endif
   
+  bindex bx;
   if (mesh->freeboundary_xin && mesh->firstX() && !mesh->periodicX) {
     for (bx.jx=mesh->xstart-1; bx.jx>=0; bx.jx--)
       for (bx.jy=mesh->ystart; bx.jy<=mesh->ystart; bx.jy++)
```
d7919: I realise this line hasn't been changed, but is this really correct -- we just have the `jy=ystart` iteration?

src/mesh/index_derivs.cxx
bendudson: I think this was to keep the same behavior as the old code. Agree it is inconsistent, but I think it's for taking mixed X-Z derivatives without having to communicate. Perhaps the region should be an optional input?

src/mesh/index_derivs.cxx
bendudson: This whole freeboundary_* code needs to be removed I think, but is probably another PR. I don't think it's used anywhere, and it significantly complicates the (already over-complicated) index_derivs code. 

Based on a switch in mesh, it applies different numerical methods at the first and last points (hence only jy=ystart). A better way is to apply the free boundary conditions on variables where needed.

bendudson:
applyXdiff(Field2D) now modified too, and an optional REGION argument added to all apply*diff functions.
Only use of calc_index is now in the free boundary parts of index_derivs, which should be removed soon.

d7919:
Could this be extended to include the upwind and flux methods as well? If so then (in combination with removing the free boundary stuff https://github.com/boutproject/BOUT-dev/tree/deleteFreeBoundaryInDiffs) we would probably be able to remove quite a bit of code (all the `set?stencil` code etc.)

PR #634, opened 2017-07-30
Operators with dataiterator arg

bendudson:
Adding some operators which are passed a DataIterator to calculate a result on a single point. This can be used to move the loop over the domain outside, reducing the number of loops and maybe improving cache efficiency. 

Adds:

* Header file bout/operators_di.hxx
* Example blob-outerloop
* Small manual section


include/bout/operators_di.hxx
```diff
@@ -0,0 +1,108 @@
...
+/// \brief Arakawa bracket in X-Z [f,g]
+///
+/// @param[in] f  Field to be differentiated
+/// @param[in] g  Field to be differentiated
+/// @param[in[ i  The point where the result is calculated
+///
+BoutReal bracket_arakawa(const Field3D &f, const Field3D &g, const DataIterator &i) {
+  Coordinates *metric = mesh->coordinates();
+  
+  BoutReal fxp = f[i.xp()];
```
d7919: Is there any benefit to making these values `const`? Likewise, I wonder if it would be beneficial to store `const` versions of `i.xp()`, `i.zp()`, etc.? Certainly I could imagine the `z` offset being non-trivial due to the periodicity (i.e. modulus).

include/bout/operators_di.hxx
```diff
@@ -0,0 +1,108 @@
...
+  BoutReal fmp = f[i.offset(-1,0,1)];
+  BoutReal fmm = f[i.offset(-1,0,-1)];
+  
+  BoutReal gxp = g[i.xp()];
+  BoutReal gxm = g[i.xm()];
+  BoutReal gzp = g[i.zp()];
+  BoutReal gzm = g[i.zm()];
+
+  // J++ = DDZ(f)*DDX(g) - DDX(f)*DDZ(g)
+  BoutReal Jpp = 0.25*( (fzp - fzm)*
```
d7919: Certainly minor, but could the 1/4 factor be merged into the final factor (3-->12) in the return statement?

PR #637, opened 2017-08-08
Overhaul testing framework

ZedThree:
Apologies for enormous PR...

This is a major change to how tests are done in BOUT++, and importantly now adds an expectation that new features or functions also include tests for that feature. All tests are now gathered in the top-level directory `tests`, and are then split into unit, integrated and MMS tests. The unit tests are very fast to compile and run, so can be run very often (e.g. on file save), while the integrated and MMS tests are more expensive and might only be run when commits are pushed to github. The following existing integrated tests have been replaced by faster and more comprehensive unit tests:

- test-field
- test-minmax
- test-options
- test-fieldgroup (except for the expected-fail compile-time test)

A few bugs found while making the unit tests have been fixed:

- Fix bug in SIGN: wrong sign for 0
- Bugfix: Options::set() wrong bodies for BoutReal/bool overloads
- Fix compiler warning and memory leak in test_utils
- Fix memory leak in RangeIterator
- Fix bug in OptionsReader when handling spaces around '=' on CLI
- Fix rare potential memory leak when reading/writing settings files
- Fix bug in FieldGroup::makeUnique(), missed Field3D vector

A new configure flag, `--enable-code-coverage` has been added. This will compile the code with the correct flags for gcov, and adds a make target, `code-coverage-capture`, which runs lcov and generates an HTML summary of code coverage. This could eventually be built and pushed to a coverage web service. The current coverage from just the unit tests is as follows:

Summary coverage rate:
  lines......: 9.0% (1629 of 18125 lines)
  functions..: 24.3% (607 of 2499 functions)


CONTRIBUTING.md
```diff
@@ -43,6 +43,13 @@ Here are some rules for editing the core BOUT++ code:
    add or update documentation where appropriate, we may ask you to do so before
    it is merged.
 
+-  As well as documentation for new features, you must also include a
+   representative test to ensure that it works correctly. Please see
+   the [tests README][test_readme] for more information on tests in
+   BOUT++. Prefer to write unit tests that check the feature at the function
+   level, rather than integrated tests that require setting up a whole physics
+   model.
+
```
bendudson: I agree with this: We don't want to discourage people from adding features, but also don't want to end up with unmaintainable code

dschwoerer:
I think `test-minmax` was the only test that tried to evolve a Field2D ...
Is this now part of another test?

ZedThree:
As far as I can tell, while it does try to evolve a Field2D, it doesn't actually include any checking that it does so correctly. The tests should really be orthogonal as well, with testing the solvers done separately to testing the min/max functions.

PR #638, opened 2017-08-08
Python boutdata patches

bendudson:
* collect now prints the path in error message ("no data files found")
  This is useful in scripts which read data from several directories

* restart.redistribute now reads NYPE rather than NPES since NPES is
  no longer written to restart files


ZedThree:
Merged via manual rebasing and push

PR #642, opened 2017-08-09
Remove deprecated get/setdata from fields

ZedThree:
Remove the get/setData functions that were deprecated in 4.0


d7919:
Are there any other deprecated routines that we can remove at this point (e.g. `field2d::?et?array`, `solver::addToRestart`)?

ZedThree:
`addToRestart` is the only one that currently gives me a warning, and I'm not sure how much that's used.

Other deprecated stuff can probably just go too.

bendudson:
I think I'm probably the only one who uses addToRestart, for feedback controller stuff in SD1D and Hermes. I deprecated the function, but can't quite remember what it's supposed to be replaced by...

I think it's a public function so that Solver can call it. What's deprecated is users calling Solver->addToRestart, since restart files are now in PhysicsModel not in Solver.

Perhaps addToRestart could be removed from Solver, and the PhysicsModel function made protected

d7919:
Isn't this PR duplicated in #643 now?

dschwoerer:
* I thought breaking changes would only go to major releases (4.0, 5.0 ...)
* Why is this going directly into 4.1-RC?

ZedThree:
This is just removal of functions that have been marked deprecated since 4.0, which was more than six months ago now. That should be plenty of time for people to remove them from their code.

Removing marked deprecated functions I think counts as cleaning/tidying :)

PR #643, opened 2017-08-09
Fix various warnings in v4.1-RC

ZedThree:
Along with other small tidying-up


d7919:
What does the warning table (#322) look like with this?

ZedThree:
Pretty much done for this exercise, I think. There are now 7 warnings with gcc 6.2 and -Wall. I've also run `scan-build` with clang 4 and fixed a bunch of potential bugs from that, although not all of them.

PR #649, opened 2017-08-11
Delete free boundary in diffs

d7919:
Remove broken/suspicious code not supported.

May want to double check how this interacts with the `bndry=free` approach, which is still supported (I think)


bendudson:
Since these methods require entirely different stencils and operators to the standard ones, I think it makes sense to separate them out. By simplifying this code, further improvements and simplifications become easier.

If something like this functionality is needed, then it can be put into the boundary condition code. For 2nd order convergence the supported method of setting free boundaries should be sufficient; it's only for higher order that more sophisticated treatment is needed.

PR #650, opened 2017-08-12
Making integrated test suite quiet by default

bendudson:
Style now similar to (copied from) pytest. The output from each test is collected by redirecting stdout, and only printed if the test fails.

If there are no failures then the output looks like:

======= Starting integrated tests ========
Running 17 tests
.................

======= All tests passed in 574.53 seconds =======

Adding some failures (first and third test), and running only 4 tests for demonstration:

======= Starting integrated tests ========
Running 4 tests
F.F.

======= FAILURES ========

----- test-io -----
Warning: No str in builtins
Warning: No range in builtins
WARNING: netcdf4-python module not found
         expect poor performance
  => Using Scientific.IO.NetCDF instead
Making I/O test

----- test-restarting -----
Warning: No str in builtins
Warning: No range in builtins
WARNING: netcdf4-python module not found
         expect poor performance
  => Using Scientific.IO.NetCDF instead
-> Making restart test
-> Testing restart append
  File "./runtest", line 32, in <module>
    if f3d_1.shape != f3d_0.shape:
AttributeError: 'int' object has no attribute 'shape'

======= 2 failed in 4.96 seconds ========


d7919:
Could this go in 4.1?

PR #652, opened 2017-08-14
Remove unused boundary files and small bug fix in Dirichlet non-staggered

### LINKED ISSUE #646: Extra boundary files
In `src/mesh` there is `boundary_standard.cxx`, `orig_boundary_standard.cxx` and `nonorthog_boundary_standard.cxx`. Only this first is included in the build.

It looks like `nonorthog` is a superset of `orig` -- containing a few additions (fixes/generalisations/not relevant?).

The `boundary_standard` version is the version we actually use and is essentially an up to date version of things, but doesn't seem to include some of the additions that are in `nonorthog`.

I think these extra files can probably be removed but we might want to see if the extra things in `nonorthog` should be merged first or not.

### LINKED ISSUE #651: Boundary non-staggered dirichlet with Field3D
Commit 1fd2315c48741365c48a73b6d538d193cba22794 replaced extrapolation in the boundary cells with just setting all boundary cells to the Dirichlet value. I think the extrapolation should be the correct thing to do. 

I've just committed 519ebbfac6d5547a02333c30671ac5eed7f9250c which reverts this change and the MMS tests past.

d7919:
Removes two large unused historical files. Fixes #646 

Adds some comments to explain why we don't use extrapolation in part of the Dirichlet boundary approach. Note, includes a commit that enables extrapolation and another which reverts this commit --
 I've left these in as I hope this, along with the comments in the code should help with making it clear why the proposed bug fix is not actually a bug fix. Closes #651


src/mesh/boundary_standard.cxx
```diff
@@ -1018,10 +1034,8 @@ void BoundaryDirichlet_O3::apply_ddt(Field2D &f) {
     (*dt)(bndry->x,bndry->y) = 0.; // Set time derivative to zero
 }
 
-
-
 void BoundaryDirichlet_O3::apply_ddt(Field3D &f) {
-  output.write("\nI'm getting called \n") ; 
+  output.write("\nI'm getting called \n") ; //Looks like this may just be experimental -- should we warn/remove?
```
ZedThree: I think this line can just go

PR #657, opened 2017-08-16
STORM3D

xujq:
STORM3D module in the framework of BOUT++, as described by Luke Easy.


bendudson:
Please contact Fulvio Militello at CCFE for access to the STORM module, since it is not publicly available.
Pull Requests on github have a different use, though the name can be confusing: They are for requesting that we pull (merge) changes into BOUT++, not requests to pull things out.

PR #658, opened 2017-08-16
make install and uninstall targets

bendudson:
This allows BOUT++ to be installed to a directory, and maintain links to include files, libraries etc. 

* There is currently no documentation or proper testing, so this is not really ready to merge, but here for discussion/tweaking. 
* I have tried to follow the GNU coding standards https://www.gnu.org/prep/standards/html_node/Makefile-Conventions.html#Makefile-Conventions but probably missed some things. 
* It's up for discussion where we install some things like python, IDL scripts. Experience or similar project examples welcome
* Hopefully this will make packaging BOUT++ for distributions easier

To test it, run
$ configure
$ make
$ make install prefix=$HOME/tmp/bout++/

This should install everything needed under $HOME/tmp/bout++ in bin, include, lib, share subdirectories. Finer control over installation directories should work though configure in a GNU-ish way.

Then add $HOME/tmp/bout++ to your PATH and try

$ bout-config --all

The examples/make-script code should then compile anywhere, linking against the installed library.

You can also try replacing the examples/conduction makefile with

    SOURCEC         = conduction.cxx
    include $(shell bout-config --config-file)

which should find make.config using the bout-config script.


bendudson:
Note: The include files are installed under a "bout++" subdirectory, to avoid conflicts with system libraries. This doesn't affect any code because the paths (BOUT_INCLUDE_PATH) is included in the flags, and updated by make install.

bendudson:
Now merged in PR #550 

Tested that the following ways to install work after installing BOUT++ with 
    $ make install prefix=$HOME/tmp/

Move examples/conduction outside the source directory (so BOUT_TOP is not a parent directory)

1. With the current makefile unmodified
    $ make BOUT_TOP=$HOME/tmp/share/bout++

2. With makefile

    SOURCEC         = conduction.cxx
    include $(shell bout-config --config-file)

    $ make

3. The examples/make-script example, using bout-config to get flags

bendudson:
Passing "--prefix=/some/install/path"  to configure also works, same as passing to "make install".

PR #659, opened 2017-08-16
Travis tweak parallel build

d7919:
This is a<s>n experimental</s> PR opened for comment. It makes a number of changes, we might not want all:

1. Remove clang openmp build -- the travis version of clang doesn't support openmp. It should be possible to get this to work but we probably don't need multiple openmp tests.
2. Adds a gcc build in which lots of features are disabled (check, track, backtrace etc.) and optimisations are enabled -- the idea behind this is that we should try to catch problems relating to different signatures etc. depending on what gets defined.
3. Use Travis "addons" to specify the packages that we want to install (see https://docs.travis-ci.com/user/installing-dependencies/ and https://github.com/travis-ci/apt-package-whitelist for details). This removes the need for `sudo` which means we can use the container based environment (see [here](https://docs.travis-ci.com/user/reference/overview/#Virtualization-environments)). This should have performance advantages : the containers take a few seconds to boot compared to up to a minute for the VM approach currently used and each container gets exactly 2 cores whilst the VM gets ~2 (presumably meaning it varies somewhat).
4. Explicitly set the number of openmp threads to 2 in the openmp build.
5. Use two parallel jobs in the make commands to speed up compilation (i.e. `-j 2`).
6. Try to use dependency caching to store the conda packages and apt packages. This involved pulling the conda setup out into a separate script.
7.  Change the `sudo` state from required to false so that we use the container environment.

I think most of these are useful apart from number 6, which seems to be a bit temperamental and possibly slower than the original approach. In testing 7 seems to have helped the openmp build go quite a bit faster, but seems to have potentially slowed the other builds slightly, but I think more experience is required.


d7919:
Somebody might want to play with the caching, but in general it seems like it might not be very helpful for us.

.travis_script.sh
```diff
@@ -17,6 +17,7 @@ then
     exit $conf
 fi
 export PYTHONPATH=$(pwd)/tools/pylib/:$PYTHONPATH
+export MAKEFLAGS="-j 2"
 time make || exit
 time make check-unit-tests || exit
```
dschwoerer: should the test run as well with `-j 2`?
Some of them are anyway parallel ...

.travis.yml
```diff
@@ -1,41 +1,39 @@
...
+    - libopenmpi-dev
+    - libfftw3-dev
+    - libnetcdf-dev
+    - libhdf5-serial-dev
+    - netcdf-bin
+    - hdf5-tools
+
 matrix:
   include:
-  - env: CONFIGURE_OPTIONS='--enable-openmp'
```
dschwoerer: maybe one additional run with default options?

.travis_script.sh
d7919: I think the integrated tests target in the makefile doesn't list each job independently -- it just runs the `test_suite_make` and `test_suite` scripts with no dependencies, so I don't think it will actually end up with multiple jobs running at the same time. The `-j 2` on all the `make` test commands is essentially pointless -- it only impacts the main compilation I think.

.travis.yml
d7919: Yes that sounds like a good idea.

.travis_script.sh
ZedThree: It should help on the unit tests though

d7919: I wasn't sure if it would pass the `-j 2` onto the child `make` -- not sure if the child inherits the parent environment or not.

d7919: [Looks like](https://www.gnu.org/software/make/manual/html_node/Options_002fRecursion.html) it should do "the right thing"

ZedThree: Should we also add `-k`? If for some reason we have multiple errors during make, it might be handy to see as many of them as possible

d7919: That is a good idea.

d7919:
I've now removed the conda caching, added a default job to the matrix and set the `-k` flag for make.

May be worth revisiting the caching in the future (particularly if we want to expand testing to include external packages such as sundials and petsc, which we could cache to save the build costs).

d7919:
Possible that container build is a bit slower because it uses a different file system (AUFS) which might not be brilliant for reading/opening lots of files ([for example](https://sthbrx.github.io/blog/2015/10/30/docker-just-stop-using-aufs/))

d7919:
Reverting to the VM based approach speeds up most of the jobs in the matrix, except the openmp case, which is slower than the others (but is at least faster than it was).

d7919:
In the current state this seems to speed up the CI time and makes the travis configuration ready for switching to containers should the suspected I/O issues be resolved. I've therefore removed the "invalid" tag to indicate that I think this now might actually be worth merging.

d7919:
I've created two branches which further extend these travis changes

1. [This branch](https://github.com/boutproject/BOUT-dev/tree/travisTestGcc6) demonstrates how we can test with a different compiler version (`gcc-6` here).
2. [This branch](https://github.com/boutproject/BOUT-dev/tree/travisTestExternals) adds `cvode`, `ida`, `hypre` and `petsc` packages and enables them in one of the jobs. This allows us to test that certain optional features are compilable (in fact this branch shows that there appears to be a problem with the petsc solver).

PR #661, opened 2017-08-16
V4.1 rc into next

dschwoerer:
RC changes into next


ZedThree:
See comment on #660 about merge order

PR #664, opened 2017-08-16
V4.1 Intel warnings/remarks fixes

ZedThree:
Fixes lots of warnings/remarks from Intel 17.0 compiled with the following flags:

```
BOUT_FLAGS		= $(CXXFLAGS) -g -O0 -fno-inline -hipa1 -DCHECK=3 -DSIGHANDLE \
   -DTRACK -DNCDF4 -DHDF5 -DLAPACK -DBOUT_HAS_IDA -DBOUT_HAS_CVODE \
   -DCVODEINT=long -DIDAINT=long -DBOUT_HAS_PVODE  -std=c++11  -w3 \
   -wd3180,981,383,1418,3280,2282,3346,177
```
where I have disabled the following unhelpful warnings:

- 177: `handler parameter "e" was declared but never referenced`
    - These are from catching exceptions. These could probably be fixed by just removing the name, but I'm a little unsure if that causes problems with rethrowing them (probably not?)
- 383: `value copied to temporary, reference to temporary used`
    - These are mostly from passing string literals, but some might need looking at properly
- 981: `operands are evaluated in unspecified order`
    - I don't think we ever rely on operands being evaluated in a specific order. A lot of these are from the MPI library
- 1418: `external function definition with no prior declaration`
    - Generally unhelpful. Catches templated functions. Some things could maybe declared `static` or put in anonymous namespaces
- 2282: `unrecognized GCC pragma`
    - This is from the HDF5 library and not us
- 3180: `unrecognized OpenMP #pragma`
    - Disappears if compiled with OpenMP
- 3280: `declaration hides member`
    - Incorrectly catches constructor arguments with the same name as members when used with initialiser lists. This is actually idiomatic C++, and not a problem. Unfortunately, the unhelpful instances drown out any useful ones.
- 3346: `dynamic exception specifications are deprecated`
    - Fixed all instances in our code, rest are from third-party libraries

Disabling these warnings leaves 67 warnings/remarks unfixed by this PR. I've gone for the low-hanging fruit, so there's definitely more that can be fixed.


include/bout/sys/expressionparser.hxx
```diff
@@ -127,7 +127,7 @@ private:
     signed char curtok;  ///< Current token. -1 for number, -2 for string, 0 for "end of input"
     double curval; ///< Value if a number
     std::string curident; ///< Identifier, variable or function name
-    signed char LastChar;   ///< The last character read from the string
+    char LastChar;   ///< The last character read from the string
```
d7919: Is there a reason this has lost signed but `curtok` hasn't?

ZedThree: This was originally part of me trying to fix silent conversions with minimal changes, but I ended up having to explicitly cast every call to `stringstream::get()` anyway, so this change can be undone.

d7919: Ah ok, guess the casts will need changing as well.

PR #667, opened 2017-08-17
Travis include externals

d7919:
Adds an entry to the travis matrix that includes petsc and hypre external packages.


d7919:
The new job currently fails due to a bug in the petsc solver implementation. This is fixed by #665

Issue #653, opened 2017-08-15
Unused source files


d7919:
There are a number of unused (i.e. not included in the build) source files within the source tree

~~~~
makefiles=$(find ./ -name makefile) ; for j in $(find ./src -name '*.cxx'|awk -F"/" '{print $NF}') ; do MATCH=$(grep ${j} ${makefiles} | wc -l) ; if [ ${MATCH} -eq 0 ] ; then echo -e "${j}\t: ${MATCH}\t\n($(wc -l $(find ./src -name ${j})))\n---------------------------------------" ; fi; done
inverter.cxx	: 0	
(449 ./src/invert/inverter.cxx)
---------------------------------------
invert_parderiv_hypre.cxx	: 0	
(176 ./src/invert/parderiv/impls/hypre/invert_parderiv_hypre.cxx)
---------------------------------------
invert_parderiv_simple.cxx	: 0	
(279 ./src/invert/parderiv/impls/simple/invert_parderiv_simple.cxx)
---------------------------------------
full_gmres.cxx	: 0	
(80 ./src/invert/full_gmres.cxx)
---------------------------------------
multigrid_gmres.cxx	: 0	
(227 ./src/invert/laplace/impls/multigrid/multigrid_gmres.cxx)
---------------------------------------
invert_gmres.cxx	: 0	
(477 ./src/invert/invert_gmres.cxx)
---------------------------------------
jstruc.cxx	: 0	
(384 ./src/precon/jstruc.cxx)
---------------------------------------
unit_tests.cxx	: 0	
(  43 ./src/mesh/impls/unit_tests.cxx
  59 ./src/sys/unit_tests.cxx
 102 total)
---------------------------------------
unit_tests.cxx	: 0	
(  43 ./src/mesh/impls/unit_tests.cxx
  59 ./src/sys/unit_tests.cxx
 102 total)
---------------------------------------
bout_facets.cxx	: 0	
(224 ./src/facets/bout_facets.cxx)
---------------------------------------

~~~~

Can any of these be removed and should any of these actually be included in the build?

bendudson:
I think they can all be removed. They'll still be in the history if needed later

d7919:
fixed in v4.1 #654

d7919:
Reopening to note removal of header files associated with these files. 

This removes all references to facets in the source code, but it remains a configuration option -- we may wish to remove these references as well? Or do we want to restore the facets files for use in other projects?

bendudson:
[...]

PR #672, opened 2017-08-18
Remove header files associated with files removed earlier

d7919:
Files discussed in #653 and removed in #654.

Note this removes references to facets, so we may wish to remove that
from the configure script as well.


ZedThree:
Yes, should be removed from configure script and scrubbed from documentation (I think there's only one line)

PR #673, opened 2017-08-18
Convert old-style casts to static_cast

ZedThree:
Use the "safer" C++ `static_cast` instead of C-style casts (except when using `malloc`).
One benefit is that it's now easier to find casting, without using `-Wold-style-cast`.

### DIFF
--- a/src/invert/laplace/impls/multigrid/multigrid_alg.cxx
+++ b/src/invert/laplace/impls/multigrid/multigrid_alg.cxx
@@ -429,7 +429,7 @@ void MultigridAlg::pGMRES(BoutReal *sol,BoutReal *rhs,int level,int iplag) {
 
   if(level == 0) {
     if((rProcI == 0) && (pcheck == 1)) {
-      rederr = log(error / ini_e) / static_cast<double>(num);
+      rederr = log(error / ini_e) / static_cast<BoutReal>(num);
       rederr = exp(rederr);      
       printf("The average error reduction of GMRES %d: %14.8f(%18.10f)\n",num,rederr,error);
       fflush(stdout);
@@ -437,7 +437,7 @@ void MultigridAlg::pGMRES(BoutReal *sol,BoutReal *rhs,int level,int iplag) {
   }
   else {
     if((rProcI == 0) && (pcheck == 1)) {
-      rederr = log(error / ini_e) / static_cast<double>(num);
+      rederr = log(error / ini_e) / static_cast<BoutReal>(num);
       rederr = exp(rederr);      
       printf("The average error reduction of PGMRES %d: %14.8f(%18.10f)\n",num,rederr,error);
       fflush(stdout);
@@ -697,7 +697,7 @@ void MultigridAlg::solveMG(BoutReal *sol,BoutReal *rhs,int level) {
   }
 
   if((rProcI == 0) && (pcheck == 1)) {
-    rederr = log(error / ini_e) / (static_cast<double>(m) + 1.0);
+    rederr = log(error / ini_e) / (static_cast<BoutReal>(m) + 1.0);
     rederr = exp(rederr); 
     if(m == MAXIT) 
       printf("Reached maximum iteration: %14.8f\n",error);
--- a/src/invert/laplace/impls/multigrid/multigrid_alg.cxx
+++ b/src/invert/laplace/impls/multigrid/multigrid_alg.cxx
@@ -429,7 +429,7 @@ void MultigridAlg::pGMRES(BoutReal *sol,BoutReal *rhs,int level,int iplag) {
 
   if(level == 0) {
     if((rProcI == 0) && (pcheck == 1)) {
-      rederr = log(error/ini_e)/((double)num);
+      rederr = log(error / ini_e) / static_cast<double>(num);
       rederr = exp(rederr);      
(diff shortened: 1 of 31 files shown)


src/invert/laplace/impls/multigrid/multigrid_alg.cxx
```diff
@@ -429,15 +429,15 @@ void MultigridAlg::pGMRES(BoutReal *sol,BoutReal *rhs,int level,int iplag) {
 
   if(level == 0) {
     if((rProcI == 0) && (pcheck == 1)) {
-      rederr = log(error/ini_e)/((double)num);
+      rederr = log(error / ini_e) / static_cast<double>(num);
```
d7919: Whilst perhaps not relevant for this PR, should this really be `BoutReal` for consistency instead of `double`?

src/invert/laplace/invert_laplace.cxx
```diff
@@ -66,7 +66,7 @@ Laplacian::Laplacian(Options *options) {
   OPTION(options, filter, 0.0);
   int ncz = mesh->LocalNz;
   // convert filtering into an integer number of modes
-  maxmode = ROUND((1.0 - filter) * ((double) (ncz / 2)));
+  maxmode = ROUND((1.0 - filter) * static_cast<double>(ncz / 2));
```
d7919: `double` --> `BoutReal`?

src/mesh/coordinates.cxx
```diff
@@ -56,7 +56,7 @@ Coordinates::Coordinates(Mesh *mesh) {
     if (options->isSet("zperiod")) {
       OPTION(options, zperiod, 1);
       ZMIN = 0.0;
-      ZMAX = 1.0 / (double)zperiod;
+      ZMAX = 1.0 / static_cast<double>(zperiod);
```
d7919: `double` --> `BoutReal`?

src/mesh/impls/bout/boutmesh.cxx
```diff
@@ -227,7 +227,7 @@ int BoutMesh::load() {
 
     NXPE = -1; // Best option
 
-    BoutReal ideal = sqrt(MX * NPES / ((double) ny)); // Results in square domains
+    BoutReal ideal = sqrt(MX * NPES / static_cast<double>(ny)); // Results in square domains
```
d7919: `double` --> `BoutReal`?

src/solver/impls/arkode/arkode.cxx
```diff
@@ -427,11 +427,11 @@ int ArkodeSolver::run() {
                    nsteps, nfe_evals, nfi_evals, nniters, npevals, nliters);
       
       output.write("    -> Newton iterations per step: %e\n", 
-                   ((double) nniters) / ((double) nsteps));
+                   static_cast<double>(nniters) / static_cast<double>(nsteps));
```
d7919: `double` --> `BoutReal`?

src/solver/impls/cvode/cvode.cxx
```diff
@@ -348,14 +348,13 @@ int CvodeSolver::run() {
 
       output.write("\nCVODE: nsteps %ld, nfevals %ld, nniters %ld, npevals %ld, nliters %ld\n", 
                    nsteps, nfevals, nniters, npevals, nliters);
-      
-      output.write("    -> Newton iterations per step: %e\n", 
-                   ((double) nniters) / ((double) nsteps));
+
+      output.write("    -> Newton iterations per step: %e\n",
+                   static_cast<double>(nniters) / static_cast<double>(nsteps));
```
d7919: `double` --> `BoutReal`?

src/solver/impls/karniadakis/karniadakis.cxx
```diff
@@ -109,12 +109,12 @@ int KarniadakisSolver::init(int nout, BoutReal tstep) {
...
   // Number of sub-steps, rounded up
-  nsubsteps = (int) (0.5 + tstep / timestep);
-  
+  nsubsteps = static_cast<int>(0.5 + tstep / timestep);
+
   output.write("\tNumber of substeps: %e / %e -> %d\n", tstep, timestep, nsubsteps);
 
-  timestep = tstep / ((float) nsubsteps);
-  
+  timestep = tstep / static_cast<float>(nsubsteps);
```
d7919: `float` --> `BoutReal`?

d7919:
Whilst we can't get any extra benefit from using C++ casting with `malloc` (as it returns `void*` so has no type data for checking -- at least I think this is the reason for not changing the `malloc` casts ) it doesn't prevent us from using C++ style casts presumably? As such we could/could we adopt a style of using C++ style casts everywhere? Are there any downsides to this and have I missed a reason for sticking with C style casts with `malloc` (indeed is there a reason we can't replace these `malloc` uses with `new`)? Otherwise I think this looks fine (`double`-->`BoutReal` issues aside).

ZedThree:
Good catch on all the wrong types!

I was wondering whether we should just convert all the `malloc`s to `new`s? No casting required then.

ZedThree:
For the `fftw_malloc`s, there are convenience functions that return the correct type already: http://www.fftw.org/fftw3_doc/Memory-Allocation.html

Issue #655, opened 2017-08-15
Commit #626 breaks fci_slab


bshanahan:
#626 breaks fci_slab example.  I have a [branch](https://github.com/boutproject/BOUT-dev/tree/fci_slab_poloidal_grids) which is currently looking to fix this, but fci_slab still fails, and I haven't fixed all of the files such as generate.

d7919:
What is the mode of failure? Wrong result, doesn't compile, etc.?

bendudson:
Does it get first order convergence (in which case it might be a boundary-off-by-half-cell issue) or no convergence?

bshanahan:
No convergence.

bshanahan:
[fci-norm.pdf](https://github.com/boutproject/BOUT-dev/files/1228260/fci-norm.pdf)

d7919:
Can this be closed now?

bshanahan:
Yep sorry.  all fixed.

PR #626, opened 2017-07-21
Zoidberg poloidal grids

bendudson:
Zoidberg can now create 3D FCI grids using more general grids, in particular grids for the simulation of the edge of stellarators.

* Elliptic grid generator for annular regions
* Flux surface tracing to align grid boundaries to 3D shaped flux surfaces
* Improved handling of field lines which leave the domain
* Unit tests using pytest
* Examples under examples/zoidberg
* Manual section on Zoidberg

PR #681, opened 2017-08-23
Fci slab poloidal grids

bshanahan:
Fixes [Issue 655](https://github.com/boutproject/BOUT-dev/issues/655), ensuring fci_slab works for with the new poloidal grid system.

*also changes pass/fail criteria if the convergence is a nan (default fail instead of pass).  
*only works for y-periodic grids


bendudson:
Great! Does this pass now, or still failing?

bshanahan:
This passes with grids which are periodic in the y-direction.

bshanahan:
Despite the fact that I've asked for the merge, I'm wondering if this (and indeed [626](https://github.com/boutproject/BOUT-dev/pull/626)) should be merged since it does not converge for non-periodic grids (which was previously the default case).

dschwoerer:
I get this error if I try to run the test case:
```
Making fci-slab test
  File "./runtest", line 57, in <module>
    field = zb.field.Slab(Bz=0.05, Bzprime=0.1)
TypeError: __init__() got an unexpected keyword argument 'Bzprime'
```

Travis is complaining because there is some requirement for matplotlib ...

bshanahan:
First error: python path needs to lead to the newest zoidberg updates (PR 626) and I can't help with the second one, as I didn't add any matplotlib lines. I think any matplotlib stuff has been there as long as the test has been.

ZedThree:
Looks like only ones that pass are with `--enable-checks`, the rest hang on one of the tests (I'm guessing fci-slab?)

bendudson:
Last merge before 4.1...

bshanahan:
Would approve this but I can't as I've opened this PR (I think).

ZedThree:
I've just changed a bit how matplotlib gets imported, and how plotting functions get disabled if it can't be imported

PR #694, opened 2017-10-03
Bugfix: Checkout googletest if not available

bendudson:
When the unit tests are run, this makes sure that the googletest submodule is available. This uses git submodule --init to fetch googletest when needed.

A quirk of makefiles mean that the behaviour of this makefile depends on dependencies being checked from left to right.

This could go into "next", but users downloading master currently need to remember to check out the googletest submodule, or they will get an error when running "make check". Hence classing this as a bug and putting in master.

### DIFF
--- a/tests/unit/makefile
+++ b/tests/unit/makefile
@@ -40,10 +40,7 @@ clean ::
 # Download Google Test
 $(GTEST_SENTINEL) :
 	echo "Downloading Google Test"
-	curdir=$(PWD)
-	cd $(BOUT_TOP)/googletest
 	git submodule update --init --recursive
-	cd $(curdir)
 
 # For simplicity and to avoid depending on Google Test's
 # implementation details, the dependencies specified below are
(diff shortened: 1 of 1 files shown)


tests/unit/makefile
```diff
@@ -27,27 +27,39 @@ GTEST_HEADERS = $(GTEST_DIR)/include/gtest/*.h \
...
 # Builds gtest.a and bout_test_main.a.
 
-# Usually you shouldn't tweak such internal variables, indicated by a
-# trailing _.
-GTEST_SRCS_ = $(GTEST_DIR)/src/*.cc $(GTEST_DIR)/src/*.h $(GTEST_HEADERS)
+# Download Google Test
+$(GTEST_SENTINEL) :
+	echo "Downloading Google Test"
+	curdir=$(PWD)
+	cd $(BOUT_TOP)/googletest
```
d7919: I don't think you have to cd as the `submodule` command will update all submodules. Indeed I'm not sure if the `cd` does actually have any impact as each line in the recipe is executed in its own shell (as far as I understand).

bendudson: Good point, now removed.

PR #696, opened 2017-10-05
Multigrid solver memory leak bugfix

johnomotani:
Fixes memory leak in MultigridAlg::pGMRES which happened when the method
returned due to norm of right-hand-side being smaller than tolerance.


johnomotani:
This is a quick fix for a bug that Daniel Hoare found when using the multigrid solver. It came up because he is solving for the A_parallel with a symmetry plane where the current vanishes, so there is one point in y where the right-hand-side always vanishes. When that happens, the solver returns early, and did not reach the 'delete []' statements.

In the long run, it might be better to refactor the multigrid solver to remove the 'new's and 'delete's within each solve and instead allocate the memory in the constructor and delete it in the destructor. But that is more than just a bug-fix.

bendudson:
Thanks! I agree that the long-term solution is to remove new and delete calls. Preferably replaced by a container like vector (or the BOUT++ Array) to handle the memory management.

Issue #698, opened 2017-10-06
Field3D default move constructor leading to segfaults


d7919:
Just found an odd seg fault with code of the form

~~~~
a = func(b);
~~~~
where `a` and `b` are `Field3D` type and `func` returns a `const Field3D`. Depending on exactly what happens in `func` a seg fault arises. A backtrace suggests this is in a `Field3D` destructor when attempting to delete the `yup`/`ydown` fields. Suspect this is related to a move constructor being used leading to multiple pointers pointing at the `yup` etc. so end up with a double delete. Removing the `default` move constructor in `Field3d.hxx` seems to solve the issue.

dschwoerer:
fixed by #699

PR #699, opened 2017-10-06
Removing Field3D move constructor

bendudson:
Relates to issue #698, but is only a temporary fix.

The "default" move constructor seems to do something involving the yup/down fields which can in some cases lead to segfaults.

The longer-term solution is probably to implement a move constructor which takes into account pointers to
yup/down fields etc.

### DIFF
--- a/include/field3d.hxx
+++ b/include/field3d.hxx
@@ -184,7 +184,7 @@ class Field3D : public Field, public FieldData {
   /// Constructor from 2D field
   Field3D(const Field2D& f);
   /// Constructor from value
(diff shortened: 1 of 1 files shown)


d7919:
There's also a move assignment operator, I'm not sure if this causes similar issues but I imagine it could.

d7919:
Should the `Field2D` and `FieldPerp` equivalents be removed as well?

bendudson:
Probably, but Field3D is the most complicated of the three since it has the yup/down fields as well as the ddt field.

PR #703, opened 2017-10-10
Make shiftedmetric use const refs when possible

d7919:
The `const` qualifier on some arguments to `shiftZ` was missing in the header.

Also passing these const fields by reference to avoid a copy operation.


d7919:
This could be classed as a bug fix (inconsistent header) so should perhaps also go into master?

bendudson:
Yes, agree this is a bugfix bugfix, and should go into master. Thanks David!

d7919:
OK thanks, I've updated the base.

PR #705, opened 2017-10-10
Ensure VarStr entries are initialised in all cases

d7919:


d7919:
I set the vector location to `CELL_DEFAULT` which is a semi-forbidden value I think. Really I think something like a `CELL_NULL` or `CELL_UNDEFINED` might be more appropriate (or if we require the components to be at the same location, possible sensible, then we could get the location of `v.x` instead).

PR #707, opened 2017-10-10
set PYTHONPATH for tests

### LINKED ISSUE #695: PYTHONPATH in tests
PYTHONPATH has to be set correctly for make check-integrated-tests to work. I can imagine a potential for confusion if someone has multiple copies of BOUT on the same system, and tries to test one while PYTHONPATH points to the tools/pylib of a different (e.g. older) version. Is there a way to override it in the makefile so that the version of tools/pylib belonging to the copy of BOUT being tested is used, instead of whatever is in the user's PYTHONPATH, or at least print some sort of warning if PYTHONPATH doesn't point to the copy of BOUT being tested?

dschwoerer:
based on fix proposed by David Dickinson
fixes #695

### DIFF
--- a/makefile
+++ b/makefile
@@ -22,6 +22,8 @@ check-mms-tests:
 	@cd tests/MMS; ./test_suite
 
 check-integrated-tests:
-	@cd tests/integrated; ./test_suite_make && ./test_suite
+	@cd tests/integrated; ./test_suite_make
+	@cd tests/integrated; PYTHONPATH=${PWD}/../../tools/pylib/:${PYTHONPATH} ./test_suite
+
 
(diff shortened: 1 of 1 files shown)


bendudson:
Should this go into master? It's not quite a bug, but may prevent some problems

ZedThree:
It's sufficiently bug-like that I reckon it should go into master

ZedThree:
Changing the base to master brings in some extra commits. This was happening to me on the PRs I just submitted too -- not sure what's happened. I'll rebase it manually so it just moves the one commit, which will involve a separate PR.

ZedThree:
Replaced by #713

Issue #594, opened 2017-07-04
-DCHECK changes memory layout


dschwoerer:
After hours of debugging, I noticed that `-DCHECK=...` can change the memory layout of bout objects.
There are some other flags as well `-DTRACK`.

While this seem perfectly fine in most current use cases, it is an issue if you want to build anything that wants to link against bout++. In which case, the specific set of `-D` flags needs to be provided for every project down the line, and everything needs to be recompiled, if they change.

One option would to put all optional elements at the end of an object, but that doesn't help:
* if there is more than one optional element
* if there are derived objects
Therefore I think they should all be included always ...

Another option, which wouldn't solve the recompile issue, would be to put all `-D` flags in an file (maybe `include/bout/config.hpp`) that is always included.
This would ensure that if bout is included from another project, it doesn't have to add all the `-DCHECK` and so on from the bout library as flags.

<s>As part of the decisions are header files only, different projects could be compiled with different levels of checking (not sure this causes no bugs - probably not a good idea ...)</s>

d7919:
Should these flags be provided by `bout-config`? 

Could you expand a little on where this is an issue; presumably if you're just linking against the BOUT++ library then you don't care what these flags are, so is this a case of trying to link specific object files (possibly individually compiled with a different set of flags) into a different project or am I missing something (likely)?

dschwoerer:
If you link against the bout library, you probably want to use functions, so if you define "linking against the bout library" to include using (and therefore compiling) functions like `Field3D::getNx()` - then, this is an issue.
For that you need the header files, which is the issue, as they provide inline functions which are flags dependent.

Using `bout-config` works - so this is not a bug, but I guess rather a feature request.

Background: I am trying to call these functions from cython - which means in the future, a change in the flags requires to rebuild + regenerate the cython files, which is possible, but I don't think it is ideal ...

d7919:
This is a little similar to common usage for PETSc, where there may be several different configurations available (the PETSC_ARCH options) but users of the library don't have to worry too much about the configuration beyond saying what PETSC_ARCH is. They get around this by making you (or providing) things that can be included and contain all the relevant settings, as you've suggested earlier. To do this I think we'd probably want to "namespace" all the BOUT flags to ensure no conflicts etc.

[...]

PR #709, opened 2017-10-12
create settings.hxx during configure

dschwoerer:
Possible solution for #594 

I think it is much easier to include this header file, then making sure anyone using BOUT++ as a library drags along all flags and options.
Also probably easier if BOUT++ should be normally installed.

I am not sure it is sufficient to include the `settings.hxx` from `bout.hxx` - so this probably needs more thought - but before that I wanted to know whether you think this might be a viable solution.

Note there is no "namespace" for the flags - so they could interfere with other projects.

### DIFF
--- a/configure
+++ b/configure
@@ -5773,6 +5773,12 @@ fi
 
     { $as_echo "$as_me:${as_lineno-$LINENO}: checking for libfftw3" >&5
 $as_echo_n "checking for libfftw3... " >&6; }
+    ac_ext=cpp
+ac_cpp='$CXXCPP $CPPFLAGS'
+ac_compile='$CXX -c $CXXFLAGS $CPPFLAGS conftest.$ac_ext >&5'
+ac_link='$CXX -o conftest$ac_exeext $CXXFLAGS $CPPFLAGS $LDFLAGS conftest.$ac_ext $LIBS >&5'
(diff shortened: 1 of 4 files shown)


ZedThree:
Not sure about this. I think if we're going to have this kind of header, it should be generated by an existing tool (e.g. autoheader). See, e.g. https://stackoverflow.com/a/19694763/2043465
Or full blown autotools, see https://www.lrde.epita.fr/~adl/autotools.html

I'm currently investigating [meson](http://mesonbuild.com/index.html) as a wholesale replacement for the current build system.

dschwoerer:
I had a look at autoheader, and it seems like an overkill.
Futher, it requires a few changes, as e.g. HAVE_HDF5 would be defined instead of HDF5.
The SO answer also does not recommend autoheader, if I read it correctly.

I don't know meson, so cannot say anything about that ...

ZedThree:
Good idea, but not in this form. I think we should move towards a settings header, but I recommend the following:

1. Generated through "usual" means, i.e. with `AC_CONFIG_FILES`
2. Used consistently throughout the project
3. Change all the current macros to be prefixed with `BOUT`, e.g. `BOUT_HAS_HDF5`

We could do this as part of moving all the headers to a consistent namespace.

PR #712, opened 2017-10-13
A few small fixes and tests for Output/ConditionalOutput

### LINKED ISSUE #693: Unit test seg fault (`Output::cleanup`) -- order dependent
I've recently come across a unit test error (a seg fault, rather than a fail) when running on a local system which displayed some unusual features, mainly that it wasn't always reproducible.

It turns out the google test framework doesn't guarantee a particular order for the individual tests to be carried out and that for certain orders there was no issue but for other orders the seg fault would occur (the flags `--gtest_shuffle` and `--gtest_repeat=10000` are helpful, see later). This indicates that there is some form of memory/dependency between the unit tests which is not ideal.

After a lot of experimentation I think I've narrowed the underlying issue to the `Output` and `ConditionalOutput` classes. Essentially the global/static `ConditionalOutput` instances (e.g. `output_info`) are created/initialised with a base of `Output::getInstance()`, i.e. the global `Output` object. This `base` pointer is then used to pass strings to the underlying object for writing. However, in the test `CleanupAndGetInstance` of the `Output` tests we call `Output::cleanup()` which deletes this base and then it produces a new instance by calling `Output::getInstance()`. Whilst (I think) this is sufficient to ensure that all `Output` objects should work, the global `ConditionalOutput` objects end up in a state where they have an invalid pointer. This then ends up causing memory corruption which leads to the unpredictable seg fault at some other random point in the code.

I'm not sure what the best fix is. One thing that does seem to work is to add a function like

~~~~
void ConditionaOutput::resetBase(){
    base = Output::getInstance();
};
~~~~

(and equivalent for `DummyOutput`) which can be called to ensure the `base` pointer is pointing to the current `Output` instance. I then added the following to the end of the cleanup test:

~~~~
  output.resetBase();
  output_info.resetBase();
  output_progress.resetBase();
  output_warn.resetBase();
  output_debug.resetBase();
~~~~

Obviously this is less than ideal, with us having to reset each instance of a conditional/dummy output and having to remember to do this after cleanup as long as a new instance has been created etc.

Whilst we're unlikely to hit this sort of bug in current use cases, I think trying to ensure that we handle issues like this where objects are in an invalid state during/after cleanup is a good idea. It improves the robustness of the code and should enable alternative use designs (e.g. having a `BOUT++` kernel inside a larger code, being able to cleanup and restart can be helpful for ensuring a clean workspace etc.).

I'd also propose that we try to catch such types of problem by adding a travis build object (or modifiying an existing one) that runs the unit tests with the additional flags `--gtest_shuffle --gtest_repeat=1000`. With the optimised build the unit tests take around `20ms` to run once, so this should only add around 20s to the build.

ZedThree:
- Squash some unused parameter warnings for `DummyOutput`
- Add tests for `ConditionalOutput` and `DummyOutput`
- Previously, it was an error to `cleanup` an `Output` object and then use a `ConditionalOutput` object that had the `Output` singleton as the base. Rather than adding lots of machinery to check the base, or to refactor `ConditionalOutput`, just remove `Output::cleanup` and change how the `Output::instance` is created/stored. Fixes #693

The last problem is really only an issue for the unit tests that call `Output::cleanup` and then try to use the singleton afterwards, I doubt anybody's doing this in practice.

### DIFF
--- a/include/output.hxx
+++ b/include/output.hxx
@@ -216,7 +216,7 @@ template <typename T> DummyOutput &operator<<(DummyOutput &out, const T *UNUSED(
 // Function pointer so we can apply unused macro to pf in function below
 using stream_manipulator = std::ostream &(*)(std::ostream &);
 
-inline DummyOutput &operator<<(DummyOutput &out, stream_manipulator *UNUSED(pf)) {
+inline DummyOutput &operator<<(DummyOutput &out, stream_manipulator UNUSED(pf)) {
   return out;
 }
 
--- a/include/output.hxx
+++ b/include/output.hxx
@@ -220,7 +220,7 @@ inline DummyOutput &operator<<(DummyOutput &out, stream_manipulator UNUSED(pf)) 
   return out;
 }
 
-inline ConditionalOutput &operator<<(ConditionalOutput &out, stream_manipulator *pf) {
+inline ConditionalOutput &operator<<(ConditionalOutput &out, stream_manipulator pf) {
   if (out.isEnabled()) {
     *out.getBase() << pf;
   }
--- a/include/output.hxx
+++ b/include/output.hxx
@@ -103,11 +103,8 @@ public:
   }
 
   static Output *getInstance(); ///< Return pointer to instance
-  static void cleanup();        ///< Delete the instance
 
 private:
-  static Output *instance; ///< Default instance of this class
-
   std::ofstream file;                 ///< Log file stream
   static const int BUFFER_LEN = 1024; ///< default length
   int buffer_len;                     ///< the current length
--- a/include/output.hxx
+++ b/include/output.hxx
@@ -32,7 +32,10 @@ class Output;
 #include "multiostream.hxx"
 #include <iostream>
 #include <fstream>
+#include <functional>
 #include "boutexception.hxx"
+#include "unused.hxx"
+
 using std::endl;
 
 /// Class for text output to stdout and/or log file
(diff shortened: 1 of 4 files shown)


include/output.hxx
```diff
@@ -195,32 +195,32 @@ public:
...
-template <typename T> DummyOutput &operator<<(DummyOutput &out, const T *t) {
+template <typename T> DummyOutput &operator<<(DummyOutput &out, const T *UNUSED(t)) {
   return out;
 }
 
-inline DummyOutput &operator<<(DummyOutput &out, std::ostream &(*pf)(std::ostream &)) {
+// Function pointer so we can apply unused macro to pf in function below
+using stream_manipulator = std::ostream &(*)(std::ostream &);
+
+inline DummyOutput &operator<<(DummyOutput &out, stream_manipulator *UNUSED(pf)) {
```
dschwoerer: This is now broken.
Not quite sure why, but I think it is one * more then before.

tests/unit/sys/test_output.cxx
```diff
@@ -125,16 +126,184 @@ TEST_F(OutputTest, DisableEnableStdout) {
...
+  EXPECT_FALSE(local_output.isEnabled());
+  local_output.enable(true);
+  EXPECT_TRUE(local_output.isEnabled());
+}
+
+TEST_F(OutputTest, ConditionalJustStdOutCpp) {
+  Output local_output_base;
+  ConditionalOutput local_output(&local_output_base);
+
+  local_output << "Hello, world!" << 1 << "\n";
```
dschwoerer: `"\n"` could be replaced by `std::endl` to make sure that is working as well.

tests/unit/sys/test_output.cxx
```diff
@@ -125,16 +126,184 @@ TEST_F(OutputTest, DisableEnableStdout) {
...
+  EXPECT_FALSE(dummy.isEnabled());
+  EXPECT_NO_THROW(dummy.enable(false));
+  EXPECT_FALSE(dummy.isEnabled());
+  dummy.disable();
+  EXPECT_FALSE(dummy.isEnabled());
+}
+
+TEST_F(OutputTest, DummyOutputStdOut) {
+  DummyOutput dummy;
+  dummy << "Vanish to the void\n";
```
dschwoerer: std::endl test is missing.

tests/unit/sys/test_output.cxx
ZedThree: Oh, to check those overloads. Good catch, thanks.

PR #713, opened 2017-10-13
Set PYTHONPATH for tests

### LINKED ISSUE #695: PYTHONPATH in tests

ZedThree:
Fixes #695. Based on fix proposed by David Dickinson. Replaces #707

### DIFF
--- a/makefile
+++ b/makefile
@@ -22,6 +22,8 @@ check-mms-tests:
 	@cd tests/MMS; ./test_suite
 
 check-integrated-tests:
-	@cd tests/integrated; ./test_suite_make && ./test_suite
+	@cd tests/integrated; ./test_suite_make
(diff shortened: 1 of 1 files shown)


dschwoerer:
Thanks for the rebase

Issue #584, opened 2017-06-21
Vector applyBoundary


bendudson:
The Vector2D class doesn't have an applyBoundary(string) member, so calls to

Vector2D v;
...
v.applyBoundary("neumann")

seem to instead call the Vector2D::applyBoundary(bool) function:

https://github.com/boutproject/BOUT-dev/blob/next/include/vector2d.hxx#L149

This means that either no boundary is applied, or probably a different boundary is applied than intended. At the moment the solution is

v.x.applyBoundary("neumann")
v.y.applyBoundary("neumann")
v.z.applyBoundary("neumann")

which calls the Field2D functions

PR #716, opened 2017-10-14
Adding Vector applyboundary(string)

bendudson:
Fixes issue #584, and includes a unit test for this problem.

* Added unit tests for Vector2D::applyBoundary(char*)
* Fixed small bug in setting GlobalN* in FakeMesh constructor
* Added method addBoundary() to FakeMesh, to allow testing of
  boundaries


ZedThree:
This just needs to wait for 4.1.1 to be merged back into next, to fix the one failing test

PR #717, opened 2017-10-16
Update user manual: --enable-checks is the configure option for checking

JosephThomasParker:
--enable-checks has replaced --with-checks as the configure option for checking.

The world's most minor PR.


bendudson:
Can this change go into the sphinx docs too, since that's the live manual? Should the tex version be removed?

d7919:
If sphinx docs are complete then I'd vote that we remove the tex version.

ZedThree:
Oops, I didn't catch that it was going into the tex docs. So much for code review!

+1 for removing tex version altogether.

PR #721, opened 2017-10-17
Manual fix

dschwoerer:
+ fixes from master into next

### DIFF
--- a/manual/sphinx/user_docs/advanced_install.rst
+++ b/manual/sphinx/user_docs/advanced_install.rst
@@ -132,12 +132,12 @@ To build PETSc without SUNDIALS, configure with
       --with-scalar-type=real \
       --with-shared-libraries=0
 
-Add ``–with-debugging=yes`` to ``./configure`` in order to allow
+Add ``--with-debugging=yes`` to ``./configure`` in order to allow
 debugging.
 
 | To build PETSc with SUNDIALS, install SUNDIALS as explained in section
   :ref:`sec-sundials`, and append ``./configure`` with
-  ``–with-sundials-dir=$HOME/local``
+  ``--with-sundials-dir=$HOME/local``
 |  
 | It is also possible to get PETSc to download and install MUMPS (see
   :ref:`sec-MUMPS`), by adding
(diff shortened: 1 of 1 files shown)


dschwoerer:
yes, but the fixes should anyway be in next, as they are (important) bug fixes.

It wasn't really intentionally, but I think they should be in next anyway, so I don't see any need to rebase.

If you shouldn't like the mixes, I can open a second PR, just for the bug fixes ...

ZedThree:
4.1.1 will hopefully be out today, and then I'll merge master into next. You can wait till then and just rebase this if you like

dschwoerer:
I thought fixes could be merged immediately into next (as it is used for production runs), but I don't mind waiting a bit ...

ZedThree:
Are the changes in this PR just in cab3722? If so, feel free to add that commit to #722.

dschwoerer:
Yes, that is the only new one.

I don't mind having that sit around or being ignored - what I don't get is why there should be any issues with merging master into next?
Why should I not benefit from the fixes in master?
I know, this has been discussed before, but it still seems very confusing to me, why I should manually merge, or produce potentially wrong results.

d7919:
_This is just my understanding and this may not be policy._
I think the idea is that tagged releases are what should be used for any science production -- using anything else, such as next, may be possible but it is not a promise. In practice we do/should try to get bug fixes into next reasonably regularly so that using next is usually not a problem (i.e. no known fatal bugs), but rather than doing it with each individual bug fix we update when there's an update to the tagged release (then available in master). This way next can be thought of as "tagged release PLUS new features MINUS old features" rather than "current master PLUS new features MINUS old features".

dschwoerer:
I think there are benefits of fixing bugs in next (why should anyone use knowingly broken code).

The point is I don't see any reason to not merge master into next.

So I have zero understanding as to why we shouldn't merge.
If you want, I can elaborate on the benefits.

ZedThree:
Most people want to be using master rather than next. There's actually very few people who both develop and use BOUT++, and so want to be using next. Therefore, we definitely want bug fixes to go into master.

In principle, it would be great if bug fixes went into both master and next. In practice, this means having to open two PRs, one for each branch, and I think most of us just forget to open both. So our policy is for bug fixes to just go straight into master, everything else into next, and then we periodically merge master into next.

After new releases, there's also usually a short period where we catch loads of bugs we didn't notice, so there's a lot of activity for a couple of weeks before it dies down again. So, in order to cut down on noisy commits to next, we can hold off on merging master into next until it dies down (i.e. once we've released the next minor version).

dschwoerer:
So merging without an extra commit would be perfectly fine?

Wouldn't a merge like this then be perfect? No extra commit message for the master to next merge?

PR #725, opened 2017-10-19
Update showdata

AdamD94:
- Modified showdata to allow fps, dpi and colormap to be passed as parameters.
- Allowed the animation object to be returned for further manipulation, eg. conversion to html5 video.
- Changed the cla() calls slightly by allowing default behavior to be overridden by setting a parameter when calling showdata.
- Also showdata now searches for ffmpeg by string if it isn't initially found.

### DIFF
--- a/tools/pylib/boututils/showdata.py
+++ b/tools/pylib/boututils/showdata.py
@@ -19,7 +19,7 @@ except:
 #import numpy as np
 from mpl_toolkits.mplot3d import axes3d
 from matplotlib import pyplot as plt
-from matplotlib import animation
+from matplotlib import animation, cm
 from numpy import linspace, meshgrid, array, min, max, floor, pi
 from boutdata.collect import collect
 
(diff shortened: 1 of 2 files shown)


tools/pylib/boututils/showdata.py
```diff
@@ -63,17 +63,30 @@ def showdata(vars, titles=[], legendlabels = [], surf = [], polar = [], tslice =
...
 
     global_colors = True: if "vars" is a list the colorlevels are determined from the mximum of the maxima and and the minimum of the  minima in all fields in vars.
 
     symmetric_colors = True: colorlevels are symmetric.
+
+    clear_between_frames: None all plots except line plots will clear between frames
+                          True all plots will clear between frames
+                          False no plots will clear between frames
+
+    return_animation = True matplotlib animation object will be returned.
```
dschwoerer: I guess there is no reason to not return it ...
If you don't want it, you just ignore the return value ...

AdamD94: You might be right. I was trying to keep the default behavior consistent. Previously if the animation was saved to a file the plot wouldn't be shown. I thought it might make sense to only show the plot on screen if it wasn't being otherwise saved or returned.

PR #726, opened 2017-10-20
Avoid unknown omp pragma warnings

d7919:
Adds a macro, `BOUT_OMP`, that can be used instead of `#pragma omp .....`. This allows us to remove these pragmas when building without openmp support, and hence avoid lots of unknown pragma warnings. I think this would make the build quiet enough that warnings could be enabled by default.

An example of usage is

To replace
~~~~
#pragma omp critical(store)
~~~~

use 
~~~~
BOUT_OMP(critical(store))
~~~~


dschwoerer:
My PR was rejected (#351)
Maybe you have more look - I am in favour of the PR 👍 

PS: I guess your comment should read `BOUT_OMP(...)`

d7919:
Thanks for spotting the typo. 

I didn't realise that #351 included a fix for the OMP pragma warnings. Looking at the comments there it looks like it was closed because it was thought that the elements of this PR had been split into separate PRs that had been accepted, but it doing so it looks like this OMP pragma fix was missed.

dschwoerer:
Note that the tests are broken:
https://travis-ci.org/boutproject/BOUT-dev/jobs/290533452

Not sure wheter before `-Wunknown-pragams` was disabled or this is introduced by this, see e.g.
https://travis-ci.org/boutproject/BOUT-dev/jobs/290487844

d7919:
That's odd -- nothing should have changed with building the unit tests but it does suggest that they are being built without passing the appropriate flags for openmp. It looks like `-Wall` and `-Wextra` (amongst other flags) are forced to be on for the unit test builds -- in `/tests/unit/makefile` they're appended to `CXXFLAGS`. I'd guess this isn't ideal as it may cause problems when using other compiler families. 

I've had some local issues with the `include` statement in `tests/unit/makefile` which were solved by moving this nearer to the top of that file. I wonder if that might help here.

Ensuring `-fopenmp` is in both `BOUT_FLAGS` and `LDFLAGS` I don't see these warnings on a local machine with gcc v5.4.

I also think the `bout_test_main` targets probably need to have a dependency on the unit test makefile as currently they're not rebuilt when the makefile changes, unlike the other files involved.

d7919:
The other motivation for this is that it's not possible to include pragmas inside macros without using this sort of approach.

d7919:
This should now be ready, removing in progress.

d7919:
Resolved conflicts.

d7919:
Is there anything I can do to make this suitable for merging or is this approach to the omp pragmas a "won't merge"?

I should note that as well as avoiding the large number of unknown pragma warnings, this also enables easy use of omp inside macros (the statements are identical inside and outside macros). Without this have to adopt `_Pragma("omp ....")` type statements inside macros.

PR #732, opened 2017-10-24
allow collect to suceed if not all writes to the files have finished

dschwoerer:
Sometimes the output files can be at different time slices, e.g. if the simulation was killed during writing the last write or if it is still running.

This patch reads all input files, and extracts the minimum of available time slices, if `tind_auto` is set to `True` in `collect`.

### DIFF
--- a/tools/pylib/boutdata/collect.py
+++ b/tools/pylib/boutdata/collect.py
@@ -64,7 +64,7 @@ def findVar(varname, varlist):
         print("Variable '"+varname+"' not found, and is ambiguous. Could be one of: "+str(v))
     raise ValueError("Variable '"+varname+"' not found")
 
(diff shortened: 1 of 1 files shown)


tools/pylib/boutdata/collect.py
```diff
@@ -86,6 +86,8 @@ def collect(varname, xind=None, yind=None, zind=None, tind=None, path=".",yguard
                            definition of nx)
     info    = True         Print information about collect?
     strict  = False        Fail if the exact variable name is not found?
+    tind_auto = False      Read all files, to get the shortes length of time_indices
```
d7919: Typos "shortes" --> "shortest", "usefull" --> "useful".

PR #738, opened 2017-11-01
Delete boundary regions in test_vector?d when tearing down test

ZedThree:
This makes them clean for AddressSanitizer


bendudson:
Longer term, the boundary regions will probably be changed to use index ranges, and should probably be tidied up in the Mesh destructor

PR #740, opened 2017-11-06
Nonuniform single option

bendudson:
As pointed out by Seto-san, these settings should be consistent, and having two settings is not useful in most cases.

This change keeps both settings, but by default the Laplacian setting follows the global one. If a user wants to override this default then they still can set Laplacian:nonuniform.


manual/sphinx/user_docs/differential_operators.rst
```diff
@@ -153,6 +154,9 @@ The correction factor is then calculated from ``d2x`` using
 
    {{\frac{\partial }{\partial i}}}(\frac{1}{\Delta x}) = -\frac{1}{\Delta x^2} {{\frac{\partial \Delta x}{\partial i}}}
 
+**Note**: There is a separate switch in the :ref:`Laplacian inversion code <sec-laplacian>`,
```
d7919: Should this include something like "overrides" or mentions that is defaults to the global/mesh setting?

Issue #741, opened 2017-11-06
Setting derivatives


dschwoerer:
Setting derivatives for the staggered derivatives seems to be somewhat broken in the bout mesh:
```
  output_info.write("\tSecond      : ");
  derivs_set(options, SecondDerivTable, "second", fd2d);
  if (StaggerGrids) {
    output_info.write("\tStag. Second: ");
    derivs_set(options, SecondStagDerivTable, "second", sfd2d);
  }
```
This prevents from using C4 - as there is no C4 second derivative implementation with only 2 guard cells.
For some reason the initialisation works never the less ...

Earlier I used in the aiolos mesh (e.g. in d7e592da32e032adf1c49653da698ca77588beaa):
```
  Options *dirOption;
  Options *defOption = option->getSection("diff");
  output.write("\tSetting derivatives for direction x:\n");
  dirOption = option->getSection("ddx");

  // This file is auto-generated - do not edit!
  // Setting derivatives for ddx and FirstStag
  if (dirOption->isSet("FirstStag")) {
    dirOption->get("FirstStag", name, "");
  } else if (dirOption->isSet("First")) {
    dirOption->get("First", name, "");
  } else if (dirOption->isSet("all")) {
    dirOption->get("all", name, "");
  } else if (defOption->isSet("FirstStag")) {
    defOption->get("FirstStag", name, "");
  } else if (defOption->isSet("First")) {
    defOption->get("First", name, "");
  } else if (defOption->isSet("all")) {
    defOption->get("all", name, "");
  } else {
    name = "C2";
  }
```
which is a slightly longer, but allows to set 'default' options:
```
[mesh:diff]
all=c4
secondStag=c2
flux=U1
[mesh:ddx]
first=c2
```
As there are `3*2*4=24` possible derivatives, it comes otherwise rather tedious to set them ...

dschwoerer:
Fixed by #1034

PR #742, opened 2017-11-07
Aiolos mesh

dschwoerer:
- [x] Depends on #815

--------------------------------

In case there is general interest in this implementation.

I hoped to clean-up the generating python code, but I am not sure I will find the time. As the python code is only used to generate the new code, I hope it is fine, as there is still the old bout-mesh available, so there should be no drawback ...


dschwoerer:
I was surprised to see just how many changes have collected over the time since I started this branch.

I added the generated code - to help with review ...

ZedThree:
The Travis tests have failed because it's using python 3. Quick fix would be to run `2to3` over the python stuff -- it's mostly choking on `print` as a keyword rather than a function, which `2to3` should be able to fix.

You might also try `flake8` (`pip3 install --user flake8`) which can catch lots of other errors/warnings

dschwoerer:
Seems like the OpenMP dataiterator was broken all along - but the test was disabled ...

bendudson:
I wouldn't put too much effort into fixing it at the moment. There's work at the moment on the iterators, including lots of changes for OpenMP in various branches.

dschwoerer:
By the time I noticed it had been broken all along I had found the bug, fixing then only took seconds ....

I think [hpx](http://stellar.cct.lsu.edu/projects/hpx/) [sounds](https://github.com/STEllAR-GROUP/hpx) interresting as a replacement for OpenMP - but I didn't found the time to have a try with it so far ...

ZedThree:
What's the status of this? New commits keep going in, and I can't tell if they're features or bug fixes. Have all the independent bits been pulled out into separate PRs? Ideally, the PR for this feature would add only the new mesh and nothing else.

It's still very hard to understand, both the generated code and the code generation. I assume it's correct seeing as the tests pass, and the speed test shows that it can be faster than BoutMesh, so I'm looking at this from a maintainability point of view. It needs to be possible for other people to understand and be able to make changes. Obviously you're keen to get it in, but it's no good if you're the only person who understands it.

You might even think about completely rewriting it in a new branch with the aim of simplifying everything, now that you know what end result should look like -- this is fairly standard practice anyway. If you can use Jinja for the code gen bits, even better.

Some specific comments:
- Is there a reason `AiolosMesh::derivs_init` needs to be generated? Could it not just use `lookupFunc` (or your own version) from `index_derivs.cxx` to set the default methods? That would *massively* simplify at least `derivs_init`. I'm not sure there's any particular need for the setup to be blazing fast
- It's not clear what "on" and "off" versions of functions are supposed to do
- In some places you've used rational numbers as the weights and in others the decimal form -- I would probably just stick to rational numbers. It's easier to see if `9. / 16.` is right compared to `5.62500e-01`. The compiler should be clever enough to deal with this as well

Some general comments:
- Please try to document how the different parts fit together -- what order the different python files are called in, what each part is supposed to generate, how the different generated functions call each other, etc.
- In Python code, try to keep all the classes and functions together, with the executable bits at the bottom, preferably wrapped up in `if __name__ == "__main__"`. Mixing function definitions and executable statements makes for hard to read code. 
- Try not to have so many nested if-statements and for-loops, it gets very difficult to work out what's going on
- Try to avoid abbreviating variable names needlessly. It just increases the mental energy required to read the code
- Try to stick to PEP8 for Python code. Whitespace around equals signs and commas makes code easier to read. There are lots of tools out there to format Python code for you

The Zen of Python is useful https://www.python.org/dev/peps/pep-0020/ :)

dschwoerer:
Some of the commits that have been coming in where fixes after merging.

I have started to cleanup the code - but that is plenty of work.
Not everything has been pulled out.
For example I added virtual to many things in BoutMesh - as without the derived mesh they are not useful for the BoutMesh. I could copy them out, to make it easier to see that this does not change anything.
Besides that I think #815 has been merged into this branch (which simplified this code) and #840 has been copied out. If the two got merged, this should reduce the changes - but besides that I think the only thing that could be pulled out is a change to python3 in `tests/integrated/test_suite_make`

I will keep working on the generating code, as it is still quite confusing ... even for me :(

-----------------------

Concerning the on and off version - the on version differentiates from a non-staggered mesh **on** the staggered mesh, and the off version from the staggered mesh **off** (onto to the non-staggered)

The generated_init could be simplified - there is probably no need to have the code three times.
But the code is significantly smarter than what the BoutMesh is doing. It might have been broken by some changes introduced in the mean time, but it allowed to set all second derivatives with a single command, rather then 6. As there are 24 possible things to set, I think this is needed, as discussed in #741 
Speed was here not the motivation. By keeping it auto-generated, this will always keep in sync with the available stencils.

Note that if the number is calculated by the code, it is normally a decimal number How do I convert that to a fraction?.
However, most stencils are just `copied` from `derivs.cxx` - so you would need to change it there.

dschwoerer:
I have changed the `generated_init.cxx` so that the for loops are all done in C++.
Can we get something similar for `BoutMesh`?

https://github.com/boutproject/BOUT-dev/pull/948 made #741 much much worse.

Before I could set `second` to `C4` and `secondStaggered` to `C2`.
With this change this became impossible, as the `BoutMesh` throws, as it does not check the `SecondStaggered` - and thus even tough this is never used, it throws.

dschwoerer:
The templated derivatives are of similar performance, so I am going to close this ...

Issue #262, opened 2016-09-27
Laplace (Field3D) broken


JensMadsen:
Running e.g. **examples/bout_runners_example/3D_diffusion**: 

'Laplace( Field3D )
Sorry, D2DXDY not implemented'

This error appeared after git pull this morning (do not know tag of latest working version, sorry)

bendudson:
This is because D2DXDY used to just return zero, but now throws this error message. The possibilities are:

a) The metric tensor components g12 (x-y) are zero for all coordinates used in practice (as far as I know), so we could remove this term from the Laplace operator. This might then surprise someone later if they want g12 != 0

b) We could implement D2DXDY, which would currently require taking derivatives in one direction, communicating, and taking derivatives the other way. Most of the time this work would be done, and then set to zero anyway.

Would it be useful to give metric tensors a flag (in Field2D?) to specify that they are zero. This could then be used in operators like Laplace to decide whether to evaluate a term. 

bool Field2D.isNonZero();     Return true if not zero

then have functions in Field2D

void Field2D.markZero();  // Mark this field as zero. The code could check values, and throw an error if not zero, or maybe set itself to zero.

void Field2D.markNonZero(); // Set back to non zero. This would be the default

JensMadsen:
**Laplace** should do exactly a Laplace operation so solution a) opens up for future bugs. Thererefore, I vote for a solution in line with Ben solution "c"): 

> Would it be useful to give metric tensors a flag (in Field2D?) to specify that they are zero. This could then be used in operators like Laplace to decide whether to evaluate a term.

> bool Field2D.isNonZero(); Return true if not zero

> then have functions in Field2D

> void Field2D.markZero(); // Mark this field as zero. The code could check values, and throw an error if not zero, or maybe set itself to zero.

> void Field2D.markNonZero(); // Set back to non zero. This would be the default

This kind of solution could also save a small amount of computational time in shifted metric cases where g^{xz} is also zero but is calculated in e.g. 'Laplace()'

dschwoerer:
Concerning a)
It would be possible to check if g12 is zero, and throw an error if not ... it still needs implementing in the future ....

Concerning b)

It would be possible to communicate the corner cells without much overhead. At the X-point there is more than one option what to communicate, but then the D2DXDY has also this issue, as near the X-Point the symmetry of second derivatives is broken ...

Concerning markZero:
[...]

PR #743, opened 2017-11-14
Const fields

dschwoerer:
Potential solution for #262 

It adds isConstant() to fields, to check e.g. if the field is everywhere zero

It still needs:

- [ ] tests
- [ ] be used in the code
- [ ] documentation

Known issues:
- [ ] `is_const` may be different on different nodes, if set with `makeAutoConstant()` (e.g. used in `coordinates.cxx` - if this is info is used to conditionally call MPI-active functions, e.g. `D2DXDY` this may lead to unexpected behaviour.


dschwoerer:
This is implemented on top of PR #458 
Otherwise it would require to many changes.

I am however not convinced this will result in a speedup. So more testing is needed ...

dschwoerer:
closed as there are to many conflicts

PR #447, opened 2017-01-30
Y derivatives transform toFieldAligned if no yup/down

bendudson:
o Added a function Field3D::hasYupYdown() which returns true
  if the field has valid yup() and ydown() fields

o indexDDY checks hasYupYdown(). If true, then uses them
  to calculate Y derivatives. If false, then uses mesh->toFieldAligned
  to transform into field aligned coordinates, perform derivative,
  and transform back.

This commit fixes many cases where a compound variable is differentiated
as discussed in issue #293

Grad_par(A + B) will now transform to field aligned coordinates,
failing if this can't be done (e.g. FCI).

Hopefully this reduces the number of additional communicate() calls required
in v4 compared to v3

Issue #293, opened 2016-10-11
FieldAligned operators in next


bendudson:
The next branch has changed the default coordinates, from being field aligned to being in orthogonal X-Z space. This removes FFT shifts when taking radial (X) derivatives, but adds them when taking parallel (Y) derivatives. It also adds overheads when taking derivatives of expressions:

```
Grad_par(A + B) 
```

Now either A+B must also operate on the yup() and ydown() fields, or an additional communication step and boundary condition is needed. 

1) In many cases this is just a matter of being more careful about which things are communicated, so writing Grad_par(A) + Grad_par(B). We should make sure that Grad_par(A+B) fails with a useful warning message, rather than producing a wrong answer

2) In some cases it is useful to have the field in field-aligned coordinates, particularly for flux conserving schemes. Perhaps it would be good to have a set of operators which use the toFieldAligned() and fromFieldAligned() functions:

Grad_par(f)   -- Use f.yup() and f.ydown(), needs communication 

FieldAligned::Grad_par(f)  -- Shift f to field aligned coordinates, don't use yup() or ydown(). No communication required 

for naming, we could either use namespaces or longer names like Grad_par_FA()?

ZedThree:
This should probably be documented in the change notes for v4.

I think that if the argument is a [*prvalue*][1], that's when we have problems. We could maybe put in some nature of detection for prvalues and warn/throw exception until this is fixed. See [here][2] for one way of doing this.

[1]: http://en.cppreference.com/w/cpp/language/value_category#prvalue
[2]: http://stackoverflow.com/a/16638081/2043465

ZedThree:
I say prvalue, that might be too specific - maybe just rvalues?

JosephThomasParker:
It's worth noting that this can have reproducibility implications for physics models as they change from v3 to v4.  While `Grad_par(U-V)` will be the same as `Grad_par(U)-Grad_par(V)`, numerically `Grad_par(n*T)` is not the same as `n*Grad_par(T)+T*Grad_par(n)`.  The two can even be qualitatively different, like having different [stability properties](http://www.ams.org/journals/mcom/1973-27-121/S0025-5718-1973-0395249-2/).

It would be nice to have a natural way of doing `Grad_par(n*T)` without having to make and communicate another `Field3D`.

Actually, I'm not clear what the correct way of doing `Grad_par(n*T)` now is.  Is it:
`Field3D nT = n*T;
mesh->communicate(nT);
[...]

PR #746, opened 2017-11-17
WIP: Fixes to enable staggered y-grids in field aligned coordinates

johnomotani:
This pull request continues the changes implemented in PR #447, addressing issue #293.

Update VDDY to follow the pattern of applyYdiff: (i) check whether yup and ydown fields are created and transforming to field aligned coordinates if not; also (ii) use DataIterator instead of bindex.

Update interp_to similarly.

Questions:
- should this PR be to master or next? To me it is a bug fix, because I think I need the updates to get STORM to work without adding extra communications, etc.
- It is nice to be able to use Field instead of Field3D/Field2D since then we have one VDDY(Field,Field) instead of four combinations, but the changes I made to get it work with 'Field' seem a bit forced?
    - add 'const Field3D toFieldAligned(const Field &f)' to mesh.hxx
    - add is3D(), yup() and ydown() to field.hxx
- Is there any test of interp_to anywhere? I made an error in it while implementing these changes (got stencils for staggered->unstaggered and unstaggered->staggered the wrong way around) that was not caught by unit, integrated or MMS tests...

### DIFF
--- a/include/field.hxx
+++ b/include/field.hxx
@@ -68,6 +68,7 @@ class Field {
 
   // Data access
   virtual const BoutReal& operator[](const Indices &i) const = 0;
+  virtual const BoutReal& operator[](const DataIterator &i) const = 0;
 
   virtual void setLocation(CELL_LOC loc) {
     if (loc != CELL_CENTRE)
@@ -143,6 +144,37 @@ class Field {
    * Return the number of nz points
    */
   virtual int getNz() const;
+
+  // Utility method to get result of Mesh::toFieldAligned for different derived classes of Field
+  virtual Field* toFieldAligned() const {
+    throw BoutException("not implemented in Field base class");
+  }
+
+  /*
+   * Interface for yup and ydown fields
+   */
+  /// Check if this field has yup and ydown fields
+  /// Should be overridden in derived classes
+  virtual bool hasYupYdown() const {
+    return false;
+  }
+  /// Return reference to yup field
+  virtual Field& yup() {
+    throw BoutException("This is not implemented: should be overridden in derived classes");
+  }
+  /// Return const reference to yup field
+  virtual const Field& yup() const {
+    throw BoutException("This is not implemented: should be overridden in derived classes");
+  }
+  /// Return reference to ydown field
+  virtual Field& ydown() {
+    throw BoutException("This is not implemented: should be overridden in derived classes");
+  }
+  /// Return const reference to ydown field
+  virtual const Field& ydown() const {
+    throw BoutException("This is not implemented: should be overridden in derived classes");
+  }
+
  protected:
   Mesh * fieldmesh;
   /// Supplies an error method. Currently just prints and exits, but
--- a/src/mesh/index_derivs.cxx
+++ b/src/mesh/index_derivs.cxx
@@ -1970,7 +1970,7 @@ const Field2D Mesh::indexVDDY(const Field2D &v, const Field2D &f, CELL_LOC outlo
 }
 
 // general case
-const Field3D Mesh::indexVDDY(const Field &v, const Field &f, CELL_LOC outloc, DIFF_METHOD method) {
+const Field3D Mesh::indexVDDY(const Field &v, const Field &f, CELL_LOC outloc, DIFF_METHOD method, REGION region) {
   TRACE("Mesh::indexVDDY(Field, Field)");
 
   ASSERT1(this == v.getMesh());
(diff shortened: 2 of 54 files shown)


ZedThree:
Only glanced over this for now, but on your last point: missing tests are bugs :)

dschwoerer:
There are tests in the aiolos mesh #742

include/field.hxx
```diff
@@ -143,6 +144,36 @@ class Field {
    * Return the number of nz points
    */
   virtual int getNz() const;
+
+  virtual bool is3D() const {
+    throw BoutException("not implemented in Field base class");
```
dschwoerer: why not set this to false?
then it only needs to overwritten once ...

src/mesh/interpolation.cxx
```diff
@@ -65,37 +65,93 @@ const Field3D interp_to(const Field3D &var, CELL_LOC loc)
...
+        if(var.hasYupYdown() &&
+            ( (&var.yup() != &var) || (&var.ydown() != &var) )) {
+          // Field "var" has distinct yup and ydown fields which
+          // will should be used to calculate interpolation along
+          // the magnetic field
+          throw BoutException("interp_to not implemented for fields with yup/ydown");
+        } else {
+          Field3D var_fa = mesh->toFieldAligned(var);
+          if (loc == CELL_YLOW) {
+            for(const auto &i : result) {
```
dschwoerer: You need handling of the boundaries.
This way it fails if the `DCHECK>2` and otherwise produces wrong results.

I had the same issue with the aiolos mesh ...

src/mesh/index_derivs.cxx
```diff
@@ -2038,14 +2138,65 @@ const Field3D Mesh::indexVDDY(const Field &v, const Field &f, CELL_LOC outloc, D
...
+          fval.pp = f[i.offset(0,2,0)];
+          fval.mm = f[i.offset(0,-2,0)];
+
+          result[i] = func(v[i], fval);
+        }
+      } else {
+        for (const auto &i : result) {
+          // Set stencils
+          stencil fval;
+          fval.c = f[i];
```
dschwoerer: setting the stancils seems like it involves copy/paste.
Can this be changed to autogenerated code, or macros?

src/mesh/interpolation.cxx
johnomotani: Thanks David! This is fixed by e801fc3

dschwoerer: e801fc3 doe only partially fix the issue:
 * I can still call with `RGN_ALL` to get an error.
 * The bondaries are kept unintialised, which may be ok, but I think STORM relies (or did) on the boundary cells being set. At least my version of STORM wouldn't work reliable, as the recycling depends on the flux on a staggered cell, which should be the first outside the domain. The old code copied the field first, so the boundaries where initialised with the unstaggered values, which hides the bug sufficiently well for most cases. The aiolos has forward and backward stencils, to set the boundaries properly, which I think is what the code should be doing ...

include/field.hxx
johnomotani: Setting to false as you say would make sense. The version in FieldData is

    virtual bool is3D() const = 0; ///< True if variable is 3D

So the function has to be implemented in every derived class.

But I think the design here needs some thought. There are both Field and FieldData base classes; Field3D and Field2D inherit from both, while FieldPerp only inherits from Field. The code in

    Field3D toFieldAligned(const Field &f)

basically needs to know if the Field f is a Field3D, Field2D or FieldPerp. At the moment it works 'correctly' because:
- for Field3D it calls toFieldAligned
- for Field2D it casts the Field to a Field2D and then creates a Field3D by assignment from the Field2D, then calls toFieldAligned on the Field3D
- for FieldPerp throws an exception because FieldPerp does not implement is3D()
but especially for FieldPerp, this is a strange and confusing way to do things. Maybe it would be better to make 'toFieldAligned' be a method of Field/Field3D/etc.? Then it would be easy to get the right method for each type from a generic Field.

johnomotani:
> Did you make sure non of these function might be overwritten?
> 
> I agree that it makes sense in a time critical place, but in general I am worried it may introduce bugs ...

I don't know. Apologies, I shouldn't really have added this to a PR I guess. Some change that I made (I think adding the 'virtual const BoutReal& operator[](const DataIterator &i) const = 0' to field.hxx) slowed things down quite a bit. Adding the final keyword in f7e1a54 restored the performance in test-interchange on my desktop machine, but STORM on Marconi was still running slower by about 2x (compared to the BOUT++ v3 version of STORM). So the last commit is just an attempt to track down what's causing that...

johnomotani:
PS if I understand correctly, 'final' should fail at compile-time if it is not correct. From http://en.cppreference.com/w/cpp/language/final:
> When used in a class definition, final specifies that this class may not appear in the base-specifier-list of another class definition (in other words, cannot be derived from). The program is ill-formed (a compile-time error is generated) otherwise.

dschwoerer:
You where quoting its use in a *class* definition. We are using it on members.
If you therefore add the keyword to the class itself, where you used them, you should get an error, if the member is tried to be overwritten, as in this case the class its self needs to be overwritten, which then would result in an compile time error.

johnomotani:
Sorry, quoted wrong part of that page. It says about methods:
> When used in a virtual function declaration or definition, final ensures that the function is virtual and specifies that it may not be overridden by derived classes. The program is ill-formed (a compile-time error is generated) otherwise.

dschwoerer:
Thanks for explaining 👍 
In this case it makes sense to replace override with final - as it is not ignored like a missing virtual ...

bendudson:
Replaced by #792

PR #747, opened 2017-11-17
Fix bug in infinite recursion in configure when netcdf not found

### LINKED ISSUE #744: ./configure deadlock
I tried to run configure on archer, but instead of failing, it called itself.
The configure command run is then:
```
  $ ./configure script found. Specify path using --with-netcdf option 5937 5
```
I guess the offending line is:
```
       [AC_MSG_ERROR([No configure script found. Specify path using --with-netcdf option])],
```
which results in:
```
test -n "$NCCONF" || NCCONF="as_fn_error $? "No configure script found. Specify path using --with-netcdf option" "$LINENO" 5"
```

Not quite sure why the configure script calls it self ...

ZedThree:
This was due to wrong use of AC_PATH_PROGS -- if the program isn't
found in path, the third argument is the *value* the variable is set
to, and not an action to be undertaken. Actually, we don't even want
to error at this stage, just disable netcdf.

Fixes #744

### DIFF
--- a/configure
+++ b/configure
@@ -5884,7 +5884,7 @@ if test "x$with_netcdf" != "xno"; then :
   # check the path provided by --with-netcdf, appending "/bin" if need
   # be, then check system path
   # Set NCCONF to the full path of the found scropt
-  case `basename $with_netcdf` in #(
+  case `basename $with_netcdf 2> /dev/null` in #(
   "ncxx4-config") :
     NCCONF=$with_netcdf ;; #(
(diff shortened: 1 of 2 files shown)


dschwoerer:
fixes #744 👍 
Thanks

dschwoerer:
Only thing: should it go into master, as it is a bug fix?

ZedThree:
Yes, that's true. I just need to rebase it properly

Issue #745, opened 2017-11-17
BOUT.log is hardcoded


dschwoerer:
If I use BOUT++ for my post processing, the BOUT.log.* are overwritten.
I can partailly prevent this by running with less cores, but it would be nice to disable or change the path of the log.

ZedThree:
Fixed in #751

PR #751, opened 2017-11-22
Adding -l,--log command-line argument

bendudson:
Allows the name of the log files to be changed. The default is set to "BOUT.log" in the DEFAULT_LOG variable. This can be changed, for example

    --log  log

would produce log files in the input/output directory ("data" by default) called "log.0", "log.1" etc.

This fixed issue #745

Includes an integrated test which checks -l, --log and -d options.
The test is quick, so added to the test suite.


tests/integrated/test-command-args/runtest
```diff
@@ -0,0 +1,87 @@
+#!/bin/bash
+
+# Compile command-args
+make || exit 1
+
+# Remove directories which may be left from previous run
+rm -r data
```
dschwoerer: we should add the `-f` flag to prevent interactive mode

tests/integrated/test-command-args/runtest
```diff
@@ -0,0 +1,87 @@
...
+rm -r test
+
+# Run without any directory, check that an exception is thrown
+echo "=========== Running without input directory =============="
+./command-args 2> stderr.log
+if ! grep -i "exception thrown" stderr.log > output.log ; then
```
dschwoerer: I would prefer to check the exit status.
The "exception thrown" may be added for a strange reason at some point in the future to the normal output.

tests/integrated/test-command-args/runtest
```diff
@@ -0,0 +1,87 @@
...
+echo
+echo "=========== Running without args in data directory =============="
+
+mkdir data
+cp BOUT.inp data
+./command-args 
+
+# Check that we have a BOUT.settings and BOUT.log.0 file
+
+if ! [ -e "data/BOUT.settings" ] ; then
```
dschwoerer: `-f` would be a tighter check then `-e` - making sure this is not a directory, symlink, etc ...

PR #752, opened 2017-11-22
Global mesh

dschwoerer:
Reduce dependency on the global mesh object.
Is part of #458 
This should not break anything. The global object is still there, but if a second mesh it used, the fields will stay on that mesh.

### DIFF
--- a/src/field/field2d.cxx
+++ b/src/field/field2d.cxx
@@ -45,7 +45,7 @@
 
 #include <bout/assert.hxx>
 
-Field2D::Field2D(Mesh *msh) : Field(msh), deriv(nullptr) {
+Field2D::Field2D(Mesh *localmesh) : Field(localmesh), deriv(nullptr) {
 
   boundaryIsSet = false;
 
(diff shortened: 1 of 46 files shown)


src/field/field2d.cxx
```diff
@@ -89,7 +89,7 @@ Field2D::Field2D(const Field2D& f) : Field(f.fieldmesh), // The mesh containing
   *this = f; //This line is probably not required as we init data from f.data above.
 }
 
-Field2D::Field2D(BoutReal val) : Field(nullptr), deriv(nullptr) {
+Field2D::Field2D(BoutReal val, Mesh * msh) : Field(msh), deriv(nullptr) {
   boundaryIsSet = false;
 
   fieldmesh = mesh;
```
ZedThree: This reverts `fieldmesh` back to the global `mesh` -- should probably be updated to mirror the other `Field2D` ctors

bendudson: Ah well spotted!

dschwoerer: Thanks, fixed in 227c93a

Issue #748, opened 2017-11-17
Test fci-slab


dschwoerer:
The FCI test is causing issues:
```
======= Starting integrated tests ========
Running 19 tests
...............F...

======= FAILURES ========
Traceback (most recent call last):
  File "./test_suite", line 68, in <module>
    print(u"\n----- {0} -----\n{1}".format(test, output))
UnicodeEncodeError: 'ascii' codec can't encode character u'\u2588' in position 223: ordinal not in range(128)
```

This can be reproduced by setting the encoding to 'ascii':
```
import sys
import codecs

if sys.stdout.encoding != 'ascii':
      sys.stdout = codecs.getwriter('ascii')(sys.stdout.buffer, 'strict')
```

dschwoerer:
And another error:
```
Traceback (most recent call last):
  File "./runtest", line 134, in <module>
    plt.figure()
  File "/usr/lib64/python3.6/site-packages/matplotlib/pyplot.py", line 535, in figure
    **kwargs)
  File "/usr/lib64/python3.6/site-packages/matplotlib/backends/backend_tkagg.py", line 81, in new_figure_manager
    return new_figure_manager_given_figure(num, figure)
  File "/usr/lib64/python3.6/site-packages/matplotlib/backends/backend_tkagg.py", line 89, in new_figure_manager_given_figure
    window = Tk.Tk()
  File "/usr/lib64/python3.6/tkinter/__init__.py", line 2017, in __init__
    self.tk = _tkinter.create(screenName, baseName, className, interactive, wantobjects, useTk, sync, use)
_tkinter.TclError: no display name and no $DISPLAY environment variable
```
A workaround would be to uninstall matplotlib ...  but this seems a bit extreme ...

dschwoerer:
Setting the IO encoding explicitly resolved the first issue:
```
export PYTHONIOENCODING=utf8
```

The more general question is, do we need non-ascii characters to be printed? And if - shouldn't we make sure the IO can handle it?

To reproduce set the enconding to anything that doesn't support the characters used in fci code ...
examples are ascii, iso-8859-1,iso-8859-15 etc. (i.e. basically anything that isn't utf-8)

bendudson:
The handling of encoding in python is awful, particularly if both python 2 and python 3 need to be supported. Setting everything to utf8 seems to be the only sensible option http://utf8everywhere.org/

dschwoerer:
[...]

PR #753, opened 2017-11-24
disable plotting in test-fci-slab

dschwoerer:
Fixes one part of #748 

Maybe not the nicest solution, but I think the tests shouldn't plot by default. It is a waste of cpu time (and disk space). If there is an issue I want to investigate, and want to see the plots, enabling them shouldn't be an issue.


bendudson:
Agree this is probably ok for now. It would be nice to have a consistent command-line option to pass to the tests to enable plotting (--show?, --plot?)

PR #757, opened 2017-11-24
Fieldops rebase

ZedThree:
Code generation for the various Field operators using python

This is #458 rebased on to next with conflicts fixed


ZedThree:
I've started tidying up the generator script to hopefully make it clearer what the functions will actually look like. This should also make it easier to change the templates in future, without having to change very much of the rest of the script

dschwoerer:
As the part that reduces the global mesh dependency already got merged separately, should the last commit be a different PR as it is a bug fix?

ZedThree:
For fun, I've also converted this to [jinja][1]: https://github.com/boutproject/BOUT-dev/blob/fieldops-jinja/src/field/gen_fieldops.jinja
This has some benefits: more of the logic is in the templates, rather than spread out.

Unfortunately, the big downside is adding a dependency on a python package. As it stands, we've got the generated code under git, so we could use it and then only require jinja for development.

[1]: http://jinja.pocoo.org/docs/2.10/

bendudson:
Nice! I like this as a solution. Depending on jinja only for development seems reasonable, especially as I don't think these low level routines will change often

src/field/gen_fieldops.py
```diff
@@ -0,0 +1,338 @@
...
+        rhs = copy(rhs)
+        lhs = copy(lhs)
+
+        # The output of the operation. The `larger` of the two fields.
+        out = returnType(rhs, lhs)
+        out.name = 'result'
+        lhs.name = 'lhs'
+        rhs.name = 'rhs'
+
+        for operator, operator_name in operators.items():
```
dschwoerer: `operators` is a dict, and does not guarantee that the order remains the same for different runs. Thus the generated code is not reproducible.

src/field/gen_fieldops.py
```diff
@@ -0,0 +1,338 @@
...
+        'result_op': out.getElement(data=elementwise),
+        'lhs_op': lhs.getElement(data=elementwise),
+        'rhs_op': rhs.getElement(data=elementwise),
+    }
+
+    return template_args
+
+
+if __name__ == "__main__":
+    for_gcc = True
```
dschwoerer: note that this might lead to compatibility issues.
If I remember correctly this has lead to a significant speed-up.
Should we aim to detect this at compile time in C++ using macros?
Otherwise, could we compile this using a C99 compiler, and use restrict? This would probably be the clean solution, and a C99 compiler should be basically always available as well ...

ZedThree:
@dschwoerer returning `NotImplemented` is really only for binary infix magic methods (e.g. `__add__`) and lets the interpreter know to try another function in a predetermined list of functions (e.g. `__radd__` on the other operand, see [here][1]). It doesn't do anything in the general case because python doesn't know what to try next.

I think here, actually `ValueError` is probably a better exception to raise, as [`NotImplementedError`][2] is for "abstract methods ... when they require derived classes to override the method". 

[1]: https://docs.python.org/3/library/numbers.html#implementing-the-arithmetic-operations
[2]: https://docs.python.org/3/library/exceptions.html#NotImplementedError

src/field/gen_fieldops.py
ZedThree: Good point, I've been using python 3.6 where this was an implementation detail. I've switched to using `OrderedDict` so this will always work

src/field/gen_fieldops.py
ZedThree: We've done quite a lot of testing on various looping schemes, and the fastest in the majority (all?) cases is the "consecutive-blocked region-based single data iterator" scheme, beating even C-loops on `__restrict__` pointers. I think that scheme is about ready to go in, so we'll be switching to that soon anyway.

ZedThree:
The PR test fails due to a test that's only in `next`, and due to the change introduced in 
f57b799: `Field3D::setLocation` is essentially a no-op for non-staggered grids. This can be fixed with a single line in `test_field3d.cxx`:

```diff
modified   tests/unit/field/test_field3d.cxx
@@ -274,6 +274,8 @@ TEST_F(Field3DTest, ConstYnext) {
 TEST_F(Field3DTest, SetGetLocation) {
   Field3D field;
 
+  field.getMesh()->StaggerGrids = true;
+
   field.setLocation(CELL_XLOW);
   EXPECT_EQ(field.getLocation(), CELL_XLOW);
```
so that could go in after this is merged.

I did wonder if `setLocation` should warn if used with a non-staggered mesh, but that would make things very noisy, as we usually unconditionally use `setLocation` without checking if it the mesh is staggered. If we did warn, should we then check if it the mesh is staggered before calling `setLocation`? With f57b799, we pay for the check regardless (though it just checks a bool).

dschwoerer:
I would prefer to always enable staggered grids. That would simplify the code in some parts, and also avoids a pit-fall when trying to use it.
I think that the test checks what a user would expect `setLocation` to do.
If we want to make switching staggered grids on and off easier, I would prefer being able to read locations from input files, and also read `location` for all fields, which then defaults to `CELL_CENTRE`.

src/field/gen_fieldops.py
dschwoerer: Thanks for fixing my bug 👍

dschwoerer:
@ZedThree Ah, you are right. I was thinking of the use of returning `NotImplemented` in the python interface. In this case `NotImplementedError` would be the right error if new fields are going to be (partially) added, but otherwise `ValueError` is better 👍

d7919:
If the generated code is committed as a part of the repository then could we add a travis task that regenerates this using the python code and checks that this doesn't lead to a difference with the committed version (i.e. git diff reports no changes)? This will help guard against the situation where either the committed generator or generated code are out of sync with each other.

d7919:
The field3d op fieldperp macro has been removed from field3d.cxx but it seems that there is no treatment of fieldperp in the code generator.

ZedThree:
Replaced by #818

PR #759, opened 2017-11-27
Fieldops merge

dschwoerer:
Another try at merging.
Alternative to #757 and #458


dschwoerer:
It seems setLocation sets actually the location independent of whether the mesh has the flag set.
This broke `test-interchange-instability`

PR #769, opened 2017-12-04
Fix off by one error in FFT based DDZ loop

### LINKED ISSUE #768: Off by one error in filtering for DDZ::FFT?
The loop at 
https://github.com/boutproject/BOUT-dev/blob/next/src/mesh/index_derivs.cxx#L1403
should go over the elements from `kmax+1` to the end of `cv`, setting them to zero. However the upper limit of the loop is `ncz/2` exclusive, whilst `cv` has length `ncz/2 + 1` so the upper index is `ncz/2`  so I think we miss this off this loop.

d7919:
Fixes #768


d7919:
Just to note -- whilst this was a bug fix this does not impact master.

PR #774, opened 2017-12-06
Report the value of BOUT_FLAGS used to compile.

d7919:
Can be helpful in some situations (e.g. bug reports).

There may be a nicer way to present this information but this should work ok.


dschwoerer:
We could also write a script that gathers all sort of information, including, `make.config`,  `config.log`, modules loaded etc.
I think of something like fedoras [fpaste --sysinfo](https://pagure.io/fpaste)

I cannot see a big advantage of having the flags printed at runtime - most infos should anyway already be printed like OpenMP status etc.
Further, important flags might be set outside of BOUT_FLAGS, so I think the best would be to include the whole make.config - but that is most likely an overkill ...

d7919:
Yes it's difficult to get the balance right. My main motivation was that when people report bugs it's not unusual to get a paste of the log file, but usually the first question after this is how did you configure and what are the flags, so I thought adding this might help.

PR #775, opened 2017-12-08
SingleDataIterator

ZedThree:
Flexible iterator for Fields. The `SingleDataIterator` (SDI) iterates over a `vector` of `int`s which are used to index a Field.                                                        
                                                                              
Three classes are defined here:                                               
  1. `SIndexRange` 
  2. `SingleDataIterator`
  3. `SIndices`
                                                               
plus a helper function, createRegionIndices, for creating the indices vectors.
RegionIndices is just a typedef for `std::vector<int>`                        
                                                                              
`SIndexRange` is the utility class that should be used in `for` loops, etc.  `begin` and `end` can be used on a `SIndexRange`, returning a `Singledataiterator`, which in turn dereferences into a `SIndexRange`, which finally can be used to index a Field.

---

This PR adds the SDI and replaces the current DataIterator (DI) only in a couple of places. The DI will be completely replaced everywhere by v4.2. This might involve renaming SDI -> DI. The current structure of the SDI is not set in stone either -- there's still some questions about location and form of the offset functions. This will be formalised before v4.2

Because the SDI is much more flexible than DI, the `REGION` `enum` will be replaced by `string`s. The current PR adds a stop-gap `REGIONtoString` `map` for converting between the `enum` and the `string` name. The `enum` will probably be deprecated in v4.2 and removed entirely in a later version.

### DIFF
--- a/include/bout/mesh.hxx
+++ b/include/bout/mesh.hxx
@@ -69,9 +69,12 @@ class Mesh;
 #include "paralleltransform.hxx" // ParallelTransform class
 
 #include "unused.hxx"
+#include "singledataiterator.hxx"
 
 #include <list>
+#include <map>
 #include <memory>
+#include <utility>
 
 /// Type used to return pointers to handles
 typedef void* comm_handle;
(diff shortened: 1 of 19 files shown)


include/bout/mesh.hxx
```diff
@@ -619,10 +642,10 @@ class Mesh {
   /// Calculates the size of a message for a given x and y range
   int msg_len(const vector<FieldData*> &var_list, int xge, int xlt, int yge, int ylt);
   
-  // Initialise derivatives
+  /// Initialise derivatives
   void derivs_init(Options* options);
   
-  // Loop over mesh, applying a stencil in the X direction
+  /// Loop over mesh, applying a stepppncil in the X direction
```
dschwoerer: typo?

ZedThree: Probably me misfingering `C-p` in Emacs...

ZedThree:
Superseded by #841

PR #776, opened 2017-12-08
Some small tests for Field2D/3D

ZedThree:
Adds some simple tests for Field2D/Field3D:

- check fieldmesh gets set correctly when copying fields
- check yup/ydown/deriv fields get created correctly

Also adds `UNUSED` to some arguments for the new `setAttributes` methods in `DataFormat`

### DIFF
--- a/tests/unit/field/test_field2d.cxx
+++ b/tests/unit/field/test_field2d.cxx
@@ -117,6 +117,54 @@ TEST_F(Field2DTest, CreateOnGivenMesh) {
   delete fieldmesh;
 }
 
+TEST_F(Field2DTest, CopyCheckFieldmesh) {
+  int test_nx = 2;
+  int test_ny = 3;
+  int test_nz = 5;
+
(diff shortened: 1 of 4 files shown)


tests/unit/field/test_field2d.cxx
```diff
@@ -117,6 +117,56 @@ TEST_F(Field2DTest, CreateOnGivenMesh) {
   delete fieldmesh;
 }
 
+TEST_F(Field2DTest, CopyCheckFieldmesh) {
+  int test_nx = 2;
```
d7919: I assume these values are different from the "global" mesh so we're testing that it's definitely using fieldmesh? Would it be more robust to explicitly define these in terms of some difference from the global mesh (e.g. test_nx = global_nx + 2 or similar)?

ZedThree: Yes, that would probably make it more obvious what's happening, thanks

PR #777, opened 2017-12-08
Default to not building with scorep

### LINKED ISSUE #720: Scalasca/scorep support enabled if present in path
The `--with-scorep` flag appears to be effectively ignored; when `scorep` is found in the path then support for it is enabled automatically no matter what the status of `--with-scorep`.

ZedThree:
Fixes #720


configure.ac
```diff
@@ -1152,7 +1152,7 @@ HAS_SCOREP="no"
 AS_IF([test "$with_scorep" != "no"], [
   AC_MSG_NOTICE(["Searching for Scorep executable"])
 
-  AS_IF([test "$with_scorep" != ""], [
+  AS_IF([test "$with_scorep" != "yes"], [
```
bendudson: What happens here if $with_scorep is "no"?

bendudson: Ah wait, it's inside the != "no" check... please ignore

PR #778, opened 2017-12-12
.so support

dschwoerer:
Enable building shared objects.

To my surprise this speeded up the test suite from 100s to 43s.
Main cause seems to be the much faster throw-catch due to the smaller binary.

Besides reducing the influence of that "bug" - Not sure this is extremely useful right now.

Default is to disable it - so there should be no side effects, if not enabled.
Probably mostly interesting for system wide installs, as it allows much smaller binaries.
On HPC systems probably only of interrest if bout++ would be installed, as otherwise one needs to make sure LD_LIBRARY_PATH includes the `.so` location ...

### DIFF
--- a/configure
+++ b/configure
@@ -775,6 +775,7 @@ enable_debug
 enable_optimize
 enable_sigfpe
 enable_backtrace
+enable_shared
 enable_openmp
 with_gcov
 enable_code_coverage
@@ -1416,6 +1417,7 @@ Optional Features:
   --enable-optimize=1/2/3 Enable optimization
   --enable-sigfpe         Enable FloatingPointExceptions
   --disable-backtrace     Disable function backtrace
+  --enable-shared         Enable building bout++ into an shared object
   --disable-openmp        do not use OpenMP
   --enable-code-coverage  Whether to enable code coverage support
 
@@ -2602,6 +2604,13 @@ if test "${enable_backtrace+set}" = set; then :
   enableval=$enable_backtrace;
 else
   enable_backtrace=yes
+fi
+
+# Check whether --enable-shared was given.
+if test "${enable_shared+set}" = set; then :
+  enableval=$enable_shared;
+else
+  enable_shared=no
 fi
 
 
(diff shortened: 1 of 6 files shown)


makefile
```diff
@@ -11,19 +11,35 @@ endif
...
 
 check-mms-tests:
-	@cd tests/MMS; ./test_suite
+	@cd tests/MMS; export LD_LIBRARY_PATH=${PWD}/lib:${LD_LIBRARY_PATH} ; ./test_suite
 
 check-integrated-tests:
-	@cd tests/integrated; ./test_suite_make
-	@cd tests/integrated; PYTHONPATH=${PWD}/../../tools/pylib/:${PYTHONPATH} ./test_suite
+	@cd tests/integrated; LD_LIBRARY_PATH=${PWD}/lib:${LD_LIBRARY_PATH}./test_suite_make
+	cd tests/integrated; export LD_LIBRARY_PATH=${PWD}/lib:${LD_LIBRARY_PATH} ; \
```
d7919: The `@` has been lost from this line resulting in this command being echoed. Also, instead of `${PWD}` should we actually use `$(BOUT_TOP)`?

configure
```diff
@@ -5670,14 +5679,31 @@ fi
...
+then
+    # compile as position independent code.
+    # -fpic is apparently faster then -fPIC, but -fPIC works always.
+    # From a SO comment (https://stackoverflow.com/a/3544211/3384414):
+    #  What's more: I did a little experiment here (on x86_64
+    #  platform), -fPIC and -fpic appears to have generated the same
+    #  code. It seems they generate a different code only on m68k,
+    #  PowerPC and SPARC.
+    # Therfore use -fPIC for now
+    CXXFLAGS+=" -fPIC"
```
d7919: This requests position independent code in both the static and shared builds. This could lead to a performance hit for the (still default) static build according to the post [here](https://gcc.gnu.org/ml/gcc-help/2009-05/msg00259.html). 

Hopefully the person typing `--enable-shared` knows what they're doing but it might be nice to make the default target the shared library when configured with `--enable-shared` so the user has to go out of their way to generate the static library when configured with shared support?

PR #786, opened 2017-12-15
merge ./test_suite and be more verbose to prevent timeouts

dschwoerer:
Many tests fail currently on travis, due to time-outs.
While the best solution would be to make them faster, preventing the time-out is still beneficial in the mean time.


ZedThree:
I don't quite see how this prevents timeouts. Is it just that travis is looking for newlines when it says "output"? Will this still fail if a single test takes longer than 10 minutes? 

A more robust method might be prefixing `"./.travis_script.sh"` with [`travis_wait`][1].

Merging the two `test_suite`s is definitely good though.

[1]: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

PR #791, opened 2018-01-04
Use specialised output object for warnings/errors/info

### LINKED ISSUE #719: output.write still used in lots of places
I notice there's still loads of `output.write`s scattered through the code, notable in the DataFile stuff, and lots that print "WARNING" or "INFO". Should these just be changed to `output_warn`, `output_info`, respectively, and those words deleted?

Actually, there's some "ERROR"s in there too.

ZedThree:
Changes all instances of `output.write("WARNING|ERROR|INFO...` to use `output_warn|error|info`

Fixes #719

### DIFF
--- a/examples/6field-simple/elm_6f.cxx
+++ b/examples/6field-simple/elm_6f.cxx
@@ -378,11 +378,11 @@ int physics_init(bool restarting) {
 
   // Load metrics
   if(mesh->get(Rxy,  "Rxy")) { // m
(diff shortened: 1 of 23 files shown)


ZedThree:
Also, I noticed that we use `#ifdef NCDF_VERBOSE; output...; #endif` in the NetCDF format files a fair bit. This could either be replaced with a netcdf-specific output unit that can be enabled separately from the other output units, or just removed entirely (and then either keeping the outputs or not).

PR #792, opened 2018-01-05
Next remove bindex

johnomotani:
Can we merge this branch? It fixes issues I have with VDDY/interp_to on staggered fields in STORM. Could then delete the staggered_ygrid_fixes branch and PR #746.


d7919:
I'm not sure but I think this should probably go into next when ready rather than master.

bendudson:
@JosephThomasParker (and I) have made modifications to these functions too, to remove bindex and introduce better ways to iterate in these operators. I think we need to have a go at merging together what has been done. I agree that the changes should then go into `next` rather than `master`.

johnomotani:
@bendudson I realised after my attempt that this next-remove-bindex branch you and @JosephThomasParker made had already done it. next-remove-bindex seems to run faster than mine, hence proposing to merge this and abandon staggered_ygrid_fixes.

I'll have a go at merging into next and testing...

PR #797, opened 2018-01-16
Travis update gcc6 to gcc7

d7919:
One of the Travis build jobs is used to check support for upcoming compiler changes -- this was done with gcc-6 but this PR changes this to gcc-7 as this is now available.

May want to consider enabling all compiler warning flags here as this is one way we can get early notice/make use of new/emerging checks.


ZedThree:
I'm in favour of making the gcc-7 job turning on all the warnings, and just compile + unit tests

PR #802, opened 2018-01-17
Remove old configuration files for specific machines

d7919:
Likely that most of these are no longer valid and could confuse new users (e.g. if they wonder why there isn't a file for the machine they want to use). The information would still be available in the repository history if required.


bendudson:
Most of these are one-liners which call configure. Would it be worth putting some of them here http://bout-dev.readthedocs.io/en/latest/user_docs/machine_install.html (at least for machines which still exist)

d7919:
Yes that's probably a good idea

d7919:
Done

PR #804, opened 2018-01-17
Next remove bindex

JosephThomasParker:
Remove the final occurrences of bindex and bvalue.

This completes #792.  The last commit probably replicates #801.


bendudson:
Awesome! Thanks Joseph!

include/bout/mesh.hxx
```diff
@@ -522,13 +522,13 @@ class Mesh {
...
   /// @param[in] method  The differencing method to use
-  const Field3D indexVDDZ(const Field &v, const Field &f, CELL_LOC outloc, DIFF_METHOD method);
+  const Field3D indexVDDZ(const Field3D &v, const Field3D &f, CELL_LOC outloc, DIFF_METHOD method, REGION region=RGN_NOZ);
 
   const Field2D indexFDDX(const Field2D &v, const Field2D &f, CELL_LOC outloc, DIFF_METHOD method, REGION region=RGN_NOX);
   const Field3D indexFDDX(const Field3D &v, const Field3D &f, CELL_LOC outloc, DIFF_METHOD method, REGION region=RGN_NOX);
   const Field2D indexFDDY(const Field2D &v, const Field2D &f, CELL_LOC outloc, DIFF_METHOD method, REGION region=RGN_NOX);
   const Field3D indexFDDY(const Field3D &v, const Field3D &f, CELL_LOC outloc, DIFF_METHOD method, REGION region=RGN_NOX);
-  const Field3D indexFDDZ(const Field3D &v, const Field3D &f, CELL_LOC outloc, DIFF_METHOD method);
+  const Field3D indexFDDZ(const Field3D &v, const Field3D &f, CELL_LOC outloc, DIFF_METHOD method, REGION region=RGN_NOX);
```
d7919: Is this region consistent with the old behaviour?

include/difops.hxx
```diff
@@ -193,7 +193,7 @@ const Field3D Grad2_par2(const Field3D &f, CELL_LOC outloc=CELL_DEFAULT);
  * These are a simple way to do staggered differencing
  */
 const Field3D Grad_par_CtoL(const Field3D &var);
-const Field3D Vpar_Grad_par_LCtoC(const Field &v, const Field &f);
+const Field3D Vpar_Grad_par_LCtoC(const Field &v, const Field &f, REGION region=RGN_NOY);
```
d7919: Is this region consistent with the old behaviour? I couldn't see it in the old implementation but didn't look that closely.

include/bout/mesh.hxx
```diff
@@ -522,13 +522,13 @@ class Mesh {
   /// @param[in] f  The field being advected
   /// @param[in] outloc The cell location where the result is desired. The default is the same as \p f
   /// @param[in] method  The differencing method to use
-  const Field3D indexVDDZ(const Field &v, const Field &f, CELL_LOC outloc, DIFF_METHOD method);
+  const Field3D indexVDDZ(const Field3D &v, const Field3D &f, CELL_LOC outloc, DIFF_METHOD method, REGION region=RGN_NOZ);
```
ZedThree: It would be good to mention `REGION` in the doxygen comment

include/bout/mesh.hxx
JosephThomasParker: Good point!  Without specifying a region, `start_index(&bx)` in the old code uses `RGN_NOBNDRY`.  This also applies in indexFDDY, and probably in the other functions. I'll check...

d7919:
Some of the tests seem to have [failed](https://travis-ci.org/boutproject/BOUT-dev/jobs/329846525#L1841).

include/bout/mesh.hxx
JosephThomasParker: Thanks, will do.

JosephThomasParker:
The last update should sort everything apart from @d7919's point about regions.  The region for the old `bindex` loop depends on `bindex`'s region property, which was set to `RGN_NOBNDRY` in `start_index`.  In all the examples I've seen, this default is never changed, so AIUI all the regions should be `RGN_NOBNDRY` not `RGN_NOX`.  Have I understood correctly?

dschwoerer:
Shouldn't we keep `bvalue` as a deprecated class until 5.0.0?
While it certainly shound't be used in the library, we should give users time to adjust - or are there good reasons that we cannot keep it as deprecated?

d7919:
Perhaps mark as deprecated in master as this is the current release and the last in which `bvalue` is available?

dschwoerer:
Is next going to be published as 5.0 or 4.2?
Not all projects have been adjusted for 4.x - therefore we shouldn't push to hard for breaking changes ...

But the main question in my opinion should be:
Are there any (significant) draw backs from keeping `bvalue`?
I mean, any draw backs besides having more dead code?

d7919:
Probably nothing major, however as it is not used in any part of the library it is straightforward (and possibly more appropriate) for the `bvalue` class to be provided by the model that wants to use it.

ZedThree:
While they haven't been explicitly marked as deprecated, they've been [noted as being deprecated since v4.0.0](https://github.com/boutproject/BOUT-dev/blob/master/include/stencils.hxx#L6). Out of the examples, only `non-local_1d` uses `bindex`, and nothing uses `bvalue`. Does STORM use either `bindex`/`bvalue`?

I think this is ok to go in to v4.2, unless there really is lots of user code we're going to be breaking.

PR #807, opened 2018-01-19
Set location of fields only if StaggerGrids is enabled

ZedThree:
Pulls f57b799 out of #757, fixes relevant test

### DIFF
--- a/src/field/field3d.cxx
+++ b/src/field/field3d.cxx
@@ -240,6 +240,12 @@ void Field3D::setLocation(CELL_LOC new_location) {
     }
     location = new_location;
   } else {
+#if CHECK > 0
+    if (new_location != CELL_CENTRE && new_location != CELL_DEFAULT) {
+      output_warn.print("WARNING: Trying to set Field3D location on non-staggered grid. "
+                        "Setting to CELL_CENTRE instead\n");
+    }
+#endif
     location = CELL_CENTRE;
   }
 }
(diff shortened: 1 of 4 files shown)


src/field/field3d.cxx
```diff
@@ -229,14 +229,19 @@ const Field3D& Field3D::ynext(int dir) const {
...
+    if (new_location == CELL_VSHIFT) {
+      throw BoutException(
+          "Field3D: CELL_VSHIFT cell location only makes sense for vectors");
+    }
+    if (new_location == CELL_DEFAULT) {
+      new_location = CELL_CENTRE;
+    }
+    location = new_location;
+  } else {
+    location = CELL_CENTRE;
```
d7919: Should we check that we don't try to set a non-centre location in the non-staggered case and if we do then throw? This would possibly help us catch places where we try to set staggering but don't check that we want to stagger? (I don't think this is my idea, I'm sure it's been mentioned previously)

ZedThree: I agree at least a warning here would be appropriate, along with modifying all current calling sites to check if we're using staggered grids.

ZedThree:
Now warns if trying to call `setLocation` to something other than `CELL_CENTRE` on a non-staggered grid.

I've not modified the calling sites, so we can just happily call `result.setLocation(otherfield.getLocation())`, which is done in a lot of places.

ZedThree:
The push tests will fail, as this breaks the drift/interchange instability tests. The PR test should pass though, as #819 fixes them.

PR #808, opened 2018-01-19
 Consistently return fields from index derivs functions

ZedThree:
Pulls b4f7836f out of #757

### DIFF
--- a/src/mesh/index_derivs.cxx
+++ b/src/mesh/index_derivs.cxx
@@ -948,8 +948,9 @@ const Field2D Mesh::applyYdiff(const Field2D &var, Mesh::deriv_func func, CELL_L
   return result;
 }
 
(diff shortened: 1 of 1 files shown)


src/mesh/index_derivs.cxx
```diff
@@ -1407,9 +1409,7 @@ const Field3D Mesh::indexDDZ(const Field3D &f, CELL_LOC outloc, DIFF_METHOD meth
 
 const Field2D Mesh::indexDDZ(const Field2D &f) {
   ASSERT1(this == f.getMesh());
```
d7919: Do we need/want similar inserts in the other routines that are changed here or is this already handled?

ZedThree: Just checked, this is done for all the routines here, just not consistently the first thing.

PR #809, opened 2018-01-19
Add overload for checkData(BoutReal)

ZedThree:
Partially pulls 262c4c3c out of #757 

This overload is useful in order to avoid some logic in the fieldops templates for deciding when to call `checkData` on the arguments.

Not sure about this overload being in `field3d.hxx` though -- might be better off in `utils.hxx`?


d7919:
I agree that this probably doesn't belong in field3d.

PR #812, opened 2018-01-23
Improve makefile:

dschwoerer:
* only rebuild bout++.o if any other file was rebuild (or it anyway
  needs to be rebuild)
* only rebuild libbout++.a if bout++.o is newer. Together with above
  this assures that libbout is, and only if, any files changed.
* libfast can be added as dependency to the checks
* make -q is still not working

If the target is up to date, `make` does nothing.
Therefore this allows to add it as a prerequisite.

Now it is sufficient to run `make -j 8 build-check`  to build the library (if needed) and the checks.


dschwoerer:
I know, nobody likes it if I change the makefile, but this way, we can add the library as a dependency, without triggering rebuilds if it is not needed.

Running make takes with my i7 ~200 ms, and goes down to ~50 ms with 8 threads.
Compiling bout++.o and recreating libbout++.a takes around 2 s.

Adding it therefore to tests or other projects that depend on bout++.
As the library is normally not changed, no further rebuilds are triggered, however if there are changes, no issues with out of date BOUT++ will be experienced.

Should I add more comments, on what is going on?

Compatibility shouldn't be an issue, gnu make is anyway required.

ZedThree:
The end goal is definitely good -- I'd like to just run `make check-unit-tests` and get everything recompiled as needed

...but there's already a fair bit of black magic in the build system as is, so I'm loath to add more. More comments or some form of documentation about how the whole thing now works would be great. Especially why certain bits are needed -- e.g. why `.dummy`?

dschwoerer:
I tried to explain - if there are still parts unclear - please ask.

dschwoerer:
I also added a bit to give a better error if `./configure` hasn't been run to update the `make.config`

PR #816, opened 2018-01-25
Allow no openmp for pvode

d7919:
Adds `--enable-pvode-omp` to configure. If not set then we don't include
the `OpenMP` flags when building `PVODE`, if set then we do include
these flags (this is the previous default behaviour).

Still requires `OpenMP` to be active (i.e. `--disable-openmp
--enable-pvode-omp=yes` will lead to a serial only `PVODE`).

This greatly helps the time to run the tests on Travis and it may be helpful in other scenarios where we want control over where we use openmp for parallelisation.

### DIFF
--- a/configure
+++ b/configure
@@ -732,6 +732,7 @@ infodir
 docdir
 oldincludedir
 includedir
+runstatedir
 localstatedir
 sharedstatedir
 sysconfdir
@@ -776,6 +777,7 @@ enable_optimize
 enable_sigfpe
 enable_backtrace
 enable_shared
+enable_pvode_omp
 enable_openmp
 with_gcov
 enable_code_coverage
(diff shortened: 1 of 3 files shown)


ZedThree:
Is it better to keep the current default and have this be `--disable-pvode-omp`?

d7919:
I changed the default mainly because I've observed OpenMP to hurt here but I haven't seen it provide benefit in this place (haven't really gone looking for this though). Could (should) add something to the documentation to clear this up?

dschwoerer:
Shoudn't `--disable-openmp --enable-pvode-omp=yes` fail, if it doesn't enable openmp?

d7919:
I've added some logic which should hopefully abort configure if openmp is disabled but the user has asked for openmp in PVODE.

d7919:
Those changes should be addressed now.

Issue #64, opened 2015-05-07
Metrics in staggered grids


johnomotani:
In difops.cxx, the CELL_CENTRE metrics are used regardless of the CELL_LOC of the derivative. This will introduce an error into the results when using staggered grids if the metric varies (in the direction of the stagger).

The solution, as it seems to me, would be to keep CELL_XLOW and CELL_YLOW versions of the metric in Mesh (either provided by the grid generator or calculated by interpolation), which can then be used as needed in difops.cxx.

An alternative, if we want to make staggered grids really first-class citizens in BOUT++, might be to associate fields at different locations with the Field*D object and introduce a method like

```
Field*D & operator()(CELL_LOC loc);
```

which would return the Field at `loc` (then `Field*D::setLocation` might just set the 'default' location). We could either interpolate from the default location when this operator is called (if the field has changed since it was last called), or add some method to the Field that the user must call to calculate the non-default versions. The differential operators might then change very simply, e.g.

```
result = DDY(var, outloc, method)/sqrt(mesh->g_22);
```

becomes

```
result = DDY(var, outloc, method)/sqrt( mesh->g_22(outloc) );
```

This seems like it might be an elegant solution, but would probably require quite a bit of work to implement...

dschwoerer:
Don't non-uniform meshes produce anyway errors, as the stencils (to my understanding) all assume a uniform mesh?
If the mesh spacing varies that strongly, I am not sure this is your most important problem ...

dschwoerer:
Partially fixed by the [newfield-branch](https://github.com/boutproject/BOUT-dev/tree/newfield)

bendudson:
Is this now fixed with the new coordinates at different locations?

johnomotani:
I think it should be fixed... I'm working on an MMS test for differential operators to verify the staggered versions.

johnomotani:
Fixed in #1198/#1271

PR #818, opened 2018-01-29
Use Jinja2 to generate code for Field* arithmetic operators

ZedThree:
Replaces #757 

Uses the Python Jinja2 templating system to replace (most of) the various macros for arithmetic operations on `Field2D/Field3D` ("fieldops").

The template is in `src/field/gen_fieldops.jinja`, and the python driver code is in `src/field/gen_fieldops.py`. Most of the driver code could be reused for future code generation of other parts of the codebase.

Since the generated code is committed to the repo, this PR only requires the use of Jinja2 in order to modify the fieldops code, and not to build BOUT++.


ZedThree:
I think this is currently failing because the interchange-instability now throws an exception due to trying to add two fields with different locations, at [`2fluid.cxx:361`][1]: `Apar` has its location set to `CELL_YLOW` although staggered grids aren't used.

#807 makes `Field3d::setLocation` a no-op if staggered grids aren't used, so this would at least patch over this particular bug. Presumably test-interchange-instability is currently wrong though -- either it needs modifying to use staggered grids correctly, or to just not use them at all.

[1]: https://github.com/boutproject/BOUT-dev/blob/master/tests/integrated/test-interchange-instability/2fluid.cxx#L361

dschwoerer:
I had a look at this a while ago. They only set the location, but there is no *real* progress towards staggered grids, so they should be removed.

Is fixed in https://github.com/boutproject/BOUT-dev/commit/492a11d1b0f45ffff5e8efbd1e134a8d17413b83

manual/sphinx/developer_docs/code_generation.rst
```diff
@@ -0,0 +1,40 @@
...
+neither is `BoutReal`.
+
+To install Jinja:
+
+.. code-block:: console
+
+   $ pip install --user Jinja2
+
+To re-generate the code, there is a `make` target for `gen_fieldops.cxx`
+in `src/field/makefile`. This also tries to apply `clang-format` in
```
d7919: Do we want/provide any instructions for ensuring clang-format is available?

manual/sphinx/developer_docs/code_generation.rst
```diff
@@ -0,0 +1,40 @@
...
+appropriate code. There is some logic in the template to handle certain
+combinations of the input fields: for example, for the binary infix
+operators, only check the two arguments are on identical meshes if
+neither is `BoutReal`.
+
+To install Jinja:
+
+.. code-block:: console
+
+   $ pip install --user Jinja2
```
d7919: Presumably this should be `pip3` on systems where the default python version is not 3? The gen_fieldops.py specifically requests v3. It may be worth noting this.

src/field/gen_fieldops.py
```diff
@@ -0,0 +1,207 @@
...
+try:
+    from builtins import object
+except ImportError:
+    pass
+
+from collections import OrderedDict
+from copy import deepcopy as copy
+import itertools
+
+import jinja2
```
d7919: Is it worth putting this in a try/except block so that we can give some BOUT++ related instructions (e.g. you need to install jinja2 see this page in the manual for how)?

src/field/makefile
```diff
@@ -1,8 +1,16 @@
...
+		  initialprofiles.cxx vecops.cxx vector2d.cxx vector3d.cxx \
+		  where.cxx globalfield.cxx generated_fieldops.cxx
 SOURCEH		= $(SOURCEC:%.cxx=%.hxx) field_data.hxx
 TARGET		= lib
 
+
 include $(BOUT_TOP)/make.config
+
+generated_fieldops.cxx: gen_fieldops.py gen_fieldops.jinja
+	./$< > $@
```
d7919: Is it worth noting what this "magic" line is doing (doesn't take too much to work it out, but might be helpful).

src/field/generated_fieldops.cxx
```diff
@@ -0,0 +1,1668 @@
...
+                                                    const BoutReal *__restrict__ rhs,
+                                                    int len) {
+
+  for (int i = 0; i < len; ++i) {
+    result[i] = lhs[i] * rhs[i];
+  }
+}
+
+// Provide the C++ wrapper for multiplication of Field3D and Field3D
+Field3D operator*(const Field3D &lhs, const Field3D &rhs) {
```
d7919: Whilst we haven't typically made use of it, it was possible with the old form to only apply these operators over a specific region (only by hacking the internals though). Probably not a big deal if we lose this ability for now but just to clarify -- is there any way to limit these operations to a region at the moment with this approach?

src/field/generated_fieldops.cxx
```diff
@@ -0,0 +1,1668 @@
...
+  Field3D result(localmesh);
+  result.allocate();
+  checkData(lhs);
+  checkData(rhs);
+
+  autogen_Field3D_Field3D_Field3D_multiplication(&result[i], &lhs[i], &rhs[i],
+                                                 localmesh->LocalNx * localmesh->LocalNy *
+                                                     localmesh->LocalNz);
+
+#if CHECK > 0
```
d7919: Shouldn't this check come _before_ we do the operation?

src/field/generated_fieldops.cxx
```diff
@@ -0,0 +1,1668 @@
+// This file is autogenerated - see gen_fieldops.py
+#include <bout/mesh.hxx>
+#include <field2d.hxx>
+#include <field3d.hxx>
+#include <globals.hxx>
+#include <interpolation.hxx>
+
+// Do the actual multiplication of Field3D and Field3D
+void autogen_Field3D_Field3D_Field3D_multiplication(BoutReal *__restrict__ result,
+                                                    const BoutReal *__restrict__ lhs,
```
d7919: Whilst probably not an issue as we only use read access for `lhs` and `rhs` presumably if we're doing `a = b * b` then we're not strictly avoiding aliasing. I may be misunderstanding what we're promising by using `__restrict__` and I'm not sure it can cause any problems anyway.

src/field/generated_fieldops.cxx
```diff
@@ -0,0 +1,1668 @@
...
+    checkData(*this);
+    checkData(rhs);
+
+    autogen_Field3D_Field3D_multiplication(&(*this)[i], &rhs[i],
+                                           fieldmesh->LocalNx * fieldmesh->LocalNy *
+                                               fieldmesh->LocalNz);
+
+#if CHECK > 0
+    if (this->getLocation() != rhs.getLocation()) {
+      throw BoutException("Error in multiplication: fields at different locations. lhs "
```
d7919: Is there anyway to distinguish that this has failed in a `*=` operation rather than the `*` operation?

src/field/generated_fieldops.cxx
```diff
@@ -0,0 +1,1668 @@
...
+// Do the actual multiplication of Field3D and Field2D
+void autogen_Field3D_Field3D_Field2D_multiplication(BoutReal *__restrict__ result,
+                                                    const BoutReal *__restrict__ lhs,
+                                                    const BoutReal *__restrict__ rhs,
+                                                    int nx, int ny, int nz) {
+
+  for (int x = 0; x < nx; ++x)
+    for (int y = 0; y < ny; ++y)
+      for (int z = 0; z < nz; ++z) {
+        result[z + nz * (y + ny * x)] = lhs[z + nz * (y + ny * x)] * rhs[y + x * ny];
```
d7919: Performance wise has the use of a counter been tested (e.g. `uint i=0` outside the loops and then `result[i] = lhs[i]*rhs[i2d]; i++;` where `i2d` can either be calculated outside the innermost loop or could be a second counter that is incremented one level up from `i`)? This may or may not help.

src/field/generated_fieldops.cxx
```diff
@@ -0,0 +1,1668 @@
...
+  result.allocate();
+  checkData(lhs);
+  checkData(rhs);
+
+  autogen_Field3D_Field3D_Field2D_multiplication(&result[i], &lhs[i], &rhs[i],
+                                                 localmesh->LocalNx, localmesh->LocalNy,
+                                                 localmesh->LocalNz);
+
+  result.setLocation(lhs.getLocation());
+
```
d7919: No location checks? Whilst Field2D doesn't have a location do we need to ensure the Field3D is therefore at the cell centre?

src/field/generated_fieldops.cxx
```diff
@@ -0,0 +1,1668 @@
...
+  ASSERT1(localmesh == rhs.getMesh());
+
+  Field3D result(localmesh);
+  result.allocate();
+  checkData(lhs);
+  checkData(rhs);
+
+  autogen_Field3D_Field3D_Field3D_multiplication(&result[i], &lhs[i], &rhs[i],
+                                                 localmesh->LocalNx * localmesh->LocalNy *
```
d7919: I could be wrong but doesn't mesh contain the total number of local grid points already so could avoid this multiplication?

src/field/generated_fieldops.cxx
```diff
@@ -0,0 +1,1668 @@
...
+  // otherwise just call the non-inplace version
+  if (data.unique()) {
+    Indices i{0, 0, 0};
+
+    checkData(*this);
+    checkData(rhs);
+
+    autogen_Field3D_BoutReal_multiplication(
+        &(*this)[i], rhs, fieldmesh->LocalNx * fieldmesh->LocalNy * fieldmesh->LocalNz);
+
```
d7919: There's no check data of the result here unlike other approaches.

src/field/generated_fieldops.cxx
ZedThree: Not currently, but we'll get that back (and hopefully better?) when we move to the SDI-block form

src/field/generated_fieldops.cxx
ZedThree: I wasn't sure, but this is defined behaviour: http://en.cppreference.com/w/c/language/restrict
It's ok as long as neither `lhs` or `rhs` are modified.

I think this form is not the most efficient either. We probably want something like:

```cpp
for (int i = 0; i < len; ++i) {
    const BoutReal *__restrict__ lhs_value = lhs[i];
    const BoutReal *__restrict__ rhs_value = rhs[i];
    BoutReal *__restrict__ result_value = result[i];
    result_value = lhs_value + rhs_value;
}
```

See e.g. http://cellperformance.beyond3d.com/articles/2006/05/demystifying-the-restrict-keyword.html 
which talks about separating load, update, store operations (at least for gcc).

src/field/generated_fieldops.cxx
ZedThree: Yes, I've moved it to top of each operator

src/field/generated_fieldops.cxx
ZedThree: I don't think so -- I couldn't see it in either `Mesh` or `BoutMesh`

src/field/generated_fieldops.cxx
dschwoerer: I tried to use inline restrict pointer, but gcc didn't provide the performance, only after putting it in a separate function it worked.

further, the example does not work, as you only take a copy of the value, so there is no vectorization.

`a = b * b` is no issue, `a = a * b` is violating the promise, but I think it should still work.
Only times I had issues was something like `a = A * a` where `A` was a matrix, and some parts of `a` where written and then read again, but I cannot see that that would happen with a vector.

Another advantage of having separate function is, that we can put it in a separate file, and compile it with a C compiler, and use `restrict` rather then a gcc extension.

src/field/generated_fieldops.cxx
dschwoerer: No. Otherwise the staggered grids are unusable.

See e.g. #64 - that would require us to extend the staggered grids support to field2D - but that is non-trivial.

d7919: OK, but does this mean they are usable but not necessarily correct?

ZedThree:
Fixed most of these, will push tomorrow

src/field/generated_fieldops.cxx
dschwoerer: To be honest, I am using it for a slab geometry, so dx etc are all constant, and the non-staggered-ness of Field2D does not bother me.

I don't know whether anybody quantified the issues for other meshes, that have changes along the staggered direction.

To solve this in a nice way would probably require to wrap the mesh quantities in an object, that contains them at several locations, and depending on the field's location they are operated with, to choose the appropriate field. A potentially remaining issue is, if one of the fields is changed, to update the other ones.

The other approach would be having a different mesh for the staggered quantities - but that results in challanges with taking derivatives from one mesh, on to the the other mesh, and vice versa, i.e. we need to add relations between meshes, both in terms of interpolating, and differentiating.

d7919: Thanks for the detailed response and some interesting ideas for how to potentially improve this. 

I'm slightly concerned that if users try to use staggered fields in their physics model and are also using Field2D for some things they may not realise the situation may not be as they think. It's now obvious when using two 3D field at different locations that this is bad (at least when CHECK is >0) but this may convince some users that their use of staggered Field3D and a non-staggered Field2D is fine as it doesn't throw an exception even with all checks turned on. I'm not suggesting there's an easy fix for this (as you've noted) and breaking staggered grids for everyone isn't a good idea either -- is there anything we can do to alert/warn the user to this beyond the manual?

src/field/gen_fieldops.jinja
```diff
@@ -90,16 +100,7 @@ void autogen_{{lhs}}_{{rhs}}_{{operator_name}}(
...
 
-    {% if lhs == rhs == "Field3D" %}
-#if CHECK > 0
-    if (this->getLocation() != rhs.getLocation()) {
-      throw BoutException(
-        "Error in {{operator_name}}: fields at different locations. lhs is at %s, rhs is at %s!",
-        strLocation(this->getLocation()), strLocation(rhs.getLocation()));
-    }
-#endif
     checkData(*this);
```
d7919: Just to check -- this was previously inside the `if lhs == rhs == "Field3D"` template block but no longer is, is this an intentional change alongside moving the CHECK block? I realise it only makes sense to check data on `*this`  after the operation, but do we still need to ensure that we only do the check if we're working with Field3D values?

ZedThree: There is now a `checkData` overload for BoutReal/Field2D/Field3D, so we should just be able to call this unconditionally.

ZedThree:
`test-interchange-instability` appears to be hanging. I'm currently having trouble reproducing this on my own machine...

src/field/generated_fieldops.cxx
dschwoerer: I think this conversation is off topic.
I does not get worse with this PR, it is just not jet perfect.
I will open an issue to continue the discussion there.

src/field/gen_fieldops.py
```diff
@@ -0,0 +1,211 @@
...
+import itertools
+
+try:
+    import jinja2
+except ImportError:
+    raise ImportError('Missing Python module "jinja2". See "Field2D/Field3D Arithmetic '
+                      'Operators" in the BOUT++ user manual for more information')
+
+# The arthimetic operators
+# OrderedDict to (try to) ensure consistency between python 2 & 3
```
d7919: Presumably order just matters to prevent issues with committing the generated code and this triggering diffs even when the routines aren't changed (just the order)?

ZedThree: Yep, exactly

src/field/gen_fieldops.py
```diff
@@ -0,0 +1,211 @@
...
+    for_gcc = True
+    print(header)
+
+    env = jinja2.Environment(loader=jinja2.FileSystemLoader('.'),
+                             trim_blocks=True)
+
+    template = env.get_template("gen_fieldops.jinja")
+
+    for lhs, rhs in itertools.product(fields, fields):
+        # We don't have define real real operations
```
d7919: Typo in comment

src/field/gen_fieldops.py
```diff
@@ -0,0 +1,211 @@
...
+            return self.name
+        return "&{}[i]".format(self.name)
+
+    def getElement(self, data=True):
+        """Get an element in the field, either using C-loop style index (default),
+        or with an `Indices` object (if `data=False`)
+
+        """
+
+        if self.field_type == 'BoutReal':
```
d7919: This section looks like it's very much tied to the specific template being used. I'm guessing there's not really much option here, but if we're thinking of expanding the templating a little further we might want to think about expanding the flexibility (e.g. adding arguments to allow us to specify the name of the x index, the size of x etc.  which can have sensible default values).

src/field/gen_fieldops.py
```diff
@@ -0,0 +1,211 @@
...
+        return copy(f2)
+    elif f2 == 'BoutReal':
+        return copy(f1)
+    else:
+        return copy(field3D)
+
+
+if __name__ == "__main__":
+    for_gcc = True
+    print(header)
```
d7919: We print to stdout and then redirect to file. Could this cause problems if something develops a warning/we want to provide updates on the process to the user via messages? Could we perhaps take an optional file to write to instead (e.g. using `sys.arg` or similar).

src/field/gen_fieldops.py
ZedThree: This is true. I don't think it would be too hard to make it more generic, but I think we'd need to see the other use-cases first.

src/field/gen_fieldops.py
ZedThree: That's sensible. The `for_gcc` argument could be passed in as well. I'm not entirely sure why I didn't do that already.

src/field/gen_fieldops.py
d7919: I was thinking a bit like https://github.com/boutproject/BOUT-dev/pull/818/files#diff-151f2a716d86ab78dc82ba9d24cd456cR184 where you use the defined dimension strings to form the variable names to use (which I think must match exactly what is used here).

ZedThree: Ah, ok. I'll see what can be done.

d7919: It's not a big issue from my view and changing it may just make things more complicated, but I do think that whilst things are mostly separated there are some dependencies in the choice of these variable names spread across the file which might be non-obvious when trying to extend/reuse.

src/field/gen_fieldops.py
```diff
@@ -0,0 +1,251 @@
...
+# OrderedDict to (try to) ensure consistency between python 2 & 3
+operators = OrderedDict([
+    ('*', 'multiplication'),
+    ('/', 'division'),
+    ('+', 'addition'),
+    ('-', 'subtraction'),
+])
+
+header = """// This file is autogenerated - see gen_fieldops.py
+#include <bout/mesh.hxx>
```
d7919: `utils.hxx` is required for `checkData` on `BoutReal` (otherwise can get ambiguous errors as tries to cast to either 2d/3d if these versions are available). I assume something you're including is including this for us at the moment as this obviously compiles fine as it passes the tests.

ZedThree: Ah, this is true. I'm going to merge this now, but I'll revisit this when we start using Block-Region/SDI

PR #819, opened 2018-01-31
Remove shifted grids from test-interchange-instability

ZedThree:
Shifted grids aren't currently used in this test case, so the `setLocation`s are unnecessary (and cause problems with #818)


d7919:
Should this also happen for the test-drift-instability case (which is essentially the same model)?

PR #821, opened 2018-01-31
Change default for include_yguards in Laplace inversion to false

johnomotani:
include_yguards was previously set to true by default, but this is probably not the behaviour expected or usually needed. Usually the inversion should be performed in the domain and the solution extended to guard cells by boundary conditions.

I set this many years ago, and changed the previous default behaviour. Apologies, I think now I shouldn't have done that. If include_yguards is set to true, it will do unnecessary laplace inversions in the guard cells at the targets (if there are any), potentially slowing down a simulation significantly since the extra work is only on the processors containing the y-boundary.

### DIFF
--- a/manual/sphinx/user_docs/laplacian.rst
+++ b/manual/sphinx/user_docs/laplacian.rst
@@ -150,7 +150,7 @@ condition on both AC and DC components.
    |                          | conditions. See :ref:`Laplace flags<tab-laplaceflags>` or               |                                              |
    |                          | :doc:`invert_laplace.cxx<../_breathe_autogen/file/invert__laplace_8cxx>`|                                              |
    +--------------------------+-------------------------------------------------------------------------+----------------------------------------------+
-   | ``include_yguards``      | Perform inversion in :math:`y`\ -boundary guard cells                   | ``true``                                     |
+   | ``include_yguards``      | Perform inversion in :math:`y`\ -boundary guard cells                   | ``false``                                    |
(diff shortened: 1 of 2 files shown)


ZedThree:
Please could you also update the documentation here:
http://bout-dev.readthedocs.io/en/latest/user_docs/laplacian.html?highlight=laplacian#id3

Found in:
https://github.com/boutproject/BOUT-dev/blob/invert_laplace_include_yguards_default_false/manual/sphinx/user_docs/laplacian.rst

johnomotani:
Thanks for the pointer! Manual updated now.

d7919:
I think this may have (re)introduced some incorrect behaviour in some of the tests.

In particular in test-interchange-instability we calculate phi at the start of the rhs and then use this in the bracket operator. Unfortunately the invert_laplace solve routine only solves for y slices that don't include the guards/boundaries if `include_yguards` is not true. This means that the DDY(phi) that is a part of the full bracket operator ends up using values from the guards of phi that have not been set. Whilst we have a communicate which we might hope to set the guards, we don't apply boundary conditions anywhere so I think the y-boundaries are uninitialised. This is an issue in the invalidateGuards branch which initialises all guards to Nan. As the boundary is not set this remains Nan and causes the run to fail.

bendudson:
For now the tests should be fixed I guess, so that include_yguards is true. It's a bit worrying that the tests don't catch this error. I think invalidating the guard cells would be a useful debugging tool.

PR #822, opened 2018-01-31
Python tools minor upgrades

johnomotani:
This PR adds caching to the BoutOutputs class and fixes a couple of minor issues with the showdata function.

### DIFF
--- a/tools/pylib/boutdata/data.py
+++ b/tools/pylib/boutdata/data.py
@@ -213,8 +213,10 @@ class BoutOutputs(object):
     """
     Emulates a map class, represents the contents of a BOUT++
     dmp files. Does not allow writing, only reading of data.
-    Currently there is no cache, so each time a variable
-    is read it is collected.
+    By default there is no cache, so each time a variable
+    is read it is collected; if caching is set to True variables
+    are stored once they are read.
+    Extra keyword arguments are passed through to collect.
     
     Example
     -------
@@ -222,16 +224,30 @@ class BoutOutputs(object):
     d = BoutOutputs(".")  # Current directory
     
     d.keys()     # List all valid keys
+
+    d.dimensions["ne"] # Get the dimensions of the field ne
     
     d["ne"] # Read "ne" from data files
+
+    d = BoutOutputs(".", prefix="BOUT.dmp", caching=True) # Turn on caching
+
+    Options
+    -------
+    prefix - sets the prefix for data files (default "BOUT.dmp")
+
+    caching - switches on caching of data, so it is only read into memory when first accessed (defaulte False)
+    
+    verbose - switches on printing from collect (default True)
     
     """
-    def __init__(self, path=".", prefix="BOUT.dmp"):
+    def __init__(self, path=".", prefix="BOUT.dmp", caching=False, **kwargs):
         """
         Initialise BoutOutputs object
         """
         self._path = path
         self._prefix = prefix
+        self._caching = caching
+        self._kwargs = kwargs
         
         # Label for this data
         self.label = path
@@ -243,16 +259,32 @@ class BoutOutputs(object):
         
         # Available variables
         self.varNames = []
+        self.dimensions = {}
+        self.evolvingVariableNames = []
+
+        # Private variables
+        self._datacache = {}
         
         with DataFile(file_list[0]) as f:
             # Get variable names
             self.varNames = f.keys()
+            for name in f.keys():
+                dimensions = f.dimensions(name)
+                self.dimensions[name] = dimensions
+                if name != "t_array" and "t" in dimensions:
+                    self.evolvingVariableNames.append(name)
         
     def keys(self):
         """
         Return a list of available variable names
         """
         return self.varNames
+
+    def evolvingVariables(self):
+        """
+        Return a list of names of time-evolving variables
+        """
+        return self.evolvingVariableNames
         
     def __len__(self):
         return len(self.varNames)
@@ -260,12 +292,18 @@ class BoutOutputs(object):
     def __getitem__(self, name):
         """
         Reads a variable using collect.
+        Caches result and returns later if called again, if self._caching=True
         
         """
 
-        # Collect the data from the repository
-        data = collect(name, path=self._path, prefix=self._prefix)
-        return data
+        if self._caching:
+            if name not in self._datacache.keys():
+                self._datacache[name] = collect(name, path=self._path, prefix=self._prefix, **self._kwargs)
+            return self._datacache[name]
+        else:
+            # Collect the data from the repository
+            data = collect(name, path=self._path, prefix=self._prefix, **self._kwargs)
+            return data
 
     def __iter__(self):
         """
@@ -286,12 +324,13 @@ class BoutOutputs(object):
         return text
     
 
-def BoutData(path=".", prefix="BOUT.dmp"):
+def BoutData(path=".", prefix="BOUT.dmp", caching=False, **kwargs):
     """
     Returns a dictionary, containing the contents of a BOUT++
     output directory. Does not allow writing, only reading of data.
-    Currently there is no cache, so each time a variable
-    is read it is collected.
+    By default there is no cache, so each time a variable
+    is read it is collected; if caching is set to True variables
+    are stored once they are read.
     
     Example
     -------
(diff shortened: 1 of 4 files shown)


johnomotani:
Have implemented a maximum size for the cache. If the caching option is set to a number, that is the maximum size in GB, if it's True then the cache is unlimited, if it's False there is no caching.

Is it worth now setting the default to a sensible cache size instead of False. 4GB maybe? I guess most people have at least that much memory these days.

tools/pylib/boutdata/data.py
```diff
@@ -235,7 +235,9 @@ class BoutOutputs(object):
     -------
     prefix - sets the prefix for data files (default "BOUT.dmp")
 
-    caching - switches on caching of data, so it is only read into memory when first accessed (defaulte False)
+    caching - switches on caching of data, so it is only read into memory when first accessed (default False)
+              If caching is set to a number, it gives the maximum size of the cache in GB, after which entries will be discarded in first-in-first-out order to prevent the cache getting too big.
+              If the variable being returned is bigger than the maximum cache size, then the variable will be returned without being added to the cache, and the rest of the cache will be left.
```
dschwoerer: Could you fix that to be not longer than 80 chars?
In emacs e.g. you can enable line wrapping with `M-x auto-fill-mode`

dschwoerer:
I got my scripts killed for using to much memory, so potentially wasting 4 GB on caching would be an issue.

tools/pylib/boututils/showdata.py
```diff
@@ -325,6 +325,11 @@ def showdata(vars, titles=[], legendlabels = [], surf = [], polar = [], tslice =
 
     # Generate grids for plotting
     # Try to use provided grids where possible 
+    # If x and/or y are not lists, apply to all variables
+    if not isinstance(x, (list,tuple)):
```
ZedThree: What happens if `x` is a numpy `ndarray`? We probably want all the `isinstance(..., list)` checks to be `isinstance(..., (list, tuple, np.ndarray))`

tools/pylib/boutdata/data.py
```diff
@@ -308,6 +375,17 @@ def BoutData(path=".", prefix="BOUT.dmp"):
 
     d["outputs"]["ne"] # Read "ne" from data files
     
+    d = BoutData(".", prefix="BOUT.dmp", caching=True, verbose=False) # Turn on caching and turn off verbosity
+
+    Options
+    -------
+    prefix - sets the prefix for data files (default "BOUT.dmp")
+
+    caching - switches on caching of data, so it is only read into memory when first accessed (defaulte False)
```
ZedThree: Typo ("defaulte")

tools/pylib/boututils/showdata.py
johnomotani: The use case I have in mind is that x=Rxy and y=Zxy so that showdata will make contour plots in the R-Z domain like this:
animation
So my aim was to check if we have a single ndarray, or different ones for different variables. It will fail if you try to pass in an 'array of arrays', so more checking the sanity of inputs would be useful.

I think it's too hard to tell the difference between a 2d array for the x-coordinate and a 2d array of 1d coordinates for each variable, so if the user wants a different x and y for each variable, we'll have to make them pass a list not an array, but I will try to add some sanity checking with helpful error messages if dimensions don't match, etc.

johnomotani:
> I got my scripts killed for using to much memory, so potentially wasting 4 GB on caching would be an issue.

I'd like to set the defaults for interactive use. I think it's not too inconvenient in a script to have to put something like BoutOutputs(caching=False), especially as you'll probably only need to type it once; but then you can type data=BoutOutputs() in an interactive session and have it 'just work' sensibly.

It is true that changing the default away from caching=False could, as you say, cause problems for existing scripts...

johnomotani:
If there's nothing else I should fix/upgrade, should I rebase this branch to squash the commits before it gets merged?

ZedThree:
No need to bother with that. I would only recommend squashing if there's lots of WIP commits.

For future reference though, you can leave out the filename from the commit message -- it's very easy to see that information in the commit itself. This is a super-useful blog post on writing commit messages that's well worth reading: https://chris.beams.io/posts/git-commit/

tools/pylib/boutdata/data.py
```diff
@@ -243,29 +264,78 @@ def __init__(self, path=".", prefix="BOUT.dmp"):
...
+                self._datacachesize = 0
+                self._datacachemaxsize = self._caching*1.e9
         
         with DataFile(file_list[0]) as f:
             # Get variable names
             self.varNames = f.keys()
+            for name in f.keys():
+                dimensions = f.dimensions(name)
+                self.dimensions[name] = dimensions
+                if name != "t_array" and "t" in dimensions:
```
dschwoerer: Why exclude `t_array`?
The time is also evolved ...

johnomotani: because I was using `evolvingVariables()` to plot all the output, and don't want to plot `t_array`. Also, since `t_array` is always present, we don't need to query the data files to find out its name.

The use case I mean would be something like:
```
data = BoutOutputs()
showdata( [ data[x][:,:,0,:] for x in data.evolvingVariables() ] )
```

tools/pylib/boututils/showdata.py
```diff
@@ -585,19 +641,25 @@ def init():
...
+            except Exception:
+            #Try specifying writer by string if ffmpeg not found
                 try:
-                    anim.save(movie+'.mp4',writer = 'ffmpeg', fps=fps, dpi=dpi, extra_args=['-vcodec', 'libx264'])
+                    anim.save(movie,writer = 'ffmpeg', fps=fps, dpi=dpi, extra_args=['-vcodec', 'libx264'])
                 except Exception:
                      print('Save failed: Check ffmpeg path')
                      raise
+        elif movietype == 'gif':
+            anim.save(movie+'.gif',writer = 'imagemagick', fps=fps, dpi=dpi, extra_args=['-vcodec', 'libx264'])
```
ZedThree: `extra_args` can be dropped here -- `vcodec` is an ffmpeg option, and `libx264` is the mpeg-4 codec

tools/pylib/boututils/showdata.py
```diff
@@ -377,14 +396,21 @@ def showdata(vars, titles=[], legendlabels = [], surf = [], polar = [], tslice =
 
 
         if not (global_colors):
-            clevels.append(linspace(fmin[i], fmax[i], Ncolors))
+            if not float32(fmax[i])>float32(fmin[i]):
```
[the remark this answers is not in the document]
johnomotani: No it wasn't required. I put it in as a fix for an error contour() throws if the maximum and minimum are too close together, but checking again now, it only seems to need a reasonable tolerance like 2.e-15 for rounding errors.

tools/pylib/boututils/showdata.py
```diff
@@ -396,8 +396,8 @@ def showdata(vars, titles=[], legendlabels = [], surf = [], polar = [], tslice =
 
 
         if not (global_colors):
-            if not float32(fmax[i])>float32(fmin[i]):
-                clevels.append([fmax[i],fmax[i]+1.])
+            if (fmax[i]-fmin[i])/abs(fmax[i]+fmin[i]) <= 2.e-15:
```
ZedThree: Could this be done then with:

```python
from numpy import isclose
...
if isclose(fmin[i], fmax[i]):
    # Would be better to use a warning here
    print("Contour levels too close, using ", fmin[i])
    clevels.append(fmin[i])
else:
...
```
This will cause `contour` to plot only a single contour, but that's essentially what's being asked.

johnomotani: Agreed isclose() is a nicer test :thumbsup:

clevels actually ends up being used by contourf(), not contour(). I was wrong when I wrote contour() above. contourf() needs at least two levels, which must be far enough apart. Keeping Ncolors contours as I have in the latest change allows us to see the difference between a totally constant field and one with O(1.e-15) variations. Maybe not much practical use, but it seems to work without throwing errors, so why not?

ZedThree: It reads much better now, it's much clearer what's going on :+1: 

I'm not sure variations on the 1.e-15 level mean anything -- this is very close to the machine precision level -- but this looks fine as is.

johnomotani:
I added a boututils.warnings module with wrappers for warnings.warn because I wanted a warning that would always be printed.

If this is a good idea, we should probably add some documentation somewhere suggesting to use them. Do we have some developer documentation for python routines? How to add warnings isn't something a user should need to see...

ZedThree:
We have some documentation for them [here][1] ([source][2]), but it could easily stand to be improved...

[1]: http://bout-dev.readthedocs.io/en/latest/user_docs/python.html
[2]: https://github.com/boutproject/BOUT-dev/blob/next/manual/sphinx/user_docs/python.rst

johnomotani:
Just changed the last commit. Naming the module 'warnings' has a strong danger of a name conflict, so I've renamed it 'boutwarnings'.

PR #824, opened 2018-02-01
Dataiterator improvements

dschwoerer:
part of #742


bendudson:
These changes look reasonable, but Joseph and Peter should probably also pass a beady eye over them. It would help understand the fixes if the internal variables had some comments to explain exactly what they are e.g. start, end vs min, max

dschwoerer:
Do the comments make sense?

ZedThree:
The code all looks fine and makes sense -- the comments are nice and clear, thanks!

I'm not sure about the test though. It's not clear to me what it's doing, and how it's different from the existing unit tests, aside from the loop over mpi+openmp processes. If it's just the hybrid parallel loop, we could rely on the different Travis builds for that.

bendudson:
Yes, happy to merge this without test-dataiterator2 (especially as it doesn't pass in this branch due to FieldPerp not having region)

dschwoerer:
`test-dataiterator2` is removed.
It might be worth activating `test-dataiterator` to catch issues with mixing xstart/xmin etc. in the future.
This might be done after #843 is merged, as this fixes that the test is run with the expected amount of openmp-threads.
Sorry, I must have missed @ZedThree comment last week :/

PR #825, opened 2018-02-01
prevent some issues with -Ofast

dschwoerer:
part of #742


dschwoerer:
These flags are only added if the compiler does not complain.
See [configure.ac L240 ff](https://github.com/boutproject/BOUT-dev/pull/825/files#diff-67e997bcfdac55191033d57a16d1408aR240)

Issue #823, opened 2018-02-01
Flux stencils


dschwoerer:
`VDDX_U4` is only 3rd order accurate - is this intentionally?
`FDDX_U1` has a sign error.
`FDDX_NND` is only 1st order accurate - is this intentionally?

The tests to reproduce are [here](https://github.com/dschwoerer/BOUT-dev/tree/my_v4/tests/MMS/derivatives_flux)

dschwoerer:
Fixed by #827

PR #827, opened 2018-02-01
fix sign error in FDDX_U1

dschwoerer:
fix for #823


dschwoerer:
Any thoughts on this one?

src/mesh/index_derivs.cxx
```diff
@@ -233,6 +233,8 @@ BoutReal FDDX_C4(stencil &v, stencil &f) {
 /// Non-oscillatory, containing No free parameters and Dissipative (NND) scheme
 /// http://arxiv.org/abs/1010.4135v1
 BoutReal FDDX_NND(stencil &v, stencil &f) {
+  throw BoutException("This converges only first order.\n"
```
d7919: What's the expected order (I couldn't see this in a skim of the reference)? Aren't limiter methods like this a combination of method orders, dropping back to first order near steep gradients to ensure we don't introduce extra oscillations?

src/mesh/index_derivs.cxx
```diff
@@ -219,7 +219,7 @@ BoutReal FDDX_U1(stencil &v, stencil &f) {
   vs = 0.5*(v.c + v.p);
   result -= (vs >= 0.0) ? vs * f.c : vs * f.p;
 
-  return result;
+  return - result;
```
d7919: Looks reasonable to me, but I could be wrong.

Should we also change the staggered version (https://github.com/boutproject/BOUT-dev/blob/41d05ad89252993a271aa411265bf31857a77e90/src/mesh/index_derivs.cxx#L386)?

src/mesh/index_derivs.cxx
dschwoerer: I tested only smooth functions (sin,cos) - and I thought it should be higher.
If you say that first order for smooth functions is the expected behaviour - I am happy to remove the warning.

d7919: I don't really know what the expected behaviour is for this specific operator.

bendudson: I think this function can be removed; it was added as a test many many years ago, and never really used for anything. Perhaps it can be added again later if tested

dschwoerer: I thought keeping it wouldn't hurt ...
If somebody was using it, this way they get a clear error message, and it is easy to re-enable, if this works for that use case ... but I have no strong opinion - I would also be happy to remove it

src/mesh/index_derivs.cxx
dschwoerer: @d7919 `FDDX_U1_stag` should be fixed now.

src/mesh/index_derivs.cxx
```diff
@@ -419,7 +421,7 @@ static DiffNameLookup DiffNameTable[] = { {DIFF_U1, "U1", "First order upwinding
 					  {DIFF_W2, "W2", "Second order WENO"},
 					  {DIFF_W3, "W3", "Third order WENO"},
 					  {DIFF_C4, "C4", "Fourth order central"},
-					  {DIFF_U4, "U4", "Fourth order upwinding"},
+					  {DIFF_U3, "U3", "Third order upwinding"},
```
ZedThree: We should keep the `"U4"` name as well, but make it a synonym for `"U3"` till after 4.2

src/mesh/index_derivs.cxx
```diff
@@ -162,8 +162,8 @@ BoutReal VDDX_U2(BoutReal vc, stencil &f) {
   return vc>=0.0 ? vc*(1.5*f.c - 2.0*f.m + 0.5*f.mm): vc*(-0.5*f.pp + 2.0*f.p - 1.5*f.c);
 }
 
-/// upwind, 4th order
-BoutReal VDDX_U4(BoutReal vc, stencil &f) {
+/// upwind, 3th order
```
ZedThree: Typo: 3th -> 3rd

src/mesh/index_derivs.cxx
```diff
@@ -390,7 +392,7 @@ BoutReal FDDX_U1_stag(stencil &v, stencil &f) {
   // Upper cell boundary
   result -= (v.p >= 0) ? v.p * f.c : v.p * f.p;
 
-  return result;
+  return - result;
```
ZedThree: Fix whitespace

src/mesh/index_derivs.cxx
ZedThree: Fix whitespace

src/mesh/index_derivs.cxx
ZedThree: I'm with @bendudson on this. If it was experimental in the first place and we're not certain that it works, let's remove it entirely for now. We can put it back if we get it working/tested.

ZedThree:
Looks good, just some minor things to tidy up

PR #833, opened 2018-02-05
WIP: Fieldops jinja nogcc

dschwoerer:
removes the `for_gcc` workaround.
Still needs work on the makefiles and configure.

Is that something worth continuing?


ZedThree:
Can you show some results comparing this to the other methods in `examples/performance/iterator`? You can try using #835 to get more consistent results.

We've done a lot of investigation into improving the Field loops, and the `SingleDataIterator` and "Block-Region" approaches are currently our preferred methods. They'll be going into `next` soon. 

In terms of performance, both these methods scale well with problem size and number of OpenMP threads, and are competitive with flat C loops. They also have some advantages in terms of flexibility -- we can change what regions they loop over at compile time, possibly even run time.

dschwoerer:
This shows roughly the same speed as the current implementation - but would be more portable.
If this is however anyway replaced by the single data iterator - than this is not worth pursuing...

PR #834, opened 2018-02-05
prefer make if it is the same as gmake

dschwoerer:


ZedThree:
If they are the same, why prefer `make`?

dschwoerer:
Because `make` is shorter.
Further, if we suggest `gmake` only when it is actually needed, the message will not be ignored.
Do you actually type `gmake`?

bendudson:
Closing since doesn't seem to be a strong case for this.

PR #839, opened 2018-02-06
More region

dschwoerer:
add the region to the interface of the min/max functions.

Part of #742 

Also add dataiterator interface to fieldperp

### DIFF
--- a/include/field.hxx
+++ b/include/field.hxx
@@ -137,6 +137,9 @@ class Field {
    * Return the number of nz points
    */
   virtual int getNz() const;
+
+  /// Make region mendatory for all fields
+  virtual const IndexRange region(REGION rgn) const = 0;
  protected:
   Mesh * fieldmesh;
   /// Supplies an error method. Currently just prints and exits, but
--- a/include/field2d.hxx
+++ b/include/field2d.hxx
@@ -302,29 +302,31 @@ const Field2D abs(const Field2D &f);
 
 /*!
  * Calculates the minimum of a field, excluding
- * the boundary/guard cells. 
+ * the boundary/guard cells by default.
  * By default this is only on the local processor,
  * but setting allpe=true does a collective Allreduce
  * over all processors.
  *
  * @param[in] f  The field to loop over
  * @param[in] allpe  Minimum over all processors?
+ * @param[in] rgn  the boundaries that should be ignored
  * 
  */
-BoutReal min(const Field2D &f, bool allpe=false);
+BoutReal min(const Field2D &f, bool allpe=false, REGION rgn=RGN_NOBNDRY);
 
 /*!
  * Calculates the maximum of a field, excluding
- * the boundary/guard cells. 
+ * the boundary/guard cells by default.
  * By default this is only on the local processor,
  * but setting allpe=true does a collective Allreduce
  * over all processors.
  *
  * @param[in] f  The field to loop over
  * @param[in] allpe  Minimum over all processors?
+ * @param[in] rgn  the boundaries that should be ignored
  * 
  */
-BoutReal max(const Field2D &f, bool allpe=false);
+BoutReal max(const Field2D &f, bool allpe=false, REGION rgn=RGN_NOBNDRY);
 
 /*!
  * Test if all values of this field are finite
--- a/include/fieldperp.hxx
+++ b/include/fieldperp.hxx
@@ -80,6 +80,8 @@ class FieldPerp : public Field {
   const DataIterator begin() const;
   const DataIterator end() const;
 
+  const IndexRange region(REGION rgn) const;
+
   /*!
    * Direct data access using DataIterator indexing
    */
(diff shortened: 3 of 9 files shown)


dschwoerer:
Ok, i will make smaller PRs 👍 
Wasn't sure how small they should be.

bendudson:
No problem; keep them coming ;)

ZedThree:
Please could you add unit tests for the min/max functions? Should be easy enough to copy the existing ones in [`tests/unit/field/test_field3d.cxx`][1] but add regions and just check that they get the values in the boundaries.


[1]: https://github.com/boutproject/BOUT-dev/blob/master/tests/unit/field/test_field3d.cxx#L960

src/field/fieldperp.cxx
```diff
@@ -236,3 +236,20 @@ const FieldPerp sliceXZ(const Field3D& f, int y) {
...
+        0, 0,
+        0, nz-1};
+    break;
+  case RGN_NOX:
+    return IndexRange{getMesh()->xstart, getMesh()->xend,
+        0, 0,
+        0, nz-1};
+    break;
+  default:
+    throw BoutException("Field3D::region() : Requested region not implemented");
```
d7919: Exception message should use field perp rather than 3D I think.

include/fieldperp.hxx
```diff
@@ -80,6 +80,8 @@ class FieldPerp : public Field {
   const DataIterator begin() const;
   const DataIterator end() const;
 
+  const IndexRange region(REGION rgn) const;
```
d7919: This seems separate from the addition of regions to min/max and doesn't seem to be used, is this in preparation for something else?

dschwoerer: This is unrelated.
It allows to iterate over a fieldperp - but I don't think I use it in the aiolos mesh.
It is rather to unify the fields.

d7919:
Is it worth doing a similar thing for `checkData`?

dschwoerer:
I am not sure - checkData is not a member function - so not sure that would work.
But we could implement checkData for Field, rather for the specific fields, if they all have the needed functions ...

d7919:
I think min and max aren't member functions either.

bendudson:
Generally I have tried to reduce the number of member functions: wherever possible code should use the class public API, and not depend on private implementation details. Reducing the size of the class helps make it more comprehensible. Proof by counterexample: the giant mess that is Mesh...

dschwoerer:
oh, you meant adding a REGION arguments?
Sorry, I thought you meant specifiing it in field.

But I think min, max, checkData and maybe more functions could equally well be defined over `Field`, rather than over the specific fields.

d7919:
Sorry I should have been clearer!

bendudson:
We could define Field versions, but then would need an efficient way to iterate over them without just casting back to Field2D/Field3D etc. If just the base class then indexing would be a virtual function call. We could perhaps make them template functions, but I don't think that has much advantage over the current scheme of using a consistent interface for the overloaded functions.

dschwoerer:
@bendudson How do the z boundaries work?
I cannot see zstart or zend in the mesh.

include/field.hxx
```diff
@@ -137,6 +137,9 @@ class Field {
    * Return the number of nz points
    */
   virtual int getNz() const;
+
+  /// Make region mendatory for all fields
```
ZedThree: Typo: "mandatory"

include/field2d.hxx
```diff
@@ -302,29 +302,31 @@ const Field2D abs(const Field2D &f);
...
  * Calculates the minimum of a field, excluding
- * the boundary/guard cells. 
+ * the boundary/guard cells by default.
  * By default this is only on the local processor,
  * but setting allpe=true does a collective Allreduce
  * over all processors.
  *
  * @param[in] f  The field to loop over
  * @param[in] allpe  Minimum over all processors?
+ * @param[in] rgn  the boundaries that should be ignored
```
ZedThree: I think this should be "rgn  Calculate the minimum inside this region".

Currently it sounds like `rgn` should contain the points you want to ignore, rather than the points you want to consider

dschwoerer: rgn can only be used to ignore the boundaries - no general region.
maybe
```
rgn  The boundaries to be included in the calculation
```
?

ZedThree: What about `RGN_ALL`? We're also moving towards supporting very general regions which may even be some subset of the interior points, so this should be

    * @param[in] rgn  The region to calculate the minimum over

and likewise for `max`

dschwoerer: `RGN_ALL` means to include all boundaries - so matches the description as far as I understand.

I wrote the documentation for what is currently done - not what will be possible in the future.
I can change of course - but I am not the expert on the generalized region thing, am not sure whether or when it is going to get merged, when that function is adapted, so I didn't match it.

Issue #389, opened 2016-11-25
xend is not an C++ end


dschwoerer:
From [v4-RC](https://github.com/boutproject/BOUT-dev/blob/v4.0.0-RC/include/bout/mesh.hxx#L233)
```
  /// Local ranges of data (inclusive), excluding guard cells
  int xstart, xend, ystart, yend;
```
As @JensMadsen pointed out, the C++ way would be that the end is outside the domain ...

On a related issue, it would be nice if the speeling would be more consistent, i.e. `localNx` instead of `LocalNx` - maybe also change to `xend` to `xEnd` or even `endX`?

JensMadsen:
I think that we should write down a style guide and add to the repo?

ZedThree:
That's the plan :)

There's going to be more information on the mailing list soon!

dschwoerer:
I would like to fix this in the 5.0 release ...

PR #841, opened 2018-02-06
Block region iterating

d7919:
Adds a Region class that can be used for describing regions over which to iterate. This should be compatible with the SDI PR (with minor tweaks) and also adds in a "contiguous block" approach which represents the loop over a region as a two level loop, the outer loop being suitable to OpenMP parallelise and the inner loop vectorising. A couple of macros have been added to assist this.

A couple of examples of use are in `tests/integrated/test_region_iterator` and there are also unit tests.

Enhancements of this approach are possible (e.g. being able to add `Region`s would be nice and there could be routines to calculate offset indices for use in indexing) but I think can come later.


include/bout/region.hxx
```diff
@@ -0,0 +1,236 @@
...
+#if MAXREGIONBLOCKSIZE <= 0
+#error "In region.hxx must set MAXREGIONBLOCKSIZE > 0"
+#endif
+
+//The following loops vectorise well well index is type int, needs testing
+//with current approach where index is ind2D or ind3D
+#define BLOCK_REGION_LOOP_SERIAL(region, index, ...)			\
+  {const auto blocks = region.getBlocks();				\
+  for(auto block = blocks.begin(); block < blocks.end() ; ++block){ \
+  for(auto index = (*block).first ; index <= (*block).second; ++index){ \
```
ZedThree: Could this just use `block->first`, `block->second`?

include/bout/region.hxx
```diff
@@ -0,0 +1,236 @@
...
+  __VA_ARGS__ \
+    }}}
+
+#define BLOCK_REGION_LOOP(region, index, ...)			\
+  {const auto blocks = region.getBlocks();			\
+  BOUT_OMP(parallel for)				    \
+  for(auto block = blocks.begin(); block < blocks.end() ; ++block){ \
+  for(auto index = (*block).first ; index <= (*block).second; ++index){ \
```
ZedThree: Same as above?

include/bout/region.hxx
```diff
@@ -0,0 +1,236 @@
...
+};
+
+template<typename T = ind3D>
+class Region {
+  //Following prevents a Region being created with anything other
+  //than ind2D or ind3D as template type
+  static_assert(std::is_base_of<ind2D,T>::value || std::is_base_of<ind3D,T>::value, "Region must be templated with either ind2D or ind3D");
+  
+public:
+  typedef T data_type;
```
ZedThree: This isn't used anywhere, is it needed?

tests/integrated/test-region-iterator/test_region_iterator.cxx
```diff
@@ -0,0 +1,64 @@
...
+  if(nerr != nerrExpected ) throw BoutException("Unexpected values found in 'e', count %d",nerr);
+
+  SOLVE_FOR(n);
+  return 0;
+}
+
+int physics_run(BoutReal t) {
+  ddt(n) = 0.;
+  return 0;
+}
```
ZedThree: I think most of this test is also done in the unit test. Can we get away with just the unit tests?

include/bout/region.hxx
```diff
@@ -0,0 +1,236 @@
...
+  {const auto blocks = region.getBlocks();			\
+  BOUT_OMP(parallel for)				    \
+  for(auto block = blocks.begin(); block < blocks.end() ; ++block){ \
+  for(auto index = (*block).first ; index <= (*block).second; ++index){ \
+  __VA_ARGS__ \
+    }}}
+
+class specificInd{
+public:
+  int ind;
```
ZedThree: Personally, I would be happier with `index`, but I can live with this :)

include/bout/region.hxx
d7919: Sure can do that.

include/bout/region.hxx
d7919: It's not currently used but I thought it might be useful to be able to query what a given Region is templated with (say if we start using `auto` in places like return values with C++14).

tests/integrated/test-region-iterator/test_region_iterator.cxx
dschwoerer: Parallel unit tests are not working or at least not enabled.
This is tested in parallel.

include/bout/region.hxx
```diff
@@ -0,0 +1,236 @@
...
+  specificInd(){};
+  specificInd(int i): ind(i){};
+  specificInd& operator++(){++ind;return *this;}
+  specificInd& operator++(int){++ind;return *this;}
```
dschwoerer: This is extremely dangerous!
Rather than doing it wrong - throw an exception?

tests/integrated/test-region-iterator/test_region_iterator.cxx
d7919: Yes, I'd like to keep this test if possible (at least for now). If this starts getting used then we may want to test/demo some more sophisticated uses that aren't suitable for unit tests. It's also a handy case should we wish to test scaling etc.

include/bout/region.hxx
d7919: Could you expand, not sure which line(s) you're referring to?

dschwoerer: line 71.
The postfix operator is not a prefix operator.
If I write `i++` I certainly don't expect to get `++i`.
This is an example that would be currently broken:
```
for (;i != i.end;){
  auto fu=in[i];
  r[i++]=fu;
}
```
`i++` should return the non-incremented value - not the incremented one.
See https://stackoverflow.com/questions/3846296/how-to-overload-the-operator-in-two-different-ways-for-postfix-a-and-prefix
https://stackoverflow.com/questions/24853/what-is-the-difference-between-i-and-i

d7919: Ah yes, thanks! I should note that this error is already in the code inside `include/bout/sys/range.hxx` (where I quickly grepped for the syntax for this as I didn't have internet at the time)

include/bout/region.hxx
```diff
@@ -0,0 +1,236 @@
...
+  bool operator==(const specificInd &x) const {
+    return ind == x.ind;
+  }
```
dschwoerer: Would it be worth to make sure that both iterators are on the same mesh/region/...?
I am not sure whether this is an issue, but we could enable additional checking e.g. if `CHECK=3` to make sure everything is the same as expected, and throw an exception instead.
similar for the other bool operators, to make sure the comparisons are sane.

include/bout/region.hxx
```diff
@@ -0,0 +1,236 @@
...
+};
+//Make identical subclasses with different names
+class ind3D: public specificInd{
+public:
+  ind3D(): specificInd(){};
+  ind3D(int i): specificInd(i){};
+  //Note operator= from base class is always hidden
+  //by implicit method so have to be explicit
+  ind3D& operator=(const int& i){ind=i;return *this;}
```
dschwoerer: why take an int as const ref?
Taking by value should be faster.

include/bout/region.hxx
```diff
@@ -0,0 +1,236 @@
...
+  
+private:
+  regionIndices indices;  //Flattened indices
+  contiguousBlocks blocks;//Contiguous sections of flattened indices
+
+  /// Helper function to create a regionIndices, given the start and end
+  /// points in x, y, z, and the total y, z lengths
+  inline regionIndices createRegionIndices(int xstart, int xend, int ystart, int yend,
+  					   int zstart, int zend, int ny, int nz){
+    int len = (xend - xstart + 1) * (yend - ystart + 1) * (zend - zstart + 1);
```
dschwoerer: no check that `xend > xstart +1` etc.

src/mesh/mesh.cxx
```diff
@@ -322,3 +322,57 @@ ParallelTransform& Mesh::getParallelTransform() {
 Coordinates *Mesh::createDefaultCoordinates() {
   return new Coordinates(this);
 }
+
+
+Region<> & Mesh::getRegion3D(const std::string &region_name){
```
dschwoerer: why do you prefer strings over enums?
This changes a compile time check, to a run time check, and adds runtime overhead.

include/bout/region.hxx
d7919: Yes, those would be sensible tests. I'll add them.

src/mesh/mesh.cxx
d7919: My understanding is that an enum can't be extended (https://stackoverflow.com/a/1804924) and we may want to allow users/algorithms to register their own regions.

tests/integrated/test-region-iterator/runtest
```diff
@@ -0,0 +1,55 @@
...
+from sys import exit
+
+MPIRUN=getmpirun()
+
+print("Making Region Iterator test")
+s, out = shell_safe("make > make.log",pipe=True)
+if s:
+  print("Failed to run make")
+  exit(s);
+
```
dschwoerer: The `_safe` versions are throwing.
No need to test the output, `s` is only returned if it `0`.
See:
https://github.com/boutproject/BOUT-dev/blob/next/tools/pylib/boututils/run_wrapper.py#L196
Same goes for the calls below.

src/mesh/mesh.cxx
bendudson: An advantage is that it makes adding custom regions in user code easier. A user can add regions and refer to them with strings. 

An alternative might be to have a vector of regions (rather than a map): regions could then be referred to by an integer index. Adding a new region would return the index, probably wrapped in a class type rather than a raw integer. We could hard-wire built in regions to have particular indices perhaps. 

I'd like to see if this really is a significant overhead, before giving up the convenience of strings and maps.

tests/integrated/test-region-iterator/runtest
d7919: Ah yes, thanks.

include/bout/mesh.hxx
```diff
@@ -562,7 +565,35 @@ class Mesh {
...
+    return getRegion3D(region_name);
+  }
+  Region<ind3D> &getRegion3D(const std::string &region_name);
+  Region<ind2D> &getRegion2D(const std::string &region_name);
   
+  /// Add a new region to the region_map for the data iterator
+  ///
+  /// Outputs an error message if region_name already exists
+  void addRegion(const std::string &region_name, Region<> region){
+    return addRegion3D(region_name, region);
```
dschwoerer: why return in void function?

include/bout/region.hxx
d7919: No good reason.

include/bout/mesh.hxx
d7919: In case we decide that these routines should actually return something/to demonstrate that these two routines have the same return type. Does this harm anything?

src/mesh/mesh.cxx
dschwoerer: Ok, that makes sense.
I am not so much worried about the runtime cost, but having checks at compile time are awesome ...
Would it be viable to have both? Some hard coded ones as enum, and additional the interface with strings, such that users can define there own?
I don't like integers - they are not user friendly ...

tests/integrated/test-region-iterator/runtest
```diff
@@ -0,0 +1,49 @@
...
+cmd = "./test_region_iterator"
+code = 0 # Return code
+pipe=True
+
+for nproc in [1,2]:#Number of mpi procs
+  for mthread in [1]:#Number of omp threads (not yet supported)
+    print("\t{n} processors and {m} threads".format(n=nproc,m=mthread))
+
+    for f in flags:
+        #print("\tflags '"+f+"'")
```
ZedThree: Not sure if it's just github, but it looks like you're using 4 spaces for indentation here, but 2 elsewhere

include/bout/region.hxx
```diff
@@ -0,0 +1,251 @@
...
+  // Following prevents a Region being created with anything other
+  // than ind2D or ind3D as template type
+  static_assert(std::is_base_of<ind2D, T>::value || std::is_base_of<ind3D, T>::value,
+                "Region must be templated with either ind2D or ind3D");
+
+public:
+  typedef T data_type;
+
+  /// Indices to iterate over
+  typedef std::vector<T> regionIndices;
```
ZedThree: typedefs should start with a capital, `RegionIndices`, etc.

include/bout/region.hxx
```diff
@@ -0,0 +1,251 @@
...
+#error "In region.hxx must set MAXREGIONBLOCKSIZE > 0"
+#endif
+
+// The following loops vectorise well well index is type int, needs testing
+// with current approach where index is ind2D or ind3D
+#define BLOCK_REGION_LOOP_SERIAL(region, index, ...)                                     \
+  {                                                                                      \
+    const auto blocks = region.getBlocks();                                              \
+    for (auto block = blocks.begin(); block < blocks.end(); ++block) {                   \
+      for (auto index = block->first; index <= block->second; ++index) {                 \
```
bendudson: Should `second` be a C++ end, so that `index < block->second`? It's not a big deal, and this code is similar to the current xstart, xend indices. Is it better to be consistent with existing BOUT++ code, or to be consistent with C++ conventions?

include/bout/region.hxx
```diff
@@ -0,0 +1,251 @@
...
+    const auto blocks = region.getBlocks();                                              \
+  BOUT_OMP(parallel for)                                                                 \
+    for (auto block = blocks.begin(); block < blocks.end(); ++block) {                   \
+      for (auto index = block->first; index <= block->second; ++index) {                 \
+        __VA_ARGS__                                                                      \
+      }                                                                                  \
+    }                                                                                    \
+  }
+
+class specificInd {
```
bendudson: Can we make this `SpecificInd` (capital first letter) to be consistent with other classes.

include/bout/region.hxx
```diff
@@ -0,0 +1,251 @@
...
+  }
+  bool operator==(const specificInd &x) const { return ind == x.ind; }
+  bool operator!=(const specificInd &x) const { return ind != x.ind; }
+  bool operator>(const specificInd &x) const { return ind > x.ind; }
+  bool operator>=(const specificInd &x) const { return ind >= x.ind; }
+  bool operator<(const specificInd &x) const { return ind < x.ind; }
+  bool operator<=(const specificInd &x) const { return ind <= x.ind; }
+};
+// Make identical subclasses with different names
+class ind3D : public specificInd {
```
bendudson: `Ind3D`?

include/bout/region.hxx
```diff
@@ -0,0 +1,251 @@
...
+  ind3D() : specificInd(){};
+  ind3D(int i) : specificInd(i){};
+  // Note operator= from base class is always hidden
+  // by implicit method so have to be explicit
+  ind3D &operator=(int i) {
+    ind = i;
+    return *this;
+  }
+};
+class ind2D : public specificInd {
```
bendudson: `Ind2D`?

include/bout/region.hxx
d7919: Sure, although this is not at all consistently adopted throughout the code!

d7919: Agree we should be consistent (I think as my specificInd started as a struct I hadn't noted this).

include/bout/region.hxx
dschwoerer: I would prefer if BOUT++ would be consistent with C++ - therefore C++ and change everything else for the 5.0 release to C++ stile.
#389 is not about documentation

d7919: This should be addressed now (in this place).

tests/integrated/test-region-iterator/runtest
dschwoerer: further `flags` is not used and `f` is overwritten - sounds just like what I fixed in #843 :)

PR #842, opened 2018-02-06
make performance/arithmetic more reliable

dschwoerer:


examples/performance/arithmetic/arithmetic.cxx
```diff
@@ -14,57 +14,61 @@ typedef std::chrono::duration<double> Duration;
...
-      rd++;
-      ad++;
-      bd++;
-      cd++;
+
+    Duration elapsed1,elapsed2,elapsed3,elapsed4;
+    for (int ik=0;ik<1e2;++ik){
+      SteadyClock start1 = steady_clock::now();
+      result1 = 2.*a + b * c;
+      elapsed1 = steady_clock::now() - start1;
```
d7919: Doesn't this end up just recording the time for the last iteration of the outer loop (or is that what's intended, with 99 trial runs)?

dschwoerer: yes, this is intentionally.
100 is maybe an overkill ... but it doesn't hurt.
Maybe switch to `min`?
`+=` would include the cache issues from the old version.

d7919: Ah OK I see, so it is intended for warmup rather than averaging. A comment may help here. 

If we wanted to also add averaging in how would you recommend we do it in this case? A separate loop after the warmup?

dschwoerer:
I think minimum is better, in case the thread gets suspended during one operation, as that shouldn't affect the reported result.

d7919:
Would both perhaps be useful? Minimum is of course useful for giving some indication for how the approaches compare in ideal conditions but average may be more of a reflection of real world behaviour? I guess the user could always just launch the job many times and do the averaging themselves (of the minima), but this may be overly optimistic (or am I missing something in how this relates to expected real world performance)?

ZedThree:
A useful set of stats for performance is the minimum, average and maximum values, and out of those I'd prefer to see the average. Reporting just the minimum is a bit misleading. For proper benchmarking, you should be running on a machine with nothing else running on it, so you don't need to worry about threads being suspended.

dschwoerer:
Now with min/max/avg.
Feel free to replace the C-style print statement with a C++ stream one - but I don't like the C++ style.
Now the user has the choice what she cares about.

dschwoerer:
Replaces #924 

@bendudson - feel free to push this branch, that is why I am pushing to boutproject, rather than my repo ...

Would it be worth putting the timeit macros into `example/performance/timeit.hxx` rather then maintaining several copies?

PR #843, opened 2018-02-06
small cleanup

dschwoerer:


tests/integrated/test-dataiterator/runtest
```diff
@@ -17,9 +17,6 @@ MPIRUN=getmpirun()
 
 print("Making DataIterator test")
 s, out = launch_safe("make > make.log",pipe=True)
```
d7919: Note that `launch _safe` launches an mpi job with the max number of procs (if nproc is not given). I think you want `shell_safe` here.

dschwoerer: fixed 👍

tests/integrated/test-dataiterator/runtest
```diff
@@ -6,59 +6,34 @@
...
-          #  code = 1
+            shell("rm data/BOUT.dmp.* 2> err.log")
 
-if code == 0:
-    print(" => All data iteration tests passed")
-else:
-    print(" => Some failed tests")
+            # Run the case
+            _, out = launch_safe(cmd+" "+f, runcmd=MPIRUN, nproc=nproc, pipe=True)
+            with open("run.log."+str(nproc)+"."+str(mproc), "w") as f:
```
d7919: Probably not really an issue but you're redefining `f` here which is initially the current flag value that we're iterating over.

tests/integrated/test-dataiterator/runtest
```diff
@@ -6,59 +6,34 @@
...
+        cmd = "./test_dataiterator"
 
-    print("   %d processors...." % (nproc))
-    r = 0
-    for f in flags:
-        stdout.write("\tflags '"+f+"' ... ")
+        print("   %d processors...." % (nproc))
 
-        shell("rm data/BOUT.dmp.* 2> err.log")
+        for f in flags:
```
ZedThree: Seeing as `flags` just has one element, the empty string, we don't need to iterate over it at all

tests/integrated/test-dataiterator/runtest
```diff
@@ -6,59 +6,31 @@
...
-    cmd = "./test_dataiterator"
+    for mproc in [1,2,4,8]:
+        cmd = "./test_dataiterator"
+
+        print("   %d processors...." % (nproc))
 
-    print("   %d processors...." % (nproc))
-    r = 0
-    for f in flags:
         stdout.write("\tflags '"+f+"' ... ")
```
d7919: `f` will not be defined on the first iteration and on subsequent iterations it will be a (closed) file object.

tests/integrated/test-dataiterator/runtest
```diff
@@ -6,59 +6,31 @@
...
-          print("PASSED")
-          #else:
-          #  print("FAILED")
-          #  code = 1
-
-if code == 0:
-    print(" => All data iteration tests passed")
-else:
-    print(" => Some failed tests")
+        _, out = launch_safe(cmd+" "+f, runcmd=MPIRUN, nproc=nproc, mthread=mproc pipe=True)
```
ZedThree: Missing comma after `mproc`.

I'm not sure about using `launch_safe` here, as it will immediately bail out and not run the further tests. Let's say threading is broken, but MPI works ok -- as it is, it will fail on `nproc=1, mproc=1`, and we won't see that `nproc=2, mproc=1` still works. Obviously, the whole test should fail, but it would be nice to run as many tests as possible to see if it's only a subset that's broken.

ZedThree:
Other than my comment, this is good

tests/integrated/test-dataiterator/runtest
dschwoerer: Thanks - fixed the missing `,`

Now it should runs all tests :+1:

PR #845, opened 2018-02-06
Fix prefix operator in RangeIterator header

d7919:
See some reference in discussion of #841

Note this was used in imex-bdf2, so could possibly explain some of the
odd behaviour observed with this solver (occasional segfaults).


d7919:
Actually, maybe doesn't explain the imex-bdf2 issue but shouldn't hurt.

PR #847, opened 2018-02-07
Fix missing overrides

d7919:
Just adds override in lots of places.

Squashes a very large number of warnings produced by clang.

### DIFF
--- a/src/fileio/impls/netcdf4/ncxx4.hxx
+++ b/src/fileio/impls/netcdf4/ncxx4.hxx
@@ -65,47 +65,47 @@ class Ncxx4 : public DataFormat {
   using DataFormat::openw;
   bool openw(const char *name, bool append=false) override;
   
-  bool is_valid();
+  bool is_valid() override;
   
-  void close();
+  void close() override;
   
-  void flush();
+  void flush() override;
 
   const char* filename() { return fname; };
 
-  const vector<int> getSize(const char *var);
-  const vector<int> getSize(const string &var);
+  const vector<int> getSize(const char *var) override;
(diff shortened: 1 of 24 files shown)


src/fileio/impls/netcdf4/ncxx4.hxx
```diff
@@ -65,47 +65,47 @@ class Ncxx4 : public DataFormat {
...
-  const vector<int> getSize(const string &var);
+  const vector<int> getSize(const char *var) override;
+  const vector<int> getSize(const string &var) override;
   
   // Set the origin for all subsequent calls
-  bool setGlobalOrigin(int x = 0, int y = 0, int z = 0);
-  bool setLocalOrigin(int x = 0, int y = 0, int z = 0) { return setGlobalOrigin(x,y,z); }
-  bool setRecord(int t); // negative -> latest
+  bool setGlobalOrigin(int x = 0, int y = 0, int z = 0) override;
+  bool setLocalOrigin(int x = 0, int y = 0, int z = 0) override { return setGlobalOrigin(x,y,z); }
```
dschwoerer: This does not seem to override
is eb66f0d the appropriate fix?

d7919: Ah yes, thanks for spotting that. I think the `nc_format.hxx` version is basically what's in the commit you suggest (just with `UNUSED` around the last three arguments). Would you be able to make a PR to merge this single commit? I'm happy to do this if you prefer.

This also highlights why having more diverse Travis cases might be useful to actually test the different implementations of various things.

PR #848, opened 2018-02-07
Improve build of autogenerated text files

dschwoerer:
If generated_fieldops.cxx is not succesfully generated, it needs to be
cleaned, otherwise make will on the next try not try to generate it
again, and the build might fail later in harder to debug ways.

### DIFF
--- a/src/field/makefile
+++ b/src/field/makefile
@@ -17,5 +17,5 @@ include $(BOUT_TOP)/make.config
 # to apply clang-format to this file in-place, but don't worry if the
 # formatting fails
 generated_fieldops.cxx: gen_fieldops.py gen_fieldops.jinja
-	./$< --filename $@
+	./$< --filename $@ || (rm $@ ; exit 1)
 	@clang-format -i $@ || echo "Formatting failed"
--- a/src/field/makefile
+++ b/src/field/makefile
@@ -17,5 +17,7 @@ include $(BOUT_TOP)/make.config
 # to apply clang-format to this file in-place, but don't worry if the
 # formatting fails
 generated_fieldops.cxx: gen_fieldops.py gen_fieldops.jinja
-	./$< --filename $@ || (fail=$?;rm $@ ; exit $fail)
+	@echo "  Generating $@"
+	@./$< --filename $@.tmp || (fail=$?; echo "touch $@ to ignore failed generation" ; exit $fail)
+	@mv $@.tmp $@
 	@clang-format -i $@ || echo "Formatting failed"
--- a/src/field/makefile
+++ b/src/field/makefile
@@ -17,5 +17,5 @@ include $(BOUT_TOP)/make.config
 # to apply clang-format to this file in-place, but don't worry if the
 # formatting fails
 generated_fieldops.cxx: gen_fieldops.py gen_fieldops.jinja
-	./$< --filename $@ || (rm $@ ; exit 1)
+	./$< --filename $@ || (fail=$?;rm $@ ; exit $fail)
 	@clang-format -i $@ || echo "Formatting failed"


dschwoerer:
Is there a way to fix this in travis:
```
E: Unable to locate package g++-7
E: Couldn't find any package by regex 'g++-7'
```

d7919:
I'm not sure why this intermittent issue seems to arise -- I suspect it might be related to some of the messages at the start of the build regarding timeouts etc. in trying to access various repos which seems to be a relatively recent issue with Travis. I've noticed the anaconda install sometimes fails due to what looks like network issues.

dschwoerer:
I didn't know travis would restart the job - in that case it is not that bad, and probably out of our control ...

d7919:
It doesn't automatically restart the job -- I clicked restart job in this case and have been doing this for the past week or so when I notice jobs failing because of these issues.

dschwoerer:
In this case: Thanks for restarting 👍
After logging in, I found the button to restart :)

ZedThree:
This is unfortunately a Travis problem, nothing we can do :(

src/field/makefile
```diff
@@ -17,5 +17,5 @@ include $(BOUT_TOP)/make.config
 # to apply clang-format to this file in-place, but don't worry if the
 # formatting fails
 generated_fieldops.cxx: gen_fieldops.py gen_fieldops.jinja
-	./$< --filename $@
+	./$< --filename $@ || (rm $@ ; exit 1)
```
ZedThree: What does an error look like here? Just something like "Make: Error 2"? Do any warnings/errors from Python get printed to stderr/stdout?

dschwoerer: Don't know, and I shouldn't have to care.
Maybe jinja2 not available for python3?
Maybe no space left on device?
Maybe the moon is in a bad position?
What ever the reason - `generated_fieldops.cxx` should get deleted.
IO is not redirect, just by checking the error status and conditinally deleting the output.
If you want to retain the error code - then we would need to store that - but I am not sure make prints that, or returns that, so 1 should be fine.

d7919: Should we make the first statement in this target be `rm $@` and then try to generate? If the generation fails then make will pick this up automatically as it does for any other build failure and we don't need to worry about handling things ourself?

dschwoerer: That doesn't work, if the file gets partially created.
It still will be there - and things can go wrong later on.
This will only work, if the generation fails before the file is opened for writing - which is rather early in the creation process.
Not sure whether this is possible in python, but maybe trap all commands in python, such that you can delete the file before python exits ... but that sounds far more difficult to me ...

d7919: True, but make aborts so it should be clear something has gone wrong. I take the point that a user could then just type `make` again and they'd probably get a different error later on.

I guess the issue is that removing the generated file puts a hard dependency on jinja2 etc. whilst at the moment we only require jinja if we really want to regenerate (i.e. we've changed the template or similar).

d7919: Perhaps horrible but I guess we could revert to checked in version in the case where jinja2 is not available.

dschwoerer: make will only try to rebuild the `.cxx` if any of the dependencies are newer.
I think that should only happen, if we make a mistake and ship a non updated `cxx` - or the user changed the template.

I often try first make in parallel, and then run make again, to see the error less scrambled.
If I now get a different error, because the generated file does not get deleted, I am having troubles ...

Further, I think, the user shouldn't have, after installing dependencies, remove the failed output to trigger a successful rebuild.

If you are worried about the wrong exit code, then I will fix that ...

d7919: I was thinking along the lines of a curious person poking around and they end up effectively touching the template in some way. What about when they checkout, can we guarantee the files will be updated in and will this mean that sometimes we try to regenerate even if the files are in sync? 

I think some sort of fallback when the failure is due to not having dependencies would be a user friendly option. 

How about we keep the checked in version, we try to generate and store this is a temporary file. If successful we rename the temporary to take the place of the final file. If it fails we remove the temporary (or leave it in place if we want to see what's going on?) and give instructions for what to do next (i.e. how to proceed with the checked in version or where to go to fix the issue)?

dschwoerer:
Now with:
* Failing repeatedly with the same error (no hiding of errors).
 * Instruction on how to avoid failing generation as fallback.
 * Old version is kept if unsuccessful.

PR #849, opened 2018-02-07
Prevent further divergence: merge master into next

dschwoerer:
Could we get these bug fixes in next?


dschwoerer:
Replaced by #930

PR #851, opened 2018-02-07
Localmesh fixes

dschwoerer:


src/field/field_factory.cxx
```diff
@@ -97,7 +97,13 @@ FieldFactory::~FieldFactory() {
 }
 
 const Field2D FieldFactory::create2D(const string &value, Options *opt, Mesh *m, CELL_LOC loc, BoutReal t) {
-  Field2D result = 0.;
+
+  if(m == NULL)
```
ZedThree: It looks like this was copied from L113, but the original code is still in place -- delete L113-116?

Also, I know it's nitpicking, but this should be:

```cpp
if (m == nullptr) {
  m = fieldmesh;
}
```

src/field/field_factory.cxx
```diff
@@ -97,7 +97,13 @@ FieldFactory::~FieldFactory() {
...
+    m = fieldmesh;
+  if(m == NULL)
+    throw BoutException("Not a valid mesh");
+
+  Field2D result(0.,m);
 
   if(mesh->StaggerGrids == false){
```
ZedThree: We should probably be consistent and use the passed in `Mesh` everywhere in this function, rather than the global `mesh`

src/field/field_factory.cxx
```diff
@@ -97,7 +97,13 @@ FieldFactory::~FieldFactory() {
 }
 
 const Field2D FieldFactory::create2D(const string &value, Options *opt, Mesh *m, CELL_LOC loc, BoutReal t) {
```
ZedThree: While you're fixing this, can we also change `m` to something like `localmesh`?

Ideally, we'd rename the global `mesh` to `global_mesh`, and then we could just use plain `mesh` everywhere

src/field/field_factory.cxx
dschwoerer: No, copied from the Field3D function - didn't notice that the check was not missing but just a few lines down.

PR #641, opened 2017-08-09
Cython

dschwoerer:
I now added some documentation ...

It still needs some cleaning ...

I have changed a few more things, so that MMS-ing works (i.e. multi mesh). For example fieldops have been merged.

PR #852, opened 2018-02-07
Use python3 as python interpreter

dschwoerer:


tests/MMS/elm-pb/makegrids.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
d7919: Isn't this one fixed in #843?

dschwoerer: I tried to remove differences between #641 and next - while #843 came from #742 
I will close this PR for now ...

PR #861, opened 2018-02-14
Pylib dimensions

dschwoerer:
Needed to fix a bug in the #641

https://github.com/dschwoerer/BOUT-dev/issues/5


dschwoerer:
It adds the function `dimensions(varname, path=".", prefix="BOUT.dmp")`

In order to do it without repetetion, I moved the search for files from `collect` to its own function, which is used by `collect`, `dimensions` and `attributes`.

It's not really a big change, it feels more like diff is not doing a great job as I moved the stuff from `collect` before `collect` ...

I will move the function `findFiles` to the end, as it is intended rather as an internal function ...

Issue #865, opened 2018-02-14
LaplaceXZ Petsc inner boundary


bendudson:
The inner X boundary for the PETSc implementation of LaplaceXZ defaults to neumann, rather than dirichlet. 

Described in comments here:
https://github.com/boutproject/BOUT-dev/blob/master/src/invert/laplacexz/impls/petsc/laplacexz-petsc.cxx#L85

and implemented here:
https://github.com/boutproject/BOUT-dev/blob/next/src/invert/laplacexz/impls/petsc/laplacexz-petsc.cxx#L340

To be consistent with the other solvers, a boundary flag of 0 should be Dirichlet.

ZedThree:
Fixed in #868

PR #868, opened 2018-02-14
Change LaplaceXZ-PETSc to Dirichlet boundary conditions by default.

bshanahan:
This fixes issue [#865](https://github.com/boutproject/BOUT-dev/issues/865).


d7919:
The failing push test is due to a broken commit to next -- if you update your branch the test should pass, but I'm not sure this is necessary given that we know why the test fails.

PR #872, opened 2018-02-15
Fix two potential bugs: out-of-bounds access and use-after-scope

ZedThree:
AddressSanitizer spotted the following two bugs:

1. When creating the contiguous blocks for `Region`s, if the last block doesn't have exactly `MAXREGIONBLOCKSIZE` points in it, `index` will be past the end of `indices`

2. When cleaning the `Array` store, we should use `a.reset()` instead of `a = nullptr`

(I'm not sure I completely understand 2, but ASan complains about it)

### DIFF
--- a/include/bout/array.hxx
+++ b/include/bout/array.hxx
@@ -370,17 +370,17 @@ private:
     BOUT_OMP(single)
     {
       for (auto &stores : arena) {
-	for (auto &p : stores) {
-	  auto &v = p.second;
-	  for (dataPtrType a : v) {
-	    a = nullptr; //Could use a.reset() if clearer
-	  }
-	  v.clear();
-	}
-	stores.clear();
+        for (auto &p : stores) {
+          auto &v = p.second;
+          for (dataPtrType a : v) {
+            a.reset();
+          }
+          v.clear();
+        }
+        stores.clear();
       }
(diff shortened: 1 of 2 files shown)


d7919:
I'm surprised by 1 as I tested exactly that case! It's possible this occurred in the change from a "mesh end" to a "C++ end". Do you have a unit test that would catch this?

ZedThree:
I had a quick stab at a unit test, but because it's going past the end of `indices`, I struggled to get something. Compiling with ASan, or changing the `break` to a `throw` catches it immediately,

d7919:
It certainly looks like it is needed, just can't understand why my test didn't work. Presumably an alternative, to avoid the branch in the loop would be to replace

~~~~
// Consider if the next point should be added to this block
      for (index++; count < MAXREGIONBLOCKSIZE; index++) {
        if (index >= npoints) {
          break;
        }
~~~~

with

~~~~
// Consider if the next point should be added to this block
int blockMax = min (npoints-index, MAXREGIONBLOCKSIZE);
      for (index++; count < blockMax; index++) {
~~~~

d7919:
I don't understand the asan warning here but I'm happy to go with reset. There are other places in array which use the nullptr approach (I think), which I think is why I didn't use reset in this instance.

PR #875, opened 2018-02-16
Extend Region class

d7919:
Adds functionality for the Region class (masking, adding, sorting, making unique, offsetting) along with unit tests.

Adds `Ind3D` and `Ind2D` indexing routines to `Field2D`.

Makes Region look like an iterator so we can have range based loops with a Region.

Extend performance/iterator test to include BLOCK_REGION_LOOP and OMP versions of a couple of other loops.

<s>Use BLOCK_REGION_LOOP in templated fieldops code. Also fixes a couple of potential future bugs in this templating code (assumed variable names replaced with templated values).</s>


d7919:
Now makes Region look like an iterator so we can have range based loops with a Region. Thanks to @ZedThree

d7919:
@ZedThree is going to look at what documentation is required. After this I think the "in progress" label can be removed.

ZedThree:
Added documentation, some more tests, parameterised some tests over `Ind2D`, `Ind3D`

When we start using this, we'll want to revisit `manual/sphinx/user_docs/iterating.rst`

include/bout/region.hxx
```diff
@@ -125,14 +125,66 @@ public:
...
+}
+inline bool operator>=(const SpecificInd &lhs, const SpecificInd &rhs) {
+  return !operator<(lhs, rhs);
+}
+inline bool operator<=(const SpecificInd &lhs, const SpecificInd &rhs) {
+  return !operator>(lhs, rhs);
+}
+
+/// Arithmetic operators with integers
+inline SpecificInd operator+(SpecificInd lhs, const SpecificInd &rhs) { return lhs += rhs; }
```
d7919: Won't this cause problems in uses like the below?

~~~~
Ind3D lhs(4);
Ind3D rhs(5);
auto newLhs = lhs+rhs;
output<<newLhs.ind<<" and "<<lhs.ind<<" should be 9 and 4, but may be 9 and 9 now?\n";
~~~~

ZedThree: It takes `lhs` as a copy, and returns the copy. There's a [test][1] explicitly for this :)

[1]: https://github.com/boutproject/BOUT-dev/pull/875/files/896f7d8fc26d48884902d25df444154c09e205de#diff-749772ea70cb4dc76e059e4671c05235R1303

d7919: Ah yes, didn't spot the absence of &

include/bout/region.hxx
```diff
@@ -22,36 +22,74 @@
...
+///
+/// can be converted to a block region loop like so:
+///
+///     BLOCK_REGION_LOOP(region, index,
+///        A[index] = B[index] + C[index];
+///     )
 #define BLOCK_REGION_LOOP_SERIAL(region, index, ...)                                     \
   {                                                                                      \
     const auto blocks = region.getBlocks();                                              \
     for (auto block = blocks.begin(); block < blocks.end(); ++block) {                   \
```
bendudson: Should this be `!=` rather than `<`? I'm not sure if there is any performance difference (maybe one instruction), but != seems to be standard. I guess it has fewer requirements on the comparison operator.

include/bout/region.hxx
```diff
@@ -62,55 +100,160 @@
...
+/// be parallelised with OpenMP. Iterating using a "block region" may
+/// be more efficient, although it requires a bit more set up. The
+/// helper macro BLOCK_REGION_LOOP is provided to simplify things.
+///
+/// Example
+/// -------
+///
+/// Index ranges can be defined manually:
+///
+///     Region<Ind3D>::RegionIndices region {0, 2, 4, 8};
```
bendudson: Does this create a region with only these four indices, or one with two ranges: 0,1,2  and 4 5 6 7 8 ? (or maybe 0,1 and 4,5,6,7?)

include/bout/region.hxx
d7919: The comments on https://stackoverflow.com/a/17853547 suggests `!=` may not work. Certainly all examples I found used `<`, should be easy enough to test though!.

include/bout/region.hxx
bendudson: Ah ok. Maybe just a note in the comments then; the "Index ranges" part of the above comment had me a bit confused. Thanks!

d7919: I've exapanded this comment now, so it's hopefully a bit clearer.

Issue #876, opened 2018-02-16
DataIterator with OpenMP


d7919:
I just came across an issue with the OpenMP version of the iterator. I had a loop of the form

~~~~
Field3D x;
x.allocate()
BOUT_OMP(parallel for)
for(int jy=ys; jy<=ye ; jy++){
    FieldPerp ss = sliceXZ(b,jy);
    x = solve(ss);
}
~~~~

The `Field3D::operator=(FieldPerp rhs)` has a loop like

~~~~
for (const &i: rhs){
     (*this)[i] = rhs[i];
}
~~~~

The dataiterator for the range based loop tries to parallelise itself, assuming that all threads will take part in this loop, which isn't the case here as each thread must do the full XZ assignment for the thread specific `jy` and hence means `x` will only get ~`ny/nthread` of the data being set.

I think this is one of the reasons why trying to OpenMP parallelise the loop at https://github.com/boutproject/BOUT-dev/blob/master/src/invert/laplace/invert_laplace.cxx#L199 has caused things to fail in previous attempts. I now have a version of this loop working by forcing the use of the serial DI.

dschwoerer:
I think only the treads of the inner most parallel region are assumed to be taking part.
I wasn't aware of any way to determine whether the all inner most threads are takeing part - or not.

I think putting
```
BOUT_OMP(parallel)
for (const &i: rhs){
     (*this)[i] = rhs[i];
}
```
should basically serialize it:
https://github.com/boutproject/BOUT-dev/blob/master/include/bout/dataiterator.hxx#L339
https://gcc.gnu.org/onlinedocs/gcc-4.5.4/libgomp/omp_005fget_005fnum_005fthreads.html

d7919:
Unfortunately that doesn't seem to help here, but it could be that a different issue arises, will need to dig further.

There's a branch that I'm using that should reference this issue.

d7919:
I think I may have just needed to put `BOUT_OMP(parallel)` on more loops. Adding parallel to all of the assignment loops does indeed help avoid this issue, thanks! I'm not sure that there is much else we can do without leaving the splitting of work to OpenMP, but it does mean that at the moment we have to either OpenMP "everything" or nothing.

d7919:
[...]

PR #877, opened 2018-02-16
Has openmp

dschwoerer:
Trying to reproduce #876 - without success.


ZedThree:
This is really two separate features, it would be better in two PRs.

Also, the test seems like it would be better suited to a unit test -- could it be rewritten as one?

dschwoerer:
Note quite sure what the two prs would be ... the test and the changes to configure?

I needed configure changes to run the test conditionally - so the one pr would depend on the other ...

ZedThree:
If the test were a unit test, it could just be surrounded by `#ifdef OPENMP_` which would do this automatically.

dschwoerer:
I force pushed without the last 2 commits :+1:

configure
```diff
@@ -1431,6 +1420,7 @@ Optional Features:
   --enable-sigfpe         Enable FloatingPointExceptions
   --disable-backtrace     Disable function backtrace
   --enable-shared         Enable building bout++ into an shared object
+  --enable-openmp         Enable building with OpenMP support
```
d7919: There's also an entry for disable-openmp a few lines below. I think we should probably just have one of these.

dschwoerer: as the default is to disable it - the one below should be removed.
However, I am not sure why it is there.

d7919: My impression is that this is a "builtin" with autoconf (e.g. see [here](https://www.spinics.net/lists/ac/msg06330.html) and [here](https://www.gnu.org/software/autoconf/manual/autoconf-2.67/html_node/Generic-Compiler-Characteristics.html)) but I can't find any good documentation making this clear.

dschwoerer: I guess that cannot be disabled (from /usr/share/autoconf/autoconf/c.m4):
```
AC_DEFUN([AC_OPENMP],
[
  OPENMP_[]_AC_LANG_PREFIX[]FLAGS=
  AC_ARG_ENABLE([openmp],
    [AS_HELP_STRING([--disable-openmp], [do not use OpenMP])])
  if test "$enable_openmp" != no; then
    AC_CACHE_CHECK([for $[]_AC_CC[] option to support OpenMP],
      [ac_cv_prog_[]_AC_LANG_ABBREV[]_openmp],
```

Should this be reported to `autoconf`: https://www.gnu.org/software/autoconf/autoconf.html#bug ?

ZedThree: That's not a particularly great choice from the autoconf maintainers there.

Easiest thing to do is to copy the `AC_DEFUN` from that file into `m4/bout.m4`, rename it `BOUT_OPENMP`, change the `AC_ARG_ENABLE` to be what we actually want, then use our new macro instead of `AC_OPENMP`.

PR #878, opened 2018-02-19
Better matrices/tensors

d7919:
Use of multi-dimensional pointer arrays with helper functions `matrix`/`free_matrix` and `*3tensor`/`free_*3tensor` in various places in the code. These are usually not ideal for a number of reasons and are often created in some form of persistent manner (either static or persistent class level allocation) to avoid repeated cost of  allocate/deallocate. 

This PR introduces  `Matrix` and `Tensor` templated containers that provide logically 2d/3d access to flat `Array` backed data. As `Array` has a number of nice features `Matrix` and `Tensor` also have a number of nice features, such as being cheap to create, some memory management, attempts to be thread safe etc. These features should allow declaration at a much more local scope (i.e. not at class level) without resorting to static and will hopefully help make more routines compatible with OpenMP.

Also deprecates existing `*3tensor` routines as all uses removed from framework (replaced with `Tensor`).

Doesn't yet remove all uses of the pointer based matrix as these occur in a large number of places and I didn't want to touch too much at once.,

Two `Matrix` methods are also marked as deprecated, `operator[]` and the const version as these should not be used in the long term, but provide backwards compatibility with common usage of the pointer based version. By marking as deprecated it is clear where we still use the old form even once the variable type has changed (from `T**` to `Matrix<T>`).

I've not changed any function signatures (except `gaussj`) -- places that expect a (single/double/triple) pointer to a block of memory still get this. To aid this I've exposed the `std::begin` of the `Array` being used for data storage. In the long run I would prefer to change the signatures to expect `Array`/`Matrix`/`Tensor` arguments.

### DIFF
--- a/include/bout/coordinates.hxx
+++ b/include/bout/coordinates.hxx
@@ -135,7 +135,7 @@ public:
   
 private:
   /// Matrix inversion by Gauss-Jordan elimination
-  int gaussj(Matrix<BoutReal> a, int n);
+  int gaussj(Matrix<BoutReal> &a, int n);
   vector<int> indxc, indxr, ipiv;
   int nz; // Size of mesh in Z. This is mesh->ngz-1
   Mesh * localmesh;
--- a/include/bout/coordinates.hxx
+++ b/include/bout/coordinates.hxx
@@ -37,6 +37,7 @@ class Coordinates;
 
 #include "mesh.hxx"
 #include "datafile.hxx"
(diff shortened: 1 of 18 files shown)


d7919:
Changes to the `gaussj` interface should have made sure Matrix passed in by reference but was passing a copy. This is not what is wanted as we want to modify the Matrix in place, @bendudson this should now be addressed.

PR #880, opened 2018-02-20
Report if Lapack is available in this build

d7919:
Can be helpful to spot when lapack is not available.


d7919:
I'd be tempted to make `--with-lapack` default to `yes` as well.

PR #881, opened 2018-02-20
Exclude guard cells when checking finite values in field functions

johnomotani:
Functions in field3d.cxx and field2d.cxx like sqrt(), etc., check whether their result is finite, if `CHECK>3`. This caused me a problem with `abs(Grad_perp(phi))` because some guard cells were NaN, but the `sqrt()` inside `abs()` checked for finite values and threw an exception.

This PR fixes this issue, but I'm not sure it's the most elegant way...
I thought about adding a new ASSERT macro that would take the index object and check if it is in a guard cell, but I think we need the Field*D to get the right mesh to get xstart, etc. from.
I guess passing the mesh as an extra argument would be one possibility, then we could have something like
```
#if CHECKLEVEL >= 3
#define ASSERTWITHINDEX3(condition,d,thismesh)     \
  if(!(condition) && !(d.x<thismesh->xstart) && !(d.x>thismesh->xend) && !(d.y<thismesh->ystart) && !(d.y>thismesh->yend)) {           \
    throw BoutException("Assertion failed in %s, line %d: %s. Failed at position (%d,%d,%d)", __FILE__, __LINE__, #condition, d.x, d.y, d.z);  \
  }
#else // CHECKLEVEL >= 3
#define ASSERT3(condition)
#endif
```

### DIFF
--- a/src/field/field2d.cxx
+++ b/src/field/field2d.cxx
@@ -458,7 +458,8 @@ bool finite(const Field2D &f) {
  * If CHECK >= 1, checks if the Field2D is allocated
  *
  * Loops over the entire domain, applies function,
- * and if CHECK >= 3 then checks result for non-finite numbers
+ * and uses checkData() to, if CHECK >= 3, check
+ * result for non-finite numbers
  *
  */
 #define F2D_FUNC(name, func)                                                             \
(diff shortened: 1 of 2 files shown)


d7919:
The appearance of NaNs suggests it could be related to #871 which would in turn suggest the fields that are being passed in have not been communicated/had boundary conditions applied.

I think the best ways to avoid this are 

1. Make sure these functions take in a `REGION` that can prevent them from operating in the boundaries.
2. Replace the `ASSERT3` inside the loop with a `checkData(result)` call after the loop (which by default excludes the guards/boundaries).

bendudson:
Perhaps to be explicit in the code, in cases where we know the boundaries are not valid (like this, following a differential operator), we could have a different function? checkDataNoGuards?

johnomotani:
Thanks David! Using checkData() seems the simplest solution. I've implemented it now.
Adding REGION arguments to the functions seems like it could be a useful thing to do, I'll have a go and make separate PR for it.

d7919:
No problem. There may be some overlap with #839.

Issue #132, opened 2015-12-15
High condition number in Mesh::gaussj


loeiten:
The ```int Mesh::gaussj(BoutReal **a, int n)``` in [mesh.cxx](https://github.com/boutproject/BOUT-dev/blob/master/src/mesh/mesh.cxx) may have a high condition number, which scales badly with increasing number of grid points. It is used to find the contravariant components of the metric tensor (if no covariant metric tensor is found) given the covariant components, and vice versa, 

The problem becomes big if one wants to use the Christoffel symbols, which are using the derivatives of the metric tensor components.

The enhance request is therefore to change the direct solver into something else which is not so sensitive to the condition number.

d7919:
This is a Gauss-Jordan elimination implementation which inverts the metric tensor. Presumably it would be easy enough to write something that uses lapack to do this inversion (dgetrf and dgetri I think) for each mesh point. The GJ implementation uses pivoting, which should help avoid some issues; if there's a large variation in the output for small changes to the input then there may be a region where some of the metric elements are very small/large which is causing this. It's not clear that an alternative approach would improve on this in all circumstances but it's probably worth a quick study.

Have you got a test case that demonstrates the issue?

For reference there's a nice simple example of inverting with lapack at http://stackoverflow.com/questions/3519959/computing-the-inverse-of-a-matrix-using-lapack-in-c

loeiten:
Hmm...indeed. I don't have too much experience with solutions to these kind of problems, I just remembered vaguely something about changing to an iterative solver could reduce the problem of the condition number (but I might be wrong).

I have something which reproduces the problem, but not a dedicated test case. I stumbled upon the problem when trying to MMS the Christoffel symbols in the [vector_operators branch](https://github.com/boutproject/BOUT-dev/tree/vector_operators).

With the following parameters in the ```BOUT.inp``` file

```
[mesh]
(...)
# ------------------------------
# Contravariant metric components
# ------------------------------
g11 = 2 
g12 = 4*geom:xl
g13 = 4*geom:xl - 2 
g22 = 8*geom:xl^2 + 1 
g23 = 8*geom:xl^2 - 4*geom:xl + 1 
g33 = 8*geom:xl^2 - 8*geom:xl + 5 
# ------------------------------
# Covariant metric components
# ------------------------------
# g_11 = 4*geom:xl^2 + 1 
# g_12 = -2*geom:xl - 1/2 
# g_13 = 1/2 
# g_22 = 3/2 
[...]

PR #883, opened 2018-02-21
Replace gaussj with direct invert 3x3

d7919:
Replace gaussj with direct invert of a 3x3 `Matrix` object. 

Templated so can be used for different numeric types.

Takes an optional parameter, `small` that is used to decide if the matrix is too poorly conditioned. If `small` is `<0` then we just return 1 when poorly conditioned, otherwise we throw.

Comes with basic unit tests.


d7919:
#132 references a (now deleted) branch that implemented a similar approach.

PR #884, opened 2018-02-21
Remove lapack replacement code

bendudson:
Take out code from NR, which was there as a fallback if lapack was not used. The main place these routines are used is in the Lapacian serial_tri solver so:

- The serial_tri and band solvers will throw an exception if used and Lapack has not been compiled in
- The default Laplacian solver is now "cyclic" rather than "serial_tri"

### DIFF
--- a/manual/sphinx/user_docs/advanced_install.rst
+++ b/manual/sphinx/user_docs/advanced_install.rst
@@ -204,17 +204,17 @@ To configure BOUT++ with PETSc and SUNDIALS, type instead::
     $ ./configure --with-petsc --with-sundials
 
 
+.. _sec-lapack:
+    
 LAPACK
 ------
 
 BOUT++ comes with linear solvers for tridiagonal and band-diagonal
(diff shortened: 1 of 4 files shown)


ZedThree:
Does this put a hard-dependency on LAPACK for some features? Or is it just for some implementations of those features?

The documentation probably needs updating to reflect this (at least [here][1], [source][2]). I'm not sure we have a comprehensive list of what implementations of different features are available.

[1]: http://bout-dev.readthedocs.io/en/latest/user_docs/advanced_install.html#lapack
[2]: https://github.com/boutproject/BOUT-dev/blob/next/manual/sphinx/user_docs/advanced_install.rst

bendudson:
No hard dependency, I don't think. The cyclic solver is always available, and works in serial or parallel. It restricts the choice of solver. I'll update the docs

d7919:
Possibly worth noting #895 before this is merged.

PR #886, opened 2018-02-21
Add OpenMP dataiterator test to unit tests

dschwoerer:
Is this correct?
Should I make sure the test is run in parallel with an acceptable number of processors?

### DIFF
--- a/tests/unit/field/test_field3d.cxx
+++ b/tests/unit/field/test_field3d.cxx
@@ -1239,3 +1239,36 @@ TEST_F(Field3DTest, DC) {
 
   EXPECT_TRUE(IsField2DEqualBoutReal(DC(field), 3.0));
 }
+
+#ifdef _OPENMP
+TEST_F(Field3DTest, OpenMPIterator) {
+  const int fields=10;
+  Field3D * d3= new Field3D[fields];
+  for (int i=0;i<fields;++i){
+    d3[i]=1;
+  }
+
+  BoutReal exp=1;
+
+  BOUT_OMP(parallel for)
+  for (int j=0;j<fields;++j){
+    BOUT_OMP(parallel)
+    for (auto i: d3[j]){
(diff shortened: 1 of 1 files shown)


tests/unit/field/test_field3d.cxx
```diff
@@ -1239,3 +1239,36 @@ TEST_F(Field3DTest, DC) {
...
+    d3[i]=1;
+  }
+
+  BoutReal exp=1;
+
+  BOUT_OMP(parallel for)
+  for (int j=0;j<fields;++j){
+    BOUT_OMP(parallel)
+    for (auto i: d3[j]){
+      EXPECT_TRUE(d3[j][i] == exp);
```
ZedThree: You should be able to use `EXPECT_DOUBLE_EQ(d3[j][i], exp)` here, which compares floats with some tolerance

tests/unit/field/test_field3d.cxx
```diff
@@ -1239,3 +1239,36 @@ TEST_F(Field3DTest, DC) {
...
+
+#ifdef _OPENMP
+TEST_F(Field3DTest, OpenMPIterator) {
+  const int fields=10;
+  Field3D * d3= new Field3D[fields];
+  for (int i=0;i<fields;++i){
+    d3[i]=1;
+  }
+
+  BoutReal exp=1;
```
ZedThree: `exp` -> `expected`

Non-abbreviated variable names are good :)

ZedThree:
It'll get run with a couple of threads in the OpenMP build, that's probably sufficient for now. When we've got more OpenMP-specific tests in there, we should run the unit tests at least with a different numbers of threads.

Not required, but if you've got clang-format installed, could you run it over this test?

dschwoerer:
I should have addressed all 3 changes.
Thanks for the feedback :+1:

PR #887, opened 2018-02-21
Fix some minor warnings

ZedThree:
Fixes most of the warnings on the Travis gcc-7 build

### DIFF
--- a/src/fileio/impls/netcdf4/ncxx4.cxx
+++ b/src/fileio/impls/netcdf4/ncxx4.cxx
@@ -179,9 +179,9 @@ bool Ncxx4::openw(const char *name, bool append) {
     }
 
     /// Test they're the right size (and t is unlimited)
-    
-    if((xDim.getSize() != mesh->LocalNx) || (yDim.getSize() != mesh->LocalNy) || (zDim.getSize() != mesh->LocalNz)
-       || (!tDim.isUnlimited()) ) {
+    if ((xDim.getSize() != static_cast<size_t>(mesh->LocalNx)) ||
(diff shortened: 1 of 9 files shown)


src/fileio/impls/netcdf4/ncxx4.cxx
```diff
@@ -179,9 +179,9 @@ bool Ncxx4::openw(const char *name, bool append) {
     }
 
     /// Test they're the right size (and t is unlimited)
-    
-    if((xDim.getSize() != mesh->LocalNx) || (yDim.getSize() != mesh->LocalNy) || (zDim.getSize() != mesh->LocalNz)
-       || (!tDim.isUnlimited()) ) {
+    if ((xDim.getSize() != static_cast<size_t>(mesh->LocalNx)) ||
+        (yDim.getSize() != static_cast<size_t>(mesh->LocalNy)) ||
+        (zDim.getSize() != static_cast<size_t>(mesh->LocalNz)) || (!tDim.isUnlimited())) {
```
dschwoerer: shouldn't `mesh->LocalN?` be unsigned?
The static_cast only hides the warning - but does not fix the underling bug (potential overflow)

ZedThree: Maybe in the next major release. We'd have to change a lot more than just those variables.

It might be better to cast the `?Dim.getSize()` values to `int`s, but I don't think this is currently a problem, unless your mesh has a very large number of points.

dschwoerer:
It might not be perfect, but I think it is overall still an improvement :+1:

PR #888, opened 2018-02-21
Move all headers to bout/ subfolder

dschwoerer:
It was rather arbitrary (at least for me) where a specific header was - whether in bout or in the root folder.

Further this could cause conflicts with other packages.

This should avoid this.

Note: Merge conflicts are a pain with that many changes.

A way around that is, to reapply the patch on your branch before merging:
```
git cherry-pick 53589c
(cd include ; sh gen_compat_headerfiles.sh)
bin/bout_fix_headerlocation.sh include/ src/ examples/ tests/
git add include/bout/
git commit -a -m 'Move all headers to bout/ subfolder'
```


bendudson:
Agree this would be very nice. Changing header locations shouldn't be a big deal, but it has caused problems in the past: people told me they hadn't upgraded BOUT++ because the header broke and it didn't compile. Definitely a tidy up for v5.0

ZedThree:
Revisit for 5.0

dschwoerer:
It is a big change - but that shouldn't break anything.
Additionally to prividing the script[[0]](https://github.com/boutproject/BOUT-dev/pull/888/commits/53589c766f5eb0549979ee3dfb4e96233779d6cd#diff-5911cd6c13a3f4695158d2033056ef8e) there are forwards from the old location to the new one: [[1]](https://github.com/boutproject/BOUT-dev/pull/888/files#diff-5acc0a9cae59d77aa49606b02aea257eR1) - generated by [[2]](https://github.com/boutproject/BOUT-dev/pull/888/commits/53589c766f5eb0549979ee3dfb4e96233779d6cd#diff-a6e94d950da9b67416ebc7f073f625f5)

(Just a note: `#pragma once` is perfectly fine here, as including the file several times does not cause harm - except of more warnings)

PR #896, opened 2018-02-22
Add ASSERT3 checks to Array indexing to check bounds

d7919:
At top level of checking we will now check all accesses to `Array` data blocks (which account for a large 
 and increasing proportion of all our data blocks).

### DIFF
--- a/include/bout/array.hxx
+++ b/include/bout/array.hxx
@@ -37,6 +37,7 @@
 #include <valarray>
 #endif
 
+#include <bout/assert.hxx>
 #include <bout/openmpwrap.hxx>
 
 /*!
@@ -271,6 +272,7 @@ public:
    * so the user should perform checks.
    */
   T& operator[](int ind) {
+    ASSERT3(0 <= ind && ind < size());
 #ifdef BOUT_ARRAY_WITH_VALARRAY
     return ptr->operator[](ind);
 #else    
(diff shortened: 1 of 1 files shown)


ZedThree:
This will result in a bounds-check being performed twice for some operations. I note that this is turned on at the same `CHECK` level as the bounds-checking in `Field3D::operator()` for instance. While this won't be a problem for production use, we should consider the impact and other possibilities first.

For example, we could require a higher check level or an orthogonal flag to enable these checks, or we could add a separate bounds-checking method on array.

These checks are only really needed for bare-`Array` use. How much of that do we expect in user-code? If `Matrix` and `Tensor` wrap up 2D and 3D usage, should we also wrap up 1D usage and enable checking in that wrapper class instead?

d7919:
I'd be happy to enable it at a stricter checking level, we just don't support one at the moment (need changes to configure as well as introducing the new ASSERT definition).

I could expect Array to be used in some user code that's doing something more sophisticated. I also anticipate it will be used to replace many instances of `new` in the framework, for which checking would be beneficial.

My view was that sometimes you want to be very sure there are no issues so this is a "belt and braces" approach. For example if a bug meant that the nx/ny/nz in a Field object was modified the bounds checking on field may not work, but the checking on Array may still help catch this etc.

ZedThree:
Looking at the tests, it looks like the practical (performance) impact is negligible, and I was just being overly cautious!

Maybe @bendudson wants to weigh in? Otherwise, I think this is fine

Issue #864, opened 2018-02-14
inc_xbndry ignored in DDZ if method is not FFT


d7919:
The input `inc_xbndry` has no effect in `indexDDZ` if not method is not FFT.

I'm aware there's been some work on this in some open PRs but I thought it useful to explicitly point out that there is a bug that can be fixed with these PRs or other similar. 

I'm tempted to push a branch which just fixes this issue to allow people to test their code in isolation from the more substantial changes in the relevant PRs (I think #815 is one).

dschwoerer:
I guess you mean *if method is not fft` ...

dschwoerer:
Yes, is fixed in #815:
https://github.com/boutproject/BOUT-dev/pull/815/files#diff-9ed7a9439d954291d98f479aa56e8e99L1709

d7919:
Excellent.

If anyone wants to test this change in isolation there's a branch at https://github.com/boutproject/BOUT-dev/tree/dont_ignore_inc_xbndry_in_ddz

PR #898, opened 2018-02-23
Fix D2DXDZ

johnomotani:
In indexDDZ, previously the ``inc_xbndry`` option was only implemented for the FFT derivative. This meant that ``D2DXDZ`` would give wrong results at x boundaries (found due to exception thrown by invalidated guard cells) when using non-FFT schemes for z-derivatives. This PR implements the ``inc_xbndry`` option to avoid this issue.

An alternative solution, more changes but possibly neater: implement region arguments for more of the derivative operators like ``DDZ``, etc. Then ``D2DXDZ`` could call ``DDZ`` with a ``RGN_NOY`` argument instead of needing an explicit ``bool inc_xbndry`` option. Any votes for doing this instead? If it sounds like a good idea I can do it.

### DIFF
--- a/src/mesh/index_derivs.cxx
+++ b/src/mesh/index_derivs.cxx
@@ -1402,7 +1402,10 @@ const Field3D Mesh::indexDDZ(const Field3D &f, CELL_LOC outloc, DIFF_METHOD meth
 
   } else {
     // All other (non-FFT) functions
-    result = applyZdiff(f, func);
+    if (inc_xbndry)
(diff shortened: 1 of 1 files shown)


d7919:
There's an issue for this #864 a sample fix in https://github.com/boutproject/BOUT-dev/tree/dont_ignore_inc_xbndry_in_ddz and another fix in PR #815 !

dschwoerer:
seems like 30 pull requests is to much ...

You can also help by trying the pull request (merge them for testing) and giving feedback ...

PR #900, opened 2018-02-23
Replace matrix with Matrix in various cyclic inversions

d7919:
Helps enable bounds checking for these various inversion classes built on `cyclic_reduction.hxx`. 

Typically we use `new` to create a new cyclic_reduction instance. Now all of the allocation is associated with Matrix or Array so should be fast. We may be able to avoid using `new` and instead just create the solver at the scope in which it is needed. Similar could be done with the storage arrays used in these solvers which are typically currently created in the constructor.

### DIFF
--- a/include/cyclic_reduction.hxx
+++ b/include/cyclic_reduction.hxx
@@ -46,6 +46,8 @@
 #include "utils.hxx"
 #include "msg_stack.hxx"
 #include <lapack_routines.hxx>
+
+#include "bout/assert.hxx"
 #include "boutexception.hxx"
 
 #include "output.hxx"
@@ -116,6 +118,28 @@ public:
 
   }
 
+  void setCoefs(Array<T> &a, Array<T> &b, Array<T> &c) {
+    ASSERT2(a.size() == b.size());
+    ASSERT2(a.size() == c.size());
+    ASSERT2(a.size() == N);
+
+    Matrix<T> aMatrix(1, N);
+    Matrix<T> bMatrix(1, N);
+    Matrix<T> cMatrix(1, N);
+
+    //Copy data into matrices
+    for(int i = 0; i < N; ++i){
+      aMatrix(0, i) = a[i];
+      bMatrix(0, i) = b[i];
+      cMatrix(0, i) = c[i];
+    }
+
+    setCoefs(aMatrix, bMatrix, cMatrix);
+    // Don't copy ?Matrix back into ? as setCoefs
+    // doesn't modify these. Could copy out if we really wanted.
+
+  }
+
   /// Set the entries in the matrix to be inverted
   ///
   /// @param[in] a   Left diagonal. Should have size [nsys][N]
@@ -173,10 +197,45 @@ public:
 
   /// Solve a set of tridiagonal systems
   /// 
+  /// @param[in] rhs Array storing Values of the rhs for a single system
+  /// @param[out] x  Array storing the result for a single system
+  void solve(Array<T> &rhs, Array<T> &x) {
+    ASSERT2(rhs.size() == x.size());
+    ASSERT2(rhs.size() == N);
+    
+    int nrhs = rhs.size();
+    Matrix<T> rhsMatrix(1, N);
+    Matrix<T> xMatrix(1, N);
+
+    //Copy input data into matrix
+    for(int j = 0; j < nrhs; ++j){
+      for(int i = 0; i < N; ++i){
+	rhsMatrix(j, i) = rhs[j][i];
+      }
+    }
+    
+    // Solve
+    solve(rhsMatrix, xMatrix);
+
+    //Copy result back into argument
+    for(int j = 0; j < nrhs; ++j){
+      for(int i = 0; i < N; ++i){
+	x[j][i] = xMatrix(j, i);
+      }
+    }    
+  };
+  
+  /// Solve a set of tridiagonal systems
+  /// 
   /// @param[in] rhs Matrix storing Values of the rhs for each system
   /// @param[out] x  Matrix storing the result for each system
   void solve(Matrix<T> &rhs, Matrix<T> &x) {
     TRACE("CyclicReduce::solve");
+    ASSERT2(std::get<0>(rhs.shape() == Nsys);
+    ASSERT2(std::get<0>(x.shape() == Nsys);
+    ASSERT2(std::get<1>(rhs.shape() == N);
+    ASSERT2(std::get<1>(x.shape() == N);
+
     // Multiple RHS
     int nrhs = std::get<0>(rhs.shape());
     
--- a/include/cyclic_reduction.hxx
+++ b/include/cyclic_reduction.hxx
@@ -48,6 +48,8 @@
 #include <lapack_routines.hxx>
 #include "boutexception.hxx"
 
+#include "output.hxx"
+
 template <class T>
 class CyclicReduce {
 public:
--- a/include/cyclic_reduction.hxx
+++ b/include/cyclic_reduction.hxx
@@ -89,7 +89,7 @@ public:
   void setPeriodic(bool p=true) {periodic=p;}
   
   /// Set up a single equation to invert
-  void setCoefs(T a[], T b[], T c[]) {
+  void setCoefs(Matrix<T> &a, Matrix<T> &b, Matrix<T> &c) {
     setCoefs(1, &a, &b, &c);
   }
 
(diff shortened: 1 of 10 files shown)


include/cyclic_reduction.hxx
```diff
@@ -100,30 +103,33 @@ public:
...
+        coefs(j, 4*i + 1) = b(j, i);
+        coefs(j, 4*i + 2) = c(j, i);
         // 4*i + 3 will contain RHS
       }
   }
 
   /// Solve a single triadiagonal system
   /// 
-  void solve(T rhs[], T x[]) {
+  void solve(Matrix<T> &rhs, Matrix<T> &x) {
```
bendudson: I think this should be Array, rather than Matrix. This function solves a single system (Ax=b), so `x` and `rhs` here should be Arrays. On line 128 the address of `rhs` and `x` were used, to convert T* to T** . Is there a way to convert an Array to a single-column (or single-row) Matrix?

include/cyclic_reduction.hxx
```diff
@@ -100,30 +103,33 @@ public:
...
     // Solving single system
     solve(1, &rhs, &x);
   }
 
   /// Solve a set of tridiagonal systems
   /// 
-  void solve(int nrhs, T **rhs, T **x) {
+  void solve(int nrhs, Matrix<T> &rhs, Matrix<T> &x) {
```
bendudson: Can the `nrhs` argument be removed? This should be the number of columns in `rhs`? I guess it is bounds checked anyway, but shouldn't be needed.

include/cyclic_reduction.hxx
d7919: There's not currently a way to convert an Array to a `Matrix`, but it should be easy enough to add a constructor to Matrix that does this. In general there isn't a unique way to do this, but we could always make the choice that either the matrix is 1xN or Nx1.

I've been thinking about the possibility of making `Matrix::operator[]` return an `Array` slice of the data (either by creating a new Array and copying in values, or making `Matrix::data` actually be a vector of `Array`). WIth this I think it would make sense to get a `1xN` shape Matrix when constructing from an Array of size N.

This interface to `solve` isn't actually used anywhere at the moment (at least not in any of the framework or the tests) so we could remove it and revisit if/when this is needed/the Matrix<-->Array conversion is sorted?

d7919: I should note there's a similar argument for setcoefs.

include/cyclic_reduction.hxx
```diff
@@ -88,43 +91,44 @@ public:
...
   void setPeriodic(bool p=true) {periodic=p;}
   
-  /// Set up a single equation to invert
-  void setCoefs(T a[], T b[], T c[]) {
-    setCoefs(1, &a, &b, &c);
-  }
-
   /// Set the entries in the matrix to be inverted
   ///
   /// @param[in] nsys   The number of independent matrices to be solved
```
ZedThree: Doxygen comment is now out of date

include/cyclic_reduction.hxx
```diff
@@ -88,43 +91,44 @@ public:
   /// By default not periodic
   void setPeriodic(bool p=true) {periodic=p;}
   
-  /// Set up a single equation to invert
-  void setCoefs(T a[], T b[], T c[]) {
```
ZedThree: Should this be deprecated instead of removed? Or is the signature compatible with `Matrix<T>`?

include/cyclic_reduction.hxx
```diff
@@ -88,43 +91,44 @@ public:
...
-  void solve(T rhs[], T x[]) {
-    // Solving single system
-    solve(1, &rhs, &x);
-  }
+  // /// Solve a single triadiagonal system
+  // /// 
+  // void solve(Matrix<T> &rhs, Matrix<T> &x) {
+  //   // Solving single system
+  //   solve(1, &rhs, &x);
+  // }
```
ZedThree: Commented-out code should be removed instead

include/cyclic_reduction.hxx
```diff
@@ -88,43 +91,44 @@ public:
...
 
   /// Solve a set of tridiagonal systems
   ///
```
ZedThree: Could the doxygen comment be expanded here?

include/cyclic_reduction.hxx
```diff
@@ -421,122 +424,92 @@ private:
...
-        T beta = c[4*i+2] / ic[1];
+        T beta = co(j, 4*i+2) / ifc(j, 1);
           
         // v_u <- v_i - beta * v_u
-        ic[1] = c[4*i + 1] - beta * ic[0];
-        ic[0] = c[4*i];
-	ic[2] *= -beta;
+        ifc(j, 1) = co(j, 4*i + 1) - beta * ifc(j, 0);
+        ifc(j, 0) = co(j, 4*i);
+	ifc(j, 2) *= -beta;
```
ZedThree: Whitespace issue

include/cyclic_reduction.hxx
```diff
@@ -421,122 +424,92 @@ private:
...
 	
         // alpha <- v_{i,i-1} / v_l,i-1
-        T alpha = c[4*i] / ic[1];
+        T alpha = co(j, 4*i) / ifc(j, 4+1);
           
         // v_l <- v_i - alpha*v_l
-	ic[0] *= -alpha;
-        ic[1] = c[4*i + 1] - alpha*ic[2];
-        ic[2] = c[4*i + 2];
+	ifc(j, 4+0) *= -alpha;
```
ZedThree: Whitespace

src/invert/parderiv/impls/cyclic/cyclic.cxx
```diff
@@ -147,11 +140,11 @@ const Field3D InvertParCR::solve(const Field3D &f) {
         
         //           const     d2dy2        d2dydz             d2dz2           ddy
         //           -----     -----        ------             -----           ---
-	a[k][y+y0] =            bcoef - 0.5*Im*kwave*ccoef                  -0.5*ecoef;
-	b[k][y+y0] = acoef - 2.*bcoef                     - SQ(kwave)*dcoef;
-	c[k][y+y0] =            bcoef + 0.5*Im*kwave*ccoef                  +0.5*ecoef;
-	
-	rhsk[k][y+y0] = rhs[y+y0][k]; // Transpose
+        a(k, y + y0) = bcoef - 0.5 * Im * kwave * ccoef - 0.5 * ecoef;
```
ZedThree: The whitespace changes have made the preceding comment not very useful. Should we either preserve the whitespace, or delete/modify the comment?

include/cyclic_reduction.hxx
```diff
@@ -421,122 +424,92 @@ private:
...
+    ifx = Matrix<T>(myns, 2*nprocs);       // Solution of interface equations
+    ifp = Array<T>(myns*2);     // Solution to be sent to processor p
+    // Each system to be solved on this processor has two interface equations from each processor
     
+    x1 = Array<T>(Nsys);
+    xn = Array<T>(Nsys);
   }
 
   /// Free all memory arrays allocated by allocMemory()
   void freeMemory() {
```
ZedThree: Is this needed at all now? Can it be deprecated/removed?

include/cyclic_reduction.hxx
d7919: Yes I guess this should be deprecated -- I've replaced all uses in the repository but I suppose someone could be using it in their private model. My main issue was how to then construct a `Matrix` from this to pass through to the actual routine.

The same for the solve interfaces I guess.

include/cyclic_reduction.hxx
d7919: Thanks!

include/cyclic_reduction.hxx
d7919: I thought I'd not staged this, oops.

src/invert/parderiv/impls/cyclic/cyclic.cxx
d7919: I'll revert the whitespace.

include/cyclic_reduction.hxx
d7919: It currently also resets the problem size. I don't think we really need this as it should be cheap enough to fully tear down and rebuild a new instance should we need to. May be worth deprecating?

d7919: I've marked as deprecated, but this is private so I think we should probably be free to just remove it if we decide that we don't need to reset the problem size or if we want to do this a different way.

include/cyclic_reduction.hxx
d7919: Now removed.

include/cyclic_reduction.hxx
d7919: I've added interfaces to `setCoefs` and `solve` that take `Array` instead of `Matrix`. These then copy into a `Matrix` and pass to the main routine before copying back into the `Array` (where needed).

include/cyclic_reduction.hxx
```diff
@@ -282,15 +395,15 @@ public:
...
+	    x1[sys0+i] = ifx(i, 2*p);
+	    xn[sys0+i] = ifx(i, 2*p+1);
 	  }
           req[p] = MPI_REQUEST_NULL;
 	}else if(nsp > 0) {
 #ifdef DIAGNOSE
           output << "Expecting receive from " << p << " of size " << len << endl;
 #endif
-	  MPI_Irecv(recvbuffer[p],
+	  MPI_Irecv(&recvbuffer(p, 0),
```
ZedThree: Can we use `std::begin` here, similar to how `ifp` gets passed below? Or is that not possible due to the subscripting?

d7919: Yes, I think the indexing prevents this from working with `std::begin`, as `recvbuffer(p,0)` refers to a single value I'm not sure what `begin` would do here.

d7919: Most of the places I take the address or use `std::begin` with all the changes related to arrays, matrices and tensors are places I would like to change the interface of the functions to take the bounds-checked class rather than reverting to passing pointers. I've not made such changes yet as it would touch a lot of signatures and there are a couple of things that we might want to change in the design of Matrix and similar which might change the best route forward.

PR #901, opened 2018-02-24
Remove matrix remainder

d7919:
Together with #900 this should remove all the remaining uses of the old style matrix. I split the two PRs as the cyclic changes in #900 are relatively large. 

The advantage of this is that `Matrix`/`Arrray` access has optional bounds checking and due to the relatively cheap cost of creating them we can remove the need for several static variables.

There's still some behaviour that could be improved in relation to this -- I've not changed many routine interfaces so we still expect pointers to data blocks in many places (and also pass the data size). It would be nice to pass the `Matrix`/`Array` directly and this would help allow us to avoid passing the sizes in (as these classes store the size). 

There are still a large number of `new` uses that could probably be replaced with `Array`. This could be addressed in a separate PR.

Marks old style `matrix` routine as deprecated.

### DIFF
--- a/include/lapack_routines.hxx
+++ b/include/lapack_routines.hxx
@@ -23,6 +23,8 @@
 #ifndef __LAPACK_ROUTINES_H__
 #define __LAPACK_ROUTINES_H__
 
+#include <utils.hxx>
+
 /* Tridiagonal inversion
  *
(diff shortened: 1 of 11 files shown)


bendudson:
Thanks David! This looks like a great improvement

PR #902, opened 2018-02-24
Replace new with array

d7919:
Replaces uses of `new` with `Array` in a number of places. Combined with #900 and #901 this will remove a large fraction of raw memory access and replaces it with something that supports bounds checking.

This could possibly be merged as is, but we may wish to try to remove most other uses first. Some not done here as difficult to test currently.

Changes the interface of some methods to take Array instead of a pointer. Other interfaces unchanged and instead pass `std::begin(ArrayVariable)`. Reasons for not changing include more widespread usage and in places we want to skip some of the first elements of the array, which is not possible with passing an Array (without first making a new array and copying in the relevant parts).


d7919:
One thought I've had is that it may be more consistent to provide a 1D wrapper around `Array` similar to the 2D `Matrix` and 3D `Tensor`. This would enable us to use things like copy-on-write consistently between `Matrix` and `Array1D` and means that we can try to avoid encouraging users to use "raw"  `Array`s so if we want to change the interface to `Array` we only have to update the internals of `Array1D`, `Matrix` and `Tensor` for example.

PR #907, opened 2018-02-27
Build API documentation on Readthedocs

ZedThree:
This attempts to build the documentation for the C++ and Python APIs on readthedocs, including the boututils, boutdata and zoidberg Python modules. Unfortunately, in order to actually get this all built within 15 minutes, I've had to turn off building the PDF -- note that this has turned off building the PDF for next and master too.

Also note that the builds will take longer when the service is busy, and this PR is currently pushing up against the time limit as is, so we should not be surprised if the occasional build times out.

The documentation looks like this: http://bout-dev.readthedocs.io/en/python-docs/_breathe_autogen/file/boutmesh_8hxx.html#classBoutMesh

Also add some documentation about dependencies for the Python tools.


dschwoerer:
Would it be worth switching to another server?
https://omadesala.physics.dcu.ie/bout has already one branch - it would be easy to add more branches ...

ZedThree:
Depends how badly we want PDFs. Another option would be to build the PDFs on Travis and upload to the website.

---

This PR has highlighted that we don't have a consistent style for the Python docstrings. I recommend we pick either [Google or NumPy][1], as Sphinx has support for both of these, so either one should look good. I'll then reformat the docstrings and submit a separate PR.

[1]: http://www.sphinx-doc.org/en/stable/ext/napoleon.html#google-vs-numpy

dschwoerer:
For #641 it is required to build the complete BOUT++ library to create the documentation, which was the motivation to setup the extra manual build system.
As we are now encountering issues with building the PDFs, it might be worth switching to a more complete solution ...

manual/sphinx/_apidoc/boutdata.rst
```diff
@@ -0,0 +1,96 @@
+boutdata package
```
d7919: Is it possible to auto-generate this file, it looks quite formulaic?

manual/sphinx/_apidoc/boututils.rst
```diff
@@ -0,0 +1,236 @@
+boututils package
```
d7919: Is it possible to auto-generate this file, it looks quite formulaic?

manual/sphinx/_apidoc/zoidberg.rst
```diff
@@ -0,0 +1,117 @@
+zoidberg package
```
d7919: Is it possible to auto-generate this file, it looks quite formulaic?

manual/sphinx/conf.py
```diff
@@ -30,16 +30,15 @@
...
 
 # Are we running on readthedocs?
 on_readthedocs = os.environ.get("READTHEDOCS") == "True"
 
 # readthedocs currently runs out of memory if we actually dare to try to do this
-if has_breathe and not on_readthedocs:
+if has_breathe:
     # Run doxygen to generate the XML sources
-    subprocess.call("cd ../doxygen; doxygen Doxyfile", shell=True)
+    subprocess.call("cd ../doxygen; doxygen Doxyfile_readthedocs", shell=True)
```
d7919: The readthedocs environment should persist between repeat runs (i.e. when generating the different output types) so it should be possible to only do this once per build rather than once per format. I'll just record here that you've timed this and noted that this call is a relatively small fraction of the time.

manual/sphinx/_apidoc/boutdata.rst
ZedThree: These were auto-generated by `sphinx-apidoc`, but then I edited them slightly (moved the full module docs to the top) and committed those edits.

I could add the autogeneration to `conf.py`, similar to the `breathe` docs, but I'm not 100% sure the documentation is laid out the best it could be

manual/sphinx/conf.py
d7919: Does this add any dependencies for users who want to build the docs locally?

PR #908, opened 2018-02-27
Block region iterator field ops

d7919:
Use the block region for fieldops.

Fixes a couple of potential "future bugs" in the jinja template.

Simplifies the generating code slightly.


ZedThree:
Apologies for the Codacy check suddenly appearing -- I wasn't expecting it to do that!

PR #909, opened 2018-03-01
Precompute Delp2 coefficients

JosephThomasParker:
Precompute the coefficients for `Delp2`, storing the coefficients as `Tensor`s in the `Laplacian`.

### DIFF
--- a/src/invert/laplace/invert_laplace.cxx
+++ b/src/invert/laplace/invert_laplace.cxx
@@ -112,6 +112,7 @@ Laplacian* Laplacian::create(Options *opts) {
 Laplacian* Laplacian::instance = NULL;
 
 Laplacian* Laplacian::defaultInstance() {
+  SCOREP0();
   if(instance == NULL)
     instance = create();
   return instance;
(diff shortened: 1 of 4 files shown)


src/invert/laplace/invert_laplace.cxx
```diff
@@ -704,6 +704,34 @@ void Laplacian::tridagMatrix(dcomplex *avec, dcomplex *bvec, dcomplex *cvec,
   }
 }
 
+void Laplacian::calcDelp2Coefs() {
+  TRACE("Laplacian::calcDelp2Coefs");
+
+  static bool initializedDelp2Coefs = false ;
```
d7919: I think this shadows the class level variable of the same name. I'd guess you want one or the other.

src/mesh/coordinates.cxx
```diff
@@ -638,32 +638,34 @@ const Field3D Coordinates::Delp2(const Field3D &f) {
...
+
+  static Laplacian *d2lap = Laplacian::defaultInstance();
+  d2lap->calcDelp2Coefs();
 
   // Loop over all y indices
   for (int jy = 0; jy < localmesh->LocalNy; jy++) {
 
     // Take forward FFT
 
+    #pragma omp parallel for
```
d7919: If we declared `delft` and `ft` inside the `jy` loop could we move the omp parallelisation to the outer `jy` loop?

src/mesh/coordinates.cxx
```diff
@@ -698,8 +700,10 @@ const FieldPerp Coordinates::Delp2(const FieldPerp &f) {
   int ncz = localmesh->LocalNz;
 
   // Allocate memory
-  auto ft = Matrix<dcomplex>(localmesh->LocalNx, ncz / 2 + 1);
-  auto delft = Matrix<dcomplex>(localmesh->LocalNx, ncz / 2 + 1);
+  Matrix<dcomplex> delft(localmesh->LocalNx, ncz / 2 + 1);
+  Matrix<dcomplex> ft(localmesh->LocalNx, ncz / 2 + 1);
+  static Laplacian *d2lap = Laplacian::defaultInstance();
```
d7919: Does this need to be static if we're getting the default instance -- I think it should be the same instance each time.

dschwoerer:
This does adds code that relies on the global mesh.
Would it be possible to avoid this and use the field mesh?
Maybe create a map with coefficients for each mesh?

d7919:
Could you replace the "raw" pragmas with `BOUT_OMP`?

bendudson:
I think this is a good idea, but unless the performance improvement is large, this should wait until a way can be found to manage dependencies e.g. update if the metric tensor is changed.

PR #915, opened 2018-03-02
Use web service for tracking unit test coverage

ZedThree:
Adds two new Travis jobs for gathering coverage information from the unit tests and uploads to CodeCov: https://codecov.io/gh/boutproject/BOUT-dev (note: you might need to go to "branches" to see anything useful!)

One of these jobs is supposed to be "maximal", i.e. try to configure with everything we support. Unfortunately, it's not possible to use SUNDIALS with it at the mo, due to the version in Ubuntu 14. At some point, I'll make a new PR that fixes this\*.
The other job is "minimal", i.e. with as few extra features turned on as possible (except code coverage, obviously). Unfortunately, this also doesn't work as intended due to how some of the configure options interact! Another PR that fixes that will be along as well.

Also adds config for Codacy, which does static analysis online: https://app.codacy.com/app/ZedThree/BOUT-dev/dashboard
Note that that appears to be linked to my personal account -- if you sign into Codacy with your own github account, I'm not sure how the settings interact!

---
\* The plan is to download + compile it on Travis ourselves, but cache the built version so we can use it in the separate jobs.

### DIFF
--- a/.codacy.yml
+++ b/.codacy.yml
@@ -1,0 +1,10 @@
+---
+languages:
+  cpp:
+    extensions:
+      - '*.hxx'
+coverage:
+  enabled: true
+  exclude_paths:
+    - tests/**
+    - examples/**
--- a/.codacy.yml
+++ b/.codacy.yml
@@ -3,6 +3,7 @@ languages:
   cpp:
     extensions:
       - '*.hxx'
+      - '*.cxx'
 engines:
   coverage:
     enabled: true
--- a/.codacy.yml
+++ b/.codacy.yml
@@ -3,8 +3,23 @@ languages:
   cpp:
     extensions:
       - '*.hxx'
-coverage:
-  enabled: true
-  exclude_paths:
-    - tests/**
-    - examples/**
+engines:
+  coverage:
+    enabled: true
+    exclude_paths:
+      - tests/**
+      - examples/**
+  bandit:
+    enabled: true
+    include_paths:
+      - tools/pylib/**
+  cppcheck:
+    enabled: true
+  pylint:
+    enable: true
+    include_paths:
+      - tools/pylib/**
+  prospector:
+    enable: true
+    include_paths:
+      - tools/pylib/**
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,15 +57,40 @@ matrix:
     #to remove this if a newer openmpi version is introduced.
   - addons:
       apt:
-        - openmpi-bin
-        - libopenmpi-dev
-        - libfftw3-dev
-        - libnetcdf-dev
-        - libhdf5-serial-dev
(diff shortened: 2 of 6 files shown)


.codacy.yml
```diff
@@ -0,0 +1,26 @@
+---
```
d7919: Is it worth commenting some of this to say what the different sections control?

.travis.yml
```diff
@@ -55,6 +55,55 @@ matrix:
...
+          - libpetsc3.4.2-dev
+          - libslepc3.4.2-dev
+          - libmumps-dev
+          - libscalapack-mpi-dev
+          - libparmetis-dev
+          - libmetis-dev
+          - liblapack-dev
+          - libparpack2-dev
+    env:
+      - CONFIGURE_OPTIONS='--enable-code-coverage --enable-debug --enable-checks=3 --with-petsc --with-lapack --with-slepc --with-mumps --enable-openmp'
```
d7919: I think --enable-debug should automatically set --enable-checks=3

d7919:
This seems to also include #911 -- can we close that PR or should we remove the commit from this one?

ZedThree:
Coverage reports were a bit inaccurate as they only included files that were "touched" by the tests. The latest commit tries to fix this by ensuring that there is a corresponding .gcda file for each .gcno file (see https://github.com/codecov/support/issues/292).

Unfortunately, we still miss stuff that doesn't get compiled in the first place, e.g. ARKODE stuff. This can be solved eventually by having enough different builds that compile with all our possible dependencies, as codecov is clever enough to merge coverage reports from multiple jobs.

d7919:
Could we merge the two different configures for codecov into a single Travis job? This would save the ~5 minutes of environment setup time from the second case, we'd just have to make distclean in between.

dschwoerer:
Currently I can only see that some unit tests are checked (coverage 100%).
This info is rather useless - I want to know what in src/ and include/ etc is not checked ...
Also if I click on the other 11 files, e.g. [here](https://github.com/boutproject/BOUT-dev/pull/924#issuecomment-371894191) I only see an empty page with no changes.
How can I use this?

d7919:
If you go to the [landing page](https://codecov.io/gh/boutproject/BOUT-dev) and then click branches you can explore the coverage for the branches that currently have coverage enabled. For example [next](https://codecov.io/gh/boutproject/BOUT-dev/branch/next). At the bottom of each of these pages you get a file explorer like tool that lets you dig down by directory/file to see the coverage of different bits.

There still seems to be some issues with the accuracy/utility of the PR comments etc. but at least the branch file summary is helpful, I think.

dschwoerer:
Yes, the branches view seems much more useful.
Thanks for the help :+1:

PR #916, opened 2018-03-02
Extend field perp

d7919:
Extend FieldPerp to behave more like 2D and 3D fields.

Also adds unit tests for most FieldPerp routines

### DIFF
--- a/src/field/fieldperp.cxx
+++ b/src/field/fieldperp.cxx
@@ -35,9 +35,9 @@
 #include <msg_stack.hxx>
 
 FieldPerp::FieldPerp(Mesh *localmesh) : Field(localmesh), yindex(-1) {
-  if (localmesh) {
-    nx = localmesh->LocalNx;
-    nz = localmesh->LocalNz;
+  if (fieldmesh) {
+    nx = fieldmesh->LocalNx;
+    nz = fieldmesh->LocalNz;
   }
   
 #if CHECK > 0
--- a/src/field/fieldperp.cxx
+++ b/src/field/fieldperp.cxx
@@ -65,6 +65,9 @@ void FieldPerp::allocate() {
       nz = fieldmesh->LocalNz;
     }
     data = Array<BoutReal>(nx * nz);
+#if CHECK > 2
+    invalidateGuards(*this);
+#endif
   } else
     data.ensureUnique();
 }
--- a/src/field/fieldperp.cxx
+++ b/src/field/fieldperp.cxx
@@ -23,6 +23,7 @@
  *
  **************************************************************************/
 
+#include <boutcomm.hxx>
 #include <globals.hxx>
 
 #include <stdlib.h>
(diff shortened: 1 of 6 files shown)


src/field/fieldperp.cxx
```diff
@@ -254,3 +341,156 @@ const IndexRange FieldPerp::region(REGION rgn) const {
     break;
   };
 }
+
+//////////////// NON-MEMBER FUNCTIONS //////////////////
```
dschwoerer: aren't the above also non-member functions?

src/field/fieldperp.cxx
```diff
@@ -33,7 +34,7 @@
 #include <boutexception.hxx>
 #include <msg_stack.hxx>
 
-FieldPerp::FieldPerp(Mesh *localmesh) {
+FieldPerp::FieldPerp(Mesh *localmesh) : Field(localmesh), yindex(-1) {
   // Get mesh size
   fieldmesh = localmesh;
```
dschwoerer: fieldmesh should already be set.

src/field/fieldperp.cxx
d7919: Yes, sorry was thrown by the `FieldPerp::region` which came after non-member stuff.

src/field/fieldperp.cxx
d7919: Ah yes thanks, it didn't used to initialise Field automatically.

src/field/fieldperp.cxx
d7919: Should be addressed now

src/field/fieldperp.cxx
d7919: Should be addressed now

src/field/fieldperp.cxx
```diff
@@ -143,8 +168,28 @@ FPERP_OP_REAL(-=, -);
 FPERP_OP_REAL(*=, *);
 FPERP_OP_REAL(/=, /);
 
+const IndexRange FieldPerp::region(REGION rgn) const {
+  switch (rgn) {
+  case RGN_ALL:
```
johnomotani: It looks strange not having ``RGN_ALL`` implemented. If I understand right, it works because the switch drops into ``case RGN_NOZ`` which (at the moment) is the same as ``RGN_ALL``. If this changes at some point, we might catch a potential mistake by adding an explicit implementation of ``RGN_ALL`` now.

d7919: Yes, I'm not these regions are as good as they could be. 

For example the y indices are taken to be {0,0}, which is fine for use with other perps but no good if we wanted to use this with Field3D indexing.

I would be happy to "fix" this in this PR if others think it would make sense. (I should note whilst the diff makes it look like I've added this as new code I haven't there's a corresponding "removal" further down, I just repositioned it in the file).

PR #918, opened 2018-03-06
More unit tests

d7919:
Expand unit tests slightly. Helps increase test coverage.


tests/unit/field/test_field3d.cxx
```diff
@@ -810,13 +1081,50 @@ TEST_F(Field3DTest, AddField3DField3D) {
...
+
+  // Throw as two rhs fields at different locations
+#if CHECK > 0
+  EXPECT_THROW(c = a + b, BoutException);
+#else
+  EXPECT_NO_THROW(c = a + b);
+#endif
+
+  // No throw as updates location of a
+  EXPECT_NO_THROW(a = c + b);
```
ZedThree: Should we also check the value of `a` is what we expect in these stagger tests?

tests/unit/sys/test_options.cxx
```diff
@@ -35,6 +37,24 @@ TEST_F(OptionsTest, SetGetInt) {
...
+  int value;
+  options.get("int_key", value, 99, false);
+
+  EXPECT_EQ(value, 42);
+
+  options.set("int_key2", 12.5, "code");
+  EXPECT_THROW(options.get("int_key2", value, 99, false), BoutException);
+  // Note we expect to get the rounded value despite the throw as we
+  // pass by reference and modify the passed variable in options.get.
+  EXPECT_EQ(value, 13);
```
ZedThree: Is this what we actually *want* to happen though? This feels more like a bug to me

d7919: It's only because we catch the throw here though and carry on that we see this at all. It's slightly unexpected but I don't think there is an expected value that makes sense either.

tests/unit/field/test_field3d.cxx
d7919: We could, but we don't actually do anything different than in the non-staggered add -- it's just the checking of locations that could be different I think.

tests/unit/sys/test_options.cxx
ZedThree: True. In that case, we should maybe not test the value after, and say that it's "undefined" instead.

d7919: We could -- I mostly included this as some form of documentation.
If we want the value to be undefined then I wonder if we can make `get` force the value to be `NaN` to clearly indicate our intent?

It's vaguely possible that someone's code is relying on this (without realising it) so by including this test we'd at least catch when this assumption changes.

PR #920, opened 2018-03-06
More region arguments

johnomotani:
Add region argument to interp_to:
* Reduces calculation by not computing values in guard cells, which can
only be calculated correctly by communicating.

* Allows communications to be skipped if region==RGN_NOBNDRY.

* Also update uses of interp_to in derivs.cxx and index_derivs.cxx to use
region argument to avoid unnecessary calculation/communication.

Also add region argument to Vector2D::abs() and Vector3D::abs().


johnomotani:
Following @dschwoerer and @d7919's comments on #921, I'll remove the changes to index_derivs.cxx from this PR. The default region argument to ``interp_to()`` maintains the previous behaviour. The changes were only intended to give minor speedups in cases that are probably never used, they conflict with #815 and should be superseded by a resolution to #922 so not worth making here.

johnomotani:
This PR now includes a bugfix for z-direction interpolation. Should I pull it out into a separate PR so it can be merged into master? I'd think it's not important to add to master, because surely no one is using z-staggering.

I've added a unit test for interp_to. It seems to work, so I've removed the integrated test.

8bf0dee implemented @d7919's suggestion about copying var to result.

tests/unit/field/test_field3d.cxx
```diff
@@ -1778,6 +1779,74 @@ TEST_F(Field3DTest, DC) {
   EXPECT_TRUE(IsField2DEqualBoutReal(DC(field), 3.0));
 }
 
+TEST_F(Field3DTest, interp_to) {
```
d7919: I'm not sure if we want to test all different locations in one test or to have a separate test for each,@ZedThree can probably comment more on this. The former saves duplicating setup etc., but the later helps more clearly isolate where the problem has occurred and if it occurs for multiple cases.

tests/unit/test_extras.hxx
```diff
@@ -61,6 +65,7 @@ public:
     PE_XIND = 0;
     StaggerGrids = false;
     IncIntShear = false;
+    transform = std::unique_ptr<ParallelTransform>(new ParallelTransformIdentity());
```
d7919: I'm not sure we want this line here in general as I think we'll want to test other transforms as well.

tests/unit/test_extras.cxx
```diff
@@ -50,3 +50,15 @@ ::testing::AssertionResult IsFieldPerpEqualBoutReal(const FieldPerp &field,
 
   return ::testing::AssertionSuccess();
 }
+
+/// Is \p field equal to \p number, with a tolerance of \p tolerance?
+::testing::AssertionResult IsBoutRealEqualBoutReal(BoutReal number1, BoutReal number2,
```
ZedThree: Was `EXPECT_DOUBLE_EQ` too strict? Can you use `EXPECT_NEAR(val1, val2, abs_tol)` instead of adding this?

tests/unit/test_extras.hxx
ZedThree: Agreed. It doesn't look like we need a ParallelTransform for any of the tests so far, so it could just be left as `nullptr`, unless I've missed something?

tests/unit/field/test_field3d.cxx
ZedThree: Splitting them up would be better, for the reasons @d7919 says -- it makes it a bit clearer when there's a problem with one case vs multiple cases.

To save on the duplicating setup, I would put all the setup in a fixture -- copy/paste the `Field3DTest` class to `Field3DInterpToTest`, put all the setup in there (including the definition and creation of `input`, but not the definition of `output`), then make a bunch of tests like:

```cpp
TEST_F(Field3DInterpToTest, CellCentreToXLow) {
```
etc.

src/mesh/interpolation.cxx
```diff
@@ -178,51 +190,58 @@ const Field3D interp_to(const Field3D &var, CELL_LOC loc)
...
       
       if(dir != CELL_ZLOW) {
-	// COMMUNICATION
-	
-	mesh->communicate(result);
-
-	// BOUNDARIES
+        // COMMUNICATION
+        if ( region!=RGN_NOBNDRY ) fieldmesh->communicate(result);
+        // BOUNDARIES
```
ZedThree: While we're touching things here, both of these comments can go.

Also, if you have clang-format installed, please run it over these bits :)

tests/unit/test_extras.cxx
johnomotani: Thank you, EXPECT_NEAR was what I wanted, just didn't know it existed. Will update...

tests/unit/test_extras.hxx
johnomotani: The ``interp_to()`` test does need a ParallelTransform.
``interp_to()`` calls ``fieldmesh->toFieldAligned(var)`` when interpolating in the y-direction.
``Mesh::communicate()`` calls ``getParallelTransform()``, and communicate is called by interp_to (unless region==RGN_NOBNDRY, which I have not set in the tests as it is not the default option, although maybe we should add a test with RGN_NOBNDRY as well).
Both of these make ``serial_tests`` segfault when calling ``setParallelTransform`` because they cannot read options.

tests/unit/field/test_field3d.cxx
johnomotani: Done. Is the way I set the class up correct? I had a bit of trouble initializing static members, etc. but I think I sorted it out.

src/mesh/interpolation.cxx
johnomotani: Ran clang-format over interpolation.hxx interpolation.cxx and the new test_interpolation.cxx file.

include/interpolation.hxx
```diff
@@ -105,79 +109,84 @@ class HermiteSpline : public Interpolation {
...
   // Use precalculated weights
-  Field3D interpolate(const Field3D& f) const;
+  Field3D interpolate(const Field3D &f) const;
   // Calculate weights and interpolate
-  Field3D interpolate(const Field3D& f, const Field3D &delta_x, const Field3D &delta_z);
-  Field3D interpolate(const Field3D& f, const Field3D &delta_x, const Field3D &delta_z, BoutMask mask);
-
+  Field3D interpolate(const Field3D &f, const Field3D &delta_x, const Field3D &delta_z);
+  Field3D interpolate(const Field3D &f, const Field3D &delta_x, const Field3D &delta_z,
+                      BoutMask mask);
```
d7919: Should we pass mask by reference?

include/interpolation.hxx
```diff
@@ -105,79 +109,84 @@ class HermiteSpline : public Interpolation {
...
+  Tensor<int> k_corner; // z-index of bottom-left grid point
 
   Field3D t_x, t_z;
 
 public:
   Lagrange4pt(Mesh *mesh = nullptr) : Lagrange4pt(0, mesh) {}
   Lagrange4pt(int y_offset = 0, Mesh *mesh = nullptr);
-  Lagrange4pt(BoutMask mask, int y_offset=0, Mesh *mesh = nullptr) : Lagrange4pt(y_offset, mesh) {
-    skip_mask = mask;}
+  Lagrange4pt(BoutMask mask, int y_offset = 0, Mesh *mesh = nullptr)
```
d7919: mask by reference?

include/interpolation.hxx
```diff
@@ -105,79 +109,84 @@ class HermiteSpline : public Interpolation {
 public:
   HermiteSpline(Mesh *mesh = nullptr) : HermiteSpline(0, mesh) {}
   HermiteSpline(int y_offset = 0, Mesh *mesh = nullptr);
-  HermiteSpline(BoutMask mask, int y_offset=0, Mesh *mesh = nullptr) : HermiteSpline(y_offset, mesh) {
-    skip_mask = mask;}
+  HermiteSpline(BoutMask mask, int y_offset = 0, Mesh *mesh = nullptr)
```
d7919: mask by reference?

tests/unit/test_extras.hxx
d7919: Adding `mesh->setParallelTransform(std::unique_ptr<ParallelTransform>(new ParallelTransformIdentity()));` to `SetUpTestCase` in test_interpolation.cxx should allow you to set the transform for the interpolation tests without hardcoding it for all tests that use FakeMesh (i.e. can remove the above line).

johnomotani: Thanks @d7919! I've implemented now as you suggested.

johnomotani:
I've updated to pass all the ``mask`` arguments by const reference. Any other changes needed? I'd like to get this mereged, because at the moment ``abs(Grad(phi))`` is broken due to checking invalid data in the guard cells (since the changes to checkData), so I want to be able to call ``abs(Grad(phi), RGN_NOBNDRY)``. I don't claim using the vector operations is the best way of doing this, but that's another issue...

d7919:
What changes to `checkData` have led to problems? I wasn't aware anything had changed in `checkData` recently, other than the option to pass in a region (but this defaults to the original behaviour).

johnomotani:
Sorry, it wasn't the changes to checkdata! It was when adding region arguments to functions (``sqrt()`` in particulary) in 812d74a, where the region is passed also to checkData. The default region argument for the functions is RGN_ALL. The vector function ``abs()`` at the moment does not have a region argument, and calls ``sqrt()`` without a region argument, causing the problem.

d7919:
Is the issue because of invalid values in the guard cells? If so do these arise in `phi`, `Grad(phi)` or `Grad(phi)*Grad(phi)`? (I'm guessing `Grad(phi)`).

The original behaviour for `checkData` usage in these functions was to only check the no_bndry region, whilst I can see the logic behind passing the region used to do the operation through to checkData I can also see it potentially causing a range of issues like this, particularly if the framework is not consistent. For example the `operator*` used in `abs` will still operate over RGN_ALL but the checkData call (implicit) will use the default region (no_bndry).

johnomotani:
You're right, it's coming from invalid guard cells due to Grad(phi).

I'd like to have region arguments available wherever possible for optimization purposes, so am happy with that as a solution. Would have no objections to not passing the region through to checkData though.

bendudson:
What is the status of this PR? Is the only remaining issue checkData? If so, I suggest we just leave checkData as always operating over no_bndry for now, so we can get the other changes in.

johnomotani:
I think the only issue is the choice of what to do for checkData. At the moment in friend functions the region is passed through to checkData, e.g. https://github.com/boutproject/BOUT-dev/blob/4223355e40eec10a1317698e8a555030b3a5bb34/src/field/field3d.cxx#L864
Shall we leave as is, or reset to the original behaviour of checkData always operating over no_bndry?

d7919:
I think checkData should retain the flexibility to operate on different regions, but i think for now we probably don't want to make use of this flexibility in most places.

bendudson:
I vote to keep the original behaviour of checkData for now. This could cause problems later if the function is only operating on a part of the domain which is smaller than the domain which checkData is checking. checkData would then be checking parts of the domain which were not supposed to be set to anything.

If the problem arises when checkData is passed `RGN_ALL`, can we ensure that checkData never checks the boundaries by masking the given `rgn` with `RGN_NOBNDRY`? That way the region checked will be the intersection of `rgn` and `RGN_NOBNDRY`

bendudson:
Good point David; perhaps keeping it but not using it for now...? Unless masking works...

d7919:
I think there will be cases where we do wish to check the boundary (e.g. for checking for invalidated guard cells). I think the real issue here is that the functions don't have consistent defaults -- having a region_default that was used everywhere would probably help avoid this, but is probably non-trivial to apply consistently across the code. 

If I recall, the specific issue here is that currently there is no region argument to this version of `abs` and hence acts on the full region, but this isn't wanted as it's operating on a temporary resulting from a differential operation (and hence has invalid guards). Being explicit and saying to abs that we only want to operate on the valid region should avoid the problem (and may act as some sort of in-code documentation of this!).

johnomotani:
David's correct. Adding the region argument to Vector3D::abs is a solution to the issue I had because it allows explicitly using no_bndry. I think the argument for reverting to not passing region arguments to checkData is that doing so might cause similar issues that would throw exceptions in existing code and force people to change it, even though it gives a correct result as long as the guard cells are not used?

bendudson:
ok yes I agree: being consistent is important, and will almost certainly cause fewer nasty surprises later. There are good reasons for the inconsistency though: arithmetic and differential operators are different with respect to their use of guard cells. My main concern was breaking a lot of correctly working code, forcing people to add region arguments.  

Perhaps just making the error message more explicit? Tell users that they can add a region argument, and point to a part of the manual?

d7919:
Yes, it would be good to avoid unnecessarily breaking perfectly fine code. In that regard I think something like the following should be fine (but I'm happy with whatever choice) : just use the default behaviour for checkData in all the arithmetic code (i.e. don't pass on the region given to the operator) but maintain the ability to checkData on other regions so that the user can choose to check this if they want. We could think about using a larger region in the checkData for fields used as input to differential operators.

bendudson:
Yes, happy with this solution for now. Thanks!

dschwoerer:
Would it be easier to store what region is valid within the fields?
That way we would not need to pass a region, if it defaults to the valid region of the currently used field.

That would further potentially speed up the code, as operations are only applied where the field is valid.

When working on two or more fields, we would need a logical and for the regions - but I think that should not be to hard.
Having the field store that information, rather then passing it manually should not make a (big) performance difference, but should be much nicer form a user point of view ...

bendudson:
I've had a go at merging interpolation.cxx, but am not familiar enough with recent changes to be sure. Could someone who knows about this please merge? Sorry for the delay; once this is done I think this is ready to go in.

johnomotani:
In merging, I've kept/reinstated code for interpolation when there is only one guard cell in case we ever implement interpolation with a 3-point stencil. At the moment it can never be reached because we check that xstart/ystart is at least 2. Is it better to delete that code and leave it to be rewritten when/if needed?

bendudson:
Thanks John! I think if the fact that the code can't be reached is documented in some comments, and the reasons explained, then that would be good.

johnomotani:
I hope the code is all OK now. I guess I should also update the manual with: (i) some mention of REGION arguments (is there any description in there yet?); (ii) and the fact that RGN_NOBNDRY avoids communication in interp_to; (iii) is there anything else from this PR that needs writing up?

bendudson:
There is some mention of regions in 
http://bout-dev.readthedocs.io/en/latest/developer_docs/data_types.html#iterating-over-fields
source code: 
https://github.com/boutproject/BOUT-dev/blob/next/manual/sphinx/developer_docs/data_types.rst
A short description there would be good for now, but this probably should be described earlier. Perhaps a section near "Variable Initialisation" under the BOUT++ models section?

johnomotani:
I've added some updates to the manual now. Everyone feel free to add more if you can think of any on region arguments, field locations or staggered grids!

PR #840, opened 2018-02-06
Interp to mesh

dschwoerer:
part of #742 
Needed to overwrite interp_to by aiolos mesh.

PR #921, opened 2018-03-06
Implement interp_to(Field2D)

johnomotani:
Code is mostly copy-paste from interp_to(Field3D).

Also had to add location/getLocation()/setLocation to Field2D.

Added very basic test of interp_to. Haven't actually checked results are correct, so this is just a regression test (for now?).

Question: is this actually something we want to do now, or is there lots more stuff that should be done to enable Field2D to have a location, so this should wait for a complete job?

This PR includes #920 so that interp_to has a region argument.


src/mesh/index_derivs.cxx
```diff
@@ -1196,7 +1196,7 @@ const Field3D Mesh::indexDDX(const Field3D &f, CELL_LOC outloc, DIFF_METHOD meth
 
       } else if (inloc != CELL_CENTRE) {
         // Interpolate then (centre -> centre) then interpolate
-        return DDX(interp_to(f, CELL_CENTRE), outloc, method);
+        return DDX(interp_to(f, CELL_CENTRE, RGN_NOY), outloc, method);
```
d7919: This looks like it is changing the standard behaviour, is this the case?

src/mesh/index_derivs.cxx
```diff
@@ -1211,7 +1211,7 @@ const Field3D Mesh::indexDDX(const Field3D &f, CELL_LOC outloc, DIFF_METHOD meth
   result = applyXdiff(f, func, diffloc);
   result.setLocation(diffloc); // Set the result location
 
-  result = interp_to(result, outloc); // Interpolate if necessary
+  result = interp_to(result, outloc, RGN_NOBNDRY); // Interpolate if necessary
```
d7919: Why is this nobndry but others are noy/nox?

src/mesh/interpolation.cxx
```diff
@@ -178,10 +186,11 @@ const Field3D interp_to(const Field3D &var, CELL_LOC loc)
 	  }
 	}
 	break;
-	// Need to communicate in Y
       }
       case CELL_ZLOW: {
-        for(const auto &i : result.region(RGN_NOZ)) {
+        REGION this_region = region;
+        if ( region==RGN_ALL ) this_region=RGN_NOZ;
```
d7919: This looks a little suspicious -- we're overriding an input argument for one specific case. I think this needs a comment to explain why there's special handling here.

src/mesh/interpolation.cxx
```diff
@@ -44,12 +44,16 @@ BoutReal interp(const stencil &s)
...
+  NOTE: Since corner guard cells cannot be communicated, it never makes sense
+  to calculate interpolation in guard cells. If guard cell values are required,
+  we must communicate (unless interpolating in z).
 
   @param[in]   var  Input variable
   @param[in]   loc  Location of output values
+  @param[in]   region  Region where output will be calculated
 */
-const Field3D interp_to(const Field3D &var, CELL_LOC loc)
+const Field3D interp_to(const Field3D &var, CELL_LOC loc, REGION region)
```
d7919: The region argument is ignored in most cases which might be a bit unexpected. A comment to note this may help.

src/mesh/interpolation.cxx
```diff
@@ -222,8 +228,134 @@ const Field3D interp_to(const Field3D &var, CELL_LOC loc)
...
+  if(mesh->StaggerGrids && loc != CELL_ZLOW && (var.getLocation() != loc)) {
+
+    // Staggered grids enabled, and need to perform interpolation
+    TRACE("Interpolating %s -> %s", strLocation(var.getLocation()), strLocation(loc));
+
+    Field2D result(var.getMesh());
+
+    if (region != RGN_NOBNDRY) {
+      // result is requested in some boundary region(s)
+      result = var; // NOTE: This is just for boundaries. FIX!
```
d7919: Is now a good time to address the "fix" mentioned in the comments?

src/mesh/interpolation.cxx
```diff
@@ -222,8 +228,134 @@ const Field3D interp_to(const Field3D &var, CELL_LOC loc)
...
+            }
+            
+            result[i] = interp(s);
+          }
+        }
+	break;
+      }
+      default: {
+	// This should never happen
+	throw BoutException("Don't know what to do");
```
d7919: Could this message be more explicit, e.g. unknown stagger direction in interp_to or similar?

src/mesh/interpolation.cxx
```diff
@@ -222,8 +228,134 @@ const Field3D interp_to(const Field3D &var, CELL_LOC loc)
...
+      // For now, shift to centre then to shifted
+      result = interp_to( interp_to(var, CELL_CENTRE) , loc, region);
+    }
+    result.setLocation(loc);
+
+    return result;
+  }
+  
+  // Nothing to do - just return unchanged
   return var;
```
d7919: I think we may still want to copy var into result. As it stands returning `var` may increase the number of references to the `var` data whilst returning `result` doesn't. This may or may not be an important difference.

src/sys/derivs.cxx
```diff
@@ -210,7 +210,7 @@ const Field3D D2DY2(const Field3D &f, CELL_LOC outloc, DIFF_METHOD method) {
     result += f.getMesh()->coordinates()->d1_dy * f.getMesh()->indexDDY(f, outloc, DIFF_DEFAULT) / f.getMesh()->coordinates()->dy;
   }
 
-  return interp_to(result, outloc);
+  return result;
```
d7919: Is there a reason this interpolation isn't needed?

tests/integrated/test-interp_to/README.md
```diff
@@ -0,0 +1,7 @@
+test-interp_to
```
d7919: I wonder if this may be better as a unit test?

src/mesh/index_derivs.cxx
johnomotani: I didn't think it changed anything, because I thought the derivative invalidates all the guard cells anyway, so as long as its input has y-guard cells the result should be the same. Is DDX ever expected to give correct results in guard cells?

src/mesh/index_derivs.cxx
johnomotani: I made it nobndry because I don't think the result needs valid guard cells. Inputs to derivatives do need guard cells at least in the direction of the derivative, hence the nox/noy.

Looking again, I think this interp_to couldn't ever be correct unless it's only interpolating in the z-direction, because the interpolation would need guard cells, so the result would have to be communicated and have a boundary condition applied. It might be better to raise an error for unhandled combinations of inloc/outloc instead?

src/mesh/interpolation.cxx
johnomotani: This might well just be wrong. I meant to keep the existing behaviour with RGN_NOZ in the 'default' case. Looking again I think I'd favour just passing the region argument (removing the special handling).

This can't ever have been used, because until this PR result.region(RGN_NOZ) would have thrown an exception as it was not implemented in Field3D. So I think it's safe to change the previous behaviour.

src/mesh/interpolation.cxx
johnomotani: Agreed, I'll add something tomorrow.

The cases except RGN_NOBNDRY are all basically the same because we always need to communicate to get valid guard cells: the corner guard cells aren't ever set, so we can't calculate the interpolation in guard cells. Also, the communicate() goes in both x- and y-directions, so needs to be called for any of nox/noy/all.
If there were 1d communication routines, then e.g. RGN_NOX could communicate just y-guard cells. Maybe that would offer a small speedup?

src/mesh/interpolation.cxx
johnomotani: Is there an obvious fix? I think the only correct thing to do is to apply a boundary condition after interpolating. I don't think the input Field3D knows its boundary conditions at the moment? If it did maybe we could try to do something clever here, but otherwise I think we just have to define the guard cells of the result of interpolation as invalid, and make the user apply a boundary condition if they need one.

tests/integrated/test-interp_to/README.md
johnomotani: Could well be, but I'm afraid I haven't learned to write unit tests yet... Any offers to write one welcome :wink:

dschwoerer:
I would advise against this PR as it interferes with #840 #815 #641 
Further it is causing bugs with the staggered derivatives, as Field2D can have a location - but the geometry is not aware of this. This needs a better fix - I will have a look at this, but if some of the open PRs are merged, I would like to start working on new issues (including the one with staggered derivatives/Field2d) ....

tests/integrated/test-interp_to/README.md
d7919: I can have a look, but may not get chance for a while. The existing unit tests give some examples that can usually be modified as a starting point. I'll ping @ZedThree as well as he may be able to help.

johnomotani:
@dschwoerer ah, good point. Agree it's best to leave this for now and revisit after those other PRs are merged.

PR #924, opened 2018-03-09
Arithmetic performance test

bendudson:
Replaces #842
* Changed the update macro to TIMEIT, which calculates both start and end times. 
* Applied clang-format


d7919:
It might be helpful for us to have a similar test for {3D,2D} and {2D,2D}.

bendudson:
Replaced by #842

PR #926, opened 2018-03-09
Add performance/arithmetic_3d2d case

d7919:
Used for testing arithmetic performance in case of mixed field types.

Comparing to standard arithmetic test seems to indicate that the "field" approach is slower when mixing field types.


bendudson:
Is the code generator flexible enough to use different methods for mixed fields? Perhaps single data iterator for 3D/3D and DataIterator for 3D/2D?

d7919:
That's certainly should be possible, #925 also helps a bit.

PR #927, opened 2018-03-09
Allow runtime control of maximum size of contiguous block to use in Region

d7919:
Adds `mesh:maxregionblocksize` input option. Defaults to whatever is
defined by MAXREGIONBLOCKSIZE (i.e. defaults to 64).

Also adds an example in examples/performance/tuning_regionblocksize that
shows timing information for `r = a + b` using a BLOCK_REGION_LOOP and
using different block sizes.

This could be used as a way of picking a good value of the default
block size for a particular machine/configuration.

### DIFF
--- a/include/bout/mesh.hxx
+++ b/include/bout/mesh.hxx
@@ -568,8 +568,9 @@ class Mesh {
    */
   void setParallelTransform();
 
-    //Region related routines
-
+  //Region related routines
+  int maxregionblocksize;
+  
   /// Get the named region from the region_map for the data iterator
   ///
   /// Throws if region_name not found
--- a/include/bout/mesh.hxx
+++ b/include/bout/mesh.hxx
@@ -568,7 +568,13 @@ class Mesh {
    */
   void setParallelTransform();
 
-  //Region related routines
+  /////////////////////////
+  // Region related routines
+  /////////////////////////
+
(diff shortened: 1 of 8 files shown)


d7919:
This could be helpful for tuning the runtime performance. It should also be possible to use this to, for example, make the inner vectorised loop correspond to the z dimension and the outer OpenMP loop be over x-y points (by setting the maxregionblocksize to LocalNz). This might be helpful in some places, for example:

~~~~
Field3D field = 1.0;
Field2D zAvgField = 0.;
const auto blocks = region_all.blocks();
BOUT_OMP(parallel for)
for (auto block = blocks.begin(); block < blocks.end() ; ++block){
     BoutReal zAvg = 0;
     for ( auto index = block->first ; index < block->second; ++index){
            zAvg += field[index];
     }
    zAvg /= mesh->LocalNz;
    zAvgField[block->first] = zAvg;
}
~~~~

include/bout/mesh.hxx
```diff
@@ -568,8 +568,9 @@ class Mesh {
    */
   void setParallelTransform();
 
-    //Region related routines
-
+  //Region related routines
+  int maxregionblocksize;
```
ZedThree: Could do with a short doxygen comment here, and some separation from the existing comment

PR #929, opened 2018-03-12
Update field2d.cxx

### LINKED ISSUE #923: Bug in Field2D with TRACK
https://github.com/boutproject/BOUT-dev/blob/c1856a3d73571748bf61def0e3d0c7a72cad65ad/src/field/field2d.cxx#L80-L82
This should be ``name = f.name`` I think?
Compilation fails if tracking is enabled with ``--enable-track``: the Travis builds must not ever enable this option? Maybe we could add to the debugging build?

d7919:
Fixes #923

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,7 +75,7 @@ matrix:
           - liblapack-dev
           - libparpack2-dev
     env:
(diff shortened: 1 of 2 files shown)


ZedThree:
Please could you also add `--enable-track` to the "maximal" coverage build?

d7919:
Added track to coverage now.

Issue #933, opened 2018-03-12
Matrix/Tensor slower relative to raw pointers


d7919:
Recent profiling by @JosephThomasParker has identified a slowdown in cyclic solves. This seems to be due to the replacement of raw pointer blocks with Array backed `Matrix`. 

It seems that it's the `ensureUnique` calls in `Matrix` indexing operators that is responsible for most of this. This is being used to allow the `Array` copy-on-write mechanics to be applied to `Matrix`, but it seems the overhead in this is significant.

Removing this COW behaviour (e.g. https://github.com/boutproject/BOUT-dev/tree/remove_copyonwrite_matrix_tensor) speeds up the cyclic solve significantly (by 50% for the interchange-instability test case on a local machine).

d7919:
Should be fixed with #935 and #969

PR #935, opened 2018-03-13
Remove copy on write from matrix/tensor

d7919:
Matrix and Tensor are wrappers around Array, which provides reference counting to enable copy-on-write mechanics (as used in the Fields).

This is currently also used to provide CoW for Matrix and Tensor, however the implementation for Matrix/Tensor seems to have non-negligible overhead so this PR disables CoW here.

Addresses #933


bendudson:
Thanks for hunting down this performance hit!

Can we just remove the "data.ensureUnique();" calls from the indexing operators, without adding it to the copy constructor and assignment operator? This is what is done for the Fields, and is consistent with Numpy's behaviour so should not be very surprising: The user must call ensureUnique() once before modifying elements directly.

d7919:
That should be possible (as long as we add a call for this at the appropriate level). I did consider this but thought it may complicate the code compared to the original form with raw pointers. 

I don't think there are many (any) places where we actually do copy Matrix/Tensor at the moment beyond the tests so either way probably won't require much change to the code.

dschwoerer:
I think copy by default should be fine.
Unlike to python we can pass const references - which avoids the copy, and ensures that the data is not changed.
I would be in favour of changing the field3d/field2d as well, as now there are move operators which avoids the copying of fields.
Tracking down issues where a copy of a field modifies the original data are really annoying (also in python) and C++ gives as the tools to get good performance and safety - so I am in favour of using that.

Issue #820, opened 2018-01-31
Staggered grids meta-issue


ZedThree:
There are currently a few problems with using staggered grids (e.g. #64, #259, #411, #741, #810). It's not clear (to me, at least) if these are related issues and staggered grids are completely broken, or if they just need a bit of polish and tightening up.

We're also missing decent tests for them. There is `examples/staggered_grid`, but it doesn't really tell us if staggered grids are working correctly or not. First task, then, should be to get some test coverage that highlights any current problems and ensures that they don't creep back in.

Once we know what's a bug and what's a design issue, we can work on fixing everything.

dschwoerer:
#742 comes with some tests, and avoids some of the issues, e.g. #810 #741 and #259 

It further has tests of the derivative, to make sure they converge with the expected accuracy.

ZedThree:
:+1: Do you want to see if you can pull out as many of the tests as you can into a separate PR?

dschwoerer:
No.

Long version:
Is there no interest to merge #742 ?
That takes time to separate.
Pulling out causes further merge conflicts, wasting even more time I would prefer to spend my time otherwise.
Unless #742 is rejected, I cannot see an advantage.

ZedThree:
Yes.

Long version: no, not in its current state:

1) It currently modifies 16k lines across 67 files and 109 commits -- that's a lot to review and understand! It's no good merging something if only one person understands the code and can fix problems with it. What happens when that person moves on to other projects? See technical debt/bus factor.

2) It introduces several unrelated features/bug-fixes -- these would be useful as standalone PRs. PRs that introduce discrete (or a set of related) features/bug fixes are easier to review, easier to understand, and are more likely to get merged faster.

3) We have other priorities -- none of the maintainers actually gets to spend all of their time on BOUT++, and some of us are contractually obliged to work on particular things. While we do want to merge the Aiolos mesh (and the Cython stuff too), we, like you, want to spend our limited time on the projects we ourselves are working on. Bigger PRs mean we have to spend more time understanding them, and get less time to work on our own projects.

Honestly, we do want to merge #742 and #641, but they are *big* PRs. Pulling out unrelated features means that those features get in faster, and we'll end up with a smaller, easier to understand PR.

PR #937, opened 2018-03-14
Newfield

dschwoerer:
Fix for #64 

This allows that the metric fields are at different locations, thus dividing a `Field?D f` by `dx` does different things, depending on where `f` is.
It should be mostly non-visible to the user, unless `staggerGrids` are used (not just enabled).
I needed to increase the number of guard cells for the MMS tests, but that shouldn't be a real issue, as I assume if staggerGrids are used, interp_to is needed somewhere in the code.

This tries to make everything smooth and hidden, without causing any significant performance hits in case the fields in the staggered locations are not used, that why they are only generated on the fly.

Note, this is currently not working with clang. I cannot understand why. The code seems to be right, so might be a bug in the compiler, but I don't know what is going on, as debugging with gdb did not work for me ...


include/flexible.hxx
```diff
@@ -0,0 +1,448 @@
...
+  // FieldData stuff
+  virtual void accept(FieldVisitor &v){
+    //#warning // Using workaround for const
+    ((F*)fields[mainid])->accept(v);
+  }
+  virtual bool isReal() const{
+    return fields[mainid]->isReal();
+  }
+  virtual bool is3D() const {
+    return fields[mainid]->isReal();
```
d7919: Should this be `is3D`?

dschwoerer: thanks, fixed

include/flexible.hxx
```diff
@@ -0,0 +1,448 @@
...
+  }
+  F & getNonConst(CELL_LOC loc){
+    return *((F*)geti(loc));
+  }
+  /// Get a const reference of the field at the specific location. If
+  /// the CELL_LOC is CELL_DEFAULT the mainlocation will be returned.
+  const F & get(CELL_LOC loc){
+    return * geti(loc);
+  };
+  Flexible<F> & operator=(const F& f) {
```
d7919: I think this might be one reason your clang build fails (possibly) -- you're not actually returning anything from this routine (and other nearby ones) despite the signature saying that you will. I've seen this cause issues with clang elsewhere.

dschwoerer: Thanks, that helped indeed :+1:

include/flexible.hxx
```diff
@@ -0,0 +1,464 @@
...
+    ((F*)fields[mainid])->operator-=(rhs);
+    clean(false);
+    return *this;
+  };
+
+private:
+  // Helper function to get index of location.
+  uint getId(CELL_LOC loc_){
+    uint loc = static_cast<uint>(loc_)-1;
+    if ( loc > num_fields || loc < 0){
```
ZedThree: `loc` can never be negative

include/flexible.hxx
```diff
@@ -0,0 +1,464 @@
...
+  Flexible<F>& operator-=(BoutReal rhs) {
+    ((F*)fields[mainid])->operator-=(rhs);
+    clean(false);
+    return *this;
+  };
+
+private:
+  // Helper function to get index of location.
+  uint getId(CELL_LOC loc_){
+    uint loc = static_cast<uint>(loc_)-1;
```
ZedThree: When `loc_` is 0, this will wrap-around

include/flexible.hxx
```diff
@@ -0,0 +1,464 @@
...
+    if ( loc > num_fields || loc < 0){
+      throw BoutException("Unexpeted Fieldlocation! (Info: I got %d - %s)",loc,strLocation(loc_));
```
ZedThree: Typo: "Unexpeted" -> "Unexpected"

include/flexible.hxx
```diff
@@ -0,0 +1,464 @@
...
+	owner[mainid]=true;
+      }
+    }
+    ASSERT1(fields[mainid]!=nullptr);
+    return fields[loc];
+  };
+  // Number of field locations we support
+  static const uint num_fields=4;
+  // The pointers to the fields. Some may be null
+  const F * fields[num_fields];
```
ZedThree: Why the `const` here? Is `Flexible<F> x` not supposed to modify its fields? I assume this being `const` is why there's casts everywhere.

include/flexible.hxx
```diff
@@ -0,0 +1,464 @@
...
+        if (fields[i] != nullptr){
+	  if (owner[i]){
+	    delete fields[i];
+	  }
+	  fields[i]=nullptr;
+        }
+      }
+    }
+  };
+  const F * geti(CELL_LOC loc_) {
```
ZedThree: This is a very confusing name, what does it mean?

include/flexible.hxx
```diff
@@ -0,0 +1,464 @@
...
+    if (loc_ == CELL_DEFAULT){
+      return fields[mainid];
+    }
+    uint loc=getId(loc_);
+    if (fields[loc] == nullptr){
+      if (fields[0] == nullptr){
+	fields[0]=new F(interp_to((*fields[mainid]),CELL_CENTRE));
```
ZedThree: Why does this always put it in `fields[0]`?

include/flexible.hxx
```diff
@@ -0,0 +1,464 @@
...
+
+#pragma once
+
+#ifndef __FLEXIBLE_H__
+#define __FLEXIBLE_H__
+
+#include <bout_types.hxx>
+#include <bout/deprecated.hxx>
+//class Field2D;
+//class Field3D;
```
ZedThree: Can be removed?

src/field/gen_fieldops.py
```diff
@@ -164,17 +166,26 @@ def returnType(f1, f2):
...
     field2D = Field('Field2D', ['x', 'y'], index_var=index_var)
+    fieldperp = Field('FieldPerp', ['x', 'z'], index_var=index_var)
     boutreal = Field('BoutReal', [], index_var=index_var)
+else:
+    field3D = Field('Field3D', ['x', 'y', 'z'])
+    field2D = Field('Field2D', ['x', 'y'])
+    fieldperp = Field('FieldPerp', ['x', 'z'])
+    boutreal = Field('BoutReal', [])
+
+if __name__ == "__main__":
```
ZedThree: This looks like a copy-paste error

src/field/gen_fieldops.py
```diff
@@ -186,6 +197,10 @@ def returnType(f1, f2):
 
     template = env.get_template("gen_fieldops.jinja")
 
+    # Declare what fields we currently support:
+    # Field perp is currently missing
+    fields = [field3D, field2D, boutreal]
```
ZedThree: Copy-paste error?

src/mesh/impls/bout/boutmesh.cxx
```diff
@@ -1868,20 +1868,22 @@ int BoutMesh::pack_data(const vector<FieldData *> &var_list, int xge, int xlt, i
...
-      for (int jx = xge; jx != xlt; jx++) {
-        for (int jy = yge; jy < ylt; jy++) {
-          for (int jz = 0; jz < LocalNz; jz++, len++) {
-            buffer[len] = (*static_cast<Field3D *>(var))(jx, jy, jz);
+      ASSERT2(var->isAllocated());
+      Indices j;
+      for (j.x = xge; j.x != xlt; j.x++) {
+        for (j.y = yge; j.y < ylt; j.y++) {
+          for (j.z = 0; j.z < LocalNz; j.z++, len++) {
+            buffer[len] = var->operator[](j);
```
ZedThree: Why `var->operator[](j)` here and `(*var)[j]` below?

tests/MMS/wave-1d/wave.cxx
```diff
@@ -62,7 +62,11 @@ class Wave1D : public PhysicsModel {
     
     /*this assumes equidistant grid*/
     int nguard = mesh->xstart;
-    coord->dx = Lx/(mesh->GlobalNx - 2*nguard);
+
+    Field2D dx=Lx/(mesh->GlobalNx - 2*nguard);
+    coord->dx = dx;
+    dx.setLocation(CELL_XLOW);
+    coord->dx = dx;
```
ZedThree: Why the assignment both before and after `dx.setLocation()`?

src/mesh/interpolation.cxx
```diff
@@ -222,14 +222,131 @@ const Field3D interp_to(const Field3D &var, CELL_LOC loc)
...
+        break;
+
+      default: {
+        // This should never happen
+        throw BoutException("Don't know what to do");
+      }
+      };
+
+      if(dir != CELL_ZLOW) {
+        // COMMUNICATION
```
ZedThree: These comments can both go

src/mesh/interpolation.cxx
```diff
@@ -222,14 +222,131 @@ const Field3D interp_to(const Field3D &var, CELL_LOC loc)
...
+              // Stencil centred around a cell centre
+              s.mm = s.m;
+              s.m  = s.c;
+            }
+
+            result[i] = interp(s);
+          }
+        } else {
+          // Only one guard cell, so no pp or mm values
+          throw BoutException("Not enough boundary cells!");
```
ZedThree: Error message could be more helpful

include/vector2d.hxx
```diff
@@ -150,6 +150,12 @@ class Vector2D : public FieldData {
   }
   void applyBoundary(const char* condition) { applyBoundary(string(condition)); }
   void applyTDerivBoundary() override;
+  virtual inline const BoutReal& operator[](const Indices &i) const {
+    throw BoutException("Indices cannot be handled ...");
```
ZedThree: I'm not sure I see the point of this. Why add this function, only to get a run-time error, rather than leave it out and get a compile-time error?

dschwoerer: Short: This function is not called, but otherwise Vector?D is virtual.

Long:
Because it is a FieldData, which it requires to set the boundaries.
FieldData is however also used to communicate.
In order to communicate the flexible field, i needed direct access, as the old solution with casting did not work anymore, as there are now 2 options that return `is3D() == true`
But as I did not want to move the boundaries stuff to Field - I added the `operator[](const Indices)` here, so that the Vector can still have the boundary stuff, but as the access is not used (they are split into the fields for communication) I added them, so that BOUT++ compiles.

I guess a nicer solution would be to add stuff that all Fields should support to the Field class, but that would have been more work.

src/mesh/impls/bout/boutmesh.cxx
dschwoerer: No real reason. Do you have any preference?

src/field/gen_fieldops.py
dschwoerer: No - this is intended.
I want to use this Field object also for the flexible field.
This I want to have them set, even in the case of not being `__main__`

src/field/gen_fieldops.py
dschwoerer: No, see above

include/flexible.hxx
dschwoerer: Should I add some extra checks to prevent this underflow - or are we happy if I test in the line below for underflow?

include/flexible.hxx
dschwoerer: I thought it would be safer to keep const pointers and not modify the fields - but it turns out that does not work so well.
I have changed it to non-const ...

include/flexible.hxx
dschwoerer: Function has been removed, work is now done in `getNonConst(CELL_LOC)`

include/flexible.hxx
dschwoerer: I added a comment.
0 is assumed to be CELL_CENTRE.
If that should change - it needs to be fixed ...

include/flexible.hxx
dschwoerer: yes, removed.

tests/MMS/wave-1d/wave.cxx
dschwoerer: I added some comments in the file.
This way several locations can be set.

include/vector2d.hxx
dschwoerer: I thought of providing a default error message if it is the operator is not provided - but I think this messages might be more helpful, and further it might  prevent from accidentally not declaring this method ...

src/mesh/interpolation.cxx
dschwoerer: ok, removed.
They where just copied from the Field3D ... I will remove them also there ...

dschwoerer:
@ZedThree on the more general issue - I assume you are refering to #820 
Some of the things are fixed in the aiolos mesh branch - but I haven't fixed them in the bout mesh - as I didn't like the copy-pasting approach used.
I think they are all (mostly) unconnected, so I doubt that there is *the one magic think* that can fix them all.
The one thing that all connects them is tests - so you are right that they are strongly needed for this PR.
Before spending time on the tests, I wanted to ask for feedback, to know whether there are any general objections to this approach - so if you (read: everybody) is (somewhat) happy with this I will add them ...

include/flexible.hxx
dschwoerer: thanks, fixed

include/flexible.hxx
dschwoerer: thanks, fixed

dschwoerer:
* Now with close to 100% unit test coverage.
* More documentation

dschwoerer:
Replaced by #1060 - thanks @johnomotani

PR #938, opened 2018-03-14
Tweak travis script

dschwoerer:
Just some minor changes:
* Pass target as argument
* Move argument checking to the beginning
* Fail if we get an unexpected argument


d7919:
Thanks!

PR #940, opened 2018-03-14
Interp to fixes

dschwoerer:
Small changes (and fixes) for interp_to(Field3D)

Edit:
* Fixes the interpolation in the Z direction
* Make sure we have enough boundary cells


dschwoerer:
Many of the changes are just from the white-space change 9c1e511
Just looking at the other commits should make it much easier to review the changes ...

dschwoerer:
Test is in #956

src/mesh/interpolation.cxx
```diff
@@ -63,161 +64,179 @@ const Field3D interp_to(const Field3D &var, CELL_LOC loc)
...
+
+          result[i] = interp(s);
+        }
+        break;
       }
       default: {
-	// This should never happen
-	throw BoutException("Don't know what to do");
+        // This should never happen
+        throw BoutException("Unsupported method of interpolation\n"
```
ZedThree: "method of interpolation" -> "interpolation location"

src/mesh/interpolation.cxx
```diff
@@ -63,161 +64,179 @@ const Field3D interp_to(const Field3D &var, CELL_LOC loc)
...
+      // For now, shift to centre then to final location loc
+
       result = interp_to( interp_to(var, CELL_CENTRE) , loc);
     }
     result.setLocation(loc);
 
     return result;
   }
-  
+  if ((loc != CELL_CENTRE && loc != CELL_DEFAULT) && (mesh->StaggerGrids == false)) {
```
ZedThree: Move this to be at the top of the function. It's better to put this kinds of input argument validation early on, so the reader can see what assumptions the function is making, what is invalid input, etc.

tests/integrated/test-drift-instability/2fluid.cxx
```diff
@@ -292,6 +294,8 @@ int physics_init(bool restarting) {
   dump.add(rho_s, "rho_s", 0);
   dump.add(wci,   "wci", 0);
 
+  ylow=mesh->StaggerGrids?CELL_YLOW:CELL_CENTRE;
```
ZedThree: Some whitespace please! Making code easy to read for the next person is really important. I know I bang on about clang-format a lot, but it really means you don't need to think about formatting -- let the computer do it for you.

Also, instead of `ylow`, consider something a bit more descriptive: `cell_location` for instance. Naming things is hard, but it's worth spending a bit of time on it.
What does this variable really represent? It's value is only possibly `CELL_YLOW`, so when we see it used later on, seeing `ylow` might make us think that it is always `CELL_YLOW`.

ZedThree:
Also, a tip for ignoring whitespace changes: add `?w=1` after the url:

https://github.com/boutproject/BOUT-dev/pull/940/files?w=1

Quite useful :)

src/mesh/interpolation.cxx
```diff
@@ -63,161 +64,179 @@ const Field3D interp_to(const Field3D &var, CELL_LOC loc)
...
+            // Producing a stencil centred around a lower Z value
+            s.pp = s.p;
+            s.p  = s.c;
+          } else if(location == CELL_ZLOW) {
+            // Stencil centred around a cell centre
+            s.mm = s.m;
+            s.m  = s.c;
+          } else {
+            // We should never be here
+            throw BoutException("Please report this bug!");
```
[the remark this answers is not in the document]
dschwoerer: I have no clue why it would be there.

I tried to change the code as little as possible - but if it does take nether of the first two cases, something is wrong.

If the code would get there, that would need some diagnoses, on why we did get here ...

tests/integrated/test-drift-instability/2fluid.cxx
dschwoerer: I am not certain about `cell_location` - as some cells are always going to be at `CELL_CENTRE`
It is not worse then before - `CELL_YLOW` implied even stronger to be at `CELL_YLOW`.

`maybe_ylow`? `someone_tried_to_add_stagger_support_but_did_not_finish`? `maybe_in_the_future_ylow`?
If you want the code to be clean, it should be removed ...

src/mesh/interpolation.cxx
```diff
@@ -63,161 +67,142 @@ const Field3D interp_to(const Field3D &var, CELL_LOC loc)
...
+              // Producing a stencil centred around a lower Y value
+              s.pp = s.p;
+              s.p  = s.c;
+            } else if(location == CELL_YLOW) {
+              // Stencil centred around a cell centre
+              s.mm = s.m;
+              s.m  = s.c;
+            } else {
+              // We should never be here
+              throw BoutException("Maybe a butterfly defracted an X-ray?");
```
johnomotani: True it would be a bug to end up here, but a more informative exception message might help fix that bug if it happens. Maybe something like 'Inconsistent values of location and loc, here one should be CELL_CENTRE and the other should be CELL_YLOW'?

src/mesh/interpolation.cxx
```diff
@@ -63,161 +67,142 @@ const Field3D interp_to(const Field3D &var, CELL_LOC loc)
...
+            // Producing a stencil centred around a lower Z value
+            s.pp = s.p;
+            s.p  = s.c;
+          } else if(location == CELL_ZLOW) {
+            // Stencil centred around a cell centre
+            s.mm = s.m;
+            s.m  = s.c;
+          } else {
+            // We should never be here
+            throw BoutException("Those damn butterflies again!");
```
johnomotani: Again, maybe  'Inconsistent values of location and loc, here one should be CELL_CENTRE and the other should be CELL_ZLOW'?

src/mesh/interpolation.cxx
dschwoerer: @johnomotani better?

ZedThree: @dschwoerer Not really.

This if-statement can just be simplified to:
```cpp
            if (location == CELL_CENTRE) {
              // Producing a stencil centred around a lower Y value
              s.pp = s.p;
              s.p  = s.c;
            } else {
              // Stencil centred around a cell centre
              s.mm = s.m;
              s.m  = s.c;
            }
```
and similarly for the other one. The original code was overly cautious. As you noticed, there is no way, short of bit-flipping, for the code to reach that exception -- therefore it can be removed. Tidying up the other conditionals as above makes it clearer there is no third option.

ZedThree:
I'm on the fence about e917828. It's a nice solution to wanting both versions optionally in the test suite, on the other hand, how likely are we to actually want both?

Also, does the non-staggered version of drift-instability work? If so, e917828 is probably fine; if not, let's just drop it completely.

bendudson:
@ZedThree I think it depends if test-drift-instability is just a test, or if it has a use as an example of using staggered grids. It it has use as an example as well as a test (which I think it does), then the code should be tidied up (there is a lot of unused/redundant code in there). I agree both staggered and non-staggered versions should work, and that having a semi-realistic test case is useful.

PR #943, opened 2018-03-15
Skip Multigrid test with mpich

dschwoerer:
The issue is that the multigrid test is extremly slow with mpich and few cores.
It took me ~30 mins to run it - that is why the latest cython branch tests all timed out.

### DIFF
--- a/tests/integrated/test-multigrid_laplace/runtest
+++ b/tests/integrated/test-multigrid_laplace/runtest
@@ -19,6 +19,11 @@ from sys import exit
 
 MPIRUN = getmpirun()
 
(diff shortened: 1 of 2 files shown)


ZedThree:
I'm not sure I see the benefit of this. It adds complexity to the tests for a very specific combination of MPI implementation and environment. If a user run this on a physical machine with lots of cores using MPICH, skipping one test will be very unexpected.

Why is MPICH important for Cython that would make this necessary?

dschwoerer:
And I thought adding a general way to skip tests would allow to add test that require some libraries, and skip them otherwise. Currently we can never run these tests - if the libraries required are not core requirements.

PR #945, opened 2018-03-16
Fixing many bugs in mumps_laplace

johnomotani:
Use of Array instead of C++-style arrays of BoutReal.
()-indexing of FieldPerps.
Remove 'FieldPerp sol' as a member variable, it only needs to be a local
variable in the solve() method.
Use a local Mesh* called localmesh instead of the global mesh pointer.

I think these changes are good/useful (some of the Array stuff is copied from what @d7919 updated, taken from the configure-fixes branch), but I could only get mumps_laplace.cxx to compile, not to link with mumps so I haven't been able to test. Does anyone have a working installation of MUMPS that links?

### DIFF
--- a/src/invert/laplace/impls/mumps/mumps_laplace.cxx
+++ b/src/invert/laplace/impls/mumps/mumps_laplace.cxx
@@ -34,10 +34,11 @@
 #include <msg_stack.hxx>
 #include <cmath>
 
-LaplaceMumps::LaplaceMumps(Options *opt) : 
+LaplaceMumps::LaplaceMumps(Options *opt, Mesh *passmesh) : 
   Laplacian(opt),
   A(0.0), C1(1.0), C2(1.0), D(1.0), Ex(0.0), Ez(0.0),
-  issetD(false), issetC(false), issetE(false)
+  issetD(false), issetC(false), issetE(false),
+  localmesh(passmesh)
 {
   // Get Options in Laplace Section
   if (!opt) opts = Options::getRoot()->getSection("laplace");
@@ -58,21 +59,21 @@ LaplaceMumps::LaplaceMumps(Options *opt) :
     if (outer_boundary_flags & ~implemented_boundary_flags) {
       throw BoutException("Attempted to set Laplacian inversion boundary condition flag that is not implemented in mumps_laplace.cxx");
     }
-    if(mesh->periodicX) {
-        throw BoutException("LaplaceMumps does not work with periodicity in the x direction (mesh->PeriodicX == true). Change boundary conditions or use serial-tri or cyclic solver instead");
+    if(localmesh->periodicX) {
+        throw BoutException("LaplaceMumps does not work with periodicity in the x direction (localmesh->PeriodicX == true). Change boundary conditions or use serial-tri or cyclic solver instead");
       }
   #endif
 
   // Get communicator for group of processors in X - all points in z-x plane for fixed y.
-  comm = mesh->getXcomm();
+  comm = localmesh->getXcomm();
   
   // Need to determine local size to use based on prior parallelisation
   // Coefficient values are stored only on local processors.
-  localN = (mesh->xend - mesh->xstart + 1) * (mesh->LocalNz);
-  if(mesh->firstX())
-    localN += mesh->xstart * (mesh->LocalNz);    // If on first processor add on width of boundary region
-  if(mesh->lastX())
-    localN += mesh->xstart * (mesh->LocalNz);    // If on last processor add on width of boundary region
+  localN = (localmesh->xend - localmesh->xstart + 1) * (localmesh->LocalNz);
+  if(localmesh->firstX())
+    localN += localmesh->xstart * (localmesh->LocalNz);    // If on first processor add on width of boundary region
+  if(localmesh->lastX())
+    localN += localmesh->xstart * (localmesh->LocalNz);    // If on last processor add on width of boundary region
   
 
   // Calculate total number of points in physical grid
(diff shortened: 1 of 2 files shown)


src/invert/laplace/impls/mumps/mumps_laplace.cxx
```diff
@@ -34,10 +34,11 @@
 #include <msg_stack.hxx>
 #include <cmath>
 
-LaplaceMumps::LaplaceMumps(Options *opt) : 
+LaplaceMumps::LaplaceMumps(Options *opt, Mesh *passmesh) :
```
ZedThree: `passmesh` can just be `localmesh`, there's no name conflict when used in a initialiser list like here.

ZedThree:
There's a lot of commented out code that can presumably just be deleted.

I'm struggling to find anywhere that has MUMPS configured how we expect it to be. BLACS is now usually bundled with ScaLAPACK, so we probably don't need to link BLACS explicitly. On the other hand, most distros that package MUMPS with PTSCOTCH (which we explicitly look for) rename the library to be something like `libmumps_ptscotch.{a,so}`, which we don't look for.

MUMPS doesn't distribute any kind of help for configuring with it, so we might just have to rely on the user passing the correct flags.

The alternative is to just drop support for MUMPS. We could just use it through PETSc.

src/invert/laplace/impls/mumps/mumps_laplace.cxx
```diff
@@ -173,58 +172,58 @@ LaplaceMumps::LaplaceMumps(Options *opt) :
...
 //     
-//     rhs = new BoutReal[meshx*mesh->LocalNy*mesh->LocalNz];
-//     rhs_slice = new BoutReal[meshx*mesh->LocalNz];
+//     rhs = new BoutReal[meshx*localmesh->LocalNy*localmesh->LocalNz];
+//     rhs_slice = new BoutReal[meshx*localmesh->LocalNz];
 //   }
-  localrhssize = (mesh->xend-mesh->xstart+1)*(mesh->LocalNz);
-  if (mesh->lastX()) {
-    localrhssize += (mesh->LocalNx-mesh->xend-1)*(mesh->LocalNz);
+  localrhssize = (localmesh->xend-localmesh->xstart+1)*(localmesh->LocalNz);
```
d7919: Could you add a comment to note that this is picking out all z for the non-guard x followed by the x-boundaries on appropriate processors, but not the non-boundary guard cells?

src/invert/laplace/impls/mumps/mumps_laplace.cxx
```diff
@@ -549,91 +550,94 @@ LaplaceMumps::LaplaceMumps(Options *opt) :
...
 
-const FieldPerp LaplaceMumps::solve(const FieldPerp &b, const FieldPerp &x0) {
-  return solve(b);
+const FieldPerp LaplaceMumps::solve(const FieldPerp &b_in, const FieldPerp &x0) {
+  return solve(b_in);
 }
 
-const FieldPerp LaplaceMumps::solve(const FieldPerp &b) {
+const FieldPerp LaplaceMumps::solve(const FieldPerp &b_in) {
+  FieldPerp b = b_in;
```
d7919: If you want `b` to be unique from `b_in` then you may need a `b.allocate()` to ensure they have different data blocks.

src/invert/laplace/impls/mumps/mumps_laplace.cxx
```diff
@@ -549,91 +550,94 @@ LaplaceMumps::LaplaceMumps(Options *opt) :
...
-    for (int z=0; z<mesh->LocalNz; z++)
-      soldata[x*mesh->LocalNz+z] = localrhs[(x-xs)*(mesh->LocalNz)+z];
+    for (int z=0; z<localmesh->LocalNz; z++)
+      sol(x,z) = localrhs[(x-xs)*(localmesh->LocalNz)+z];
   
   return sol;
 }
 
-void LaplaceMumps::solve(BoutReal* rhs, int y) {
+void LaplaceMumps::solve(Array<BoutReal> rhs, int y) {
```
d7919: You may want to pass `rhs` by reference here.

src/invert/laplace/impls/mumps/mumps_laplace.cxx
```diff
@@ -173,58 +172,58 @@ LaplaceMumps::LaplaceMumps(Options *opt) :
...
-  if (mesh->lastX()) {
-    localrhssize += (mesh->LocalNx-mesh->xend-1)*(mesh->LocalNz);
+  localrhssize = (localmesh->xend-localmesh->xstart+1)*(localmesh->LocalNz);
+  if (localmesh->lastX()) {
+    localrhssize += (localmesh->LocalNx-localmesh->xend-1)*(localmesh->LocalNz);
   }
-  if (mesh->firstX()) {
-    localrhssize += mesh->xstart*(mesh->LocalNz);
+  if (localmesh->firstX()) {
+    localrhssize += localmesh->xstart*(localmesh->LocalNz);
```
d7919: Should the `firstX()` if block finish here like the `lastX()` one? As it stands `firstX` leads to a lot of setup that won't be done on other procs. This includes allocating `rhs`, `rhs_positions` etc.

src/invert/laplace/impls/mumps/mumps_laplace.cxx
```diff
@@ -173,58 +172,58 @@ LaplaceMumps::LaplaceMumps(Options *opt) :
...
     }
-    rhs_positions = new int[nxpe];
+    rhs_positions = Array<int>(nxpe);
     rhs_positions[0] = 0;
     for (int i=1; i<nxpe; i++)
       rhs_positions[i] = rhs_positions[i-1] + localrhs_size_array[i-1];
-    
-    rhs = new BoutReal[meshx*meshz];
+
+    rhs = Array<BoutReal>(meshx * meshz);
```
d7919: There's a `Matrix` class provided in `utils.hxx` that is a light wrapper around `Array` that gives logically 2d access, which might help things here.

ZedThree:
I've got this working on Travis, but it needs the configure fixes in #944 to order to build correctly.

ZedThree:
Currently, there's *no* way to check this works. It's not possible to use `test-laplace`, as that changes the flags on the fly, which [throws an exception][1]. We either need a new test to demonstrate that this works, or we remove it as being broken. Even just a basic "invert and dot product" on `gauss(x) * sin(y)` would do.

As an aside, that test probably needs rewriting or supplementing to be able to check all the different implementations -- it won't work for the ones that are serial/parallel only -- but maybe we also need to do something about this implementation. Do we want to be able to set the flags programmatically? I imagine (outside of testing) that we don't often want to change the flags at runtime, but I am speculating there.

A hacky way to do it would be to make a new options object with the flags we want before creating a new MUMPS laplacian.

[1]: https://github.com/boutproject/BOUT-dev/blob/31e748c6bdf40c1118a562838cd3495467264129/src/invert/laplace/impls/mumps/mumps_laplace.hxx#L103

d7919:
Are there any views on this? Do we still support MUMPs?

bendudson:
I'm happy to leave it in with health warnings, if anyone might actually use it. I don't think however that there's much point: PETSc can call MUMPS, though there is some overhead IIRC. I don't think there's reason to spend time on maintaining it, personally.

d7919:
> I'm happy to leave it in with health warnings, if anyone might actually use it. I don't think however that there's much point: PETSc can call MUMPS, though there is some overhead IIRC. I don't think there's reason to spend time on maintaining it, personally.

OK, I was looking at this in the context of #1448 and wondering if I should change all these loops in a branch based off next if we were likely to want to change the code in a big way. For now I'll just ignore this file.

ZedThree:
To be replaced with big "here be dragons" warning. Maybe deprecate and remove in future version, unless someone can write a test that proves it does work

johnomotani:
BTW, I think `test-multigrid_laplace` should be trivial to modify to test MUMPS, just change `laplace:type` in the BOUT.inp. I also don't think it's worth supporting the stand-alone MUMPS implementation though.

PR #948, opened 2018-03-19
Extend differencing type checking slightly to be more rigorous in checking

d7919:
Rather than just checking if the diff type is known for any type of differencing operation, also check if it is
valid for the current differencing type (first, second, upwind etc.).

Specifically this will now catch the case where `mesh:ddz:upwind=fft`,
previously this would pass the checks as `fft` is a known type, but
because it's not implemented for `upwind` this would actually (silently)
revert to `u1`.

Reuses code from #481.


ZedThree:
Just noting here that the reason the affected MMS tests can't use WENO3 for the upwind scheme is that they are also using staggered grids, and there's no staggered WENO scheme

PR #953, opened 2018-03-21
Track Headerfile changes for rebuild

### LINKED ISSUE #931: Changed header files does not trigger rebuild
That should be fixed in the makefiles.
This could probably achieved (at least for gcc) with adding `-MM -MF .file.mk` to the compile command and include that in the makefile.

Would that be useful?

dschwoerer:
Resolves #931 

Note, that makes `make` slower if nothing needs to be done (around 500 ms vs 170 ms) - but I am willing to pay that price if that means I don't have to `make clean` to make sure everything is rebuild after changing a header file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,4 @@ aclocal.m4
 *.gcno
 bout-coverage/
 src/.libfast
+.*.mk
--- a/make.config.in
+++ b/make.config.in
@@ -311,10 +311,18 @@ endif
 -include .*.mk
 
 # If it is a libfast target, track changes to rebuild library if needed
+#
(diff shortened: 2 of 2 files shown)


dschwoerer:
In case the compiler does not support them, I hope the compiler fails (i.e. returns non-zero).
This will be ignored, the missing makefiles with the additional dependencies will be ignored, and everything should work as before - just a tiny bit slower, due the extra calls to the compiler.

Intel supports the flags: https://software.intel.com/en-us/cpp-compiler-18.0-developer-guide-and-reference-alphabetical-list-of-compiler-options

PR #956, opened 2018-03-21
Interp_to MMS test

dschwoerer:
A simple interp_to MMS test

Requires
 - [x] #641 
 - [x] #940 
 - [x] #957 

Actual code is here:
https://github.com/boutproject/BOUT-dev/tree/interp_to/tests/MMS/interpolation/runtest


ZedThree:
Does this test really need to be done in Python? Could it not be implemented in C++ instead? Then this PR wouldn't have 250+ commits in it.

dschwoerer:
I hoped that the cython branch gets merged soonish - than this test will only be a two or three commits.

It can be done in C++, but it is much easier to do in python.
Especially plotting in C++ is rather difficult.

I had done a test for interp_to before, implemented in C++ and python, but it was really slow (on some systems), and I am not even sure it did everything this one is doing.

If you don't like it as test, I can close and add to the cython branch as an example - [the test it self is quite simple](https://github.com/boutproject/BOUT-dev/pull/956/commits/b3f1f7d8f43d5bb6e1300abb870ad67912550473) and I did this way also [tests for ?DD? for the aiolos mesh](https://github.com/dschwoerer/BOUT-dev/blob/my_v4/tests/MMS/derivatives_flux/runtest) - they are however only in the [my_v4](https://github.com/boutproject/BOUT-dev/pull/956/commits/b3f1f7d8f43d5bb6e1300abb870ad67912550473) branch.

------------------------------------------

I hope that one day the python interface will allow us to write more simple and fast tests like this one.

bendudson:
Failed in interpolation MMS test:
```
Assertion failed in generated_fieldops.cxx, line 195: localmesh == rhs.getMesh()
```

src/mesh/interpolation.cxx
```diff
@@ -146,7 +146,7 @@ const Field3D interp_to(const Field3D &var, CELL_LOC loc, REGION region) {
           // coordinates
 
           Field3D var_fa = fieldmesh->toFieldAligned(var);
-          Field3D result_fa;
+          Field3D result_fa(fieldmesh);
```
dschwoerer: Not sure this is the best fix. This way it does not include the fix from L70 ...

dschwoerer:
The interpolation Y->X etc are failing.
That is probably related to the above issue - and could be somewhat fixed.
The question is, should they rather throw?

Further, interp_to has some comments about a interpolation with only yup and ydown.
I tried second order interpolation - and the results where basically unusable. Are you happy if I remove that part which cannot be reached?

dschwoerer:
Now that the python interface is merged, can this be merged?

dschwoerer:
Could go into 4.2?

Not a bug fix, but I think tests should also go in ...

d7919:
The conflicts will need resolving, if this is done then this may remove most of the changes to the existing code such that the PR is really just adding an extra test. I'd be in favour of treating new tests as bug fixes (although I've not reviewed the actual test here).

tests/unit/field/test_field3d.cxx
```diff
@@ -644,11 +644,12 @@ TEST_F(Field3DTest, IterateOverRGN_NOY) {
   EXPECT_TRUE(region_indices == result_indices);
 }
 
-TEST_F(Field3DTest, IterateOverRGN_NOZ) {
+TEST_F(Field3DTest, IterateOverInvalid) {
```
d7919: Just realised this means we're now not testing `RGN_NOZ` is it possible to add a test for this? Presumably just duplicating the `RGN_ALL` test.

tests/MMS/interpolation/runtest
```diff
@@ -0,0 +1,110 @@
...
+                conv=errc/difc
+                if order-prec < conv < order+prec:
+                    pass
+                else:
+                    err="%s -- %s -> %s failed (conv: %f)\n"%(func_,inloc,outloc,conv)
+                    errlist+=err
+                    try:
+                        #if True:
+                        #if 'Y' in dirs:
+                        if False:
```
ZedThree: Can we drop this whole bit, or use a command line argument to do this?

tests/unit/field/test_field3d.cxx
```diff
@@ -644,11 +644,12 @@ TEST_F(Field3DTest, IterateOverRGN_NOY) {
...
   Field3D field = 1.0;
+  REGION invalid = (REGION) 99999;
```
ZedThree: This should be a `static_cast<REGION>` rather than C-style cast

dschwoerer:
should I resolve the conflicts once more?

bendudson:
Yes please @dschwoerer . Sorry; this should go in now

dschwoerer:
How do I add Regions? I need RGN_NOZ.

What exception should an invalid region throw?
Used to be a BoutException, now it is a different one (but gtest doesn't bother to tell which)

dschwoerer:
PS: I couldn't find any documentation - addregion only lists where the function definitions are, but not where or how I can add new ones

d7919:
> How do I add Regions? I need RGN_NOZ.
> 
> What exception should an invalid region throw?
> Used to be a BoutException, now it is a different one (but gtest doesn't bother to tell which)

Mesh has `addRegion3D`, `addRegion2D` and `addRegionPerp` which can be used to register a new region with the mesh (these take a name and the Region). For `RGN_NOZ` I think it makes sense to ensure this is always provided so I would recommend adding this in `Mesh::createDefaultRegions`.

There is some documentation in the manual https://bout-dev.readthedocs.io/en/latest/developer_docs/data_types.html#creating-new-regions but it shows up quite far down the list of results if you search for "region" so can be easy to miss.

d7919:
As far as I can see the exception thrown is still a `BoutException`.

dschwoerer:
Is broken due to #1567

dschwoerer:
Thanks @d7919 for pointing where I need to add regions :+1: 

Would it make sense to rename `createDefaultRegions` to `addDefaultRegions`?
Other functions are called `addRegion...` - thus I was looking for `addDefaultRegions` - but I am also happy with create ...

dschwoerer:
I have fixed #1567 in 430b429

0e835bd is a more important bugfix, that should go probaly stright into master, as it fixes interpolation from one ?LOW to another ?LOW.

dschwoerer:
I cannot reproduce why that still fails.
If I do a git clone, checkout ede0217, and run `bash .travis_script.sh -uim` everything passes.

dschwoerer:
Fails seem to be caused by #1560 and me accidentally updating gtest.

Should I revert the updates?

d7919:
It looks like the current fail is just for the case where we don't have BOUT's checking enabled such that we're still hitting the throw due to `std::map::at` with an invalid key and hence the unit test gets a different type of exception than expected.

Some of the possible routes forward are:

1. Only do this unit test when we have out checking enabled.
2. Decide which type of Exception to expect based on `CHECK`.
3. Try to ask for the region by `string` rather than `enum`.
4. Catch the throw from `std::map::at` and throw a `BoutException`.

I'd probably go for either 2 or 3, but there are probably other options as well.

dschwoerer:
catching and re-throwing would allow the fastest solution - with basically no performance impact by default ...

dschwoerer:
:sob: Where do all the conflicts come from?

include/interpolation.hxx
```diff
@@ -100,13 +100,26 @@ const T interp_to(const T& var, CELL_LOC loc, REGION region = RGN_ALL) {
...
+	case RGN_NOX:
+	  realregion = RGN_NOBNDRY;
+	  break;
+	default:
+	  realregion = region;
+	  break;
+	}
         if ((location == CELL_CENTRE) && (loc == CELL_XLOW)) { // C2L
-          BOUT_FOR(i, result.getRegion("RGN_NOBNDRY")) {
+          BOUT_FOR(i, result.getRegion(realregion)) {
```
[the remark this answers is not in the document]
dschwoerer: I fixed it for CELL_ZLOW, where we shouldn't be doing this, and then changed the other locations as well ... but yes, only z should be changed ...

dschwoerer:
Merge conflict in ` tests/unit/field/test_field3d.cxx ` ...

PR #958, opened 2018-03-22
Catch all exceptions by reference

ZedThree:
Caught thanks to static analysis


d7919:
Can we add this to our `test-code-style`?

ZedThree:
I think that might be a bit tricky. Luckily, this is something Codacy can catch! We just need to keep an eye on it, it should tell us when new issues are introduced in PRs

PR #962, opened 2018-03-23
Fix order of HDF5 flags

### LINKED ISSUE #961: Error in make system related to linking with HDF5
The following issue was found when trying to build BOUT++ on/for the NERSC systems (e.g., CORI Haswell, with custom build NETCDF and HDF5.

When trying to build BOUT++ codes with a BOUT++ build that uses HDF5, either with 

--with-hdf5=PATH (location of h5cc) 

or with 

--with-netcdf

where the netcdf build is one that has the netcdf4 (hdf5-based) format enabled, the make.config file produced has the items

-lhdf5 -lhdf5_hl

in that order, which is the wrong order, and produces unsatisfied externals/symbols that prevent the build of codes dependent on the associated BOUT++ library build.

This order does not seem to correspond to that given by the h5cc script pointed to by the
--with-hdf5=PATH option.

Having these in the opposite order, i.e.,

-lhdf5_hl -lhdf5

seems to work robustly. 

The simple kludge/temporary fix is just to hand edit the make.config file, but I do not understand the autoconf/make system well enough to know how to make a "permanent" fix.

ZedThree:
When iterating over flags from h5cc, the HDF5_{CPPFLAGS,LDFLAGS,LIBS} variables were assembled in reverse order. This causes problems when linking against static libraries.

Fixes #961

### DIFF
--- a/configure
+++ b/configure
@@ -7386,18 +7386,18 @@ HDF5 support is being disabled (equivalent to --with-hdf5=no).
                 for arg in $HDF5_SHOW $HDF5_tmp_flags ; do
           case "$arg" in
             -I*) echo $HDF5_CPPFLAGS | $GREP -e "$arg" 2>&1 >/dev/null \
-                  || HDF5_CPPFLAGS="$arg $HDF5_CPPFLAGS"
+                  || HDF5_CPPFLAGS="$HDF5_CPPFLAGS $arg"
               ;;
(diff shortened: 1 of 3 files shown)


ZedThree:
I'm going to merge this directly into #944 as well, so that we ensure it gets into next.

PR #964, opened 2018-03-23
Fix assignment in assert

ZedThree:
An `ASSERT1` had an [assignment instead of comparison][1]

Also:
- removes a now unnecessary `ASSERT1` -- `start_index` no longer exists.
- moves the asserts to be the first thing in function calls (just in difops.cxx for now)

[1]: https://github.com/boutproject/BOUT-dev/commit/14cedca3c440c3f3d15c53b492f33ff72b70283b#diff-f39f25f34d0d23c5829694f4545ec237L731

### DIFF
--- a/src/mesh/difops.cxx
+++ b/src/mesh/difops.cxx
@@ -724,11 +724,14 @@ const Field2D bracket(const Field2D &f, const Field2D &g, BRACKET_METHOD method,
   return result;
 }
 
-const Field3D bracket(const Field3D &f, const Field2D &g, BRACKET_METHOD method, CELL_LOC outloc, Solver *solver) {
+const Field3D bracket(const Field3D &f, const Field2D &g, BRACKET_METHOD method,
+                      CELL_LOC outloc, Solver *solver) {
   TRACE("bracket(Field3D, Field2D)");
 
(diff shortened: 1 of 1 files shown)


src/mesh/difops.cxx
```diff
@@ -74,8 +74,9 @@ const Field3D Grad_par(const Field3D &var, DIFF_METHOD method, CELL_LOC outloc)
 *******************************************************************************/
 
 const Field3D Grad_parP(const Field3D &apar, const Field3D &f) {
+  ASSERT1(apar.getMesh() == f.getMesh());
+
   Mesh *mesh = apar.getMesh();
-  ASSERT1(mesh == f.getMesh());
```
dschwoerer: note that this is (non-significantly) slower - as getMesh is called 3 times, rather than twice ...

ZedThree: I don't think it will really make any difference. If you've got optimisations turned on, the call to `getMesh` will be inlined and will be one or two instructions more at most. The worst case is 10 instructions without optimisations, including stack management from the non-inlined function call, none of which are more than a single clock cycle. If speed is really that crucial, then just turn checks off, and the whole line will vanish anyway.

PR #969, opened 2018-03-28
OpenMP cyclic-reduce solver; add ensureUnique to Matrix,Tensor

bendudson:
This looks like a large PR, but it's mostly formatting in the first commit which doesn't change any code.
The actual changes are:

* OpenMP parallelise the cyclice-reduce solver. There are three places which are trivially parallel, essentially parallelising in Y and Z for the Laplacian solve. The MPI communication is still done in OpenMP serial regions for now.
* Add ensureUnique() to Matrix and Tensor. This was because indexing a Matrix in an OpenMP parallel loop (reading from Matrix, not writing to) resulted in a call to ensureUnique which then messed around with the memory, causing problems with multiple threads. Calling ensureUnique before the parallel region solves the problem.

I think an improvement would be to remove the call to ensureUnique in the Matrix and Tensor indexing operator
https://github.com/boutproject/BOUT-dev/blob/next/include/utils.hxx#L65
This would be consistent with Array and the Fields, and would probably help with issue #933. I think it would also fix the problem I ran into: The arrays were being read, so not being unique was not a problem.

### DIFF
--- a/include/cyclic_reduction.hxx
+++ b/include/cyclic_reduction.hxx
@@ -366,7 +366,6 @@ public:
       
       BOUT_OMP(parallel for)
       for (int i = 0; i < myns; ++i) {
-        //output << i << "\n";
         //  (a  b) (x1) = (b1)
         //  (c  d) (xn)   (bn)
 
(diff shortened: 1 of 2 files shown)


include/cyclic_reduction.hxx
```diff
@@ -290,351 +300,373 @@ public:
...
-	output << "(" << c << ", " << d << ") ("<<xn[i]<<")   (" << bn << ")\n\n";
+
+      // For OpenMP, ensure that memory won't be modified inside parallel loop
+      if2x2.ensureUnique();
+      x1.ensureUnique();
+      xn.ensureUnique();
+      
+      BOUT_OMP(parallel for)
+      for (int i = 0; i < myns; ++i) {
+        //output << i << "\n";
```
d7919: Probably don't need to include this commented out line.

include/cyclic_reduction.hxx
```diff
@@ -290,351 +300,373 @@ public:
...
     // xa -- Result for each system
     // co -- Coefficients & rhs for each system
-    Array<T> gam(nloc);
-    for(int i=0;i<ns;i++) { // Loop over systems
+    BOUT_OMP(parallel for)
+    for (int i = 0; i < ns; i++) { // Loop over systems
+      Array<T> gam(nloc);
       T bet = 1.0;
-      xa(i, 0) = x1[i]; // Already know the first 
+      xa(i, 0) = x1[i]; // Already know the first
```
d7919: Is there any risk of the mentioned race/memory issue happening here such that we need to make this unique outside the parallel region?

d7919:
Do you have any feel for how well OpenMP performs here for "typical" cases?

bendudson:
No not yet; I have a 2-core laptop here so can't really test. X2go is usable, but the lag is quite annoying from Taiwan

PR #972, opened 2018-04-02
Rangeiterator to region in Solver

bendudson:
* Adds regions RGN_UPPER_Y, RGN_LOWER_Y, RGN_INNER_X, RGN_OUTER_X, RGN_UPPER_INNER_Y, RGN_UPPER_OUTER_Y, RGN_LOWER_INNER_Y, RGN_LOWER_OUTER_Y and RGN_BNDRY (all boundaries).
* Solver now uses Regions to loop over grid points and copy data to/from fields. Note that this changes the order in which evolving variables appear, if evolving boundaries are used. It has the advantage that the part of the grid which is evolved in time could potentially be modified using a custom Region.
* Region constructor now creates an empty region if given an empty range of indices (e.g. xend < xstart). This is to allow empty regions to be created, and to make porting the boundary region creation code.

Much of the remaining use of RangeIterator requires index offsets, for example when setting boundary conditions.


ZedThree:
Once #934 is merged (just waiting for tests), this can use those 2D <-> 3D conversion functions.

Also, we should add unit tests for these. I realise these are a bit more complicated as there's lots of possible combinations, but I'll have a stab at them tomorrow.

src/solver/solver.cxx
```diff
@@ -850,17 +829,17 @@ void Solver::loop_vars_op(int jx, int jy, BoutReal *udata, int &p, SOLVER_VAR_OP
...
     
-    for (jz=0; jz < mesh->LocalNz; jz++) {
+    for (int jz=0; jz < nz; jz++) {
       
       // Loop over 3D variables
       for(const auto& f : f3d) {
         if(bndry && !f.evolve_bndry)
           continue;
-        (*f.var)(jx, jy, jz) = udata[p];
+        (*f.var)[Ind3D(i2d.ind*nz + jz)] = udata[p];
```
ZedThree: I've finally got around to updating this, but it ends up looking like:

```cpp
        (*f.var)[f.var->getMesh()->ind2Dto3D(i2d, jz)] = udata[p];
```

Given that we're already looping over `mesh->LocalNz`, I'm not 100% sure this is any better. I suspect that `solver` probably assumes all the fields it's evolving are on the same `mesh` anyway...

PR #977, opened 2018-04-08
Suppress multigrid diagnostic output

johnomotani:
The multigrid laplace solver used to print out timing information every
300 solves. Now suppress this unless the multigrid solver's checking is
turned on by 'laplace:checking' being greater than 0 (this is used to
set the 'pcheck' variable in LaplaceMultigrid).

### DIFF
--- a/src/invert/laplace/impls/multigrid/multigrid_laplace.cxx
+++ b/src/invert/laplace/impls/multigrid/multigrid_laplace.cxx
@@ -417,18 +417,17 @@ BOUT_OMP(for)
   // Compute solution.
 
   mgcount++;
-  if((mgcount == 300) && (pcheck == 0)) {
-    tcheck = 1;
-  }
-  t0 = MPI_Wtime();
+  if (pcheck > 0) t0 = MPI_Wtime();
 
   kMG->getSolution(std::begin(x), std::begin(b), 0);
 
-  t1 = MPI_Wtime();
-  if((mgcount == 300) && (tcheck != pcheck)) tcheck = pcheck;
-  soltime += t1-t0;
-  if(mgcount%300 == 0) {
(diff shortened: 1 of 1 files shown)


src/invert/laplace/impls/multigrid/multigrid_laplace.cxx
```diff
@@ -417,18 +417,17 @@ BOUT_OMP(for)
...
   kMG->getSolution(std::begin(x), std::begin(b), 0);
 
-  t1 = MPI_Wtime();
-  if((mgcount == 300) && (tcheck != pcheck)) tcheck = pcheck;
-  soltime += t1-t0;
-  if(mgcount%300 == 0) {
-    output<<"Accumulated execution time at "<<mgcount<<" Sol "<<soltime<<" ( "<<settime<<" )"<<endl;
+  if (pcheck > 0) {
+    t1 = MPI_Wtime();
+    if((mgcount == 300) && (tcheck != pcheck)) tcheck = pcheck;
```
bendudson: What does tcheck do? It doesn't seem to be used in this file.

src/invert/laplace/impls/multigrid/multigrid_laplace.cxx
```diff
@@ -417,18 +417,17 @@ BOUT_OMP(for)
...
+    soltime += t1-t0;
+    if(mgcount%300 == 0) {
+      output<<"Accumulated execution time at "<<mgcount<<" Sol "<<soltime<<" ( "<<settime<<" )"<<endl;
```
bendudson: I think soltime should be reset here

src/invert/laplace/impls/multigrid/multigrid_laplace.cxx
johnomotani: As far as I can see it was not used. Don't know what the original intention was. I have removed it, and also an 'rcheck', in a new commit.

PR #978, opened 2018-04-09
Prevent shadowing

dschwoerer:


dschwoerer:
You can verify that it does make a difference, by applying this patch:
```
diff --git a/src/mesh/meshfactory.cxx b/src/mesh/meshfactory.cxx
index a3fede8d..3879e033 100644
--- a/src/mesh/meshfactory.cxx
+++ b/src/mesh/meshfactory.cxx
@@ -61,7 +61,7 @@ Mesh* MeshFactory::createMesh(GridDataSource *source, Options *options) {
   options->get("type", type, MESH_BOUT);
   
   if(!strcasecmp(type.c_str(), MESH_BOUT)) {
-    return new BoutMesh(source, options);
+    return new BoutMesh(source, (Options*)nullptr);
   }
 
   throw BoutException("Mesh type not implemented: %s",type.c_str());
```

With that patch the shadowing causes issues - and e.g. test_io segfaults (I guess all integrated and MMS tests will fail) - this is already discussed in #960 

I also wrote a unit test, to show this:
https://travis-ci.org/boutproject/BOUT-dev/jobs/364605182
And fixed with this branch:
https://travis-ci.org/boutproject/BOUT-dev/builds/364606532

PR #981, opened 2018-04-12
Improve manual

bendudson:
Mainly re-organising, some fixing and updating. Trying to reduce the number of sections, and organise the material better. 

Not yet finished; work in progress (and probably always will be...)


ZedThree:
I've added this to readthedocs so we can see what it looks like before merging: http://bout-dev.readthedocs.io/en/improve-manual/

ZedThree:
I've set the `default_role` in Sphinx to be `any`, which means you can now just use plain backticks to link to the C++ API docs, e.g. use \`Mesh\` to link to the `Mesh` docs. You can also link to the Python docs with ":py:obj:\`boutdata.collect.collect\`" -- or even ":py:obj:\`~boutdata.collect.collect\`", which just renders the last part (i.e. `collect`).

Unfortunately, I've not been good and done this everywhere, only a smattering of places. Please feel free to fix this and add links everywhere! We've used this pattern a lot:

```
You do x by calling:

::

    object->x(args);
```
which could perhaps be written better like so:

```
You can do x by calling `Object::x` like this::

    object->x(args);
```
I will try and change all of these patterns eventually. Note that the double colon has also been merged into the end of the previous line.

Also a bunch of other things:

- Don't create separate files for classes with breathe, should make readthedocs faster and time out less often (all the documentation is now on the file pages)
- Add lots of #defines to doxygen so it includes lots of optional stuff (unfortunately, this then pushed the time back up)- Add a bunch more doxygen comments (now up to ~36% doxygen coverage!)
- Fix a bunch of doxygen problems (typos, etc.)
- Fix a bunch of Sphinx problems
- Go through conduction example [here][1] line-by-line

[1]: http://bout-dev.readthedocs.io/en/improve-manual/user_docs/physics_models.html

ZedThree:
Also added an example to the [README.md][1] so people can immediately see something interesting. We should also add this to the website front page.

I realised I never actually fixed my points 2, 3, 4 from above either...


[1]: https://github.com/boutproject/BOUT-dev/blob/improve-manual/README.md

bendudson:
Thanks Peter! A big task but very much needed.

ZedThree:
There's more incoming, unfortunately... The good news is that we'll be up to 38% doxygen coverage!

PR #984, opened 2018-04-12
Optimize collect in BoutOutputs

johnomotani:
collect() spends a lot of its time opening the DataFile objects. In
BoutOutputs it makes sense to make a cache of all the DataFiles so they
only need to be opened once. This speeds up subsequent collects (I've
seen about 5x speedup).

Implemented using a method of BoutOutputs that mostly duplicates the
collect function, except for using a cache for the DataFiles.

Maybe it would be nicer to avoid code duplication by still using the collect function: we could add some optional arguments to collect to pass in filenames and DataFile objects, instead of finding them within the function.

Does this seem useful, and are there any preferences on the implementation?

### DIFF
--- a/tools/pylib/boutdata/data.py
+++ b/tools/pylib/boutdata/data.py
@@ -5,9 +5,11 @@
 # 
 
 import os
+import sys
 import glob
+import numpy
 
-from boutdata import collect
+from boutdata.collect import findVar as collect_findVar, findFiles as collect_findFiles
 
 try:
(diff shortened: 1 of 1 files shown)


bendudson:
Yes this seems useful; collecting data tends to take much longer than it should do. Reducing code duplication would be nice, but finding ways to make it work better first is important.

bendudson:
This feature already merged in PR #996

PR #989, opened 2018-04-13
Slow tests

dschwoerer:
- [x] Based on #980 
- [x] Allows to run easily all tests
- [x] Improved some MMS tests - make sure they fail properly if the convergence isn't as expected
- [x] Some MMS tests fail faster - don't waste to much CPU time

I used this branch to run all tests - [the results are not that great](https://omadesala.physics.dcu.ie/~bout-test/)

Should I open an issue for each failing MMS test?


tests/MMS/advection/runtest
```diff
@@ -32,7 +34,8 @@ options = [
     ]
 
 # List of NX values to use
-nxlist = [16, 32, 64, 128, 256, 512, 1024]
+#[16, 32, 64, 128, 256, 512, 1024]
+nxlist = [64, 128]
```
ZedThree: I'm not sure about this, I don't think two points are enough for an MMS test

tests/MMS/advection/runtest
```diff
@@ -95,15 +98,15 @@ for opts,label,sym in options:
     print("Convergence order = %f" % (order))
     
     if (label == "1st order upwind"):
-        if order < 0.75: # Should be first order accurate
+        if order < 0.75 or order > 1.25: # Should be first order accurate
```
ZedThree: A more pythonic way of writing this:

```python
if 0.75 < order < 1.25:
# success condition
else:
# fail
```

tests/MMS/hw/runtest
```diff
@@ -17,7 +19,9 @@ print("Making MMS test")
 shell_safe("make > make.log")
 
 # List of NX values to use
-nxlist = [16, 32, 64, 128, 256, 512]
+nxlist = [32, 64]
+# Old list:
+# nxlist = [16, 32, 64, 128, 256, 512]
```
ZedThree: Again, I'm not sure about this, as it makes the test less rigorous

tests/integrated/test-petsc_laplace/runtest
```diff
@@ -4,6 +4,9 @@
 # Run the test, compare results against the benchmark
 #
 
+#requires: petsc
+#requites: slow_tests
```
ZedThree: Typo: requites -> requires

tests/integrated/test_suite
```diff
@@ -12,32 +12,87 @@ from __future__ import print_function
...
+    raise RuntimeError("The make flag needs to be set by passing --make")
 
 ##################################################################
 
+if args.set_list is None:
+    # Get list of directories containing test cases
+    try:
+        # Requires python >= 3.5
+        tests=glob.iglob('**/runtest', recursive=True)
+    except:
```
ZedThree: We should catch the specific exception, which in this case is `TypeError`, I think

tests/integrated/test_suite
```diff
@@ -12,32 +12,87 @@ from __future__ import print_function
...
+    tests=[x.rsplit('/',1)[0] for x in tests]
+
+else:
+    # Take the user provided list
+    tests=[x.rstrip('/') for x in args.set_list]
+
+
+if args.get_list:
+    for test in tests:
+        req_met, req_expr = requirements.check(test+"/runtest")
```
ZedThree: For unused function results, you can use `_`, so:

```python
req_met, _ = requirements.check(test+"/runtest")
```

tests/MMS/advection/runtest
dschwoerer: I haven't found a single test that takes more than two points to evaluate to success condition.

Basically all use ` order = log(error_2[-1] / error_2[-2]) / log(dx[-1] / dx[-2])`.

tests/MMS/hw/runtest
dschwoerer: To some extend yes - as we are not going to such high grid numbers, but it does not change the outcome and makes it much faster.

dschwoerer: I checked:

`nx_max` | convergence | wall time
-----|----|---
16 | 1.593060 | ~1.5 s
32 | 1.940418 | ~ 3 s
40 | 1.964195 | ~ 4 s
48 | 1.976635 | ~ 5 s
64 | 1.988427 | ~10 s
128 | 1.998536 | ~ 45 s
256 | 2.000325 | ~ 640 s

Not sure why we should actually run with more than `nx=40`.
128 is an order of magnitude more precise then we check for in any test ...
I guess it looks nicer when you can plot a longer line showing the conversion, but I think to actually use in a test suite, this is perfectly reasonable.

dschwoerer:
Should I split this into several PRs?

tests/MMS/advection/runtest
dschwoerer: ### Arakawa

nx | time | order
--|--|--
 16 |   2.410 s | ?
 32 |   3.438 s | 3.255013
 64 |  10.307 s | 2.490309
 128 |  21.605 s | 2.060740
 256 |  45.281 s | 2.004003
 512 | 104.097 s | 1.998664

### U1

nx | time | order
--|--|--
 16 |   2.531 s | ?
 32 |   5.458 s | 2.373164
 64 |   9.895 s | 1.242753
 128 |  18.549 s | 0.972551
 256 |  49.725 s | 0.971749
 512 | 138.652 s | 0.985141


### Central

nx | time | order
--|--|--
 16 |   2.737 s | ?
 32 |   3.666 s | 3.449089
 64 |  14.249 s | 2.902876
 128 |  28.197 s | 2.220268
 256 |  47.013 s | 2.051120
 512 | 126.226 s | 2.015397

### Weno 3

nx | time | order
--|--|--
 16 |   5.916 s | ?
 32 |  15.774 s | 3.480261
 64 |  29.917 s | 3.135221
 128 |  41.231 s | 2.400913
 256 |  58.130 s | 2.106673
 512 | 140.946 s | 2.038875

tests/MMS/hw/runtest
ZedThree: The issue is not when it works -- it's when it doesn't work. What we normally see with MMS tests working correctly is that the error scaling is not right at very low grid resolution (as you would expect) as well as at high grid resolution (due to a different dominant error source), and it's only in some middle range that the error scaling holds. One way failures manifest themselves is by a shortening of this range from either or both ends. So what we want to check is that the error scales correctly over a wide range of reasonable grid resolutions.

The MMS tests are slow -- but that's fine, this PR lets us run them optionally. This is something we'll only do infrequently, possibly with a cron job on Travis or something, or just for releases.

dschwoerer: In that case we should make sure that it converges over the broad range, not just the last point.
I can have a go when I have more time ...

Right now, this might still be an improvement, as it will at least detect if the range is shortened from one end, rather than to check somewhere in the middle (I guess)

ZedThree:
Renamed `slow_tests` -> `all_tests`. Also added a synonym for running all tests (`test_suite --all`).

tests/integrated/test_suite
```diff
@@ -12,53 +12,100 @@ from __future__ import print_function
...
-    for test in testfile:
-        test=test.strip()
-        if test.startswith("#"):
-            pass
-        else:
-            tests.append(test)
+try:
+    requirements.add('make', args.make)
+except RuntimeError:
+    raise RuntimeError("The make flag needs to be set by passing --make")
```
ZedThree: I'm not sure I understand this error message. Is this for the case where you're trying to run a single test which requires `make`?

ZedThree:
`test_suite_make` now *runs* all of the tests (plus any that `#require make`, minus any that `#require not make`). This means we end up running the integrated/MMS tests *twice*

Do we still need `test_suite_make`? Is there value in having a separate compilation test suite?

tests/integrated/test_suite
```diff
@@ -69,35 +116,35 @@ for t in tests:
 
     start_time_ = time.time()
     # Run test, piping stdout so it is not sent to console
-    status,out = shell("./runtest", pipe=True)
+    status, out = shell("./runtest", pipe=True)
```
d7919: Should this be `command` rather than `./runtest` as sometimes we just want to make?

tests/integrated/test_suite
```diff
@@ -12,53 +12,100 @@ from __future__ import print_function
...
 for t in tests:
     os.chdir(t)
 
-    print(output%t,end='')
-    sys.stdout.flush() # Make sure name is flushed
+    print(output % t, end='')
+    sys.stdout.flush()  # Make sure name is flushed
+
     # Check requirements
     req_met, req_expr = requirements.check("./runtest")
```
d7919: Should this be `command` rather than `./runtest` as sometimes we just want to make?

Issue #988, opened 2018-04-12
test-petsc_laplace_MAST-grid fails


dschwoerer:
Still python2 - TabError: inconsistent use of tabs and spaces in indentation

PR #992, opened 2018-04-13
Fix test-petsc_laplace_MAST-grid

dschwoerer:
Fixes part of #988 
I have not configured with petsc - so not sure the test actually works ...


bendudson:
I get an error when running `test-petsc-laplace` along the lines of:
`[0]PETSC ERROR: Unable to find requested KSP type (j`
I thought this was just a PETSc 3.8 issue on Arch, but I get the same issue on `sawtooth` (after some other weirdness; see issue #997 ). I suspect that the type char* being given to PETSc is junk somehow.

dschwoerer:
The test exits by returning 1. This causes the launch_safe to fail.
I fixed this in 048f25047b5c4c2935d4c526a03cf1a600b1461b

With that the first 3 tests still fail (I tried petsc 3.[4-8])
https://omadesala.alfa1024.info/~bout-test/2018-04-27_14:32/

dschwoerer:
Everything got so much greener:

https://omadesala.alfa1024.info/~bout-test/2018-05-01_16:00/

Now only a few MMS tests are failing.

PR #993, opened 2018-04-13
Enable configuring for fedoras installed PETSc

dschwoerer:
* Do not require explicit location of header

I assume we don't support PETSc 3.8 yet?
The documentation seems to be slightly outdated in that regard ..


bendudson:
This does compile, run with PETSc 3.8.1 (just tested on Arch)

PR #994, opened 2018-04-13
Check result location in derivatives

johnomotani:
The interp_to in the return statement of D2DY2 is redundant and can be
removed. It is a bug, because if D2DY2 is called without an outloc
argument, CELL_DEFAULT is passed as the location to interp_to().
interp_to does not handle CELL_DEFAULT, so an exception is raised unless
staggergrids=false so that interp_to does nothing.

The interp_to() call was redundant because result already has location 'outloc' from the calls to indexD2DY2 and indexDDY (the latter relevant only if nonuniform grid corrections are enabled).

This raises the question: should interp_to handle CELL_DEFAULT? It would make sense to return the original field without interpolating in that case. From this example, and the fact that this sort of bug hasn't showed up anywhere else before, I think passing CELL_DEFAULT to interp_to is probably a sign that some location has not been handled correctly (for example the call to interp_to is redundant, as here), so it is more helpful to have this be an error.


bendudson:
In a case like this where the location should already be set correctly, I think it would be good to put an assertion like:

`ASSERT2(result.getLocation() == outloc)`

Partly as a check, and partly as documentation.

johnomotani:
Added ASSERT statements as @bendudson suggested.
Added a ``location`` member and ``getLocation()`` method to ``Vector3D``, so that Vector operations can check location too.

While doing that, I realized that the existing implementation for derivatives between different staggered grids looked incorrect: it attempted to interpolate to ``CELL_CENTRE``, then take the derivative, then interpolate to the output location, but guard cells were not set anywhere so I don't think the results could be correct. The last commit in this PR replaces the old implementation with an exception saying the requested combination of locations is not implemented. Hopefully this should be clearer if a user ever happens to call the unhandled cases. Assuming this is the way to go, should we delete the old code, which for the moment I've just commented out? It might not be silly to keep commented out code, because we could in principle make it work, for some value of work, by communicating and applying some boundary conditions, e.g. Neumann, as suggested but apparently not fully implemented (it doesn't look like boundary conditions are applied?) in https://github.com/boutproject/BOUT-dev/blob/8389878169b4e4824c2d486dad7d9d65fa0db8c5/src/sys/derivs.cxx#L296-L307

bendudson:
If Vector3D has a location, then there should be checks that all the components have the same location. Otherwise something like this could be done:
```
Vector3D v = ...;
v.setLocation(CELL_YLOW);
v.x.setLocation(CELL_XLOW);
```

It should be easy to have Vector3D.setLocation call its components to set their locations, but I don't think it's easy to have `v.x.setLocation()` check that it's consistent with its parent vector.

johnomotani:
Does this commit help? By checking for consistency in ``Vector3D::getLocation()`` at least we don't think the location of all the fields is right when it isn't.

dschwoerer:
Could you add parenthesis to the long asserts to make the more readable?

The 2fluid.cxx has already been fixed in PR #940 
I have not tested your solution - but I think Vi needs to be staggered ...

johnomotani:
I've added more parentheses to the ASSERTs and removed the change to test-drift-instability. I think it makes more sense to leave the latter for #940, and merge this PR afterward.

dschwoerer:
I am sorry - you had already done what I asked :/
Maybe adding:
```
if (location == CELL_VSHIFT) {
 ASSERT1(1st part);
} else {
 ASSERT1(2nd part);
}
```
The `if else` should be removed if the ASSERT1 is removed - thus should have no performance implication.

johnomotani:
Thanks @dschwoerer! That's much more readable, and I didn't know it was possible to do like that and get zero performance hit. Nice :smile:

src/mesh/index_derivs.cxx
```diff
@@ -1471,7 +1509,11 @@ const Field3D Mesh::indexD2DX2(const Field3D &f, CELL_LOC outloc, DIFF_METHOD me
   result = applyXdiff(f, func);
   result.setLocation(diffloc);
 
-  result = interp_to(result, outloc);
+  // result = interp_to(result, outloc);
```
ZedThree: Please remove commented out code

src/mesh/index_derivs.cxx
```diff
@@ -2852,12 +2957,21 @@ const Field2D Mesh::indexFDDY(const Field2D &v, const Field2D &f, CELL_LOC outlo
...
+      ((v.getLocation() != CELL_CENTRE) || (f.getLocation() != CELL_CENTRE))) {
+    // Staggered differencing
+    throw BoutException("Unhandled staggering");
+  }
+  else {
+    result.setLocation(CELL_CENTRE);
+  }
+
+  ASSERT1(this == v.getMesh());
+  ASSERT1(this == f.getMesh());
```
ZedThree: Please could you move these checks to be the first thing in the function call? It's a good idea to bail as early as possible if big assumptions like these aren't met

src/mesh/index_derivs.cxx
```diff
@@ -1164,18 +1170,22 @@ const Field3D Mesh::indexDDX(const Field3D &f, CELL_LOC outloc, DIFF_METHOD meth
...
-        diffloc = CELL_CENTRE;
-
-      } else if (inloc != CELL_CENTRE) {
-        // Interpolate then (centre -> centre) then interpolate
-        return DDX(interp_to(f, CELL_CENTRE), outloc, method);
-      }
+      // Derivative of interpolated field or interpolation of derivative field
+      // cannot be taken without communicating and applying boundary
+      // conditions, so throw an exception instead
+      throw BoutException("Unsupported combination of inloc/outloc in Mesh:indexDDX(Field3D).");
```
ZedThree: I've forgotten most of the discussion about this, forgive me, but is it the case that previously, we needed to communicate/apply BCs and weren't doing so? If so, this is fine, but maybe put the inloc/outloc in the exception?

src/mesh/index_derivs.cxx
```diff
@@ -1164,18 +1170,22 @@ const Field3D Mesh::indexDDX(const Field3D &f, CELL_LOC outloc, DIFF_METHOD meth
...
+      //   // Shifting
+
+      //   func = sfDDX;                // Set default
+      //   table = FirstStagDerivTable; // Set table for others
+      //   diffloc = CELL_CENTRE;
+
+      // } else if (inloc != CELL_CENTRE) {
+      //   // Interpolate then (centre -> centre) then interpolate
+      //   return DDX(interp_to(f, CELL_CENTRE), outloc, method);
+      // }
```
ZedThree: Please remove commented out code (and similarly for the other instances below)

src/mesh/index_derivs.cxx
```diff
@@ -1251,7 +1269,13 @@ const Field3D Mesh::indexDDY(const Field3D &f, CELL_LOC outloc, DIFF_METHOD meth
 
   result.setLocation(diffloc); // Set the result location
 
-  return interp_to(result, outloc); // Interpolate if necessary
+  // return interp_to(result, outloc); // Interpolate if necessary
+  // Cannot interpolate result without communicating and applying boundary
+  // conditions first, instead check that it is at outloc
+  ASSERT2(((outloc == CELL_DEFAULT) && (result.getLocation() == f.getLocation())) ||
+          (result.getLocation() == outloc));
```
ZedThree: Can these checks be moved earlier? i.e. what's the earliest we can bail out if something's wrong?

src/mesh/index_derivs.cxx
johnomotani: > is it the case that previously, we needed to communicate/apply BCs and weren't doing so?
That's right.

I'm adding inloc/outloc to the exception at the moment... Is there a better way to add an enum than as an int? I can't quickly find a nice way of printing the enum value :disappointed:

ZedThree: Ah, I'm afraid not. The easiest way is to make a `constexpr std::map<ENUM, std::string>` and look up the name in that, but that might be overkill here.

dschwoerer: There is a function for that purpose:
https://github.com/boutproject/BOUT-dev/blob/master/include/interpolation.hxx#L41

src/mesh/index_derivs.cxx
johnomotani: Looking at them again, I think these checks only replicate the inloc/outloc checks already done at the top of the functions, so I have just deleted them.

johnomotani:
In the most recent commit I pushed, I've made all the derivatives use ``result.setLocation(diffloc)`` before returning ``result``. Previously several of them used ``result.setLocation(inloc)``, which I think is incorrect, but I'm not sure. Could someone have a look over and see you think ``diffloc`` or ``inloc`` is correct please?

johnomotani:
The new ``derivatives3`` test is failing, for ``W2``, ``W3`` and ``S2`` derivatives.

* It looks like the ``S2`` derivatives are 3rd order accurate. Are we sure they are not supposed to be 3rd order?
* Using doPlot=True in the test, W2 has large errors near where the function is zero. W2 converges at second order if the derivative is never zero. Is this the expected behaviour? It looks to me as though the 'smoothness indicators' for W2 measure gradient vs. curvature, so maybe we should expect accuracy to be ~1st order when gradient->0 so (curvature/gradient) is large?
* W3 converges faster slower than expected for the parameters in the test at the moment, but faster than 3rd order if I increase the maximum resolution even further.

dschwoerer:
Maybe remove them from the tests for now, and open issues for them?

Adding specific tests for there special properties would also be nice ...

bendudson:
Testing convergence of nonlinear methods like WENO schemes is nontrivial... the switches in them make the order of convergence depend on the solution, and may vary across the domain. At high resolution when everything is smooth, I think they should recover their advertised convergence (e.g W3 -> 3rd order), but if the limiters are active then this may not be guaranteed.

For now probably best remove the test from the default test suite, and raise a few issues. We can then merge this branch.

PR #996, opened 2018-04-14
Redistribute dmp files

johnomotani:
This PR adds a method redistribute() to BoutOutputs to redistribute dump files, similar to restart.redistribute(). This allows appending to the data files when changing the number of processors.

To allow this, needed to enable writing of netCDF and HDF5 files with unlimited time dimensions.

Also:
* adds a few fixes for DataFile_HDF5.
* enables more file suffixes in collect()


johnomotani:
I don't think we have tests for the python tools at the moment? Should we?

dschwoerer:
0bc10cf  reverts the fix I made to get hdf5 working.

What python version are you using?
What is the error you get without?

johnomotani:
> dschwoerer commented
> 
> 0bc10cf reverts the fix I made to get hdf5 working.
> 
> What python version are you using?
> What is the error you get without?

In d72d044e2032ad71e6ecf6f916fdf4c546ba8d14 I changed the comparison to byte strings, ``b"Field3D_t"``, etc. which I think fixes the same error. When creating a ``'type'`` attribute, e.g. https://github.com/boutproject/BOUT-dev/blob/e948a21f626537b4e5f9ea1fef57f0034ec7349c/tools/pylib/boututils/datafile.py#L690 passing a ``'utf-8'`` string as ``vartype`` gives me an error
```
/home/john/soltransport/BOUT-dev/tools/pylib/boututils/datafile.py in write(self, name, data, vartype)
    688             self.handle.create_dataset(name, data=data)
    689 
--> 690         self.handle[name].attrs.create('type', vartype)
    691 
    692     def attributes(self, varname):

/usr/lib/python3/dist-packages/h5py/_hl/attrs.py in create(self, name, data, shape, dtype)
    167             # Make HDF5 datatype and dataspace for the H5A calls
    168             if use_htype is None:
--> 169                 htype = h5t.py_create(original_dtype, logical=True)
    170                 htype2 = h5t.py_create(original_dtype)  # Must be bit-for-bit representation rather than logical
    171             else:

/usr/lib/python3/dist-packages/h5py/h5t.cpython-35m-x86_64-linux-gnu.so in h5py.h5t.py_create (/build/h5py-nQFNYZ/h5py-2.6.0/h5py/h5t.c:16206)()

/usr/lib/python3/dist-packages/h5py/h5t.cpython-35m-x86_64-linux-gnu.so in h5py.h5t.py_create (/build/h5py-nQFNYZ/h5py-2.6.0/h5py/h5t.c:16037)()

/usr/lib/python3/dist-packages/h5py/h5t.cpython-35m-x86_64-linux-gnu.so in h5py.h5t.py_create (/build/h5py-nQFNYZ/h5py-2.6.0/h5py/h5t.c:15997)()

TypeError: No conversion path for dtype: dtype('<U6')
```
Because of this error, I thought it was better to use byte strings everywhere for consistency.

Apologies for the lack of comment when reverting in 0bc10cf. I didn't understand what your change had done and didn't think about why it was there because my code worked. My bad, sorry!

dschwoerer:
I thought everything should always be unicode.

I didn't know that there are (old?) hdf5 libraries that don't support this (yet).

Would it be nicer to use internally unicode strings, and only fall back to byte strings when needed?
It seems tools/pylib/boututils/datafile.py:690 would be such a place.

I guess both solutions would work - ether convert everything to bytes or everything to unicode ...

johnomotani:
Agreed it's nicer to use unicode strings internally. I've updated to do that now.

Also added some more checking to BoutOutputs in initialisation, to avoid odd errors I got while testing changes, which were due to having extra BOUT.dmp* files from an older run with more processors lying around in the directory.

tools/pylib/boutdata/collect.py
```diff
@@ -90,12 +90,58 @@ def collect(varname, xind=None, yind=None, zind=None, tind=None, path=".",yguard
...
+        try:
+            mxg  = f["MXG"]
+        except KeyError:
+            mxg = 0
+            print("MXG not found, setting to {}".format(mxg))
+        try:
+            myg  = f["MYG"]
+        except KeyError:
+            myg = 0
+            print("MXG not found, setting to {}".format(mxg))
```
bendudson: MYG rather than MXG

tools/pylib/boutdata/collect.py
```diff
@@ -457,30 +520,68 @@ def findFiles(path,prefix):
     """
 
     file_list_nc = glob.glob(os.path.join(path, prefix+".nc"))
-    file_list_h5 = glob.glob(os.path.join(path, prefix+".hdf5"))
-    if file_list_nc != [] and file_list_h5 != []:
+    file_list_ncdf = glob.glob(os.path.join(path, prefix+".ncdf"))
```
bendudson: Can this be replaced by shorter code like:

```python
suffixes = [".nc", ".hdf5", ".ncdf", ".cdl", ".h5", ".hdf5", ".hdf"]
file_list = None
suffix = ""
for test_suffix in suffixes:
    files = glob.glob(os.path.join(path, prefix+test_suffix))
    if files:
        if file_list: # Already had a list of files
            raise IOError("Files with both {0} and {1} extensions are present. Do not know which to read.".format(suffix, test_suffix))
        suffix = test_suffix
        file_list = files
```

tools/pylib/boututils/datafile.py
```diff
@@ -571,49 +597,100 @@ def keys(self):
...
+        if self.handle is None:
+            return None
         try:
             var = self.handle[varname]
         except KeyError:
             return None
         return var.shape
 
-    def write(self, name, data):
+    def write(self, name, data, vartype=None):
```
ZedThree: This has a different signature than the method it's overriding. Please could you bring check that this, the base class and the other derived class (`DataFile_netCDF`), all have the same function signature for this method?

tools/pylib/boutdata/processor_rearrange.py
```diff
@@ -0,0 +1,92 @@
...
+                    return False
+
+            mz = s[maxdims - 1]
+            break
+
+    # Calculate total size of the grid
+    nx = mxsub * nxpe
+    ny = mysub * nype
+
+    return nxpe, nype, npes, mxsub, mysub, nx, ny, mz, mxg, myg
```
ZedThree: Given the number of return values, it might be better to wrap these up in a `dict` or `namedtuple`

tools/pylib/boututils/datafile.py
```diff
@@ -571,49 +597,100 @@ def keys(self):
...
+    def vartype_from_array(self, data):
+        try:
+            ndim = len(data.shape)
+        except AttributeError:
+            ndim = 0
+        if ndim == 4:
+            return 'Field3D_t'
+        elif ndim == 3:
+            # not ideal, 3d field might be time-evolving 2d field,
+            # 'Field2D_t', but can't think of a good way to distinguish
```
ZedThree: Please could you put this in the docstring? It feels like it could be a bit of a "gotcha"

d7919: By itself probably no easy way to distinguish -- could possibly add an optional that represents the time array so could then check if the first dim has the same length as that, but that might start getting a bit ugly.

bendudson: I think a longer-term solution should be to make more use of metadata in the dmp files: whether a variable is time-evolving, units etc.

ZedThree: This is actually a method on `Datafile`, so we could just look up `t_array` in the file and compare with that.

+1 for attributes. Not 100% sure how best to do it in for pylib though. One way might be to wrap `numpy.ndarray` so we can add our own attributes.

d7919: It's possible to define our own `dtype` which includes names so it _might_ be possible to define the relevant dtypes for the different types of data, but this might not be the nicest.

bendudson:
Thanks John. I think this looks good, but I'm going through it and trying to tidy up a little bit.
One thing I'm not sure about is the variable attribute "type", returned by DataFile.type and in the attributes dictionary. 

The NetCDF implementation doesn't give a type, so instead infers a type string ("Field3D_t" etc.) based on the dimensions. The HDF library does give type as an attribute, but I doubt it's the same as these strings. The type returned therefore depends on whether you're reading NetCDF or HDF5. 

I think the solution may be to have a "bout_type" which is one of the strings we decide; "Field3D_t" etc. "type" would be the underlying type, which may be format specific. 

I think this would remove the circular dependency between the `type` and `attributes` methods, which is currently mitigated by using a keyword `ensureTypePresent`. Unfortunately, the caching means that this keyword won't always be effective: If `attributes` is first called with `ensureTypePresent` false, it will cache the result so that a following call with `ensureTypePresent` true may not have the `type` field. Personally I think this keyword should be removed, "bout_type" always present.

johnomotani:
Thanks @bendudson!
The HDF5 'type' attribute was set by BOUT++ to exactly the same strings as in datafile.py, but 'type' wasn't a good name to use. I've renamed the 'type' attribute to 'bout_type' in both the C++ and Python I/O code now.
Also removed the circular dependency, instead making sure the attributes always have a 'bout_type'.

johnomotani:
I'm wondering if c722066 introduces a potential backwards-compatibility issue for people with existing HDF5 output files. I think maybe it doesn't in any important way, because the only use of the 'type' attribute in the original HDF5 implementation was to enable ``DataFile.dimensions()`` to work unambiguously. If someone wants to read time-evolving Field2D variables from old HDF5 outputs and write them back to new dump files to be appended to by BOUT++, then there would be an error, but I guess that is unlikely enough not to worry about? This is the only case I can think of where DataFile.dimensions() giving the wrong result has much of a consequence (in this case the time-dependent Field2D would not be given an extendible t-dimension).

PR #1002, opened 2018-04-17
Make sure DataIterator.zp() is not used with negative numbers

dschwoerer:
Depending in the sign, different logics need to be used.
ASSERT3 should be sufficient, as zp() and the more are used in inneer
loops, and no runtime dependence is expected.

### DIFF
--- a/include/bout/dataiterator.hxx
+++ b/include/bout/dataiterator.hxx
@@ -9,6 +9,7 @@
 #include <iterator>
 #include <iostream>
 #include "unused.hxx"
+#include "assert.hxx"
 
 #ifdef _OPENMP
 #include <omp.h>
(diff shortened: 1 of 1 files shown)


bendudson:
Failure of PR tests seems to be a MUMPS issue, not connected to this PR

ZedThree:
I don't really understand why it's failing. It's like it's got the latest configure from next, but the old Travis config -- #944 is the reason trying to compile with MUMPS fails, but it also removed it from the Travis job.

I'll merge this but will watch the build of next in case we need to back it out.

ZedThree:
This went in fine, I suspect it was just Travis hanging on to out-dated config for the PR build

PR #1005, opened 2018-04-18
Remove test-ballooning

### LINKED ISSUE #985: test-ballooning fails
Not sure why, error is:
Fail, maximum difference = 1.033736e+00

Branch: test-requirements

dschwoerer:
The method has been removed since v4.0.0
It was working in v3.1
Adding checks now and throw if any of the options are set is to late, as
most code is probably ported by now.

Resolves #985


ZedThree:
Just to note that `test-initial` is missing a test for `ballooning`, because I was too lazy to implement it at the time.

bendudson:
A note on this: test-ballooning was used during the transition, to ensure that the old and new methods gave the same results. Now that the old method has been removed (as of v4), this test is indeed obsolete. More tests should of course be added to test the shifted metric and ballooning transform methods.

PR #1012, opened 2018-04-20
Add explanaiton for test and disable by default

dschwoerer:
@d7919 - is the explanation sufficient, or should I extend?


d7919:
Thanks, that's very helpful. Two small typos in the first paragraph (there-->their and ether-->either). Thanks!

dschwoerer:
Thanks for spotting, fixed them now :smiley:

PR #1014, opened 2018-04-20
Warn if object files remain after cleaning

dschwoerer:
* If some cxx files get removed, that have been build, the corresponding
object files may not get removed.
* As libfast includes any .o file, they will get included, and may cause
issues.
* To avoid deleting still needed files, this is not done automatically,
but rather as a separate target.


ZedThree:
Is this a big problem? You can always just run `find src -name "*.o" -delete` if this happens.

dschwoerer:
That is what I used in the past - but sometimes it took me quite some time to figure out why this was happening.
It is not always the case of duplicated definitions, so I thought having it caught by clean would be nice.

It only takes ~3 ms on my machine, so I thought it may be useful ...

PR #1016, opened 2018-04-20
Print names first in tests

dschwoerer:
* If something hangs, you know what it is


ZedThree:
Unicode is just in the `make` tests -- I still can't see those tests in the travis output though

PR #1019, opened 2018-04-23
Even more manual improvements

ZedThree:
Separate PR from the other manual one, as this does have some actual code changes:

- Completely removes the unused `src/precon` makefile
- Makes `SolverType` a `typedef` rather than `define`
- Change various `string` to `std::string` to make doxygen realise a function does have documentation

Also:

- Silences all most all of the worst doxygen warnings
- Brings the doxygen coverage up to 38%
- Fix lots of LaTeX in documentation
- More consistent documentation between different Field types, including making sure everything has the optional `region` parameter documented

### DIFF
--- a/include/bout/fieldgroup.hxx
+++ b/include/bout/fieldgroup.hxx
@@ -97,7 +97,7 @@ class FieldGroup {
 
   /// Add a Vector2D \p v to the group.
   ///
-  /// Pointer's to this vector's components will be stored internally,
+  /// Pointers to this vector's components will be stored internally,
   /// so the lifetime of this variable should be longer than the
   /// lifetime of this group.
   void add(Vector2D &v) {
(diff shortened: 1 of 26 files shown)


include/bout/fieldgroup.hxx
```diff
@@ -45,62 +53,64 @@ class FieldGroup {
...
+  // so the lifetime of this variable should be longer
+  // than the lifetime of this group.
   void add(Field3D &f) {
     fvec.push_back(&f);
     f3vec.push_back(&f);
   }
 
+  /// Add a Vector2D \p v to the group.
+  ///
+  /// Pointer's to this vector's components will be stored internally,
```
bendudson: Pedantic typo: `Pointer's` -> `Pointers`

PR #1020, opened 2018-04-24
Cleanup calculus.py

### LINKED ISSUE #873: Pylib error
pylib writes errors to stdout:
`ERROR: scipy.signal module not available`
Should be written to stderr / raised as error

dschwoerer:
* Avoid repeating error messages
* Import from scipy later, as most function don't need it
* Whitespace changes

Resolves #873

### DIFF
--- a/tools/pylib/boututils/calculus.py
+++ b/tools/pylib/boututils/calculus.py
@@ -8,39 +8,30 @@ from __future__ import print_function
 from __future__ import division
 
 try:
-  from builtins import range
+    from builtins import range
 except:
-  pass
+    pass
 
 try:
-  from past.utils import old_div
+    from past.utils import old_div
 except:
-  def old_div(a,b):
-    return a / b
+    def old_div(a,b):
+        return a / b
 
-try:
-    from numpy import zeros, arange, pi, ones, array, transpose, sum, where, arange, multiply
-    from numpy.fft import rfft, irfft
-except ImportError:
-    print("ERROR: NumPy module not available")
-    raise
-try:
-    from scipy.signal import convolve
-except ImportError:
-    print("ERROR: scipy.signal module not available")
-    raise
+from numpy import zeros, arange, pi, ones, array, transpose, sum, where, arange, multiply
+from numpy.fft import rfft, irfft
 
 def deriv(*args, **kwargs):
     """Take derivative of 1D array
-    
+
     result = deriv(y)
     result = deriv(x, y)
-     
+
     keywords
-    
+
     periodic = False    Domain is periodic
     """
-     
+
     nargs = len(args)
     if nargs == 1:
         var = args[0]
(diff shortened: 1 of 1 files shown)


tools/pylib/boututils/calculus.py
```diff
@@ -8,39 +8,30 @@
...
+    pass
 
 try:
-  from past.utils import old_div
+    from past.utils import old_div
 except:
-  def old_div(a,b):
-    return a / b
+    def old_div(a,b):
+        return a / b
```
bendudson: I think this is fine for now, but at some point removing old_div calls would be good.

tools/pylib/boututils/calculus.py
```diff
@@ -105,24 +96,25 @@ def deriv2D(data,axis=-1,dx=1.0,noise_suppression=True):
...
       raise RuntimeError("Data too small to use 11th order method")
     tmp = array([old_div(-1.0,512.0),old_div(-8.0,512.0),old_div(-27.0,512.0),old_div(-48.0,512.0),old_div(-42.0,512.0),0.0,old_div(42.0,512.0),old_div(48.0,512.0),old_div(27.0,512.0),old_div(8.0,512.0),old_div(1.0,512.0)])
   else:
     if s[axis] < 9:
       raise RuntimeError("Data too small to use 9th order method")
-    tmp = array([old_div(1.0,280.0),old_div(-4.0,105.0),old_div(1.0,5.0),old_div(-4.0,5.0),0.0,old_div(4.0,5.0),old_div(-1.0,5.0),old_div(4.0,105.0),old_div(-1.0,280.0)])       
-  
+    tmp = array([old_div(1.0,280.0),old_div(-4.0,105.0),old_div(1.0,5.0),old_div(-4.0,5.0),0.0,old_div(4.0,5.0),old_div(-1.0,5.0),old_div(4.0,105.0),old_div(-1.0,280.0)])
+
+  from scipy.signal import convolve
```
bendudson: I think it makes sense to move this import into the function, since most functions here do not depend on it. I think however that it should be at the start of the function to make the dependency clearer. 

PEP8 recommends putting imports at the top of the file, but there is some discussion of situations when that is not optimal (and I suspect this is one of them): https://stackoverflow.com/questions/128478/should-import-statements-always-be-at-the-top-of-a-module

dschwoerer: Yes, I thought so. I often enough don't have scipy installed - in which case I can still use the other functions.

tools/pylib/boututils/calculus.py
dschwoerer: Oh, they really need to be removed.
They fall back to normal division - which is probably not what the intention was. Having always / would be better, then the errors would at least appear more often, and thus easier to notice + fix.

Should part of https://github.com/boutproject/BOUT-dev/commit/b3ec7246599af75a026c7bfad212d931fddf31a5 be undone?

bendudson: probably part of it... it depends if any code depends on the old behavior with integer division. In most cases it's safe to just replace `old_div(a, b)` with `(a/b)` but the occasional `int(a/b)` may be needed.

bendudson:
Cancelled the push travis test, as the changed code isn't in the test suite. PR test still running

PR #1013, opened 2018-04-20
Disable dependency tracking by default

dschwoerer:
* Dependency tracking can cause issues
* Is slower than normal build
* Can be enabled at configure time

Resolves #997 #998

PR #1022, opened 2018-04-24
Make.config rewrite

dschwoerer:
Alternative to #1013 

I think it should be mostly backward compatible with the old build system.

Note that the change in `tests/integrated/makefile` is not required.
It only helps to make a rebuild of the library from that directory faster, as not every test is trying to recompile the library.

This helps in general with simplifying `make.config` - the old had at least 4 levels of nested ifs - this one has only two (for finding all `cxx` files).

Obsoletes #1014


ZedThree:
Some nice bits in here, but I think the better solution is to scrap (most of) the recursive make stuff, and just hardcode the list of source files in the top-level `makefile` (doesn't even need to be in `make.config.in`). We already do hardcode the source files, just spread across the directory-level `makefile`s. Maintaining this list is not a big deal.

Also, did you mean to remove the install/uninstall targets?

dschwoerer:
I am worried that if we break the system to much - then codes that rely on our make system, might not work any more.
I found yesterday in the docs that you can just set lib as target, to get your code added to the library.
So while it is probably easy to adopt for most of the changes, it would probably be easier for projects if we remain closer to what we have now.

I personally would prefer to have several files to include the cxx files and directories. That prevents merge conflicts, can easily be generated at configure time (around 300 ms).

One thing I haven't really figured out yet, is how to deal with file auto-generation.
Having that all moved to the main makefile might make that really long and unreadable.
Maybe we should allow to have makefiles included? That would allow to keep it quite generic, and allow abitrary rules to create files. It also allows to have the rules defined somewhat locally.

Another option would be to enforce a certain naming scheme, but that makes it quite hard to specify additional dependencies. I am thinking on the aiolos branch, where I parse some other cxx files, to generate some intermediate files, and then call some scripts called `%.in.py` to generate `%` - but if additionally some of the other files changes, they need to be regenerated as well.


About the install/uninstall targets - no, I just took some parts that I needed - the should be readded - but I would first sort out the difficult stuff:
 * To what extend do we want to break backwards compatibility?
 * What about non-standard generation rules?
 * *other stuff I forgot*

ZedThree:
Version 5.0 is likely to break lots of things, as it will include fully 3D metrics and treat all three directions in a more consistent fashion. Let's revisit the build system then, when we won't have to worry so much about not breaking existing things. 

Closing as replaced with #1031

PR #1023, opened 2018-04-25
Fix elm-pb case with compressional effects

bendudson:
Requires some changes to the Div_par_CtoL etc. functions. These are used instead of StaggerGrids in some of the terms (not self consistently).

### DIFF
--- a/examples/elm-pb/elm_pb.cxx
+++ b/examples/elm-pb/elm_pb.cxx
@@ -931,17 +931,7 @@ int physics_init(bool restarting) {
   V0net.z = -Dphi0;
 
   U0=B0vec*Curl(V0net)/B0;     //get 0th vorticity for Kelvin-Holmholtz term
-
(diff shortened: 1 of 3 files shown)


dschwoerer:
I tried this, and it still fails: [output](https://omadesala.alfa1024.info/~bout-test/2018-04-27_14:32/raw/mms.log)

I think the issue is that CtoL and LtoC set the location to the staggered location and expect the field to be at the respective location.
But the elm-pb test has staggerGrids disabled - so this cannot work.

bendudson:
Odd; that elm-pb test has compress0 set to false, so the terms "fixed" by this PR shouldn't be affected. I will try to reproduce this.

dschwoerer:
I think it was broken before ...

I will have a go at repairing it ...

dschwoerer:
Oh, the elm_pb is a different file from the example ...
Should both files be the same?

PR #1026, opened 2018-04-27
Increase the precision of stored BoutReals in Options

ZedThree:
Not quite a bug, but currently when reading in double precision numbers using `Options::get`, they are rounded to something like single precision (possibly just 8 characters?). This increases the precision when reading them in, as well as ensuring they are stored in scientific notation.

This does change how we write out `Option`s:
```diff
...
-   "real_key = 6.17e+23",
+   "real_key = 6.17000000000000006e+23",
...
```
which required modifying the `OptionsINI` test. Shouldn't break backwards compatibility, but could be a little annoying. Not 100% sure of a solution to that, as I don't think we store what type it "should" be.

### DIFF
--- a/src/sys/options.cxx
+++ b/src/sys/options.cxx
@@ -1,10 +1,11 @@
 #include <options.hxx>
 #include <boutexception.hxx>
 #include <utils.hxx>
-#include <sstream>
 #include <output.hxx>
-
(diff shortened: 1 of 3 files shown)


ZedThree:
Yeah, I'm not so keen on that either, but I'm not sure of a way round it yet. At least not one that's not vaguely horrible

PR #1029, opened 2018-04-27
Simplify configure with petsc

dschwoerer:
* I reverted e50674f as it was causing issues
* I simplified configuring. Now it should work as described in the manual
* For distributions, I thought of reverting to `pkgconf`

### DIFF
--- a/configure
+++ b/configure
@@ -650,6 +650,8 @@ LTLIBOBJS
 SCOREPPATH
 sundials_config
 SLEPC_VARS
+PETSC_ARCH
+PETSC_DIR
 PETSC
 PARALLELHDF5_TYPE
 PARALLELHDF5_LIBS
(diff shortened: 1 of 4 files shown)


ZedThree:
Thanks for having a stab at this, I know how painful configuring with PETSc is. I wouldn't be opposed to just requiring PETSC_DIR and PETSC_ARCH to be set if we can't immediately guess.

I'll give this a test on a few different systems

dschwoerer:
As long as it works, I am happy with others solution as well ... 
Before it basically ignored setting --with-petsc.

Are there any issues arising if PETSC_ARCH is empty / the wrong part or the path?

ZedThree:
Setting the whole path to PETSC_DIR and PETSC_ARCH empty is fine. PETSC_ARCH is just a convenience thing.

PR #1030, opened 2018-04-27
netCDF-cxx library needed for Ubuntu

jenuis:
Add netCDF-cxx library for installing dependencies for Ubuntu


ZedThree:
Either this package, or `libnetcdf-c++4-dev` should be sufficient. We should support both, these correspond to the two different NetCDF interfaces in BOUT++.

PR #1032, opened 2018-04-27
Make Solvers discoverable

ZedThree:
I'm happy to split this up into separate PRs if it makes it more readable. More detailed explanation to come as well.

See tests/integrated/test-solvers for end results, but basically makes Solvers somewhat discoverable, as in you can list what Solvers are available.

This PR contains:
- generic factory template
- use of Factory to re-implement SolverFactory (without changing API, although it was private before, so could update it...)
- fix a bunch of bugs and inconsistencies in some of the solvers
- add a test for all of the available solvers

TODO:
- add `isEigenSolver` and `isTimeSolver` to Solvers to automatically cull solvers from test
- add separate tests for the eigen solvers and SNES (steady-state solver)
- merge all the PETSc implementations into one working one


ZedThree:
I realise the `runtest` for test-solver could be replaced with a shell script that just does `make && ./test_solver`.

Also codacy has things to say about this PR.

Jobs for Monday.

dschwoerer:
No need for a runtest script.
Just set `TARGET='runtest'` in make.config and it will generate the runtest binary.

On more important means:
If one of the solvers throws, the `test-solver` still passes.
Is that intended?

ZedThree:
Thanks, I missed handling the exceptions correctly. That's sorted now.

Also, it turns out some nature of script is still needed to at least compile the test. A script is also useful to make sure a log gets saved. A shell script would still be fine, as `tee` is in coreutils for BSD as well, but I've stuck with the python one.

PR #1034, opened 2018-04-28
Simlify and extend Function lookup

### LINKED ISSUE #741: Setting derivatives

dschwoerer:
* Deduplicated code by templating
* Return DiffLookup and convert to appropriate type as needed
* Check %sStag first, for staggered derivatives
* Check general diff location, if nothing in ddx is specified

- [x] Resolves #741 
- [x] Checks needed
- [x] Documentation needed

### DIFF
--- a/manual/sphinx/user_docs/bout_options.rst
+++ b/manual/sphinx/user_docs/bout_options.rst
@@ -272,10 +272,13 @@ and problem.
 Differencing methods
 --------------------
 
-Differencing methods are specified in the section (``[ddx]``,
-``[ddy]``, ``[ddz]`` and ``[diff]``), one for each dimension.
-The ``[diff]`` section is only used if the section for the dimension
-does not contain an option for the differencing method.
+Differencing methods are specified in the section (``[meesh:ddx]``,
+``[mesh:ddy]``, ``[mesh:ddz]`` and ``[mesh:diff]``), one for each
+dimension. The ``[mesh:diff]`` section is only used if the section for
+the dimension does not contain an option for the differencing method.
+Note that ``[mesh]`` is the name of the section passed to the mesh
+constructor, which is most often ``mesh`` - but could have another
+name, e.g. if multiple meshes are used.
 
 -  ``first``, the method used for first derivatives
 
@@ -292,7 +295,7 @@ central, W = WENO), and the number gives the order.
 The staggered derivatives can be specified as ``FirstStag`` or if the
 value is not set, then ``First`` is checked.
 Note that for the staggered quantities, if the staggered quantity in a
-dimension is not set, first the staggered quantity in the ``[diff]``
+dimension is not set, first the staggered quantity in the ``[mesh:diff]``
 section is checked. This is useful, as the staggered quantities are
 more restricted in the available choices than the non-staggered
 differenciating operators.
--- a/manual/sphinx/user_docs/bout_options.rst
+++ b/manual/sphinx/user_docs/bout_options.rst
@@ -272,7 +272,7 @@ and problem.
 Differencing methods
 --------------------
 
-Differencing methods are specified in the section (``[meesh:ddx]``,
+Differencing methods are specified in the section (``[mesh:ddx]``,
 ``[mesh:ddy]``, ``[mesh:ddz]`` and ``[mesh:diff]``), one for each
 dimension. The ``[mesh:diff]`` section is only used if the section for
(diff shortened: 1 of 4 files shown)


manual/sphinx/user_docs/bout_options.rst
```diff
@@ -272,8 +272,10 @@ and problem.
 Differencing methods
 --------------------
 
-Differencing methods are specified in three section (``[ddx]``,
-``[ddy]`` and ``[ddz]``), one for each dimension.
+Differencing methods are specified in the section (``[ddx]``,
```
johnomotani: These sections are `[mesh:ddx]`, `[mesh:ddy]` and `[mesh:ddz]` now, would be good to update the documentation here. I guess also `[mesh:diff]`?

dschwoerer:
Should be ready for merging :+1:

src/mesh/index_derivs.cxx
```diff
@@ -570,91 +548,79 @@ Mesh::flux_func sfFDDX, sfFDDY, sfFDDZ;
...
-  printFuncName(method);                         // Print differential function name
-  f = lookupFluxFunc(table, method);
+  label = def;
+  if (staggerGrids) {
+    output_info.write("\tStag. %-6s: ", name.c_str());
+    for (auto &_name : {name + "stag", name}) {
+      for (auto &opts : options) {
+        if (opts->isSet(_name)) {
+          opts->get(_name, label, "");
+          goto endloop;
```
ZedThree: Is the `goto` necessary here? Could you not just do

```cpp
    opts->get(_name, label, "");
    sf = lookupFunc(stable, label);
    return;
```
Am I missing something?

ZedThree:
I think this is good, and if anything doesn't go far enough. I've been playing about today making the various deriv tables maps, and it generally makes things even nicer than this PR already does. It also gives the added advantage of being able to register new derivatives in the user code.

This could go in now, once my question about the goto has been addressed, and I'll submit another PR turning things into maps, and we can see how that looks.

src/mesh/index_derivs.cxx
dschwoerer: As it only duplicates one line of code, it should be fine ...
In general `goto`s are underrated - they can be quite useful - e.g. for breaking loops :)

manual/sphinx/user_docs/bout_options.rst
```diff
@@ -272,8 +272,13 @@ and problem.
 Differencing methods
 --------------------
 
-Differencing methods are specified in three section (``[ddx]``,
-``[ddy]`` and ``[ddz]``), one for each dimension.
+Differencing methods are specified in the section (``[meesh:ddx]``,
```
bendudson: typo `meesh`

PR #1037, opened 2018-04-30
Improve mpich detection

dschwoerer:
Still not good, but it is working again on travis.

### DIFF
--- a/tests/requirements/mpich
+++ b/tests/requirements/mpich
@@ -1,23 +1,31 @@
-#!/usr/bin/env python
+#!/usr/bin/env bash
 
 # Tests if mpich is being used
 # NOTE: A more robust method is needed!
 
-from boututils.run_wrapper import shell
-from sys import exit
+SCRIPTPATH="$( cd "$(dirname "$0")" ; pwd -P )"
 
-# Test working on ubuntu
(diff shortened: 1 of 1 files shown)


dschwoerer:
The old script causes unintended test skips, see e.g:
https://travis-ci.org/boutproject/BOUT-dev/jobs/373276923#L2547

bendudson:
Thanks David for all your work on this. It seems to be a frustratingly difficult thing to do, find out which MPI implementation we're using. I'm not convinced though that this is a better solution; this is probably my problem but there is too much black magic here. I would prefer to find a more robust method.

dschwoerer:
Please note that this version is working, while the current version seems to be broken. `test-multigrid_laplace` gets currently skipped on the openmpi runs, as it seems to detect that mpich is used.

bendudson:
Appreciated; I guess I don't really understand how this new version works

ZedThree:
My understanding is that we were skipping this test with mpich because there were problems with openmpi. Are those problems now sufficiently resolved that we can just use openmpi and remove this check entirely?

dschwoerer:
Should be clearer now:
https://github.com/boutproject/BOUT-dev/blob/mpich_detection_rebase/tests/requirements/mpich
It also has a fallback - using the help which should contain a link to the project website

PR #1040, opened 2018-04-30
D2dy2 bugfix test2

dschwoerer:
@johnomotani 
It seems there hasn't been a merge for a while, so I did not want to push directly.

I wrote a tests, and it seems `diffloc` is the right thing to set :+1:


johnomotani:
Cheers @dschwoerer! :+1:

PR #1041, opened 2018-04-30
Deprecate ^ operator as cross product

### LINKED ISSUE #983: Deprecating operator^
Should we deprecate (and remove this) completely?
It seems currently to be only used in vector?d as cross product.
`grep 'operator[[:space:]]*\^' include/ -r`
I guess cross(vector1,vector2) would be easier to read?
Further we could remove the note from the user manual about `^` and move it to the developer manual for future reference as to why it was removed.

dschwoerer:
Resolves #983


bendudson:
Agree this is probably for the best. It's a shame,  but the low operator precedence creates too many opportunities for bugs. Thanks!

Issue #736, opened 2017-10-26
Field location is not stored in dump files


dschwoerer:
That would be nice, so that the `boutcore.Field3D.fromCollect()` could return a Field3D with an appropriate location.

Maybe an additional char/string variable could be added, to store for each field the location, something like `<name>_location`?

johnomotani:
In both NetCDF and HDF5 files you can add attributes to variables.
http://www.unidata.ucar.edu/software/netcdf/docs/group__attributes.html
https://support.hdfgroup.org/HDF5/doc1.6/UG/13_Attributes.html

This might be neater than adding extra variables, but requires adding code to each fileio implementation.

bendudson:
Agree, this would be very nice. Attributes sounds like a good way to go, though string support is also on the list of requests (issues #222 and #426 ). 
The output system really needs changing, so that it doesn't depend on storing pointers to user variables. This would be a breaking change. For now we're stuck with the original design, so I may find time to implement this.

PR #1042, opened 2018-05-01
Save field cell location attributes

bendudson:
The datafile class now sets an attribute "cell_location" for each 2D and 3D field (and vector component).
Fixes issue #736.

The attribute is a string which is the same as the enum name e.g. CELL_CENTRE, CELL_YLOW etc.

To do this, a const map has been added to bout_types.hxx, together with an inline function CELL_LOC_STRING(). This replaces `strLocation` in interpolation.hxx, which used a switch.

Together with PR #770 (saving attributes in NetCDF) and #996 (reads attributes into a BoutArray class), this now works (in test-io):

```
> f3d = collect("f3d")
> f3d.attributes
{'cell_location': 'CELL_CENTRE', 'bout_type': 'Field3D'}
```

Added some checks to `test-io` to check that the cell_location attribute is read correctly.


dschwoerer:
Support for boutcore's fromCollect will follow when this is merged :+1:

johnomotani:
We'll need to add writing of the Vector3D::location when #994 is merged.

bendudson:
At the moment Vector3D objects are written out as separate components, each of which has a location. If the Vector3D::location is consistent with the (x,y,z) locations then this should be ok I think.

johnomotani:
@bendudson right! yes, that should be fine. Sorry, I forgot how vectors got written.

I added attribute writing for HDF5, and an updated test-io_hdf5 passes, with ``cell_location`` being written correctly.

Issue #999, opened 2018-04-17
Bout Extended Slow Tests


dschwoerer:
Currently there are many tests that fail.

I started building a CI system that runs all the tests, and reports the output:
https://omadesala.alfa1024.info/~bout-test/

It currently builds all branches in the main repository that include the patch for enabling all tests (196e809 ; #989 )

I am not sure whether adding an issue for this is the right place, but I thought it might be useful to know ...

I hope to enable PETSc in the build, to increase the coverage further ...

d7919:
Does the GBS mms test ever pass? I've tried with various branches and after many hours (per run) it never seems to work. The version reported on these tests seems to fail after a relatively short time (<200s) but when I've been running these it only fails after several hours -- are you patching these tests in any way?

bendudson:
It did at one point...  Once upon a time.

dschwoerer:
I haven't patched the test, it probably got worse since you last tested ...

Discussion of GBS test is here: #1502 

---------------

I have disabled to run all branches, as that significantly increased the load (and thus latencies) ...
I am thinking of adding support to request to run it on pull requests, I thought e.g. via adding a tag for pull requests, that can trigger the run, or a comment to request it - any preferences/thoughts?

PR #1050, opened 2018-05-03
Cleanup PETSc interface

ZedThree:
Replaces #1029 

- Removes the PETSc version specific solvers and configure. Now just checks that PETSc version is at least 3.4.0.
- Fixes various issues with `PetscSolver`
- Fixes minor issues with some other optional solvers
- Removes `EmptySolver`
- Turns off IO in test-solver
- Verified working "out of the box" with PETSc versions 3.6, 3.8, 3.9.
    - 3.7 just needs a smaller timestep
    - 3.4 and 3.5 are more tricky and possibly just broken. Note that 3.4 is the version on Travis, but test-solver is an integrated test, and we only run the unit tests with PETSc enabled

Some information about the various PETSc versions that might be useful:

PETSc Versions on various systems
---------------------------------

Operating systems:

- Fedora EPEL 6: 3.7
- openSUSE Leap 15.0: 3.8
- Ubuntu 14.04: PETSc 3.4
- Ubuntu 16.04: PETSc 3.6
- Ubuntu 18.04: PETSc 3.7

Supercomputers:

- Archer: 3.5-3.7
- Marconi: 3.7-3.8

test-solver
-------------

- 3.4:
    - euler doesn't advance
    - rk only has single scheme, does not support interpolation. This is a problem as we explicitly call `TSInterpolate`
    - beuler/theta/cn not adaptive
- 3.5:
    - euler doesn't advance, but passing `-ts_type rk -ts_rk_type 1fe -ts_adapt_type none` does, which should be equivalent
    - rk schemes don't get correct answer, although close
    - beuler/theta/cn also not correct but close
- 3.6:
    - euler does correct thing (needs higher NOUT to get accuracy)
    - rk default scheme is correct, 5dp close
    - beuler/theta/cn get close still
- 3.7:
    - euler as expected
    - rk schemes get close
    - beuler/theta/cn get close
- 3.8:
    - euler as expected
    - rk default scheme is correct, 5dp close
    - beuler close, theta/cn correct
- 3.9:
    - euler as expected
    - rk default scheme is correct, 5dp close
    - beuler close, theta/cn correct


bendudson:
Thanks @ZedThree ! 
Looks like the PETSc version check fails in configure:
```
configure: error: PETSc version must be at least 3.4.0
```
but the version installed on Travis looks like 3.4.2

ZedThree:
It's due to the PETSc headers being installed under /usr/lib/petscdir/3.4.2/include/, but `PETSC_DIR/PETSC_ARCH` is /usr/lib/petscdir/3.4.2/linux-gnu-c-opt/.

Could be fixed either with passing `CPPFLAGS=-I/usr/lib/petscdir/3.4.2/include/`, or with some configure magic `grep`ping `petscvariables` for `PETSC_CC_INCLUDES` and using that to find `petscversion.h`

dschwoerer:
This is also causing issues with my installed petsc.

It is a shame that petsc does not provide better support for installing :-1:

dschwoerer:
If the make.config is broken, that does not make configure fail:
```
config.status: creating make.config
/usr/lib64/mpich/lib/petsc-3.8.3/conf/variables:118: /lib64/mpich/lib/petsc/conf/petscvariables: No such file or directory
gmake: *** No rule to make target '/lib64/mpich/lib/petsc/conf/petscvariables'.  Stop.
/usr/lib64/mpich/lib/petsc-3.8.3/conf/variables:118: /lib64/mpich/lib/petsc/conf/petscvariables: No such file or directory
gmake: *** No rule to make target '/lib64/mpich/lib/petsc/conf/petscvariables'.  Stop.
** configure continues **
```
(but that is not caused by this change)

Fedora currently needs this fix (bug in petsc)
```
sudo ln -s /usr/lib64/mpich/lib/petsc{-3.8.3,}
```
But with that it works :+1:

dschwoerer:
```
./configure --with-petsc=/usr/lib64/mpich/lib/petsc-3.8.3
```
I have reported this bug [here](https://bugzilla.redhat.com/show_bug.cgi?id=1566824), but had so far no response

```
./configure --with-petsc=/usr/lib64/mpich/
```
Is working as expected - fails without the workaround, and works with.

dschwoerer:
I tried it for BEST #999 - and it fails:
`./configure $flags --with-petsc=$HOME/petsc/$petsc/arch-linux2-cxx-debug`
configure verifies that the version is sufficient, but then it fails to build.
Do you save the configured PETSC_DIR  and PETSC_ARCH in the make.config?

https://omadesala.alfa1024.info/~bout-test/2018-05-03_17:14/

ZedThree:
We shouldn't need PETSC_DIR/ARCH in make.config, just this line:
```make
include ${PETSC_CONFDIR}/variables
```

I think I understand the different PETSc installation scenarios a bit better now:
- `petscvariables` is installed under `$PETSC_DIR/$PETSC_ARCH/[lib/petsc/]conf`
- `variables` is installed under `$PETSC_DIR/[lib/petsc/]conf`

with `[lib/petsc/]` being PETSc version specific.

There are two different ways to "install" PETSc:
- Just `./configure .. && make ...`: compiles and installs it with `$PETSC_DIR=$(pwd)` and `$PETSC_ARCH` set to something like `arch-linux2-cxx-debug/`. This is complicated by the face that it might possibly be essentially moved wholesale somewhere else, but PETSC_DIR **and** PETSC_ARCH are both needed to use this install
- `./configure --prefix=/path/ ... && make ... && make install ...`: installs it with `$PETSC_DIR=/path/` and `$PETSC_ARCH=""`, and only PETSC_DIR is needed.

This means that `petscvariables` and `variables` may be in the same directory, or `variables` might be in a different one.

From what I can tell, different distros have installed PETSc using both methods, and possibly changed which method with different releases.

In this version of configure, we only set `PETSC_DIR` from the argument to `--with-petsc`, as anything else will just be guessing. If you've installed PETSc with the first method above, and I think that's what you've got on your system @dschwoerer, you'll need to set PETSC_DIR/ARCH in your environment, or at least on the command line for both configure and make and not supply an argument to `--with-petsc`. If this latest commit works, I'll add all this to the docs.

I have to say I think PETSc have made this overly complicated, and they should just install under prefix like literally everybody else.

dschwoerer:
Some testing:
`./configure --with-petsc=$HOME/soft/petsc-rel/petsc-3.9.0/arch-linux2-cxx-debug/lib/petsc/` fails to configure
`./configure --with-petsc=$HOME/soft/petsc-rel/petsc-3.9.0/` fails to make, which is imho not acceptable.

With #1029 I was able to configure as it (IMHO) should be done with --with-petsc=path.
Would you be happy if I made that work (again)?

Further, we should store PETSC_ARCH, as `conf/variables` includes  `$PETSC_ARCH/[lib/petsc]/conf/petscvariables` - which might set a different $PETSC_ARCH and thus cause issues.

Setting PETSC_DIR and ARCH does not work with:
`./configure  --with-petsc PETSC_DIR=$HOME/soft/petsc-rel/petsc-3.9.0 PETSC_ARCH=arch-linux2-cxx-debug`
Configure works fine, but make works.
I know, the workaround is simple, but I should not have to ...

ZedThree:
`--with-petsc PETSC_DIR=... PETSC_ARCH=...` should now work in the vast majority of cases.

Now sets `PETSC_DIR/ARCH` in `make.config` so they don't need to be set in the environment.

Also tries a little harder to find petscversion.h. This should make it work on Fedora and other distributions that don't obey the `PETSC_DIR/ARCH` setup.

`--with-petsc=$PETSC_DIR` works if and only if `PETSC_ARCH` is supposed to be empty. No attempt is made at trying to work it out. This is merely a handy shortcut.

ZedThree:
I've checked this on Fedora 28 -- it fails at the make stage because they don't use PETSC_DIR and PETSC_ARCH consistently. At (our) configure time, PETSC_DIR=/usr/lib64/openmpi but at make time, PETSC_DIR=/usr. This is ultimately because they use sed to replace some but not all instances of PETSC_DIR/ARCH in the variables/petscvariables.

I can make a note of that in the manual (the error message now directs people to the manual if configuring with PETSc fails).

configure.ac
```diff
@@ -593,86 +593,95 @@ AS_IF([test "x$with_lapack" != "xno"], [
...
-      AC_MSG_NOTICE([***********************************])
-      AC_MSG_WARN([Unrecognised PETSc version])
-      AC_MSG_NOTICE([MAJOR VERSION: $PETSC_VERSION_MAJOR])
-      AC_MSG_NOTICE([MINOR VERSION: $PETSC_VERSION_MINOR])
-      AC_MSG_NOTICE([RELEASE      : $PETSC_VERSION_RELEASE])
-      AC_MSG_NOTICE([])
-      AC_MSG_NOTICE([Trying PETSc-dev interface])
-      AC_MSG_NOTICE([***********************************])
-      CXXFLAGS="$CXXFLAGS -DBOUT_HAS_PETSC -DBOUT_HAS_PETSC_DEV"
+# PETSc changed the location of the conf direction in 3.5, so we
```
d7919: Typo, direction-->directory

d7919:
Should the changes to cvode/arkode/ida be a separate PR?

ZedThree:
They're part of removing `EmptySolver`, which should've been part of my other PR adding the generic factory. That's done here just because tidying up PETSc involved tidying up other solver bits too

ZedThree:
I've done some further testing: in 3.4, the call to `TSInterpolate` can be turned off by passing `-interpolate` on the command line. This causes the rk TS to run, but it doesn't actually work (`Very small steps: 0.000000`, I think we looked at this before, don't remember the conclusion). Euler *will* work, but needs more timesteps. 

Unfortunately, `Solver` is set to take `NOUT` and `TIMESTEP` from the global section of the `Options`, so it's actually really tricky to adjust those parameters for just one solver in `test-solver`. It might be possible to make `Solver` check it's local options section for them first I guess?

I think we definitely need to revisit the use of `TSInterpolate` here anyway, but I still don't really grok PETSc enough to do it yet.

src/solver/impls/petsc/petsc.cxx
```diff
@@ -118,10 +118,6 @@ int PetscSolver::init(int NOUT, BoutReal TIMESTEP) {
   nout = NOUT;
   tstep = TIMESTEP;
 
-  // Set the rhs solver function
-  // func = f;
-
-  PetscInt n2d = n2Dvars();       // Number of 2D variables
```
d7919: I wonder if that fact that this isn't used is actually indicating a bug -- `n3d` is used to determine how many variables there are at a given grid point etc. but presumably if we have evolving 2d fields we should include these as well? Take the extreme of a 2D problem, I think the PETSc solver may behave in a strange way in that case.

src/solver/impls/petsc/petsc.cxx
```diff
@@ -359,26 +365,13 @@ int PetscSolver::init(int NOUT, BoutReal TIMESTEP) {
...
-    /* number of z points */
-    PetscInt nz  = mesh->LocalNz;
-
     /* number of degrees (variables) at each grid point */
     PetscInt dof = n3Dvars();
 
     /* Stencil width. Hardcoded to 2 until there is a public method to get mesh->MXG */
-    PetscInt n = local_N; //mesh->xend*mesh->yend*nz*dof; //<- that doesn't seem to work. Why is n3Dvars()*nz?
-    // PetscInt n = MXSUB*MYSUB*nz*dof;
     PetscInt sw = 2;
```
d7919: Code comment above suggest the stencil width (`sw`) should be MXG, i.e. this should be `sw = mesh->xstart;`.

src/solver/impls/petsc/petsc.cxx
```diff
@@ -359,26 +365,13 @@ int PetscSolver::init(int NOUT, BoutReal TIMESTEP) {
...
-    //PetscInt MYSUB = mesh->yend - mesh->ystart + 1;
-
-    PetscInt nx = mesh->xend;//MXSUB;
-    PetscInt ny = mesh->yend;//MYSUB;
-
-    /* number of z points */
-    PetscInt nz  = mesh->LocalNz;
-
     /* number of degrees (variables) at each grid point */
     PetscInt dof = n3Dvars();
```
d7919: I think this code might need generalising to handle 2Ds as well

src/solver/impls/petsc/petsc.cxx
```diff
@@ -359,26 +365,13 @@ int PetscSolver::init(int NOUT, BoutReal TIMESTEP) {
...
 
     /* Stencil width. Hardcoded to 2 until there is a public method to get mesh->MXG */
-    PetscInt n = local_N; //mesh->xend*mesh->yend*nz*dof; //<- that doesn't seem to work. Why is n3Dvars()*nz?
-    // PetscInt n = MXSUB*MYSUB*nz*dof;
     PetscInt sw = 2;
-    PetscInt dim = 3;
     PetscInt cols = sw*2*3+1;
```
d7919: Would be nice to note what these numbers represent.

src/solver/impls/petsc/petsc.cxx
```diff
@@ -399,180 +392,40 @@ int PetscSolver::init(int NOUT, BoutReal TIMESTEP) {
...
 #endif
       output_info << "compute J by slow fd is done.\n";
       //ierr = MatView(J,PETSC_VIEWER_STDOUT_WORLD);CHKERRQ(ierr);
     } else { // get sparse pattern of the Jacobian
       throw BoutException("Path followed in PETSc solver not yet implemented in general "
                           "-- experimental hard coded values here. Sorry\n");
-
-      output_info << "get sparse pattern of the Jacobian...\n";
-
-      if(n2Dvars() != 0) throw BoutException("PETSc solver can't handle 2D variables yet. Sorry\n");
```
d7919: I think this throw is related to the earlier comments about not using n2d etc. I think we should probably fix to allow 2D vars, if not we should probably keep this throw and move it to near the start.

src/solver/impls/petsc/petsc.cxx
ZedThree: I think it probably does already handle 2Ds ok. This bit is for the preconditioner, and only if the user hasn't supplied their own. And then it's only for the preallocation of J and is just an optimisation.

Certainly it works with a Field2D in the test-solver

ZedThree:
I think it does handle evolving Field2Ds -- I just checked that test solver works with a Field2D.

The path that throws is if the user wants a preconditioner, hasn't supplied their own function and doesn't want to use the PETSc `SNESComputeJacobianDefault` ("slow finite difference") function. A general preconditioner is probably a big project in itself, but maybe we could handle the flow a bit better. I've made the error message hopefully a bit more helpful now at least

PR #1052, opened 2018-05-04
Make unit tests less verbose

### LINKED ISSUE #1039: Unit tests to noisy
The unit tests produce roughly 1200 lines.

If one fails, I am having troubles finding that one.

Without the unit tests there would be only ~500 lines on travis.

dschwoerer:
Resolves #1039


dschwoerer:
To prevent issues with awk not being present, and thus the test failing - I could add a check that awk is present, and if it is not present, fall back to cat, to print everything:
```
if which awk >/dev/null 2>&1
then
    exec awk '{ ....
else
    exec cat -
fi
```

That makes the code (slightly) more complicated, and I assume no awk being present should not happen often ...

I just testid and it runs with `gawk`, `mawk` and `nawk`

PR #1053, opened 2018-05-04
Continue travis tests after failure

dschwoerer:
Continue even if building some of the tests fail, or the unit or
integrated tests fail.


bendudson:
I'm not convinced that this is the right thing to do. In my opinion the travis tests are there to check the code, not to debug our code, and should report as soon as possible when checks fail.  Happy to merge this though if more people think it would be a useful feature.

dschwoerer:
There seems to be no interest - keeping the load on travis low is also useful to keep it fast(ish)
Should we remove the -k flag from make?

PR #1056, opened 2018-05-04
Petsc tidy simplify configure

dschwoerer:
* Accept --with-petsc=$PETSC_DIR/$PETSC_ARCH
* Accept --with-petsc PETSC_DIR=... PETSC_ARCH=...
* Accept --with-petsc=$PETSC_DIR
* Work with installed petsc on fedora
* If we cannot compile with petsc during configure, say that, rather then say the version is not supported
* fail if make.config is broken

### DIFF
--- a/configure
+++ b/configure
@@ -8833,6 +8833,8 @@ fi
 if test "x$with_petsc" != "x" && test "$with_petsc" != "no"; then :
 
 
+  _PETSC_DIR="$PETSC_DIR"
+  _PETSC_ARCH="$PETSC_ARCH"
   if test "$with_petsc" != "yes"; then :
 
     PETSC_DIR="$with_petsc"
@@ -8862,6 +8864,7 @@ eval ac_res=\$$as_ac_File
 $as_echo "$ac_res" >&6; }
 if eval test \"x\$"$as_ac_File"\" = x"yes"; then :
 
+    conf=conf
     PETSC_CONFDIR=${PETSC_DIR}/conf
 
 else
@@ -8885,6 +8888,7 @@ eval ac_res=\$$as_ac_File
 $as_echo "$ac_res" >&6; }
 if eval test \"x\$"$as_ac_File"\" = x"yes"; then :
 
+      conf=lib/petsc/conf
       PETSC_CONFDIR=${PETSC_DIR}/lib/petsc/conf
 
 else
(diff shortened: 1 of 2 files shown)


ZedThree:
This is clever, but I'd really like to keep configure as simple as possible. As soon as we start trying to manipulate paths, it becomes much more difficult to maintain, and much more fragile.

Setting `PETSC_DIR/ARCH` in `make.config` might be ok, although PETSc users might be expecting to be able to swap `PETSC_ARCH` at make-time which this would break. We could just not support that though.  
Also a better configure error message for not finding PETSc is good, as is failing if make.config is broken.

I only expect the following two situations to actually work:
- if PETSc is installed the non-prefix way (i.e. the way PETSc wants you to):
  ```
  export PETSC_DIR=...
  export PETSC_ARCH=...
  ./configure --with-petsc
  make
  ```
- if PETSc is installed using `--prefix`: 
  `./configure --with-petsc=$PETSC_DIR && make`
  This is only a convenience for the case when `PETSC_ARCH=""`

As far as I can tell, Fedora's build of PETSc is broken as `PETSC_CC_INCLUDES` is wrong in `petscvariables` (should be `-I/usr/include/petsc`, not `-I/usr/include`). That should be fixed there.

dschwoerer:
I think that while it is fragile (I know that) if it fails to do the right thing, it falls bake to the initial things, which I think should minimize the fragility. Supporting the PETSc supplied by distributions is beneficial as you don't have to deal with configuring and compiling PETSc yourself.

While it is beneficial for distributing BOUT++ as a package, where it is also beneficial if it just works - I can apply patches before compiling and configuring ...

If you think we should allow PETSC_DIR and PETSC_ARCH to be changed, we can change the assignment to a conditional assignment (`?=`) - that way it might work most of the times, unless PETSC_DIR or ARCH is set to something unintended (e.g. an empty string).

I really would prefer if I can install without export and by providing the path. I prefer if I don't have to read the documentation, as having thinks just work with as few surprises as possible is nice.

I know that PETSc is currently broken in fedora, but there is already a new build (version 3.9.0) that should fix that.
Concerning the CC_INCLUDE issue - that does not matter as it only affects the non-mpi version.

Should we maybe check at configure time that PETSc has mpi enabled and is the same mpi?
Currently using a serial PETSc configures fine (I tried the fedora one, patched) but it fails to compile with:
```
In file included from /usr/include/petsc/petscis.h:7:0,
                 from /usr/include/petsc/petscvec.h:9,
                 from /usr/include/petsc/petscmat.h:6,
                 from /usr/include/petsc/petscpc.h:6,
                 from /usr/include/petsc/petscksp.h:6,
                 from petsc_laplace.hxx:61,
                 from petsc_laplace.cxx:28:
/usr/include/petsc/petscsys.h:137:6: error: #error "PETSc was configured with --with-mpi=0 but now appears to be compiling using a different mpi.h"
 #    error "PETSc was configured with --with-mpi=0 but now appears to be compiling using a different mpi.h"
      ^~~~~
```

ZedThree:
Sorry, but this will making maintenance harder and doesn't add much over that's not in already in #1050

The checks that PVODE compiles and the makefile works could be pulled out into a separate PR, but see https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/Limitations-of-Builtins.html#Limitations-of-Builtins about using `if ! command` (i.e. use `AS_IF` instead)

PR #1059, opened 2018-05-14
Workarounds to get boutcore Cython module to work with OpenMPI

johnomotani:
There are two issues (I guess bugs with OpenMPI) that were causing segfaults:

(i) OpenMPI assumes the argv array it is passed is terminated by a null pointer; it iterates through it with a loop like 'for (p=argv; *p; ++p)'. Adding a null pointer as a final element of string_buf in
boutcore.init() fixes the segfault caused previously by this loop failing to stop.

(ii) OpenMPI requires some 'dlopen' flags to be set in order to link correctly. I don't understand the voodoo, but copied from the header files of 'mpi4py' project (https://bitbucket.org/mpi4py/mpi4py) into the new boutcore_openmpi_compat.hxx header, and at least on my desktop machine this seems to work.

### DIFF
--- a/tools/pylib/_boutcore_build/boutcore_openmpi_compat.hxx
+++ b/tools/pylib/_boutcore_build/boutcore_openmpi_compat.hxx
@@ -1,0 +1,146 @@
+// Copied from mpi4py, mpi4py/src/lib-mpi/compat/openmpi.h and
+// mpi4py/src/dynload.h (at the time of writing on 14/5/2018, mpi4py was at
+// revision 18c52a9):
+//     https://bitbucket.org/mpi4py/mpi4py
+//     Author:  Lisandro Dalcin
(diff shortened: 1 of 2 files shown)


tools/pylib/_boutcore_build/boutcore_openmpi_compat.hxx
```diff
@@ -0,0 +1,146 @@
...
+//     POSSIBILITY OF SUCH DAMAGE.
+
+#ifndef PyMPI_DYNLOAD_H
+#define PyMPI_DYNLOAD_H
+
+#if defined(OMPI_MAJOR_VERSION)
+
+bool is_openmpi = true;
+
+#if HAVE_DLFCN_H
```
dschwoerer: maybe add a comment that we do never set this flag currently ...

PR #1060, opened 2018-05-14
Newfield nextmerge

johnomotani:
This PR is a rebase of the `newfield` branch, PR #937, onto next.

The main differences compared to `newfield` are:
* ensuring that the pointers to member fields in Flexible objects never change, to be compatible with file I/O (the pointers are what are stored in Datafile, and if they change segfaults may happen).
* Removing `getNonConst()` method, replaced by a `const_cast<>` for adding Flexible fields to Datafile, where we will assume that ignoring `const`-ness is OK to read fields from a file, but everywhere else all references to members of Flexible should be `const`.


dschwoerer:
It seems that made it so slow that some tests timeout ...
This only affects the --enable-debug builds - so not when we really care about performance.
Something to prevent timeouts would be nice, that would also avoid the mpich issue ...

dschwoerer:
Not copying the fields, but rather taking const references gives quite a  speed-up:
before:
```
test-drift-instability-staggered ✓ 163.346 s
test-interchange-instability     ✓ 592.890 s
```
after:
```
test-drift-instability-staggered ✓ 112.693 s
test-interchange-instability     ✓ 154.489 s
```

johnomotani:
@ZedThree For me the major thing is having geometrical quantities at the right location (metrics, zShift).  Those don't need updating once they've been calculated the first time, and it's useful to have guard cells initialized in case we want to take a derivative, e.g. in Div_par(f) we need Grad_par(f/mesh->Bxy). For this I don't know if all the machinery that's in `Flexible<>` is absolutely necessary, it wouldn't be so bad to explicitly pass the location, e.g. instead of `g_22` have `g_22(location)` which returns the needed Field2D.

Another possible use is that in STORM we have quite a few fields with both CENTRE and YLOW versions. It might be convenient to have an object that could wrap up both of those and supply the right one where needed. In this case, there is definitely a problem with automatically updating all locations when one is changed, because say you change the CENTRE field, you need to at least communicate, and possibly also apply some custom sheath boundary conditions, before interpolating to get the YLOW field. The YLOW field might then also need communicating and boundary conditions itself. Maybe all these complications mean that this case is not suitable for a Flexible-like class because there's not much that can be done 'automagically'.

dschwoerer:
Concerning
```
A = B
assert(A == B)
```
Are you complaining that a copy-assignment operator is missing? I guess that would probably be useful ...

Assuming that the field has an `==` operator, that would still hold, as A or B's mainlocation should be the same, and if any of them is a field, that location would match the others main location, thus a cast to field, to use the operator would return the field at the main location, and the `==` should return true, or what case am I missing?

I agree with what @johnomotani said. I implemented this to hide that dx etc can have different locations, and can mostly be used as they are now.
If the field gets updated, on whether other fields should be recomputed or not, that is a question that can in general not be answered. If it is just because BCs have been applied, then no. If it is just a field that has been interpolated, BCs applied and communicated, then the answer is again no.  If the field however changed, then the answer is yes. I didn't see any way to have the flexible field decide that on it's own, that's why set takes a second argument, on whether to reset the other fields, or not.

Could you elucidate on where you see breakage of backwards-compatibility?
The flexible field should always cast to a field, so if any code is not aware of the flexibility, it should not cause issues. The only case I can see that a flexible field is modified in such a way, that the field changes, and the shifted versions aren't updated. That would e.g. require changing dx with the index operator - is this an actual use case? I thought that was a rather theoretical scenario for breakage ...

I think having to pass dx the location would rather break backwards compatibility, as dividing by dx can now fail, as that used to ignore dx's lack of location, and now BOUT++ makes sure that the location is right.

ZedThree:
The lack of copy-ctors is mostly what I meant, yes. But `Flexible<T>` sometimes acts like a container of `T` and sometimes as a `T` itself. Using `operator=(T&)` to set elements of a container is a bit weird. Multiple assignments to a `T` not "overwriting" each other is also a bit weird.

Re:backwards compatibility, consider the following:

```cpp
Field2D &g11_ref = mesh->coordinates()->g11;
```
This does an implicit conversion using `Flexible<T>::operator T()` which returns a (temporary) copy. The above won't compile because you can't take a non-const lvalue reference to an r-value (a temporary). `const&` or `&&` will both work, but that's potentially changing user-code.

Also

```cpp
auto g11_copy = mesh->coordinates()->g11;
```

won't compile, although this could be solved by providing the copy-ctor and copy-assignment operator.

There's various other problems with template type-deduction

---

If the biggest use for this kind of functionality is just for the metric elements, we could overload `Mesh::coordinates` to take a cell location and return a whole `Coordinates` object with the elements at the desired location. The interpolation could either be done first in `Coordinates::geometry`, or cached on the first call to `Mesh::coordinates(CELL_LOC)`. Interpolation would be done from the `CELL_CENTRE` location. We already have the expectation that changing the metric requires a call to `Coordinates::geometry` -- that could check if other locations have been used and if so, update them.

I think this approach has several advantages:

1. Much smaller code footprint and interface surface area, i.e. less chance for bugs
2. Much simpler to change code to take advantage of it:

```diff
  void foo(const Field3D& f) {
-     auto coords = f.getMesh()->coordinates();
+     auto coords = f.getMesh()->coordinates(f.getLocation());
     // use coords->g11, etc. as normal
```
3. Keeping the different locations in-sync would be much easier, as an explicit call to `Coordinates::geometry` is already required
4. Doesn't change the type of the metric elements

The big problem with making this more generic is making it behave "properly", e.g. I would really expect it keep the different field-locations in-sync. And as both @dschwoerer and @johnomotani have pointed out, that is tricky in the general case, especially as it might involve communication.

The current interface also allows users to add unrelated fields to the same `Flexible<T>`, as long as they are at different locations. In a "full" interface, I would expect that after the following:

```cpp
Flexible<T> a;
a = b; // b at CELL_CENTRE
a = c; // c at CELL_XLOW
```
when I access `a` at different locations, I would be getting `c` interpolated to those locations. Instead, I would get either `b` or `c` depending on the location. I feel that's a bit surprising.

johnomotani:
@ZedThree @dschwoerer I think the coordinates approach sounds good. I'll have a crack at implementing it, unless someone else has started working on it already?

dschwoerer:
@johnomotani 
No, I didn't have a go. I don't think it is going to be easier to use in the end.

@ZedThree 
I think you are mistaken. If a field is assigned with `operator=` it should become the main field.
So in your example, a requested at CELL_CENTRE would return c interpolated, which I think is what you think it should do?
To set an interpolated field, you need to call `.set()` - where I think we could come up with a better name to indicate that this is only setting an interpolated version of what already exists.

Thus if you use it 'naively' you will always get interpolated versions - as you need to know more about that.

I think that makes clear that more documentation is needed, with examples etc.
If anybody wants to implement the other approach, I am happy to wait for that, but I think the flexible field approach will do a much better job at hiding the Field2D has a location change, and thus will introduce less issues.

If the flexible field approach should be chosen, it needs:
 * a copy constructor
 * more documentation
 * examples
 * a better name for `.set()`

bendudson:
Closing this for now, though happy to reopen later. The Flexible field approach has some nice features, but risks being over-engineered for what is currently needed. #1198 and #1200 are more straightforward for now.

PR #1061, opened 2018-05-14
Add missing from field aligned

johnomotani:
In several y-derivatives where we have to convert the input to field-aligned coordinates, the result was not transformed back to the original coordinates.

Also, several advective and flux derivatives which take two fields, v and f, as inputs, had cases using yup/ydown fields for one of v and f, but not the other. These do not make sense as multiplication of a field in field-aligned coordinates with another in non-field-aligned coordinates is incorrect. Delete these cases and fall back to converting both v and f to field-aligned if either does not have yup/ydown fields.


src/mesh/difops.cxx
```diff
@@ -281,7 +283,7 @@ const Field3D Grad_par_CtoL(const Field3D &var) {
       }
     }
 
-    result = mesh->fromFieldAligned(result);
+    result = mesh->fromFieldAligned(result_fa);
```
d7919: I'm not sure that `result_fa` is actually set in the above loop. What happened previously was that `result` was used to store the field aligned result and then we did an "in-place" transform back into non-field aligned : `result = mesh->fromFieldAligned(result);`

src/mesh/difops.cxx
```diff
@@ -324,65 +326,15 @@ const Field3D Vpar_Grad_par_LCtoC(const Field3D &v, const Field3D &f, REGION reg
       result[i] -= (vval.p >= 0.0) ? vval.p * fval.c : vval.p * fval.p;
     }
   }
-  else if (vUseUpDown) {
-    // Only v has up/down fields
-    // f must shift to field aligned coordinates
-    Field3D f_fa = mesh->toFieldAligned(f);
```
d7919: We're shifting `f` to be field aligned such that it should make sense to multiply with `v`, which has field aligned data available (in yup and ydown), so I think this is ok but could be missing something.

src/mesh/difops.cxx
```diff
@@ -424,12 +379,15 @@ const Field3D Grad_par_LtoC(const Field3D &var) {
...
     // No yup/ydown field, so transform to field aligned
 
-    Field3D var_fa = var.getMesh()->toFieldAligned(var);
+    Field3D var_fa = fieldmesh->toFieldAligned(var);
+    Field3D result_fa(fieldmesh);
+    result_fa.allocate();
 
     for(auto &i : result.region(RGN_NOBNDRY)) {
-      result[i] = (var_fa[i.yp()] - var_fa[i]) / (metric->dy[i]*sqrt(metric->g_22[i]));
+      result_fa[i] = (var_fa[i.yp()] - var_fa[i]) / (metric->dy[i]*sqrt(metric->g_22[i]));
```
d7919: I don't think this is required, although it probably doesn't hurt beyond memory usage.

src/mesh/difops.cxx
johnomotani: Cheers @d7919, that was just a mistake.

src/mesh/difops.cxx
johnomotani: I may well have misunderstood. I thought that to/fromFieldAligned rotate the points in the z-direction by zShift, so the toroidal position of the point (i,j,k) would be different for field-aligned and non-field-aligned fields?

src/mesh/difops.cxx
johnomotani: Agreed it's not required. I thought it would be clearer to keep field-aligned and non-field-aligned fields separate, but am happy to revert back if people think it would be better. You make a good point about the memory usage, I'm inclined now to change it back to `result = mesh->fromFieldAligned(result);` style.

johnomotani:
More up to date version of these fixes is included in #1176

PR #1063, opened 2018-05-15
Ny interp pol slice

d7919:
Allows data to be interpolated along the field line before doing the
transform to the poloidal plane.

Also fixes for v4 onwards due to change in meaning of `nz` (i.e. we no longer have the duplicate point).

### DIFF
--- a/tools/pylib/boutdata/pol_slice.py
+++ b/tools/pylib/boutdata/pol_slice.py
@@ -32,7 +32,7 @@ def pol_slice(var3d, gridfile, n=1, zangle=0.0):
 
     nx, ny, nz = s
 
-    dz = 2.*np.pi / float(n * (nz-1))
+    dz = 2.*np.pi / float(n * nz)
 
     try:
         # Open the grid file
--- a/tools/pylib/boutdata/pol_slice.py
+++ b/tools/pylib/boutdata/pol_slice.py
@@ -20,8 +20,10 @@ except ImportError:
     print("=> Set $PYTHONPATH variable to include BOUT++ pylib")
     raise SystemExit
 
-def pol_slice(var3d, gridfile, n=1, zangle=0.0):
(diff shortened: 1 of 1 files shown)


tools/pylib/boutdata/pol_slice.py
```diff
@@ -64,6 +66,43 @@ def pol_slice(var3d, gridfile, n=1, zangle=0.0):
...
+    if nyInterp is not None:
+        if ny != nyInterp:
+            doInterp = True
+
+    #Setup interpolation if requested
+    if doInterp:
+        try:
+            from numpy import linspace, meshgrid, append, newaxis
+        except ImportError:
+            print("ERROR: NumPy module not available")
```
ZedThree: I think we require numpy for the pylib stuff, so I think we can just safely import it here

tools/pylib/boutdata/pol_slice.py
```diff
@@ -80,11 +119,18 @@ def pol_slice(var3d, gridfile, n=1, zangle=0.0):
...
     zm = (z0 - 1 + (nz-1)) % (nz-1)
 
-    # There may be some more cunning way to do this indexing
-    for x in np.arange(nx):
-        for y in np.arange(ny):
-            var2d[x,y] = 0.5*p[x,y]*(p[x,y]-1.0) * var3d[x,y,zm[x,y]] + \
-                         (1.0 - p[x,y]*p[x,y])   * var3d[x,y,z0[x,y]] + \
-                         0.5*p[x,y]*(p[x,y]+1.0) * var3d[x,y,zp[x,y]]
+    #For some reason numpy imposes a limit of 32 entries to choose
+    #so if nz>32 we have to use a different approach. This limit may change with numpy version
```
ZedThree: See here for a possible solution: https://stackoverflow.com/a/39169371/2043465

PR #1064, opened 2018-05-15
Add separate option for staggered derivatives

johnomotani:
If StaggerGrids=true, then only methods implemented for staggered derivatives could be used for unstaggered derivatives, e.g. FFT derivatives could not be used for unstaggered derivatives. This commit adds an option which, if set explicitly, allows a different method to be chosen for staggered derivatives to avoid the exception being thrown for unknown option when it is only unknown for the staggered case. If this option is used, presumably the staggered derivative should not be used,
but since the user must set the option explicitly, they must be aware that different methods would be used.

### DIFF
--- a/src/mesh/index_derivs.cxx
+++ b/src/mesh/index_derivs.cxx
@@ -568,7 +568,7 @@ Mesh::flux_func sfFDDX, sfFDDY, sfFDDZ;
  *******************************************************************************/
 
 /// Set the derivative method, given a table and option name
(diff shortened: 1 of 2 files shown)


dschwoerer:
Seems to be related to #1034
#1034 doesn't allow to ignore invalid derivatives - but should make it (hopefully) sufficiently easy to set valid ones.

johnomotani:
Yes, this is just duplicating #1034. Closing...

PR #1066, opened 2018-05-21
Boutoptionsfile update

johnomotani:
Add a method `as_dict()` to convert a `BoutOptionsFile` object to a nested dictionary. This may be useful, for example if you want to flatten the dictionary.

Fixes a bug with `BoutOptionsFile.keys()`. In Python 3 the `keys()` method of `dict` returns a `dict_keys` dictionary view object. These cannot be added together like lists. Instead use `list(self._sections) + list(self._keys)` to add lists of keys in the two dictionaries. This is compatible with Python 2 and Python 3.

Also, a question. We have two classes for reading input files, BOUTOptions in boututils.options https://github.com/boutproject/BOUT-dev/blob/bffe0bb529355e54d76ba1ef07b590f57c817e6a/tools/pylib/boututils/options.py#L14 and BoutOptionsFile in boutdata.data https://github.com/boutproject/BOUT-dev/blob/bffe0bb529355e54d76ba1ef07b590f57c817e6a/tools/pylib/boutdata/data.py#L137
Is either preferred? And should we drop/deprecate the other one to standardize?

### DIFF
--- a/tools/pylib/boutdata/data.py
+++ b/tools/pylib/boutdata/data.py
@@ -108,6 +108,14 @@ class BoutOptions(object):
         """
         return self._keys.keys()
 
+    def as_dict(self):
+        """
+        Return a nested dictionary of all the options.
+        """
+        dicttree = {name:self[name] for name in self.values()}
+        dicttree.update({name:self[name].as_dict() for name in self.sections()})
+        return dicttree
+
     def __len__(self):
         return len(self._sections) + len(self._keys)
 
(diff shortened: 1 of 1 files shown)


ZedThree:
I think it might be possible to replace most of the BoutOptions stuff with some wrapper or subclass of the built-in ConfigParser. It just needs a little fiddling to handling the "global" initial section

johnomotani:
This stackoverflow answer about the alternative ConfigObj class might also be interesting if you want to go that way (it seems to handle the global section already. Also allows subsections, but only with a particular syntax as far as I can see): https://stackoverflow.com/a/16350580

PR #1073, opened 2018-05-25
Python numpydocs

ZedThree:
Try to consistently apply the [numpydocs style](https://numpydoc.readthedocs.io/en/latest/format.html) to the python documentation.

Should only be changes to docstrings and whitespace, plus removing some unneeded module imports to help readthedocs.

I'll get this branch building on readthedocs to check it looks ok there. Should appear [here](https://bout-dev.readthedocs.io/en/python-numpydocs) when it's done building

This is still not done, but most of the way there...


ZedThree:
Vast majority of all (useful) python stuff in boutdata/boututils/zoidberg now in a consistent documentation style and generated on readthedocs. Direct link: https://bout-dev.readthedocs.io/en/python-numpydocs/api_reference.html

ZedThree:
I did very briefly looked into this. See here: https://stackoverflow.com/a/31882049/2043465

But it requires being able to build the cython bits without actually building the main BOUT++ library, which I'm not less sure about.

dschwoerer:
Building the interface without the library seems quite difficult. Implementing a fake bout++ library would be possible, but that seems to me like a lot of work, and I am wondering whether switching to another documentation server might be easier ...

PR #1077, opened 2018-05-31
Fix order of arguments in call to tridagMatrix in LaplaceSPT

johnomotani:
I think this was a typo in changes made in 72c8f8ca78273ae796589e5543aad234145a0453.


d7919:
Thanks for catching this! I'd recently noticed SPT was not agreeing with Cyclic so this may be the reason. Any idea why this compiled without complaining? It would presumably require a BoutReal-->Int conversion for the kwave argument.

dschwoerer:
Type conversion warnings can be enabled in gcc, but they are currently very noisy, as we are doing so many conversions without explicitly casting. It is good BOUT++ is not a security relevant library ...

johnomotani:
I think you're right why it compiled. I found it because of an out-of-bounds exception which must have been because kz>LocalNy while I was chasing a different bug (see #1078).

ZedThree:
4.2-rc will be branched soon, and we'll have a good old cleanup of warnings. Conversion warnings sounds like an easy win.

ZedThree:
This is also in LaplacePDD at least

d7919:
I think the other impls are ok; cyclic and serial_tri both have the right order.

PR #1078, opened 2018-05-31
Allow number of x-boundary points to change in FFT Laplace solvers

johnomotani:
In the FFT-based solvers the variables 'inbndry' and 'outbndry' were set in several places using a hard-coded number (2) of boundary points. Replace this with mesh->xstart in case the number of x-guard cells is changed, otherwise the solvers fail to give correct results if, e.g., MXG=1.

Also fix LaplaceShoot similarly.

### DIFF
--- a/src/invert/laplace/impls/serial_tri/serial_tri.cxx
+++ b/src/invert/laplace/impls/serial_tri/serial_tri.cxx
@@ -83,16 +83,16 @@ const FieldPerp LaplaceSerialTri::solve(const FieldPerp &b, const FieldPerp &x0)
 
   // Setting the width of the boundary.
   // NOTE: The default is a width of 2 guard cells
-  int inbndry = 2, outbndry=2;
+  int inbndry = mesh->xstart, outbndry=mesh->xstart;
 
   // If the flags to assign that only one guard cell should be used is set
   if (global_flags & INVERT_BOTH_BNDRY_ONE) {
-    inbndry = outbndry = 1;
+    inbndry = outbndry = mesh->xstart-1;
   }
   if (inner_boundary_flags & INVERT_BNDRY_ONE)
-    inbndry = 1;
+    inbndry = mesh->xstart-1;
   if (outer_boundary_flags & INVERT_BNDRY_ONE)
-    outbndry = 1;
+    outbndry = mesh->xstart-1;
 
   /* Allocation fo
    * bk   = The fourier transformed of b, where b is one of the inputs in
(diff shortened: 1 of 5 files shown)


ZedThree:
This looks ok to me. Also just adds another argument for a better Laplacian test

src/invert/laplace/impls/serial_tri/serial_tri.cxx
```diff
@@ -83,16 +83,16 @@ const FieldPerp LaplaceSerialTri::solve(const FieldPerp &b, const FieldPerp &x0)
 
   // Setting the width of the boundary.
   // NOTE: The default is a width of 2 guard cells
-  int inbndry = 2, outbndry=2;
+  int inbndry = mesh->xstart, outbndry=mesh->xstart;
 
   // If the flags to assign that only one guard cell should be used is set
   if (global_flags & INVERT_BOTH_BNDRY_ONE) {
-    inbndry = outbndry = 1;
+    inbndry = outbndry = mesh->xstart-1;
```
bendudson: I think this one should probably be kept, as the intention of INVERT_BOTH_BNDRY_ONE is to force the number of cells to be 1 (whether this is a good idea ever is debateable). The new code would go wrong if mesh->xstart=1.

src/invert/laplace/impls/serial_tri/serial_tri.cxx
```diff
@@ -83,16 +83,16 @@ const FieldPerp LaplaceSerialTri::solve(const FieldPerp &b, const FieldPerp &x0)
...
   }
   if (inner_boundary_flags & INVERT_BNDRY_ONE)
-    inbndry = 1;
+    inbndry = mesh->xstart-1;
```
bendudson: Similar here and outer boundary flags version

src/invert/laplace/impls/shoot/shoot_laplace.cxx
```diff
@@ -76,14 +76,14 @@ const FieldPerp LaplaceShoot::solve(const FieldPerp &rhs) {
   
   // Get the width of the boundary
   
-  int inbndry = 2, outbndry=2;
+  int inbndry = mesh->xstart, outbndry=mesh->xstart;
   if(global_flags & INVERT_BOTH_BNDRY_ONE) {
-    inbndry = outbndry = 1;
+    inbndry = outbndry = mesh->xstart-1;
```
bendudson: Same here

PR #1080, opened 2018-05-31
Fix various 'import' related issues in python

ZedThree:
- Remove unused imports
- Catch `ImportError` rather than everything where appropriate
- Reorder/consolidate some imports

And the big one: we **require** the `future` module to use Python 2. This is important, as various functions, like `str` and `range`, behave differently in 2/3, and `future` gives us identical behaviour -- but requires importing them from `builtins` (which is a builtin module in 3). In several places we `try` importing them, but just `pass` if we can't. We should fail instead. Therefore, I've done a slightly hacky thing, and `try` importing them in the module `__init__.py`s, and print a (hopefully) helpful error if we can't.


examples/blob2d/blob_velocity.py
```diff
@@ -95,5 +94,5 @@ def blob_velocity(n,**kwargs):
         import matplotlib.pyplot as plt
         plt.plot(vx)
         plt.show()
-except:
+except ImportError:
```
dschwoerer: The general except is justified, as matplotlib may fail for other reasons, e.g. no suitable X11 server available.
I have changed this in some tests in the other direction, as I got errors after installing matplotlib.
Ether we want to make sure we plot - or we don't care if it fails, in which case we should ignore for whatever reason.
We might want to add an option to blob_velocity, to only plot if it is asked to plot.

Besides that, all the changes seem good, thanks for the cleanup :+1:

ZedThree: That file will be getting a better cleanup soon anyway!

Adding an option to plot is something a few other examples could benefit from too.

ZedThree:
This is just failing due to the multigrid test hanging. #1084 should sort that out permanently though

PR #1081, opened 2018-05-31
FCI divergence operators

bendudson:
Implements a new parallel transform method, `integrateYupYdown`, which defaults to the existing `calcYupYdown`. For the FCI scheme, this performs an integration over the area of the ends of the flux-tubes, using cell corner interpolations. This requires two interpolations per cell rather than one, but results in better conservation properties e.g. mass. 

Requires making `Mesh::getParallelTransform` public.

Examples of use in `examples/fci-wave` include README and script to run and analyse output.

### DIFF
--- a/src/mesh/parallel/fci.cxx
+++ b/src/mesh/parallel/fci.cxx
@@ -110,10 +110,14 @@ FCIMap::FCIMap(Mesh &mesh, int dir, bool zperiodic)
         xt_prime_corner(x, y, z) =
             0.25 * (xt_prime(x, y, z) + xt_prime(x + 1, y, z) + xt_prime(x, y, z + 1) +
                     xt_prime(x + 1, y, z + 1));
+
+        zt_prime_corner(x, y, z) =
+            0.25 * (zt_prime(x, y, z) + zt_prime(x + 1, y, z) + zt_prime(x, y, z + 1) +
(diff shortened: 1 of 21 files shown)


src/mesh/parallel/fci.cxx
```diff
@@ -94,9 +97,43 @@ FCIMap::FCIMap(Mesh &mesh, int dir, bool zperiodic)
...
+  
+  // Cell corners
+  Field3D xt_prime_corner(&mesh), zt_prime_corner(&mesh);
+  xt_prime_corner.allocate();
+  zt_prime_corner.allocate();
 
-  interp->calcWeights(xt_prime, zt_prime);
+  for (int x = mesh.xstart; x <= mesh.xend; x++) {
+    for (int y = mesh.ystart; y <= mesh.yend; y++) {
+      for (int z = 0; z < mesh.LocalNz - 1; z++) {
```
ZedThree: Could this be range-based loop over `RGN_ALL`?

d7919: Possibly a similar issue for the `z+1` indexing as RGN_NOBNDRY visits all z currently?

ZedThree: Good point, I guess it would need a custom region! Probably better like this in that case. It's in setup, so not likely to be a pinch point requiring omp

Issue #1079, opened 2018-05-31
rk4 sets number of rhs to 0


dschwoerer:
rk4 is very slow, roughly a factor of 10 compared to pvode at the same number of rhs evaluation.

At least to me this is surprising, as the rk4 scheme should not be that complicated compared to an rhs evaluation ...
It claims that 93% of the time is spent in Calc, only ~4.4 % in solver - which I doubt.

Is there something I am missing that can explain this?

d7919:
Is the nrhs very different? Are you using it adaptively? The rk schemes do typically spend a short amount of time in the solver compared to the implicit solvers.

d7919:
As the explicit solvers are limited by the CFL condition, whilst the implicit solvers are not limited in step size but instead just seeking a particular tolerance, the explicit solvers can often be fastest for low resolution cases but much slower for high resolution runs, at least in my experience.

ZedThree:
Can you check rk4 vs rkgeneric and/or PETSc's rk solver? @d7919 what's the default order of rkgeneric? For PETSc, I think you also need to set `ts_rk_type` to 4, otherwise it uses a third order scheme.

d7919:
There isn't a default order for rkgeneric -- you have to specify which scheme you want to use (probably something worth improving on by setting a sensible default).

dschwoerer:
I tried with the blob2d example, where everything is fine.

I tested with the monitor example, and the issue is that the nrhs displayed are wrong, which explains the factor of 10 difference.

rkgeneric seems to be suffering from the same issue, i.e. resetting the number of rhs after calling the monitors, while they should be reset from the monitor that outputs the information.
Default for rkgeneric seems to be rkf45.

PR #1083, opened 2018-05-31
rhs counters are reset by Solver

dschwoerer:
The resets by the implementation where not needed, and are bad if
Monitors with different timesteps are used.
If some monitors are called more often then the output monitor, this
resulted in a wrong number of rhs shown.

This caused the confusion in #1079

### DIFF
--- a/include/bout/solver.hxx
+++ b/include/bout/solver.hxx
@@ -268,10 +268,17 @@ class Solver {
     throw BoutException("resetInternalFields not supported by this Solver");}
 
   // Solver status. Optional functions used to query the solver
-  virtual int n2Dvars() const {return f2d.size();}  ///< Number of 2D variables. Vectors count as 3
-  virtual int n3Dvars() const {return f3d.size();}  ///< Number of 3D variables. Vectors count as 3
-  
-  int rhs_ncalls,rhs_ncalls_e,rhs_ncalls_i; ///< Number of calls to the RHS function
+  /// Number of 2D variables. Vectors count as 3
+  virtual int n2Dvars() const {return f2d.size();}
+  /// Number of 3D variables. Vectors count as 3
+  virtual int n3Dvars() const {return f3d.size();}
+
+  /// Get and reset the number of calls to the RHS function
+  int resetRHSCounter();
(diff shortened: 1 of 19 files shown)


bendudson:
The rhs counts are reset in solver.cxx here: https://github.com/boutproject/BOUT-dev/blob/next/src/solver/solver.cxx#L716 but only when `(iter % freqDefault) == 0`. Is this the desired behaviour (genuine question)?

dschwoerer:
Yes. The default frequency is the one at which the output monitor runs - which I think is the only use of the counter (I just grep'ed the code and could see no other usage).

If the counter is cleared more(less) often, the numbers reported by the monitor are not with respect to the last output of the monitor, which is confusing/wrong.

Should I document this somewhere? If more details are needed, I can try to elaborate.

bendudson:
If you could add a sentence or two to the monitors section of the manual that would be excellent. Thanks!

dschwoerer:
I had another look at the code, and think it makes it more clear, if I change the interface of solver, to not expose the raw int, but an interface to get and reset the integer, similar to the timer class.
That way the checking is not required, and the code should read more natural ...

include/bout/solver.hxx
```diff
@@ -268,10 +268,17 @@ class Solver {
...
-  virtual int n3Dvars() const {return f3d.size();}  ///< Number of 3D variables. Vectors count as 3
-  
-  int rhs_ncalls,rhs_ncalls_e,rhs_ncalls_i; ///< Number of calls to the RHS function
+  /// Number of 2D variables. Vectors count as 3
+  virtual int n2Dvars() const {return f2d.size();}
+  /// Number of 3D variables. Vectors count as 3
+  virtual int n3Dvars() const {return f3d.size();}
+
+  /// Get and reset the number of calls to the RHS function
+  int resetRHSCounter();
```
ZedThree: I'm not sure about using a single method to both get the value and reset it at the same time. It's useful for how it's currently used, but what if a monitor just wants to get the number of RHS calls without changing it?

bendudson: Agree we can add `getRHScounter` or similar if there is a demand for it.

PR #1085, opened 2018-06-01
Make quieter unit tests optional, rather than default

ZedThree:
If the unit tests fail, they don't necessarily get reported as failing because the `quiet.sh` script returns 0. This means Travis won't report unit tests failures, for example.

With bash, there is `$PIPESTATUS`, but I can't find a portable way of getting the exit code of `serial_tests` and also pipe into `quiet.sh`. I've taken the easy route, and just made `check-unit-tests-quiet` as a top-level target, with `check-unit-tests` being the "noisy" version.

This (or something like this) is essential to make sure Travis reports failures with the unit tests


ZedThree:
Replaced by #1086

PR #1086, opened 2018-06-01
Fix exit status of quiet script

dschwoerer:
Alternative to #1085 
Thanks @ZedThree for noticing this bug

### DIFF
--- a/tests/unit/makefile
+++ b/tests/unit/makefile
@@ -91,7 +91,7 @@ serial_tests: makefile $(BOUT_TOP)/make.config $(OBJ) $(LIB) $(SUB_LIBS) $(TEST_
 #       target in make.config being used instead.
 check: $(GTEST_SENTINEL) serial_tests
 	@echo "Running unit test"
-	@./serial_tests | ./quiet.sh
+	@./quiet.sh ./serial_tests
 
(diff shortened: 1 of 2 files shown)


ZedThree:
I prefer #1085 just to make things as simple as possible on Travis

dschwoerer:
I prefer #1086 as it makes things as readable as possible on Travis

ZedThree:
I don't think I understand what the problem is, to be honest. I can just Ctrl+F "failed" to find the bad tests on Travis. I prefer to have as few moving parts as possible, less chance for bugs to creep in.

dschwoerer:
I think it anyway needs to be fixed, to work correctly - or remove it completely, as keeping broken stuff is worse than not having it.
Assuming we want to keep it, we have to fix it.
If it is working anyway, why should we not use it on travis as well?
I can see your point that we want to keep stuff simple on travis, but we also want to keep stuff quite.
That is why we e.g. implemented the `./test_suite` script and made the integrated tests so much less verbose.

ZedThree:
I disagree that we need to keep things quiet on Travis, they just need to be readable. Anyway, this urgently needs fixing more than anything

PR #1087, opened 2018-06-01
Evaluate expressions from input files in BoutOptions/BoutOptionsFile

johnomotani:
Adds method evaluate_scalar(name) to BoutOptions that can evaluate scalar expressions.

Adds a method evaluate(name) to BoutOptionsFile that can evaluate scalars or arrays that depend on (x,y,z). Array variables are three dimensional (with size 1 if a dimension had no explicit dependence).

This could do with some testing, but seems to work.


tools/pylib/boutdata/data.py
```diff
@@ -141,6 +144,58 @@ def __str__(self, indent=""):
...
+        Evaluate (recursively) scalar expressions
+        """
+        expression = self._substitute_expressions(name)
+
+        # replace ^ with ** so that Python evaluates exponentiation
+        expression = expression.replace("^","**")
+
+        # use numpy functions
+        for func in funcstoreplace:
+            expression = re.sub(r"\b"+func.lower()+r"\b", "numpy."+func, expression)
```
ZedThree: Could these just be `import`ed from `numpy`? That would remove these two lines, plus `funcstoreplace`

tools/pylib/boutdata/data.py
```diff
@@ -216,6 +271,67 @@ def __init__(self, filename, name="root"):
...
+        gridfile = None
+        try:
+            self.nx = self["mesh"].evaluate_scalar("nx")
+            self.ny = self["mesh"].evaluate_scalar("ny")
+        except KeyError:
+            try:
+                gridfilename = self["mesh"]["file"]
+            except KeyError:
+                gridfilename = self["grid"]
+            gridfile = DataFile(gridfilename)
```
ZedThree: You should call `gridfile.close()` when you're done with it

johnomotani:
@ZedThree, agree with all your points above. I've addressed them in the latest commit.

tools/pylib/boutdata/data.py
```diff
@@ -245,6 +298,60 @@ def __init__(self, filename, name="root"):
...
```
bendudson: Reading the grid file could easily fail, since the path is relative to where BOUT++ is run, and could give unexpected results. Since its only the grid size which is needed, I wonder if it would be more robust to read (collect) these from the output files? The dmp files are not guaranteed to be in the same place as the options file (BOUT++ `-f` flag), but probably more reliable than the grid file location.

johnomotani: Agree it's a good idea to try to be more reliable, especially with relative paths. I've pushed an update to try to address this issue as best as I could think how to. It's a bit ugly, so any suggestions welcome :smile:

PR #1088, opened 2018-06-01
Fix copy constructor bug

d7919:
Fix copy constructor bug from 6746ec8a592606ed72f2eded964c6ce26f04c619. Thanks to @ZedThree for spotting this.


ZedThree:
I'm currently writing unit tests for the basic stuff in vector2d/3d. Dot/cross products, etc. will have to wait until I sort something out for Coordinates.

ZedThree:
I think I've missed some Vector3D operators, will finish off on Monday

ZedThree:
This current set of unit tests are sufficient to catch this bug. I'll need to do some work with `Coordinates` in order to make tests for the rest of the Vector stuff, so I'll do that in a separate PR

Issue #1090, opened 2018-06-05
Parsing incomplete expressions with ExpressionParser returns null rather than exception


ZedThree:
For instance:

```cpp
auto fieldgen = parser.parseString("2 +");
```

`fieldgen` is `nullptr`, whereas other incomplete or malformed expressions raise `ParseException`. I'm hesitant to say it's a bug, because in e.g. `Options::get`, we explicitly check if we get a `nullptr` or a real `FieldGenerator` back.

Would it be better for such expressions to throw exceptions, or are there situations where returning null makes more sense?

bendudson:
There are a few places in ExpressionParser which return NULL (or 0). I think they can all be changed to throw ParseException, and I agree it makes sense to throw rather than return null.

ZedThree:
From what I can tell, we just throw a BoutException if we get null back anyway. Changing it to a more explicit ParseException might be better. One downside would be user code that catches BoutExceptions would need to be modified to catch ParseExceptions instead.

bendudson:
I think most places (all?) which use ExpressionParser use it via FieldFactory, since that adds most of the functions and the interface to the Options tree. Perhaps everywhere that ExpressionParser is called e.g.
https://github.com/boutproject/BOUT-dev/blob/master/src/field/field_factory.cxx#L359
FieldFactory could catch ParseException and convert it to BoutException?

ZedThree:
That would work. An alternative might be to make ParseException inherit from BoutException?

bendudson:
Ah yes, that might be nicer... It does make ExpressionParser less independent of BOUT++, but for error handling this probably makes sense.

bendudson:
Fixed in #1091

PR #1091, opened 2018-06-05
ExpressionParser throw ParseException rather than returning null

bendudson:
Previous behaviour was inconsistent: Some failures would result in a thrown exception, whilst others would result in NULL being returned.

The new behaviour is to always either return a valid generator, or throw an exception.

Partly fixes issue #1090


ZedThree:
I have a few more unit tests for ExpressionParser. Would you prefer them merged in here or a separate PR?

bendudson:
I guess it makes sense to merge them in here for testing before merging into next. Thanks!

src/sys/expressionparser.cxx
```diff
@@ -219,11 +219,13 @@ std::shared_ptr<FieldGenerator> ExpressionParser::parseParenExpr(LexInfo &lex) {
   lex.nextToken(); // eat '('
   
   std::shared_ptr<FieldGenerator> g = parseExpression(lex);
-  if(!g)
-    return NULL;
-  
-  if((lex.curtok != ')') && (lex.curtok != ']'))
-    return NULL;
+  if (!g)
+    throw ParseException("Incomplete parentheses; did you forget a ')'?");
```
ZedThree: I think we can just delete this completely, as `parseExpression` now returns something or throws

bendudson: Ah good spot. I think I deleted the wrong if; the check
`if((lex.curtok != ')') && (lex.curtok != ']'))` 
should be there, rather than `if (!g)`

ZedThree:
Sorry, got distracted and did something weird there!

ZedThree:
Added some commits that further tidy up ExpressionParser:

- remove some unused/unreachable code
- use a type alias for `std::shared<FieldGenerator>`
- use `std::make_shared`/`std::make_pair` where appropriate
- add some more unit tests, bringing coverage on expressionparser up to >90%

include/bout/sys/expressionparser.hxx
```diff
@@ -41,6 +41,8 @@ class ParseException;
 #include <memory>
 #include <exception>
 
+using FieldGenerator_ptr = std::shared_ptr<FieldGenerator>;
```
d7919: Is this name consistent with our conventions (I think it might not be, but not sure)?

PR #1094, opened 2018-06-05
fix isMultiple

### LINKED ISSUE #1093: Unexpected behaviour with bout_abs and isMultiple in monitor.hxx
`bout_abs` returns -1 if its argument is negative. Given the name, I would expect it to do something like return the absolute value. What is the intended behaviour? Can it just be replaced with `std::abs`?

I'm not 100% sure what `isMultiple` is supposed to do. It looks like it's supposed to return `true` if either argument is a multiple of the other one, but:
- if either argument is `int` zero, there's a divide-by-zero. For `double`, evaluates to `true`
- it evaluates to `true` for **any** `int` arguments
- it evaluates to `true` for some surprising `double` arguments, e.g. `isMultiple(3., 4.) == true`

What is the intended behaviour? What is its range of validity? Can it just be replaced with modulo?

dschwoerer:
* bout_abs was missing *a - replaced by std::abs
* isMultiple is really only needed for boutreals
* Negative (inc. 0) values are not expected, so we check now for that.
* use round() rather than casting to int - no risk of overflow
* use 1e-12 rather than 1e-5 - should be achievable using doubles

@ZedThree does this sufficiently resolve #1093 ?

### DIFF
--- a/include/bout/monitor.hxx
+++ b/include/bout/monitor.hxx
@@ -3,29 +3,40 @@
 #include "bout_types.hxx"
 
 class Solver;
-
+/// Check if two numbers are multiples of each other, with an accuracy
+/// of 1e-12
 bool isMultiple(BoutReal a, BoutReal b);
 
+/// Monitor baseclass for the Solver
+/// Can be called ether with a specified frequency, or with the
+/// frequency of the BOUT++ output monitor.
 class Monitor{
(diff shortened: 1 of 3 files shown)


ZedThree:
Looks good on first blush, but does need a doxygen comment on it please!

I was writing unit tests for them today which is how I noticed, so I will
merge them into this tomorrow

PR #1095, opened 2018-06-05
Check field location in solver

dschwoerer:
* Compare location to that of field
* Don't interpolate but throw

### DIFF
--- a/src/solver/solver.cxx
+++ b/src/solver/solver.cxx
@@ -1102,9 +1102,7 @@ void Solver::save_derivs(BoutReal *dudata) {
   // Make sure 3D fields are at the correct cell location
   for(const auto& f : f3d) {
     if(f.var->getLocation() != (f.F_var)->getLocation()) {
-      //output.write("SOLVER: Interpolating\n");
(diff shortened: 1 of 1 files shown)


src/solver/solver.cxx
```diff
@@ -1101,8 +1101,9 @@ void Solver::save_derivs(BoutReal *dudata) {
 
   // Make sure 3D fields are at the correct cell location
   for(const auto& f : f3d) {
-    if(f.location != (f.F_var)->getLocation()) {
+    if(f.var->getLocation() != (f.F_var)->getLocation()) {
       //output.write("SOLVER: Interpolating\n");
```
d7919: Can get rid of this commented out line and the call to `interp_to` below.

ZedThree:
Why throw instead of interpolating? I'm guessing you've found it leads to wrong results or something?

bendudson:
Probably best to throw rather than silently interpolate: the user probably doesn't intend to calculate time derivatives at the wrong location, and if they do they they can put an explicit interpolation call in.

PR #1098, opened 2018-06-06
Monitor tidy

ZedThree:
- Simplifies implementation of `isMultiple`, makes it inline
- Removes `MonitorFunc` and monitor.cxx
- Replaces `#pragma once` with header guards


dschwoerer:
I wanted to say that `#pragma once` is finally supported by all compilers, but it seems cray dropped the support :sob: 
https://en.wikipedia.org/wiki/Pragma_once#Portability

ZedThree:
Yeah, the `#pragma once` definitely looks nicer, but it also has its own problems unfortunately. Notably, it doesn't play nice with symlinks for example.

This is currently failing due to some of the old PETSc interfaces using `MonitorFunc`. They'll be removed with #1050, so that should go in first.

PR #1100, opened 2018-06-07
Fix some memory leaks in multigrid

ZedThree:
The multgrid objects were not being `delete`d, just set to `NULL`. This replaces them with `unique_ptr` so they'll get cleaned up automatically.

And because I was touching them, I've cleaned up the ctors/dtors a little.


src/invert/laplace/impls/multigrid/multigrid_solver.cxx
```diff
@@ -764,16 +717,4 @@ MultigridSerial::MultigridSerial(int level, int gx, int gz, int UNUSED(px),
       output<<i<<" Ser glo dim "<<gnx[i]<<","<<gnz[i]<<endl;
     }    
   }
-
-  // This could use a Matrix
```
bendudson: Where is this allocation now being done? I see matmg is deleted in the destructor, but where is it created?

ZedThree: Most of this code was done in the `MultigridAlg` constructor, which was called first, then repeated here

PR #1104, opened 2018-06-12
Add option for setting the window title in showdata

johnomotani:
Adds another optional argument for showdata that can be used to set the title of the window. For example, I'm using this to show the directory where the data being plotted has come from when calling showdata() from a script.


bendudson:
Tests don't check showdata, and only a small change.

PR #1105, opened 2018-06-12
More SI constants

dschwoerer:
I thought they might be useful to have ...
Not sure they should all be in the SI name space, or whether there should be name spaces for atom or mass or both ...

### DIFF
--- a/include/bout/constants.hxx
+++ b/include/bout/constants.hxx
@@ -22,6 +22,10 @@ namespace SI {
   const BoutReal Me  = 9.10938356e-31;  ///< Electron mass
   const BoutReal Mp  = 1.672621898e-27; ///< Proton mass
   const BoutReal kb  = 1.38064852e-23;  ///< Boltzmanns constant
+  const BoutReal u   = 1.660539040e-27;        ///< Unified atomic mass unit
+  const BoutReal M_Hydrogen = 1.008 * u;       ///< Mass of a Hydrogen atom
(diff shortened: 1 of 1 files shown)


include/bout/constants.hxx
```diff
@@ -22,6 +22,10 @@ namespace SI {
   const BoutReal Me  = 9.10938356e-31;  ///< Electron mass
   const BoutReal Mp  = 1.672621898e-27; ///< Proton mass
   const BoutReal kb  = 1.38064852e-23;  ///< Boltzmanns constant
+  const BoutReal u   = 1.660539040e-27;        ///< Unified atomic mass unit
```
bendudson: Not sure about single character names. Perhaps `amu` or `uamu`?

dschwoerer: `u` and `Da` are the  official names: https://en.wikipedia.org/wiki/Unified_atomic_mass_unit
I haven't heard `Da` often, so I chose `u`.
I thought single character might be fine, as the name space is so small, but I am happy to change it ...

ZedThree: My vote's for `amu`

bendudson: Agree; although according to that Wikipedia page `amu` is technically an obsolete unit, `amu` is used now to refer to `u` or `Da`. It probably doesn't matter, but my order of preference is `amu` > `Da` > `u`.

dschwoerer: Ok, changed :+1:

PR #1106, opened 2018-06-12
Additions to Python tools

johnomotani:
A few small upgrades for various Python tools:

* Previously, `showdata()` used `collect('t_array')` to get simulation times for the animation's titles. Now allow `t_array` to be passed in as an argument instead in case `showdata()` is not called from the same directory as the data, or in case the t-direction is sliced. Default maintains original behaviour.

* Previously when reading from a 'parallel' dump file, the slicing arguments tind, xind, yind, zind could not be used. Now implemented also for this case.

* Added a write_new() method for BoutOptionsFile to write a new BOUT.inp file, including any values that have been changed or added.

I can split these up if it is preferred, but they are all fairly small changes.

### DIFF
--- a/tools/pylib/boutdata/collect.py
+++ b/tools/pylib/boutdata/collect.py
@@ -172,10 +172,9 @@ def collect(varname, xind=None, yind=None, zind=None, tind=None, path=".",
 
     if datafile_cache is None:
         # Search for BOUT++ dump files
-        file_list, parallel, suffix = findFiles(path, prefix)
+        file_list, parallel, _ = findFiles(path, prefix)
     else:
         parallel = datafile_cache.parallel
-        suffix = datafile_cache.suffix
         file_list = datafile_cache.file_list
 
     def getDataFile(i):
--- a/tools/pylib/boutdata/collect.py
+++ b/tools/pylib/boutdata/collect.py
@@ -68,6 +68,40 @@ def findVar(varname, varlist):
               "' not found, and is ambiguous. Could be one of: "+str(v))
     raise ValueError("Variable '"+varname+"' not found")
 
+def check_range(r, low, up, name="range"):
+    """ Utility function to check index ranges"""
+    r2 = r
+    if r is not None:
+        try:
+            n = len(r2)
+        except:
+            # No len attribute, so probably a single number
+            r2 = [r2, r2]
+        if (len(r2) < 1) or (len(r2) > 2):
+            print("WARNING: "+name+" must be [min, max]")
+            r2 = None
+        else:
+            if len(r2) == 1:
+                r2 = [r2, r2]
+            if r2[0] < 0 and low >= 0:
+                r2[0] += (up-low+1)
+            if r2[1] < 0 and low >= 0:
+                r2[1] += (up-low+1)
+            if r2[0] < low:
+                r2[0] = low
+            if r2[0] > up:
+                r2[0] = up
+            if r2[1] < low:
+                r2[1] = low
+            if r2[1] > up:
+                r2[1] = up
+            if r2[0] > r2[1]:
+                tmp = r2[0]
+                r2[0] = r2[1]
+                r2[1] = tmp
+    else:
+        r2 = [low, up]
+    return r2
 
 def collect(varname, xind=None, yind=None, zind=None, tind=None, path=".",
             yguards=False, xguards=True, info=True, prefix="BOUT.dmp",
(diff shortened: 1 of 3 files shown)


bendudson:
Thanks John!

1. I'm not sure if the original behaviour makes sense really: If I pass in a subset of the array, for example keeping only every n'th time point, then the time is wrong. I support dropping the original behaviour as mixing data reading with data plotting.

2. I didn't realise the parallel dumping worked! I'm pleasantly surprised. Is this HDF5?

3. Thanks; saving BOUT.inp files sounds really useful for automating parameter scans. (see also `bout_runners` by Michael).

tools/pylib/boutdata/data.py
```diff
@@ -247,6 +249,50 @@ def __init__(self, filename, name="root"):
 
                         section[line[:eqpos].strip()] = value
 
+    def write_new(self, filename=None, force_overwrite=False):
```
ZedThree: Could just be `write` rather than `write_new`, as it doesn't necessarily make a new file

Also `force_overwrite` could just be `overwrite`

tools/pylib/boutdata/data.py
```diff
@@ -247,6 +249,50 @@ def __init__(self, filename, name="root"):
...
+        """
+        Write a new input file with the options (including any modifications
```
ZedThree: For the [numpdocs style][1], this should start on the same line as the triple quotes, with an empty line between the short summary and the rest. I would also simplify the short doc to:

```python
"""Write to BOUT++ options file

By default...
```

[1]: https://numpydoc.readthedocs.io/en/latest/format.html#numpydoc-docstring-guide

johnomotani:
Sorry, 'parallel dump files' was misleading. I collect my dump files into a single file for downloading/reading, using a python script. Then the python tools treat that file as a 'parallel' output file. I guess I might as well upload that too, might be useful to someone...

OK, I'll change the behaviour like 1.

tools/pylib/boutdata/data.py
```diff
@@ -247,6 +249,50 @@ def __init__(self, filename, name="root"):
...
+        since reading in).
+        By default this will overwrite the original input file, prompting the
+        user for confirmation.
+        Note, no comments from the original input file are transferred to the
+        new one.
+
+        Parameters
+        ----------
+
```
ZedThree: This empty line can cause problems rendering with sphinx

tools/pylib/boutdata/data.py
```diff
@@ -247,6 +249,50 @@ def __init__(self, filename, name="root"):
...
+            file, if there is one. Otherwise, just overwrite without asking.
+            (default False)
+        """
+        if filename is None:
+            filename = self.filename
+
+        if not force_overwrite and os.path.exists(filename):
+            user_input = "a"
+            while user_input not in ["y", "n", ""]:
+                user_input = input(filename+" already exists: do you want to over-write it? y/[n]")
```
ZedThree: I'm not sure about this. It's fine if it's being used interactively, but in a script, this could cause the process to just wait indefinitely. Probably better to just raise the `ValueError` instead

johnomotani:
I've changed as @ZedThree suggested.

Also re-wrote the commit adding the squashoutput.py script with some useless code cut out and adding a docstring for the function.

How do we handle Python scripts that can be called as executables? Would it be useful to symlink them from the `bin` directory or do we just leave it to the user to add `tools/pylib/boutdata`, etc., to their $PATH if they want to?

ZedThree:
They can always be run like e.g. `python -m boutdata.squashoutput`

tools/pylib/boutdata/squashoutput.py
```diff
@@ -0,0 +1,150 @@
...
+      of the output file
+"""
+
+from boutdata.data import BoutOutputs
+from boututils.datafile import DataFile
+from boututils.boutarray import BoutArray
+import numpy
+import os
+
+def squashoutput(datadir=".", outputname="BOUT.dmp.nc", format="NETCDF4", tslice=[None,None,None], xslice=[None,None,None], yslice=[None,None,None], zslice=[None,None,None], singleprecision=False):
```
ZedThree: The default argument for `tslice` etc. should just be `None` rather than a list of `None`s. See [here][1] for why. Note that you'll need to modify your checks later.


[1]: http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments

tools/pylib/boutdata/squashoutput.py
```diff
@@ -0,0 +1,150 @@
...
+    # Create single file for output and write data
+    with DataFile(fullpath,create=True,write=True,format=format) as f:
+        for varname in outputvars:
+            print(varname)
+
+            var = outputs[varname]
+            if singleprecision:
+                var = BoutArray(numpy.float32(var), var.attributes)
+            if var.attributes["bout_type"] == "Field3D_t":
+                var = var[tslice[0]:tslice[1]:tslice[2], xslice[0]:xslice[1]:xslice[2], yslice[0]:yslice[1]:yslice[2], zslice[0]:zslice[1]:zslice[2]]
```
ZedThree: I think this could be tidied up if you used `slice` objects instead of lists/tuples here. Possibly even take them as arguments in the first place? You could check if you got a `list`/`tuple` with `isinstance(tslice, (list, tuple))` and then just convert to `slice`

ZedThree:
Codacy has flagged a few issues: https://app.codacy.com/app/ZedThree/BOUT-dev/pullRequest?prid=1769564#newIssuesView

johnomotani:
Agreed slice objects would be neater. When I looked at it, I thought it would be nicer to use slice objects for the *ind arguments to `collect()` and ranges arguments to `DataFile.read()`, then in `squashoutputs()` I can just pass them through. So I've updated DataFile and collect. This could create backward compatibility issues with user Python scripts, in particular it changes the meaning of the *ind arguments to use standard Python slicing conventions for '(start, stop)' instead of inclusive '(min,max)' indices.

I've fixed the Codacy issues.

johnomotani:
This PR now contains only updates to slicing/range arguments of `DataFile` and `collect()` and the addition of the squashoutput function/script.

ZedThree:
Sorry @johnomotani, I think I've sent you off down the wrong path a bit talking about `slice`s, as this now starts breaking backwards compatibility. Unfortunately, `collect` for instance is a little sloppy with the `xind` arguments, etc., which makes maintaining backwards compatibility a bit tough:

- `collect` uses *inclusive* maximum, not *exclusive* like standard Python indexing, so `xind=[2, 2]` would need to be `xind=[2, 3]` to give you the single x index at 2
- `collect` actually sorts the `xind` arguments (after bringing into the range `[0, nx]`), so `xind=[2, 8]` and `xind=[8, 2]` give you the same result

One might think of both of these as bugs though, in which case, breaking backwards compatibility is probably fine.

Functions should still be able to take the same values though, so `DataFile.read` will probably need to convert ranges to slices (though that was using Python indexing, so should be a bit easier).

I've already started looking at this, so I'll just merge my changes straight in here

johnomotani:
@ZedThree DataFile.read does still take ranges, but not in a backward compatible way. It wouldn't be hard to add a check for the old-style list and handle that. I guess I'd assumed that the only place likely to be using the ranges argument was collect() so as long as both got updated it didn't matter too much. Apart from changing from inclusive to exclusive maximum, collect should maintain the old behaviour for *ind arguments if they are given as an int or a list, although the code could probably at least use more comments and a more informative docstring.

For DataFile I think maybe breaking backward compatibility is not too bad, because it will throw an exception if called with the old-style argument. Changing *ind from inclusive to exclusive maximum in collect worries me more, because it will (probably) run and change behaviour silently. BTW for the example you gave, you could and can still use an integer like xind=2 if you want a single index.

I agree that your second bullet is a bug, I think if I put xind=[8,2] I'd expect either an error or an empty result.

d7919:
Should we add an action to redo the python tools (or at least their interface) for the 5.0 milestone?

ZedThree:
@johnomotani `DataFile.read` I've got a fix to maintain backwards compatibility. I think I've got one too for `collect`, need to test again.

@d7919 Yes, I think the whole of pylib needs revisiting

ZedThree:
I've pushed fixes to maintain backwards compatibility for `collect` and `DataFile.read`, except `xind=[8,2]` now throws, which I think is sensible.

I've also simplified some things and added some relevant cleanup from another branch I've been working on

ZedThree:
@johnomotani Could you check that this works for your use cases in its current form?

It passes the integrated tests, so I'm assuming `collect` is still backwards compatible, but hopefully I've not broken it for your new cases

tools/pylib/boutdata/collect.py
```diff
@@ -141,63 +203,73 @@ def getDataFile(i):
...
-            xstart = 0
-            xlim = None
+            nx = f["nx"]
         else:
-            xstart = mxg
-            if mxg > 0:
-                xlim = -mxg
-            else:
-                xlim = None
+            nx = f["nx"] - 2*mxg
```
johnomotani: For this single data file case, I think xguards==False needs fixing. In the multiple file case below, the boundary guard cells get cut out if xguards==False before xind is used. But here, xind is just passed straight to `DataFile.read()` so to maintain the same behaviour as the multiple-file case, we need to offset the start and stop values of xind by mxg if xguards==False.

tools/pylib/boutdata/collect.py
```diff
@@ -141,63 +203,73 @@ def getDataFile(i):
...
-            ystart = 0
-            ylim = None
+            ny = f["ny"] + 2*myg
         else:
-            ystart = myg
-            if myg > 0:
-                ylim = -myg
-            else:
-                ylim = None
+            ny = f["ny"]
```
johnomotani: Similar issue to xguards==False case above.

tools/pylib/boutdata/collect.py
ZedThree: Could you email me a failing example so I get this right? Unfortunately, we're missing a test suite for the python tools

tools/pylib/boutdata/squashoutput.py
```diff
@@ -0,0 +1,110 @@
...
+                return None
+            else:
+                raise
+    parser.add_argument("datadir", nargs='?', default=".")
+    parser.add_argument("--outputname",default="BOUT.dmp.nc")
+    parser.add_argument("--tind", type=int_or_none, nargs='*', default=[None])
+    parser.add_argument("--xind", type=int_or_none, nargs='*', default=[None])
+    parser.add_argument("--yind", type=int_or_none, nargs='*', default=[None])
+    parser.add_argument("--zind", type=int_or_none, nargs='*', default=[None])
+    parser.add_argument("--singleprecision", type=str_to_bool, default=True)
```
ZedThree: This could just be a simple flag to turn off using single precision using `action="store_false"`

PR #1107, opened 2018-06-12
Add write method to BoutOptionsFile and update showdata

johnomotani:
I've pulled these changes out of #1106 because they are simpler. Adding write() to BoutOptionsFile is backward compatible, and the t_array change to showdata may change some animation titles but should not break anything.

### DIFF
--- a/tools/pylib/boutdata/data.py
+++ b/tools/pylib/boutdata/data.py
@@ -193,6 +193,7 @@ class BoutOptionsFile(BoutOptions):
     """
 
     def __init__(self, filename, name="root"):
+        self.filename = filename
         BoutOptions.__init__(self, name)
         # Open the file
         with open(filename, "r") as f:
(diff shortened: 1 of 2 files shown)


tools/pylib/boutdata/data.py
```diff
@@ -247,6 +248,44 @@ def __init__(self, filename, name="root"):
...
+            (defaults to path of the file that was read in)
+        force_overwrite : bool
+            If False then prompt the user whether to overwrite an existing
+            file, if there is one. Otherwise, just overwrite without asking.
+            (default False)
+        """
+        if filename is None:
+            filename = self.filename
+
+        if not force_overwrite and os.path.exists(filename):
```
bendudson: `force_overwrite` should be `overwrite` to be consistent with arguments.

tools/pylib/boutdata/data.py
```diff
@@ -247,6 +248,44 @@ def __init__(self, filename, name="root"):
...
+        user for confirmation.
+        Note, no comments from the original input file are transferred to the
+        new one.
+
+        Parameters
+        ----------
+        filename : str
+            Path of the file to write
+            (defaults to path of the file that was read in)
+        force_overwrite : bool
```
bendudson: Comments should be updated: user isn't prompted anymore but an exception is raised

johnomotani: This and also the previous comment `force_overwrite`->`overwrite` are fixed now.

ZedThree:
This does change the previous behaviour for `showdata`, though it doesn't look like it breaks anything. Does anyone have strong opinions one or the other? I do see the appeal of not calling `collect` in a plotting routine.

tools/pylib/boutdata/data.py
```diff
@@ -247,6 +248,44 @@ def __init__(self, filename, name="root"):
...
+            (default False)
+        """
+        if filename is None:
+            filename = self.filename
+
+        if not overwrite and os.path.exists(filename):
+            raise ValueError("Not overwriting existing file, cannot write output to "+filename)
+
+        def write_section(basename, opts, f):
+            if basename is not "":
```
ZedThree: Could just be `if basename:`, unless I'm missing something? Similarly for the same conditional below.

tools/pylib/boutdata/data.py
```diff
@@ -247,6 +248,44 @@ def __init__(self, filename, name="root"):
...
+                f.write("["+basename+"]\n")
+            for key, value in opts._keys.items():
+                f.write(key+" = "+str(value)+"\n")
+            for section in opts.sections():
+                if basename is not "":
+                    write_section(basename+":"+section, opts[section], f)
+                else:
+                    write_section(section, opts[section], f)
```
ZedThree: I would be tempted to write this to remove the semi-duplicated branch:

```python
section_name = basename+":"+section if basename else section
write_section(section_name, opts[section, f)
```
although I appreciate that the ternary can be a bit confusing

PR #1110, opened 2018-06-14
pol_slice tidy and simplify

ZedThree:
- Use a context manager to open grid file, so it always gets closed
- Raise exceptions rather than print errors
- Simplify conditional
- Simplify use of map_coordinates
- PEP8 formatting

### DIFF
--- a/tools/pylib/boutdata/pol_slice.py
+++ b/tools/pylib/boutdata/pol_slice.py
@@ -68,9 +68,11 @@ def pol_slice(var3d, gridfile, n=1, zangle=0.0, nyInterp=None):
 
         # Interpolate to output positions and make the correct shape
         # np.mgrid gives us an array of indices
-        var3d = map_coordinates(varTmp, coordinates=np.mgrid[0:nx, 0:nyInterp, 0:nz],
+        # 0:ny-1:nyInterp*1j means use nyInterp points between 0 and ny-1 inclusive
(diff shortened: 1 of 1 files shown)


tools/pylib/boutdata/pol_slice.py
```diff
@@ -62,45 +60,26 @@ def pol_slice(var3d, gridfile, n=1, zangle=0.0, nyInterp=None):
...
-    #Setup interpolation if requested
-    if doInterp:
+                raise ValueError("Neither qinty nor zShift found")
 
+    # Decide if we've asked to do interpolation
+    if nyInterp is not None and nyInterp != ny:
         varTmp = var3d
-        #These are the index space co-ordinates of the output arrays
-        xOut = np.linspace(0,nx-1,nx)
-        yOut = np.linspace(0,ny-1,nyInterp)
```
d7919: `yOut` here is a high resolution sample of y index space spanning 0 to `ny-1`, in the new version with `mgrid` seems to just use the presumably integer range `0` to `nyInterp-1`. I've not tested this but I'm not sure I'd expect this to work.

ZedThree: Fixed with `0:ny-1:nyInterp*1j` which gives `nyInterp` points between 0 and `ny-1` inclusive. Confirmed that this gives identically behaviour to original implementation

PR #1114, opened 2018-06-15
Add basic IndexOffset class + tests

ZedThree:
Essentially offsets from DataIterator pulled out into standalone class


ZedThree:
> You are missing the case that i is > 1, and index.ind % nz is < i.
In that case the first test is insufficient.

I've added a test for `zmm` which seems to work as-is.

include/bout/indexoffset.hxx
```diff
@@ -0,0 +1,50 @@
...
+                "IndexOffset must be templated with either Ind2D or Ind3D");
+
+  Mesh &mesh;
+  const int nx, ny, nz;
+
+  IndexOffset(Mesh &mesh)
+      : mesh(mesh), nx(mesh.LocalNx), ny(mesh.LocalNy), nz(mesh.LocalNz) {}
+
+  /// Convenience functions for converting to (x, y, z)
+  int x(T index) const { return (index.ind / nz) / ny; }
```
d7919: `nz` is initialised by whatever mesh thinks `nz` is -- presumably this is ok when templated on `Ind3D` but not when for `Ind2D` where effective `nz` is 1?

include/bout/indexoffset.hxx
```diff
@@ -0,0 +1,50 @@
...
+  const inline T xp(T index, int i = 1) const { return index + (i * ny * nz); }
+  /// The index one point -1 in x
+  const inline T xm(T index, int i = 1) const { return index - (i * ny * nz); }
+  /// The index one point +1 in y
+  const inline T yp(T index, int i = 1) const { return index + (i * nz); }
+  /// The index one point -1 in y
+  const inline T ym(T index, int i = 1) const { return index - (i * nz); }
+  /// The index one point +1 in z. Wraps around zend to zstart
+  const inline T zp(T index, int i = 1) const {
+    ASSERT3(i > 0);
```
d7919: Should we have these asserts for all directions or do we want to allow negative offsets?

include/bout/indexoffset.hxx
```diff
@@ -0,0 +1,50 @@
...
+    return (index + i) % nz == 0 ? index - nz + i : index + i;
```
d7919: We're paying the cost of a mod for each call and then doing another calculation to find the result. I think we can probably just use a calculation involving mod to find the answer without a further calculation? The ternary style is handy when we know i can only be 1 so that we can avoid the mod in the logic part.

include/bout/indexoffset.hxx
```diff
@@ -0,0 +1,50 @@
...
+  IndexOffset(Mesh &mesh)
+      : mesh(mesh), nx(mesh.LocalNx), ny(mesh.LocalNy), nz(mesh.LocalNz) {}
+
+  /// Convenience functions for converting to (x, y, z)
+  int x(T index) const { return (index.ind / nz) / ny; }
+  int y(T index) const { return (index.ind / nz) % ny; }
+  int z(T index) const { return (index.ind % nz); }
+
+  const inline T xp(T index, int i = 1) const { return index + (i * ny * nz); }
+  /// The index one point -1 in x
```
d7919: Note comment isn't consistent with ability to pass in `i`.

include/bout/indexoffset.hxx
ZedThree: Right. We'll need to specialise the whole template for `Ind2D`, which will unfortunately involve some repeated code, but I'm not sure of a better way.

include/bout/indexoffset.hxx
ZedThree: For x and y, I don't think it matters, as their implementations are symmetric, it's just for z that it seems to be important.

d7919: It was more for consistency (i.e. easier to document).

include/bout/indexoffset.hxx
```diff
@@ -0,0 +1,50 @@
...
+    return index % nz == 0 ? index + nz - i : index - i;
+  }
+  // and for 2 cells
+  const inline T xpp(T index) const { return xp(index, 2); }
+  const inline T xmm(T index) const { return xm(index, 2); }
+  const inline T ypp(T index) const { return yp(index, 2); }
+  const inline T ymm(T index) const { return ym(index, 2); }
+  const inline T zpp(T index) const { return zp(index, 2); }
+  const inline T zmm(T index) const { return zm(index, 2); }
+};
```
d7919: It might be helpful to provide a generic `offset(ix,iy,iz)` routine?

include/bout/indexoffset.hxx
ZedThree: We discussed this earlier, but for future reference: because we only have the single index, we have to pay the cost of the modulus to check if we've reached either end of z, but the modulus is not sufficient for the actual periodicity -- we need to shift by nz.

Hopefully, the z-offsetting is not performance critical. If it is, we could switch to using precalculated offset indices. This interface should still work in that case.

ZedThree:
- Made the z offsetting work correctly with periodicity
- Added a generic `offset` method
- Initialise `nz` to 1 if `std::is_base_of<Ind2D, T>` is true
- Added bounds checking for x, y offsets. This is at CHECK = 4+, so won't be on by default
- Added a ton more unit tests

include/bout/indexoffset.hxx
```diff
@@ -0,0 +1,99 @@
...
+#endif
+    return index + (dy * nz);
+  }
+  /// Negative offset \p index by \p dy in y, default to offset by one
+  const inline T ym(T index, int dy = 1) const { return yp(index, -dy); }
+  /// Positive offset \p index by \p dz in z, default to offset by one.
+  /// Periodic in z, cannot handle negative offsets
+  const inline T zp(T index, int dz = 1) const {
+    ASSERT3(dz > 0);
+    if (nz == 1) {
```
JosephThomasParker: Does branching here and in `zm` have performance implications?

include/bout/indexoffset.hxx
```diff
@@ -0,0 +1,99 @@
...
+///     auto index = std::begin(region);
+///     IndexOffset<Ind3D> offset(*mesh);
+///
+///     result = field[offset.yp(*index)] - field[offset.ym(*index)];
+///
+template <typename T = Ind3D>
+struct IndexOffset {
+
+  static_assert(std::is_base_of<Ind2D, T>::value || std::is_base_of<Ind3D, T>::value,
+                "IndexOffset must be templated with either Ind2D or Ind3D");
```
JosephThomasParker: Can we also include `IndPerp` in this?  Is it as simple as checking `ny=1` in the same way as `Ind2D` does with `nz=1`?

include/bout/indexoffset.hxx
ZedThree: Plausibly, but I suspect the branch predictor will save us. It will destroy any chance of vectorising loops, which might be the bigger deal. I'm not sure what can be done to remove it -- the offsetting can't be done for `Field2D`, and it's not possible to specialise a single member function.

JosephThomasParker: What is the expect behaviour when using `Field2D`, ie `nz == 1`?  I'd expect `zp` to be ignored, in which case `zp(index2d, dz)` should return `index2d` (the `nz==1` logic), not `index2d - 1 + dz` (the first branch of the ternary).

d7919: With the implicit assumption that `dz<=nz` we have `index2D == index2D -1 + dz`

ZedThree:
We have to make a choice about how to deal with offsetting in a direction that only has a single point, i.e. `Ind2D`, `IndPerp`. Some combination of the following:

- `ASSERT3(abs(d?) < n?)` -- note that z can be `dz <= nz` to account for the periodicity, and this will be handled correctly
- `if (n? == 1) return index`

In the first case, we're saying explicitly that it's an error to offset in a direction with a single point, but we avoid branching.

In the second case, we all offsetting in any direction by any amount, it just might not do anything. This potentially makes generic code simpler? `DDY(FieldPerp)` would automatically return `0.` for instance. But we might end up with branches in hot loops, and might also prohibit vectorisation.

Personally, I'm in favour of the first option.

bendudson:
Agreed; I think this is a good choice. If the check is not going to be on by default, using `ASSERT3` then I think the manual should more clearly point out that development should be done at higher checking levels.

Note that quite a few models evolve `Field3D` fields, but only have 2D domains e.g. `blob2d`. 

Short-cut optimisation like `DDY(FieldPerp) -> 0` should be done with a single check outside the main loop.

ZedThree:
The default `CHECK` level is 2 -- would it be better for these checks to be on at that level?

bendudson:
As long as the performance hit isn't terrible then yes I think so. Correct by default is probably better than fast.

ZedThree:
Replaced by #1211

PR #1119, opened 2018-06-18
Delete old INVERT_IN_RHS and INVERT_OUT_RHS flags

johnomotani:
These were removed in 8eacc13f09acfb756b0f7c9f697be938a8f35836 but added back in 5f86706813a601f358004a306d23afa08ec560d5. Not sure why they were added again, but they do not seem to be used anywhere now.

### DIFF
--- a/include/invert_laplace.hxx
+++ b/include/invert_laplace.hxx
@@ -99,9 +99,6 @@ const int INVERT_KX_ZERO     = 16; ///< Zero the kx=0, n = 0 component
   const int INVERT_DC_IN_GRADPARINV = 2097152;
  */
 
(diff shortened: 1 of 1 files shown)


bendudson:
It is still used, but the uses could definitely be tidied up. It's used for gyro-averaging operators:
https://github.com/boutproject/BOUT-dev/blob/next/include/gyro_average.hxx#L37
https://github.com/boutproject/BOUT-dev/blob/next/examples/gyro-gem/gem.cxx#L622
but the magic number is used rather than the symbol:
https://github.com/boutproject/BOUT-dev/blob/next/src/invert/laplace/invert_laplace.cxx#L839

PR #1124, opened 2018-06-18
Set result location in Delp2 from incoming field

AdamD94:
Delp2 doesn't currently handle staggered fields and throws when the input field is not cell centred. 
I don't think Delp2 needs to be stagger aware since it operates on xz slices. So I think setting the result location based on the input field fixes the problem.


johnomotani:
This seems sensible, and is consistent with what the rest of the code does at the moment. Delp2 should be location dependent in the metric terms in the call to tridagcoefs, which is addressed in #1060.

PR #1127, opened 2018-06-19
Add implementation of 'Naulin' Laplacian solver

johnomotani:
Based on original BOUT++ version by Michael Loiten in CELMA https://celma-project.github.io/

This is an iterative solver that uses the FFT solvers for the Delp2 part of the Laplacian inversion, and an iteration for the remaining terms that cannot be solved directly.

Includes #1122

### DIFF
--- a/src/field/field3d.cxx
+++ b/src/field/field3d.cxx
@@ -829,6 +829,44 @@ BoutReal max(const Field3D &f, bool allpe, REGION rgn) {
   return result;
 }
 
+BoutReal mean(const Field3D &f, bool allpe, REGION rgn) {
+  TRACE("Field3D::Max() %s",allpe? "over all PEs" : "");
+
+  ASSERT2(f.isAllocated());
+
+  Mesh *localmesh = f.getMesh();
+
+  int npoints;
+  switch (rgn) {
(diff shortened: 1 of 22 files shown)


src/field/field3d.cxx
```diff
@@ -829,6 +829,44 @@ BoutReal max(const Field3D &f, bool allpe, REGION rgn) {
...
+  
+  for(const auto& i: f.region(rgn))
+    result += f[i];
+
+  result /= npoints;
+  
+  if(allpe) {
+    // MPI reduce
+    BoutReal localresult = result;
+    MPI_Allreduce(&localresult, &result, 1, MPI_DOUBLE, MPI_MAX, BoutComm::get());
```
d7919: I might be failing at basic maths here but don't you need to divide this by the number of processors as well? I think that would also only work if `npoints` is uniform over processors (which might not be true for future regions). Probably the safest is to also `allreduce` `npoints` and do the division `result/=npoints` after the reduction?

src/field/field3d.cxx
```diff
@@ -829,6 +829,44 @@ BoutReal max(const Field3D &f, bool allpe, REGION rgn) {
   return result;
 }
 
+BoutReal mean(const Field3D &f, bool allpe, REGION rgn) {
```
d7919: Could you add unit tests for this to the existing `Field3D` tests?

PR #1130, opened 2018-06-20
Extend support for 1D simulation

dschwoerer:
* Delp2 takes the sum of X and Z derivatives - if both are zero, the
result is zero. Therefore and required
* bracket is the product of two derivatives - if ether is zero, the
whole derivative vanishes

### DIFF
--- a/src/mesh/coordinates.cxx
+++ b/src/mesh/coordinates.cxx
@@ -643,6 +643,10 @@ const Field2D Coordinates::Delp2(const Field2D &f) {
 const Field3D Coordinates::Delp2(const Field3D &f) {
   TRACE("Coordinates::Delp2( Field3D )");
 
+  if (localmesh->GlobalNx == 1 && localmesh->GlobalNz == 1) {
+    // copy mesh, location, etc
+    return f*0;
+  }
   ASSERT2(localmesh->xstart > 0); // Need at least one guard cell
 
   Field3D result(localmesh);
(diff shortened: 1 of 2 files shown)


src/mesh/coordinates.cxx
```diff
@@ -643,6 +643,10 @@ const Field2D Coordinates::Delp2(const Field2D &f) {
 const Field3D Coordinates::Delp2(const Field3D &f) {
   TRACE("Coordinates::Delp2( Field3D )");
 
+  if (localmesh->GlobalNx == 1 && localmesh->GlobalNz == 1) {
+    // copy mesh, location, etc
+    return f*0;
```
bendudson: Can this just be `return 0.0` or `return Field3D(0.0)` to avoid multiplication?

dschwoerer: No, it would need to be at least:
```
Field3D result(f.getMesh());
result=0;
result.setLocation(f.getLocation());
return result;
```
Maybe it is time to provide a simpler way to achieve copying all the data from the Field, except the data.

This would maybe work as well, as copying the field does not duplicate the data, only when the data is changed a new data block is allocated.
```
Field3D result = f;
result=0;
return result;
```
Anyway - I think the above multiplication shouldn't at all be an issue, as the field only has LocalNy points. In that case raw numerics is often cheap compared to all the other overheads ...

PR #1131, opened 2018-06-21
real Y

dschwoerer:
`y` in the BOUT.inp does not take dy into account, and always goes from [0:2*pi]
realy() is implemented as a function, and takes dy into account.
So when dy is 1, `realy()` goes from [-myg:ny+myg] - or something like that,
depending on the staggering etc.
Therefore realy does need to no the mesh, to get dy
MPI communication is also needed.

The question is: Is something like this going to be useful for bout?


src/field/fieldgenerators.cxx
```diff
@@ -268,13 +269,148 @@ std::shared_ptr<FieldGenerator> FieldTanhHat::clone(const list<std::shared_ptr<F
...
+      for (int y = 0; y < localmesh->LocalNy - localmesh->ystart * 2; ++y) {
+        tmp[x] += localmesh->coordinates()->dy(x, y);
+      }
+    }
+    // printf("%d: send %g\n",localmesh->getYProcIndex(),tmp[localmesh->xstart]);
+    MPI_Request out =
+        localmesh->sendToProc(localmesh->getXProcIndex(), localmesh->getYProcIndex() + 1,
+                              tmp, localmesh->LocalNx, 666);
+    MPI_Status stat;
+    MPI_Wait(&out, &stat);
```
dschwoerer: Something seems to be off with the MPI calls - during deallocation I get segfaults in
```
# 3  BoutMesh::clear_handles (this=this@entry=0x555555bfa1d0) at boutmesh.cxx:1918
1918        comm_list.pop_front();
```

Is there anything wrong with they way I have used sendToProc/receiveFromProc?
I was further surprised that both calls where non-blocking - for non blocking calls I would have expected Isend and Ireceive. That was also not mentioned in the documentation, as far as I could see ...

dschwoerer: Ignore above comment - segfault seems to be unrelated ....

ZedThree:
The general idea is definitely useful -- being able to get the real space coordinates. I'm not sure about this particular implementation of it though. It's worth thinking carefully about how to do this properly so that it doesn't bite us in the arse later.

dschwoerer:
I am not convinced myself that this is the best approach - however I have not come up with a better idea ...

As long as we only expose the realy to the user, the implementation does not matter, and we can change it any time when something better comes up.
It seems to me however quite important to use expressions on non-uniform meshes ...

Would you be willing to merge this, if I add documentation and clean up a bit?

src/field/fieldgenerators.cxx
```diff
@@ -268,13 +269,148 @@ std::shared_ptr<FieldGenerator> FieldTanhHat::clone(const list<std::shared_ptr<F
...
+                 tanh( s*(X->generate(x,y,z,t,i,localmesh) - (c - 0.5*w)) )
+               - tanh( s*(X->generate(x,y,z,t,i,localmesh) - (c + 0.5*w)) )
              );
 }
+
+BoutReal FieldRealY::generate(double x, double y, double z, double t,
+                              const DataIterator &i, Mesh *localmesh) {
+  BoutReal *offset = doCache(localmesh);
+  int gny = localmesh->GlobalNy - mesh->ystart * 2;
+  int iy = y / 2 / PI * gny * 2 + 10.5;
```
d7919: What's the origin of the 10.5 here?

ZedThree:
I'm not convinced this is the right way to go, even at the interface level. It couples `FieldGenerator` to both `DataIterator` and `Mesh`, and requires more things being passed around. Even if it doesn't tie us to an implementation, it does tie us to an interface.

Another way to do this would be to have a flag to switch between generating the coordinates in the current angular space or in real space. This could be at the global and/or individual variable level.

As `FieldX`, etc. just return the appropriate passed-in value, we would only need to modify `FieldFactory::create3D`, etc., which already know about `Mesh`.

The natural place to put the real space coordinates would be `Coordinates`, which would also make them available elsewhere in the library.

src/field/fieldgenerators.cxx
dschwoerer: 10: a sufficently large, even integer to be always positive. evenness to preserve staggering info
.5: so that the cast rounds appropriately.

dschwoerer:
I guess that would be nicer, but how do we achieve this? 

Adding the needed capabilities to coordinates would be easy.
But then to get x,y,z - should we pass some sort of call back?
An object that can be changed (switching to different coordinates) and can return x,y,z as needed?
realy could then be simple `real(y)` and usage would allow *sin(real(y))` as well as `real(sin(y))`?
That would allow more changes in the future, without further interface changes, as rather than passing x,y,z we pass an object, that can be changed in the future if needed.

Is this something along what you thought?

dschwoerer:
> I guess that would be nicer, but how do we achieve this?
> 
> Adding the needed capabilities to coordinates would be easy.
> But then to get x,y,z - should we pass some sort of call back?
> An object that can be changed (switching to different coordinates) and can return x,y,z as needed?
> realy could then be simple `real(y)` and usage would allow *sin(real(y))`as well as`real(sin(y))`?
> That would allow more changes in the future, without further interface changes, as rather than passing x,y,z we pass an object, that can be changed in the future if needed.
> 
> Is this something along what you thought?

Any thoughts @ZedThree ?

dschwoerer:
Replaced by #1705

PR #1135, opened 2018-06-25
The simple version does not work.

dschwoerer:
Take for example 10 and 1e-4 - obviously they should be multiples.
The first fmod returns 1e-4 - as expected.
The second however returns 0.00099999999999979185 - as 10 is slightly
smaller than 1e5 * 1e-4 - due to the finite accuracy of floating point
numbers.
Even if we do not compare for zero - but a finite small value, the
result will be negative, as 10 was slightly smaller - compared to the
divisor - and thus the result of fmod is in the order of the smaller
number.

### DIFF
--- a/include/bout/monitor.hxx
+++ b/include/bout/monitor.hxx
@@ -3,6 +3,7 @@
 
 #include "bout_types.hxx"
 #include "bout/assert.hxx"
+#include "utils.hxx"
 
 #include <cmath>
 
@@ -14,7 +15,12 @@ class Solver;
 inline bool isMultiple(BoutReal a, BoutReal b) {
   ASSERT2(a > 0);
   ASSERT2(b > 0);
-  return (std::fmod(a, b) == 0) || (std::fmod(b, a) == 0);
+
+  auto min = a>b?b:a;
+  auto max = a>b?a:b;
+  auto ratio = std::round(max/min);
+  auto error = ratio*min - max;
+  return (std::abs(error/max) < 1e-12);
 }
 
 /// Monitor baseclass for the Solver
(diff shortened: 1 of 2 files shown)


bendudson:
I'm unsure about this function: Is it really supposed to give the same result for `isMultiple(2,10)` as `isMultiple(10,2)`? I would expect only one of these to be true.

ZedThree:
I think the name is a bit of a misnomer, and it's really checking if one number is a multiple of the other. As far as I'm aware, it's only used in `Solver` for checking when monitors should be called. A monitor may be called either multiple times per timestep, or every n timesteps. This function attempts to check both conditions simultaneously.

A rename is probably a good idea. It might also be implemented like:

```cpp
bool isCompatibleTimestep(BoutReal a, BoutReal b) {
    return isMultipleOf(a, b) || isMultipleOf(b, a);
}
```
where `isMultipleOf(a, b)` does check if `a` is multiple of `b` but *not* vice-versa.

---

The calling sites are a little confusing too, to be honest:

```cpp
    if (!isMultiple(timestep,TIMESTEP)){
```

- `timestep` is `Solver::timestep` which just has the documentation "timestep - shouldn't be changed after init is called."
- `TIMESTEP` is the second argument to `Solver::solve` which is called `dt` in the header file, `TIMESTEP` if set in the global options section and `output_step` if set in the options section passed to `Solver`

The former could be renamed, or at least the documentation updated to be more explicit about what it is -- the smallest timestep? The output step? I'm not sure.

The latter, a bit trickier -- it would be nice if `TIMESTEP` and `output_step` were named consistently, but not much to be done now I think. Maybe just the variable renamed to `output_step` internally.

---

@dschwoerer Please could you add some cases to `TEST(MonitorTest, IsMultiple)` in `test_monitor.cxx` that don't currently work in next but should work. Unfortunately, FP maths is tricky, so I doubt we'll be able to hit all the edge cases, but what's there is apparently not enough.

PR #1138, opened 2018-07-03
Allows rkgeneric to reset for use with SLEPc

d7919:


src/solver/impls/rkgeneric/rkgeneric.cxx
```diff
@@ -78,6 +79,16 @@ int RKGenericSolver::init(int nout, BoutReal tstep) {
   return 0;
 }
 
+void RKGenericSolver::resetInternalFields(){
+  //Zero out history
+  for(int i=0;i<nlocal;i++){
```
ZedThree: Could this be a range-based loop?

d7919: It could be replaced with

~~~~
tmpState.assign(nlocal, 0); // or tmpState.assign(tmpState.size(),0);
f2.assign(nlocal, 0); // or f2.assign(f2.size(),0);
~~~~

ZedThree: I was just about to suggest `std::fill`, but `assign` with `size` looks better

d7919: Ah oops I thought these were vectors.

PR #1144, opened 2018-07-05
Use Array<BoutReal> in imex-bdf2

ZedThree:
Replaces use of `BoutReal*` with `Array<BoutReal>` in imex-bdf2 solver. This fixes some sign-compare warnings, as well removes all uses of `new` (in imex-bdf2).

Along with #1142 and #1143, this makes the main library `-Wall` clean, excluding SLEPc and MUMPS, but including PETSc and SUNDIALS. We're not quite `-Wextra` clean yet, that's mostly unused-parameters.

### DIFF
--- a/src/solver/impls/imex-bdf2/imex-bdf2.cxx
+++ b/src/solver/impls/imex-bdf2/imex-bdf2.cxx
@@ -16,33 +16,15 @@
 #include "petscsnes.h"
 #include "petscmat.h"
 
-IMEXBDF2::IMEXBDF2(Options *opt) : Solver(opt), u(nullptr) {
+IMEXBDF2::IMEXBDF2(Options *opt) : Solver(opt) {
 
   has_constraints = true; ///< This solver can handle constraints
   
 }
 
 IMEXBDF2::~IMEXBDF2() {
-  if (u != nullptr) {
-    delete[] u;
-    for (int i = 0; i < uV.size(); i++) {
-      delete[] uV[i];
-    }
-    for (int i = 0; i < fV.size(); i++) {
-      delete[] fV[i];
-    }
-
-    delete[] rhs;
-
-    VecDestroy(&snes_f);
-    VecDestroy(&snes_x);
-
-    if (have_constraints)
-      delete[] is_dae;
-
-    if (adaptive)
-      delete[] err;
-  }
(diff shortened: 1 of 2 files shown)


src/solver/impls/imex-bdf2/imex-bdf2.cxx
```diff
@@ -166,37 +148,36 @@ int IMEXBDF2::init(int nout, BoutReal tstep) {
...
+    err = Array<BoutReal>{nlocal};
     OPTION(options, adaptRtol, 1.0e-3); //Target relative error
     OPTION(options, mxstepAdapt, mxstep); //Maximum no. consecutive times we try to reduce timestep
     OPTION(options, scaleCushUp, 1.5);
     OPTION(options, scaleCushDown, 1.0);
   }
 
   // Put starting values into u
-  saveVars(u);
+  saveVars(std::begin(u));
```
bendudson: Presumably at some point saveVars et al could be modified to take an Array as input?

ZedThree:
The PETSc documentation is not exactly forthcoming on this, but looking at
the source, there is a check that it's not null, and also some refcounting,
so I don't think we need to check ourselves.

However, that is relying on an implementation detail of PETSc, and they
don't exactly respect backwards compatibility!

On 5 Jul 2018 6:27 p.m., "Ben Dudson" <notifications@github.com> wrote:

*@bendudson* commented on this pull request.
------------------------------

In src/solver/impls/imex-bdf2/imex-bdf2.cxx
<https://github.com/boutproject/BOUT-dev/pull/1144#discussion_r200430148>:

> -    for (int i = 0; i < fV.size(); i++) {
-      delete[] fV[i];
-    }
-
-    delete[] rhs;
-
-    VecDestroy(&snes_f);
-    VecDestroy(&snes_x);
-
-    if (have_constraints)
-      delete[] is_dae;
-
-    if (adaptive)
-      delete[] err;
-  }
+  VecDestroy(&snes_f);

Should these also be in a conditional so that they are not destroyed unless
created?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/boutproject/BOUT-dev/pull/1144#pullrequestreview-134746381>,
or mute the thread

PR #1145, opened 2018-07-06
Slepc configure update

d7919:
Updates SLEPc configuration treatment to be in line with recent changes to PETSc treatment.


.travis_script.sh
```diff
@@ -70,7 +70,20 @@ then
...
+    echo -e $RED_FG
+    echo "**************************************************"
+    echo "Printing make commands:"
+    echo "**************************************************"
+    echo -e $RESET_FG
+    echo
+    make -n $MAIN_TARGET
+    exit $make_exit
+fi
+
```
ZedThree: I added this to try and debug this particular branch, but it may be more generally useful. Happy to remove it if there are any objections

PR #1150, opened 2018-07-09
More unit tests msg stack

d7919:


tests/unit/sys/test_msg_stack.cxx
```diff
@@ -29,6 +29,61 @@ TEST(MsgStackTest, PopTest) {
...
+  dump = msg_stack.getDump();
+  auto second_dump = "====== Back trace ======\n";
+  EXPECT_EQ(dump, second_dump);
+}
+
+TEST(MsgStackTest, ReallocStorageTest) {
+  MsgStack msg_stack;
+
+  for (int i = 0; i < 20; i++) {
+    msg_stack.push("Message %i", i);
```
ZedThree: Should this `EXPECT_NO_THROW`? Or possibly have a check that there are the correct number of messages in the `msg_stack` at the end?

d7919: There are no throws in `push` so there's no way this could throw at the moment. We could check the number of messages but this is independent of the path through `push` so should probably be a standalone test if we want to test that bit I guess (and is essentially in `PushReturnTest`)

d7919: But yes I see the issue that there is no `EXPECT...` at the moment in this test.

I'd guess the best thing, if a bit ugly would be to check the contents of the back-trace.

d7919: Updated to actually test something now.

PR #1151, opened 2018-07-09
More unit tests expressionparser

d7919:
Increase expression parser unit test coverage

### DIFF
--- a/tests/unit/sys/test_expressionparser.cxx
+++ b/tests/unit/sys/test_expressionparser.cxx
@@ -283,6 +283,41 @@ TEST_F(ExpressionParserTest, AddNullaryFunction) {
   }
 }
 
+TEST_F(ExpressionParserTest, CloneBinaryOp) {
+  FieldBinary goodFieldBinary(nullptr, nullptr, '*');
+  std::list<FieldGeneratorPtr> args;
+  args.push_front(nullptr);
+  args.push_front(nullptr);
+  auto clonedFieldBinary = goodFieldBinary.clone(args);
+  parser.addBinaryOp('&', clonedFieldBinary, 10);
+  auto clonedFieldgen = parser.parseString("2 & (x + 3)");
+  auto actualFieldgen = parser.parseString("2 * (x + 3)");
+
+  for (auto x : x_array) {
+    for (auto y : y_array) {
+      for (auto z : z_array) {
+        for (auto t : t_array) {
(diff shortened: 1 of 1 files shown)


tests/unit/sys/test_expressionparser.cxx
```diff
@@ -213,20 +213,25 @@ TEST_F(ExpressionParserTest, BadNumbers) {
...
 }
 
 TEST_F(ExpressionParserTest, BadExpressions) {
+  parser.addGenerator("increment", std::make_shared<IncrementGenerator>());
+
   EXPECT_THROW(parser.parseString("x = 1"), ParseException);
-  EXPECT_THROW(parser.parseString("sin(x"), ParseException);
-  EXPECT_THROW(parser.parseString("sin[x"), ParseException);
+  EXPECT_THROW(parser.parseString("increment(x"), ParseException);
+  EXPECT_THROW(parser.parseString("increment"), ParseException);
```
d7919: This has also lost the `[` as well as changing `sin`-->`increment`, is that intentional?

tests/unit/sys/test_expressionparser.cxx
```diff
@@ -213,20 +213,25 @@ TEST_F(ExpressionParserTest, BadNumbers) {
...
   EXPECT_THROW(parser.parseString("2]"), ParseException);
   EXPECT_THROW(parser.parseString("4+"), ParseException);
   EXPECT_THROW(parser.parseString("+4"), ParseException);
   EXPECT_THROW(parser.parseString("\n"), ParseException);
   EXPECT_THROW(parser.parseString("(3"), ParseException);
+  EXPECT_THROW(parser.parseString("2-3[4"), ParseException);
```
d7919: Maybe this replaces the above `[` that was removed?

tests/unit/sys/test_expressionparser.cxx
ZedThree: Yes, the `[` was extraneous for hitting that exception (i.e. it was already throwing before parsing the `[`)

tests/unit/sys/test_expressionparser.cxx
ZedThree: Not quite, but it does hit a different error [here](https://github.com/boutproject/BOUT-dev/blob/next/src/sys/expressionparser.cxx#L271)

PR #1158, opened 2018-07-10
Add IndPerp to IndexOffset class

JosephThomasParker:
Expand the IndexOffset class (and the mesh) to allow indexing with IndPerp.

See https://github.com/boutproject/BOUT-dev/pull/1114


src/mesh/mesh.cxx
```diff
@@ -341,6 +341,14 @@ Region<Ind2D> & Mesh::getRegion2D(const std::string &region_name){
    }
    return found->second;
 }
+
+Region<IndPerp> & Mesh::getRegionPerp(const std::string &region_name){
+   auto found = regionMapPerp.find(region_name);
+   if (found == end(regionMapPerp)) {
+     throw BoutException("Couldn't find region %s in regionMap2D", region_name.c_str());
```
d7919: Type `regionMap2D` --> `regionMapPerp`.

src/mesh/mesh.cxx
```diff
@@ -377,9 +392,26 @@ void Mesh::createDefaultRegions(){
...
                                        maxregionblocksize));
 
+  //Perp regions
+  addRegionPerp("RGN_ALL", Region<IndPerp>(0, LocalNx - 1, 0, 0, 0, LocalNz - 1, 1, LocalNz,
+                                       maxregionblocksize));
+  addRegionPerp("RGN_NOBNDRY", Region<IndPerp>(xstart, xend, 0, 0, 0, LocalNz - 1, 1, LocalNz,
+                                           maxregionblocksize));
+  addRegionPerp("RGN_NOX", Region<IndPerp>(xstart, xend, 0, 0, 0, LocalNz - 1, 1, LocalNz,
+                                       maxregionblocksize));
+  addRegionPerp("RGN_NOY", Region<IndPerp>(0, LocalNx - 1, 0, 0, 0, LocalNz - 1, 1, LocalNz,
```
d7919: Does `NOY` make sense for `indPerp` as it only has a single y value?

src/mesh/mesh.cxx
JosephThomasParker: Thanks!

ZedThree:
Should this be merged into `region-offsets2` rather than `next`?

d7919:
Should this PR be into the existing `region-offsets2` PR #1114 to make the additions for perp support more explicit?

src/mesh/mesh.cxx
JosephThomasParker: No, but I think this gives the expected behaviour.  IndPerp has no Y guard cells, so returning the region without guard cells is just itself.

JosephThomasParker:
Sure, I'm happy with that.  Just thought I'd give you the option of keeping them separate :smile:

d7919:
Can just make this PR point towards the other branch rather than replaying these commits directly into the other branch. That way can keep separate whilst checking etc. but then merge all in together when ready

include/bout/indexoffset.hxx
```diff
@@ -58,6 +61,9 @@ struct IndexOffset {
       throw BoutException("Offset in y (%d) would go out of bounds at %d", dy, index.ind);
     }
 #endif
+    if (ny == 1) {
```
ZedThree: Maybe this should really be `ASSERT3(dy <= ny)` to avoid the branch in production code?

Actually, maybe that condition is equivalent to the `CHECK` above (L60)?

d7919: I think similar asserts in the other directions would be useful. That way we simplify production code, giving the compiler the best chance of understanding what we're trying to do.

d7919:
I think there are some templated unit tests in `tests/unit/include/bout/test_region.hxx` that can be extended to also test `IndPerp`.

include/bout/mesh.hxx
```diff
@@ -672,10 +678,32 @@ class Mesh {
...
 
+  /// Converts an IndPerp to an Ind3D using calculation
+  /// Algebra not as simple as in Ind2D case:
+  ///   Ind3D   = (jx * ny + jy )* nz + jz
+  ///   IndPerp =  jx * nz + jz
+  /// Cannot eliminate both jx and jz, so deduce jz in function.
+  ///   Ind3D   = ny * (IndPerp - jz) + jy * nz + jz
+  Ind3D indPerpto3D(const IndPerp &indPerp, int jy = 0){
+    int jz = indPerp.ind % LocalNz;
+    return Ind3D( (indPerp.ind - jz) * LocalNy + jy * LocalNz + jz);
```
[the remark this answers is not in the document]
JosephThomasParker: Thanks - done!

PR #1164, opened 2018-07-11
Replace shared_ptr<?>(new ?()) with make_shared

d7919:
Prefer use of `make_shared` over `new` with shared_ptr constructor.

### DIFF
--- a/include/field_data.hxx
+++ b/include/field_data.hxx
@@ -51,7 +51,7 @@ using std::map;
 #include "boundary_region.hxx"
 #include "parallel_boundary_region.hxx"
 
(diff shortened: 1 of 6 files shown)


d7919:
I also have a commit that replaces `std::shared_ptr<FieldGenerator>` with `FieldGeneratorPtr` everywhere that could be added to this PR but touches >10 files.

ZedThree:
Could you add those changes just for field_factory?

d7919:
I can do it just for the files already changed in the PR if you like (which is the majority of all occurrences).

ZedThree:
Yep, sounds good!

PR #1173, opened 2018-07-17
Use localmesh in boundary

dschwoerer:
Part of the push to not use the global mesh object.


bendudson:
Merged, but looks like all these changes have already been made in `next`.

PR #1174, opened 2018-07-17
Apply BoutReal updates to all flexible fields

dschwoerer:
Should be cheaper the re-interpolating ...


include/flexible.hxx.in.py
```diff
@@ -265,9 +265,15 @@ class Flexible: public FieldData {
 # Template to update a FlexibleField in place
 template_inplace = jinja2.Template("""\
 {% if field == 'BoutReal' %}\
+  // BoutReal does not has a location, so we can aplly the operation
```
johnomotani: typo aplly->apply

Issue #46, opened 2015-03-19
zShift for Python


bendudson:
To shift back into orthogonal (psi, zeta) coordinates, need a Python routine to perform FFTs and apply phase shift.

johnomotani:
Could this also help 3D visualisation? If we pre-process the data back to an orthogonal grid, it might be much easier to set up any 3D visualisation program to plot things without needing to do lots of fancy interpolation, etc.

If this is a common thing to want, it might also be helpful to have an option in BOUT++ to output data in orthogonal coordinates instead of/as well as in field-aligned? I guess it would not be hard to implement, since all the shifting routines exist already.

d7919:
I think this is possibly invalid now (or at least needs updating) -- the fields can now be written in either shifted or unshifted coordinates.

dschwoerer:
With #641 it should be easy to implement - I just wasn't sure what function was required/requested.

johnomotani:
This feature is now provided in `xBOUT` https://github.com/boutproject/xBOUT/pull/105.

PR #1175, opened 2018-07-18
getShifted/getUnshifted methods for BoutOutputs

johnomotani:
Add methods to BoutOutputs class to shift fields orthogonal->field-aligned or the reverse.

Make 'evolvingVariables' a member variable rather than a method.

Add support to DataFile for file-level attributes.

### DIFF
--- a/tools/pylib/boutdata/data.py
+++ b/tools/pylib/boutdata/data.py
@@ -501,6 +501,8 @@ class BoutOutputs(object):
     ----------
     attributes : dict
                  dictionary of attributes from the first DataFile
+    evolvingVariables : list
+                        list of time-evolving variables
 
     Examples
     --------
@@ -558,7 +560,7 @@ class BoutOutputs(object):
         # Available variables
         self.varNames = []
         self.dimensions = {}
-        self.evolvingVariableNames = []
+        self.evolvingVariables = []
 
         with DataFile(latest_file) as f:
             self.attributes = f.attributes()
@@ -579,7 +581,7 @@ class BoutOutputs(object):
                 dimensions = f.dimensions(name)
                 self.dimensions[name] = dimensions
                 if name != "t_array" and "t" in dimensions:
-                    self.evolvingVariableNames.append(name)
+                    self.evolvingVariables.append(name)
 
         # Private variables
         if self._caching:
@@ -603,12 +605,6 @@ class BoutOutputs(object):
 
         """
         return self.varNames
-
-    def evolvingVariables(self):
-        """Return a list of names of time-evolving variables
-
-        """
-        return self.evolvingVariableNames
 
     def redistribute(self, npes, nxpe=None, mxg=2, myg=2, include_restarts=True):
         """Create a new set of dump files for npes processors.
--- a/tools/pylib/boutdata/data.py
+++ b/tools/pylib/boutdata/data.py
@@ -10,6 +10,7 @@ import numpy
 import re
 
 from boutdata.collect import collect, create_cache
+from boutdata.shiftz import shiftz
 from boututils.boutwarnings import alwayswarn
 from boututils.datafile import DataFile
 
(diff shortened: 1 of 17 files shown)


tools/pylib/boutdata/data.py
```diff
@@ -516,6 +516,9 @@ class BoutOutputs(object):
 
     >>> d = BoutOutputs(".", prefix="BOUT.dmp", caching=True) # Turn on caching
 
+
+    :ivar attributes: dictionary of attributes from the first DataFile
+    :ivar evolvingVariables: list of time-evolving variables
```
ZedThree: Please could you put these in [numpydoc][1] style? These can go under an `Attributes` section (see link)

[1]: https://numpydoc.readthedocs.io/en/latest/format.html#class-docstring

tools/pylib/boutdata/data.py
```diff
@@ -554,10 +557,15 @@ def __init__(self, path=".", prefix="BOUT.dmp", suffix=None, caching=False,
...
+        self.evolvingVariables = []
+        self._shiftAngle = None # used for getShifted()/getUnshifted() methods, calculated from self['zShift'] when first needed
 
         with DataFile(latest_file) as f:
-            npes = f.read("NXPE")*f.read("NYPE")
+            self.attributes = f.attributes()
+            try:
+                npes = f.read("NXPE")*f.read("NYPE")
+            except KeyError:
+                npes = self._attributes["NXPE"]*self._attributes["NYPE"]
```
ZedThree: This is `_attributes`, but `attributes` above. Are they supposed to be different variables?
Also, are `NXPE/NYPE` actually file attributes rather than variables?

tools/pylib/boutdata/data.py
```diff
@@ -599,11 +607,47 @@ def keys(self):
...
+    def shiftAngle(self):
+        """
+        The shift angle, calculated from zShift the first time it is used
+        """
+        if self._shiftAngle is None:
+            try:
+                nz = self["MZ"]
+            except KeyError:
+                try:
+                    nz = self.attributes("mz")
```
ZedThree: I think you meant to use square brackets here

tools/pylib/boutdata/data.py
```diff
@@ -599,11 +607,47 @@ def keys(self):
...
+                except KeyError:
+                    nz = self.attributes["mesh:nz"]
```
ZedThree: I'm a bit confused about where this is coming from -- `mesh:nz` looks like an input file variable, but this class reads output files

tools/pylib/boututils/datafile.py
```diff
@@ -319,9 +319,13 @@ def __getitem__(self, name):
     def __setitem__(self, key, value):
         self.impl.__setitem__(key, value)
 
-    def attributes(self, varname):
+    def attributes(self, varname=""):
         """Return a dictionary of attributes
 
+        These are attributes for varname if varname is specified. If varname is
+        the empty string (default case) then the global attributes of the file
+        are returned.
```
ZedThree: Does this actually work? I can't see anything that handles the empty string in the NetCDF/HDF5 implementations of `attributes`. It looks like they both try and find attributes for a variable named `""`, i.e. the empty string. As there likely *isn't* a variable called this, they return `None`. I might be wrong, I don't have a file with file-level attributes to hand!

Also, I'm assuming "global attributes" means "file attributes"?

tools/pylib/boutdata/data.py
```diff
@@ -554,10 +557,15 @@ def __init__(self, path=".", prefix="BOUT.dmp", suffix=None, caching=False,
         # Available variables
         self.varNames = []
         self.dimensions = {}
-        self.evolvingVariableNames = []
+        self.evolvingVariables = []
+        self._shiftAngle = None # used for getShifted()/getUnshifted() methods, calculated from self['zShift'] when first needed
```
ZedThree: Please put long comments like this on separate lines

tools/pylib/boutdata/data.py
```diff
@@ -599,11 +607,47 @@ def keys(self):
...
+                    nz = self.attributes["mesh:nz"]
+            zlength = self["dz"]*nz
+            self._shiftAngle = 2.*numpy.pi*self["zShift"]/zlength
+        return self._shiftAngle
+
+    def getShifted(self, name):
+        """
+        Return a variable shifted by zShift: if BOUT++ output is in
+        'unshifted', i.e. toroidal, coordinates, this function returns a field
+        in field-aligned coordinates
```
ZedThree: This sounds a bit like it checks if the output file is in unshifted space, and always does the "correct" thing, but if I understand correctly, it unconditionally shifts the variable?

Always doing the "correct thing" would be best; if that's not possible currently, please could you change the docstring to say no checking is done (or something along those lines)

johnomotani:
@ZedThree thanks for the comments!

The parts using file attributes to read options I should never have put in here. They were meant to go with another project which is incomplete anyway. Sorry! The support for file attributes was indeed broken, but I've fixed and tested it now. I think it may be useful in future.

To allow getShifted/getUnshifted to do the right thing automatically, I've added another attribute to keep track of whether Field3Ds are in orthogonal or field-aligned coordinates. It required some changes to the BOUT++ code as well, so suggestions on a simpler way are welcome, but what I've added seems useful anyway. Don't know if we're ever likely to handle fields on a mix of orthogonal and field-aligned coordinates, but if we do there's some support for keeping track of it. Possibly coordinate_system should be an enum like CELL_LOC rather than a string?

johnomotani:
This PR would resolve #46.

johnomotani:
This PR should be re-written to use the `coordinate_system` attribute introduced in #1459 when that PR is merged.

johnomotani:
To resolve #46, the Python code in this PR could be modified easily to recognise the `direction_y` attribute instead of `coordinate_system` (to check whether a variable is in field-aligned or toroidal coordinates), some naming updated (e.g. `toFieldAligned` and `fromFieldAligned` instead of `getShifted` and `getUnshifted`) and the C++ changes removed. But since the feature is already provided and tested in `xBOUT`, I'd suggest just closing this PR.

ZedThree:
Ok, closing as xBOUT replaces this functionality

PR #1176, opened 2018-07-18
Shiftedmetric updates

johnomotani:
A collection of small updates:
* fixes to use of transformation to field-aligned coordinates to calculate derivatives for fields that do not have yup/ydown fields
* updates to setting of location of fields, to help with using staggered grids
* add a method that deletes yup/ydown fields (and will be used to delete field-aligned 'field_fa' member when that is added)
* write zShift to the output files to use for post-processing

These updates are in preparation for adding a shifted-metric type class that shifts variables to field-aligned coordinates (which is helpful when also using staggered grids), and for an update to ShiftedMetric to include 2 points up and down the field (if myg>1).

### DIFF
--- a/include/bout/mesh.hxx
+++ b/include/bout/mesh.hxx
@@ -610,12 +610,12 @@ class Mesh {
   typedef BoutReal (*flux_func)(stencil&, stencil &);
 
   /// Transform a field into field-aligned coordinates
-  const Field3D toFieldAligned(const Field3D &f) {
-    return getParallelTransform().toFieldAligned(f);
+  const Field3D toFieldAligned(const Field3D &f, const REGION region = RGN_NOX) {
+    return getParallelTransform().toFieldAligned(f, region);
   }
   /// Convert back into standard form
-  const Field3D fromFieldAligned(const Field3D &f) {
-    return getParallelTransform().fromFieldAligned(f);
+  const Field3D fromFieldAligned(const Field3D &f, const REGION region = RGN_NOX) {
+    return getParallelTransform().fromFieldAligned(f, region);
   }
 
   bool canToFromFieldAligned() {
--- a/include/bout/mesh.hxx
+++ b/include/bout/mesh.hxx
@@ -609,6 +609,12 @@ class Mesh {
   /// Derivative functions of a velocity field, and field stencil v, f
   typedef BoutReal (*flux_func)(stencil&, stencil &);
 
+  /// Calculate yup/ydown fields in case they can be computed for an
+  /// intermediate variable without communicating
+  void calcYUpDown(Field3D &f) {
+    getParallelTransform().calcYUpDown(f);
+  }
+
   /// Transform a field into field-aligned coordinates
   const Field3D toFieldAligned(const Field3D &f, const REGION region = RGN_NOX) {
     return getParallelTransform().toFieldAligned(f, region);
--- a/include/bout/paralleltransform.hxx
+++ b/include/bout/paralleltransform.hxx
@@ -38,11 +38,11 @@ public:
   
   /// Convert a 3D field into field-aligned coordinates
   /// so that the y index is along the magnetic field
-  virtual const Field3D toFieldAligned(const Field3D &f) = 0;
+  virtual const Field3D toFieldAligned(const Field3D &f, const REGION region = RGN_NOX) = 0;
   
   /// Convert back from field-aligned coordinates
   /// into standard form
-  virtual const Field3D fromFieldAligned(const Field3D &f) = 0;
+  virtual const Field3D fromFieldAligned(const Field3D &f, const REGION region = RGN_NOX) = 0;
 
   virtual bool canToFromFieldAligned() = 0;
 };
@@ -65,7 +65,7 @@ public:
    * The field is already aligned in Y, so this
    * does nothing
    */ 
-  const Field3D toFieldAligned(const Field3D &f) override {
+  const Field3D toFieldAligned(const Field3D &f, const REGION UNUSED(region)) override {
     return f;
   }
   
@@ -73,7 +73,7 @@ public:
    * The field is already aligned in Y, so this
    * does nothing
    */
-  const Field3D fromFieldAligned(const Field3D &f) override {
+  const Field3D fromFieldAligned(const Field3D &f, const REGION UNUSED(region)) override {
     return f;
   }
 
@@ -109,13 +109,13 @@ public:
    * in X-Z, and the metric tensor will need to be changed 
    * if X derivatives are used.
    */
-  const Field3D toFieldAligned(const Field3D &f) override;
+  const Field3D toFieldAligned(const Field3D &f, const REGION region=RGN_NOX) override;
 
   /*!
    * Converts a field back to X-Z orthogonal coordinates
    * from field aligned coordinates.
    */
-  const Field3D fromFieldAligned(const Field3D &f) override;
+  const Field3D fromFieldAligned(const Field3D &f, const REGION region=RGN_NOX) override;
 
   bool canToFromFieldAligned() override{
     return true;
@@ -142,7 +142,7 @@ private:
    * Shift a 2D field in Z. 
    * Since 2D fields are constant in Z, this has no effect
    */
-  const Field2D shiftZ(const Field2D &f, const Field2D &UNUSED(zangle)){return f;};
+  const Field2D shiftZ(const Field2D &f, const Field2D &UNUSED(zangle), const REGION UNUSED(region)=RGN_NOX){return f;};
 
   /*!
    * Shift a 3D field \p f in Z by the given \p zangle
@@ -151,7 +151,7 @@ private:
    * @param[in] zangle   Toroidal angle (z)
    *
    */ 
-  const Field3D shiftZ(const Field3D &f, const Field2D &zangle);
+  const Field3D shiftZ(const Field3D &f, const Field2D &zangle, const REGION region=RGN_NOX);
 
   /*!
    * Shift a 3D field \p f by the given phase \p phs in Z
@@ -162,7 +162,7 @@ private:
    * @param[in] f  The field to shift
    * @param[in] phs  The phase to shift by
    */
-  const Field3D shiftZ(const Field3D &f, const arr3Dvec &phs);
+  const Field3D shiftZ(const Field3D &f, const arr3Dvec &phs, const REGION region=RGN_NOX);
 
   /*!
    * Shift a given 1D array, assumed to be in Z, by the given \p zangle
--- a/include/bout/paralleltransform.hxx
+++ b/include/bout/paralleltransform.hxx
@@ -59,7 +59,10 @@ public:
    * Merges the yup and ydown() fields of f, so that
    * f.yup() = f.ydown() = f
    */ 
-  void calcYUpDown(Field3D &f) override {f.mergeYupYdown();}
+  void calcYUpDown(Field3D &f) override {
+    f.mergeYupYdown();
+    f.setHasValidYUpDown(true);
+  }
   
   /*!
    * The field is already aligned in Y, so this
--- a/include/field3d.hxx
+++ b/include/field3d.hxx
@@ -231,7 +231,15 @@ class Field3D : public Field, public FieldData {
   
   /// Check if this field has yup and ydown fields
   bool hasYupYdown() const {
-    return (yup_field != nullptr) && (ydown_field != nullptr);
+    return hasValidYUpDown && (yup_field != nullptr) && (ydown_field != nullptr);
+  }
+
+  /// Set validity state of yup/ydown fields
+  void setHasValidYUpDown(bool set) {
+    if (set) {
+      ASSERT1(yup_field != nullptr && ydown_field != nullptr);
+    }
+    hasValidYUpDown = set;
   }
 
   /// Return reference to yup field
@@ -461,6 +469,9 @@ private:
   CELL_LOC location = CELL_CENTRE; ///< Location of the variable in the cell
   
   Field3D *deriv; ///< Time derivative (may be NULL)
+
+  /// flag to record whether yup_field and ydown_field have been set correctly
+  bool hasValidYUpDown;
 
   /// Pointers to fields containing values along Y
   Field3D *yup_field, *ydown_field;
--- a/include/field3d.hxx
+++ b/include/field3d.hxx
@@ -223,6 +223,11 @@ class Field3D : public Field, public FieldData {
    * Ensure that yup and ydown refer to this field
    */
   void mergeYupYdown();
+
+  /*!
+   * Delete all yup/ydown fields and field_fa
+   */
+  void deleteYupYdown();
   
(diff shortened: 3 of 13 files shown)


include/bout/paralleltransform.hxx
```diff
@@ -65,21 +69,24 @@ public:
...
+  const Field3D fromFieldAligned(const Field3D &f, const REGION UNUSED(region)) override {
     return f;
   }
 
   bool canToFromFieldAligned() override{
     return true;
   }
+
+  /// Write out ParallelTransform variables to file
+  virtual void outputVars(Datafile &UNUSED(file)) {};
```
ZedThree: This can be deleted -- it's already in the base class

include/bout/paralleltransform.hxx
```diff
@@ -183,6 +190,9 @@ private:
    * @param[out] out  A 1D array of length mesh.LocalNz, already allocated
    */
   void shiftZ(const BoutReal *in, const std::vector<dcomplex> &phs, BoutReal *out);
+
+  /// Write out ParallelTransform variables to file
+  void outputVars(Datafile &file);
```
ZedThree: This can be marked `override`

src/field/field3d.cxx
```diff
@@ -198,16 +194,23 @@ void Field3D::mergeYupYdown() {
...
+  yup_field = nullptr;
 
-  if(ydown_field != nullptr) {
+  if (ydown_field != nullptr && ydown_field != this) {
     delete ydown_field;
   }
-
-  yup_field = this;
-  ydown_field = this;
+  ydown_field = nullptr;
```
ZedThree: A fractionally better way of writing this:

```cpp
void Field3D::clearYupYdown() {
  if (yup_field == this && ydown_field == this) {
    return;	
  }

  delete yup_field;
  yup_field = nullptr;

  delete ydown_field;
  ydown_field = nullptr;
}
```

1. Preconditions/assumptions up-front with early exit
2. No guard against deleting `nullptr` (this is fine in C++)

src/mesh/parallel/shiftedmetric.cxx
```diff
@@ -120,30 +120,29 @@ void ShiftedMetric::calcYUpDown(Field3D &f) {
...
 
   Field3D result(&mesh);
   result.allocate();
-  
-  for(int jx=0;jx<mesh.LocalNx;jx++) {
-    for(int jy=0;jy<mesh.LocalNy;jy++) {
-      shiftZ(f(jx,jy), phs[jx][jy], result(jx,jy));
-    }
+
+  for(auto i : f.region2D(region)) {
```
ZedThree: Should be `const auto &i` -- avoids making a copy each iteration

src/mesh/parallel/shiftedmetric.cxx
```diff
@@ -168,18 +167,25 @@ void ShiftedMetric::shiftZ(const BoutReal *in, const std::vector<dcomplex> &phs,
...
-    }
+  invalidateGuards(result); // Won't set x-guard cells, so allow checking to throw exception if they are used.
+
+  // We only use methods in ShiftedMetric to get fields for parallel operations
+  // like interp_to or DDY.
+  // Therefore we don't need x-guard cells, so do not set them.
+  // (Note valgrind complains about corner guard cells if we try to loop over
+  // the whole grid, because zShift is not initialized in the corner guard
+  // cells.)
+  for(auto i : f.region2D(region)) {
```
ZedThree: See above -- `const auto &i`

src/mesh/index_derivs.cxx
```diff
@@ -1190,7 +1183,7 @@ const Field3D Mesh::indexDDX(const Field3D &f, CELL_LOC outloc, DIFF_METHOD meth
       throw BoutException("Cannot use FFT for X derivatives");
   }
 
-  result = applyXdiff(f, func, diffloc, region);
+  Field3D result = applyXdiff(f, func, diffloc, region);
 
   result.setLocation(diffloc); // Set the result location
```
ZedThree: I know you've not touched this, but this looks suspicious. `applyXdiff` should be setting the location to `diffloc` -- do we need to then set the location again immediately after?

include/bout/paralleltransform.hxx
```diff
@@ -6,6 +6,7 @@
 #ifndef __PARALLELTRANSFORM_H__
 #define __PARALLELTRANSFORM_H__
 
+#include <datafile.hxx>
```
ZedThree: As we only need a reference to `Datafile` in this header, you could just forward declare it:

```cpp
class Datafile;
```

This avoids needing to `#include` the header in this header, hopefully speeding up compilation times. It will need to go in `shiftedmetric.cxx` so `ShiftedMetric::outputVars` can use `Datafile::add`

src/mesh/index_derivs.cxx
```diff
@@ -3015,12 +2950,8 @@ const Field3D Mesh::indexFDDY(const Field3D &v, const Field3D &f, CELL_LOC outlo
...
 
-  // There are four cases, corresponding to whether or not f and v
-  // have yup, ydown fields.
-
-  // If vUseUpDown is true, field "v" has distinct yup and ydown fields which
-  // will be used to calculate a derivative along
-  // the magnetic field
+  // If *UseUpDown is true, field "*" has distinct yup and ydown fields which
+  // will be used to calculate a derivative along the magnetic field
   bool vUseUpDown = (v.hasYupYdown() && ((&v.yup() != &v) || (&v.ydown() != &v)));
```
ZedThree: I know you've not touched this, but do we really need to check that `yup/ydown` are distinct from the field? Is `hasYupYdown` not sufficient here?

src/mesh/index_derivs.cxx
johnomotani: I think this is actually hiding a bug at the moment, see last commit of #1061. If/when that is merged, we should probably get rid of the `result.setLocation` after `apply*diff`.

src/mesh/index_derivs.cxx
johnomotani: Without this check I think we wouldn't be able to use 5-point stencils with `paralleltransform=identity`. I haven't checked, but without this check I guess behaviour might even be inconsistent for identity transform, depending on whether a field had been communicated (when it would use the yup/ydown branch) or not (when it would use the to/fromFieldAligned branch).

I guess we could have ParallelTransformIdentity do nothing for calcYUpDown instead of using mergeYupYdown, then hasYupYdown would return false and we'd keep the current behaviour but maybe in a simpler/clearer way? If we did this we could get rid of mergeYupYdown and all the checks for `&f.yup()!=&f`.

src/mesh/parallel/shiftedmetric.cxx
johnomotani: `for (auto i :...` is used in quite a few places in the code. Is it worth getting rid of them? I've made a patch for that in #1182.

johnomotani:
This PR addresses a few things that aren't really related:
- bugfixes for shift-to-field-aligned derivatives
- setting locations in derivatives
- tidying up deletion of yup/ydown fields
- fixing invalidation of guard cells in interp_to
- enhancements to ShiftedMetric:
  - write zShift to output files
  - region arguments for to/fromFieldAligned

I think it's ready to merge, but should I split it up?

include/bout/mesh.hxx
```diff
@@ -609,12 +609,12 @@ class Mesh {
   typedef BoutReal (*flux_func)(stencil&, stencil &);
 
   /// Transform a field into field-aligned coordinates
-  const Field3D toFieldAligned(const Field3D &f) {
-    return getParallelTransform().toFieldAligned(f);
+  const Field3D toFieldAligned(const Field3D &f, const REGION region = RGN_NOX) {
```
d7919: Is there a reason for this choice of default region?

include/field3d.hxx
```diff
@@ -233,6 +233,11 @@ class Field3D : public Field, public FieldData {
    * Ensure that yup and ydown refer to this field
    */
   void mergeYupYdown();
+
+  /*!
+   * Clear all yup/ydown fields and field_fa
```
d7919: What do you mean by clear (delete, zero out, something else)?

src/field/field3d.cxx
```diff
@@ -375,6 +378,7 @@ Field3D & Field3D::operator=(const Field3D &rhs) {
 
   setLocation(rhs.location);
 
+  clearYupYdown();
```
d7919: Would it be better to copy the yup/ydown (as we do for the data) in this case?

src/mesh/index_derivs.cxx
```diff
@@ -649,10 +649,9 @@ const Field2D Mesh::applyXdiff(const Field2D &var, Mesh::deriv_func func,
     return Field2D(0., this);
   }
 
-  CELL_LOC diffloc = var.getLocation();
-
   Field2D result(this);
   result.allocate(); // Make sure data allocated
+  result.setLocation(loc);
```
d7919: This seems to (potentially) be changing the output location -- it's now certain to be at the requested `loc` which sounds sensible, but by default I think `loc` is `CELL_DEFAULT` and I don't think this makes sense in a call to `setLocation`

src/mesh/index_derivs.cxx
```diff
@@ -872,6 +866,7 @@ const Field2D Mesh::applyYdiff(const Field2D &var, Mesh::deriv_func func, CELL_L
 
   Field2D result(this);
   result.allocate(); // Make sure data allocated
+  result.setLocation(diffloc);
```
d7919: This is different to the X functions which now use the passed `loc` rather than `diffloc` -- is there a reason for the difference?

src/mesh/index_derivs.cxx
```diff
@@ -923,10 +916,9 @@ const Field3D Mesh::applyYdiff(const Field3D &var, Mesh::deriv_func func, CELL_L
     return Field3D(0., this);
   }
 
-  CELL_LOC diffloc = var.getLocation();
-
   Field3D result(this);
   result.allocate(); // Make sure data allocated
+  result.setLocation(loc);
```
d7919: This is `diffloc` in earlier y derivative function.

src/mesh/index_derivs.cxx
```diff
@@ -1227,9 +1207,7 @@ const Field3D Mesh::indexDDY(const Field3D &f, CELL_LOC outloc,
       throw BoutException("Cannot use FFT for Y derivatives");
   }
 
-  result = applyYdiff(f, func, diffloc, region);
-
-  result.setLocation(diffloc); // Set the result location
+  Field3D result = applyYdiff(f, func, diffloc, region);
```
d7919: Could just directly return `applyYdiff(f, func, diffloc, region);` I think.

src/mesh/parallel/shiftedmetric.cxx
```diff
@@ -168,18 +169,25 @@ void ShiftedMetric::shiftZ(const BoutReal *in, const std::vector<dcomplex> &phs,
...
     return f; // Shifting makes no difference
 
   Field3D result(&mesh);
   result.allocate();
-
-  for(int jx=0;jx<mesh.LocalNx;jx++) {
-    for(int jy=0;jy<mesh.LocalNy;jy++) {
-      shiftZ(f(jx,jy), mesh.LocalNz, zangle(jx,jy), result(jx,jy));
-    }
+  invalidateGuards(result); // Won't set x-guard cells, so allow checking to throw exception if they are used.
```
d7919: This shouldn't be needed. Creating a field and allocating should give you a field with invalid guards (if check is at the right level). Let me know if this isn't the case and I can look at why/fixing it!

src/mesh/parallel/shiftedmetric.cxx
```diff
@@ -168,18 +169,25 @@ void ShiftedMetric::shiftZ(const BoutReal *in, const std::vector<dcomplex> &phs,
...
+
+  // We only use methods in ShiftedMetric to get fields for parallel operations
+  // like interp_to or DDY.
+  // Therefore we don't need x-guard cells, so do not set them.
```
d7919: I think we might want to set the x-guard cells in some future situations. Is it easy enough to support this even if the default is to not do it?

include/bout/mesh.hxx
johnomotani: I think the reason I did this is that:
- we (usually, probably) want to calculate the result in the y-guard cells, to avoid interpolation and needing to apply parallel boundary conditions before taking derivatives or interpolating
- we don't want to shift in the corner guard cells, because they contain invalid data and I think I was getting exceptions thrown when including them
- we didn't have a region that excludes just the corner guard cells, but includes x- and y-guard cells
- since radial derivatives should probably be done in the orthogonal coordinates rather than the field-aligned ones, the x-guard cells shouldn't normally be needed.

include/field3d.hxx
johnomotani: I mean delete (if they exist). It would make more sense to change the name...

src/field/field3d.cxx
johnomotani: > Would it be better to copy the yup/ydown (as we do for the data) in this case?

I don't know. I guess it only makes sense to copy yup/ydown if we also calculate yup/ydown in arithmetic operators. But if we calculate yup/ydown everywhere possible by default, we'd be doing a lot of computations that might not be needed (if we don't take a y-derivative, etc.). It would be nice to give an option for including yup/ydown fields, but I don't see a nice way to do it...

src/mesh/index_derivs.cxx
johnomotani: Good point. Fixed in 0e1adbe, and similar issues with indexVDD*(Field2D) in c334b8b.

src/mesh/index_derivs.cxx
johnomotani: It was different because staggering was not implemented for applyYdiff(Field2D). I've implemented it now in 2a35d4d.

src/mesh/index_derivs.cxx
johnomotani: Yep, updated in ce76aa2.

src/mesh/parallel/shiftedmetric.cxx
johnomotani: That's right. Sorry, I was trying to be over-cautious.

src/mesh/parallel/shiftedmetric.cxx
johnomotani: I think it's not trivial at the moment, but should be when `Region<T>` goodness is available #1245, if we define a region with everything but the corner guard cells, which is maybe what `RGN_ALL` should be?

johnomotani:
I've pulled out several of the commits, that were independent of the rest, into separate PRs (#1427, #1432, #1433). What's left is an attempt to delete, or mark as invalid, `yup_field` and `ydown_field` members of `Field3D` when the field is changed by assignment, etc. I think this code now won't go in as-is because of clashes with other changes, e.g. in #1345, but maybe still useful to refer back to if we decide to do something similar...

PR #1177, opened 2018-07-18
New shifted metric

johnomotani:
Add a ParallelTransform class that shifts variables to a field-aligned coordinate system. This means all shifted variables share the same coordinate system even when they are on staggered grids, so this method is compatible with staggered grids while the existing ShiftedMetric class is not.

Depends on updates from #1060 (hence PR based on #1060) and also #1176 (the number of files touched by this PR should go down quite a bit if #1176 is merged into the base).


johnomotani:
The `ShiftToFieldAligned` `ParallelTransform` no longer uses `Flexible<>`. I've changed the base to #1176 because it still requires the bugfixes from there.

johnomotani:
Being replaced by #1460.

PR #1178, opened 2018-07-18
ShiftedMetric with 2 yup/ydown points

johnomotani:
Add support for 2 yup/ydown points to ShiftedMetric, enabled when MYG>1. This allows higher-order derivative methods (4th order centred differences or 2nd/3rd order upwind schemes) to be used.

Also has some support for staggered grids. The original intention was to enable 4th order interpolation to be compatible with staggered grids, but this is not enough. To support staggered grids, ShiftedMetric would have to calculate interpolated values half a grid cell along the field from all the centred and staggered points. This does not seem practical, which motivated writing the ShiftToFieldAligned class in #1177.

I don't think this is actually useful to merge, since it adds quite a lot of complexity to ShiftedMetric but still does not support staggered grids. Uploading in case anyone might find it useful in future.

### DIFF
--- a/include/bout/mesh.hxx
+++ b/include/bout/mesh.hxx
@@ -614,12 +614,12 @@ class Mesh {
   typedef BoutReal (*flux_func)(stencil&, stencil &);
 
   /// Transform a field into field-aligned coordinates
-  const Field3D toFieldAligned(const Field3D &f) {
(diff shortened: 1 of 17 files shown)


ZedThree:
Replaced with #1345 

Leaving the branch for future reference -- @johnomotani feel free to delete if you don't think it's needed

PR #1182, opened 2018-07-24
Change 'auto i' to 'const auto &i' in for loops

johnomotani:
Avoids making a copy each iteration.


ZedThree:
I was a little worried at first that the `const` on the loops over `Field3D`s didn't make sense, but that applies to the *dereferenced* iterator and not the iterator itself. In our case, dereferencing a `DataIterator` returns a reference to `this` and the `const` is on that reference.

This did make me think that maybe `Solver::loop_vars_op` should take `ind2d` by `const&`, similarly `ArkodeSolver::loop_abstol_values_op`.

PR #1194, opened 2018-07-31
Block region describe

d7919:
Adds some reporting for various Region properties

### DIFF
--- a/include/bout/region.hxx
+++ b/include/bout/region.hxx
@@ -538,7 +538,11 @@ public:
 
     // Count the number of small blocks, defined as blocks less than smallSizeFrac of maxBlockSize
     const BoutReal smallSizeFrac = 0.5;
-    result.numSmallBlocks = std::count_if(std::begin(blockSizes), std::end(blockSizes), [&blockSizes, &result, smallSizeFrac](int theSize) { return theSize < smallSizeFrac * result.maxBlockSize;}) ;
+    result.numSmallBlocks =
+      std::count_if(std::begin(blockSizes), std::end(blockSizes),
+		    [&blockSizes, &result, smallSizeFrac](int theSize) {
+		      return theSize < smallSizeFrac * result.maxBlockSize;
+		    });
 
     return result;
   }
--- a/include/bout/region.hxx
+++ b/include/bout/region.hxx
@@ -496,7 +496,7 @@ public:
   }
   
   /// Returns a RegionStats struct desribing the region
-  RegionStats getStats() {
+  RegionStats getStats() const {
     RegionStats result;
 
     result.numBlocks = blocks.size();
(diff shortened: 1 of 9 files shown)


include/bout/region.hxx
```diff
@@ -485,6 +512,37 @@ public:
...
+
+    result.maxBlockSize = *(minMaxSize.second); //Note have to derefence to get actual value
+    result.numMaxBlocks = std::count(std::begin(blockSizes), std::end(blockSizes), result.maxBlockSize);
+
+    
+    result.maxImbalance = static_cast<BoutReal>(result.maxBlockSize)/static_cast<BoutReal>(result.minBlockSize);
+
+    // Count the number of small blocks, defined as blocks less than smallSizeFrac of maxBlockSize
+    const BoutReal smallSizeFrac = 0.5;
+    result.numSmallBlocks = std::count_if(std::begin(blockSizes), std::end(blockSizes), [&blockSizes, &result, smallSizeFrac](int theSize) { return theSize < smallSizeFrac * result.maxBlockSize;}) ;
```
ZedThree: Please could you clang-format this long line?

include/bout/region.hxx
```diff
@@ -230,6 +230,33 @@ inline Ind2D operator+(int n, Ind2D rhs) { return rhs += n; }
 inline Ind2D operator-(Ind2D lhs, int n) { return lhs -= n; }
 inline Ind2D operator-(Ind2D lhs, const Ind2D &rhs) { return lhs -= rhs; }
 
+/// Structure to hold various derived "statistics" from a particular region
+struct RegionStats {
+  int numBlocks = 0;           // How many blocks
```
ZedThree: These comments could all be doxygen comments

ZedThree:
I think we were talking about this before, but it might be good if we had an `output_verbose` that came on when programs are run at higher verboseness levels. At the minute, we just have `output_debug` that might not be there at all. It would be for another PR, at any rate.

d7919:
I've used `output_info` which is only on at the most verbose (which is still the default level but one higher than required for progress output).

include/bout/region.hxx
d7919: Does that mean `///` and on the line before the variable?

ZedThree: That's one way, the other way is `///<` on the same line

include/bout/region.hxx
d7919: I ran `git-clang-format-4.0` on this (after adding a trailing space so I had something to stage) and it said there was nothing to modify.

d7919: I've manually split it as shown. Not sure why it wasn't being picked up automatically.

PR #1196, opened 2018-07-31
Speedup travis by removing jobs

d7919:
Merges python and shared build jobs together.

Merges gcc-7 and debug builds together.

Removes clang non-debug build.

Moves coverage builds into separate stage.

Only runs coverage and finalise coverage stages when branch is master or next.
Can also force coverage to be built for any branch by including "build coverage" anywhere in the commit message.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,37 +70,35 @@ matrix:
 
 #COVERAGE
   - stage: Coverage
-    matrix:
-      include:
-      - addons:
-          apt:
-            packages:
-              - *standard_packages
-              - libpetsc3.4.2-dev
-              - libslepc3.4.2-dev
-              - liblapack-dev
-              - libparpack2-dev
-              - jq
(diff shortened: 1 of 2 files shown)


d7919:
A non-protected branch push now uses around 70-80 minutes of build time rather than ~120 minutes. A PR/merge to next/master (or commit with `build coverage` in commit message) should use around 90-100 minutes. 

Scope for merging "standard" build (i.e. with no explicit configure options) with the python/shared job as only difference in later adds `-fpic` and also runs the python tests. This would save another 10 minutes.

ZedThree:
I think the only plausible difference between the "standard" and python/shared jobs is some symbols not being exported in the shared library, but these would be caught when the static library is linked; so it should be fine to merge them.

d7919:
I guess we actually get the static only approach from the openmp branch as well.

PR #1197, opened 2018-07-31
Cython linking

dschwoerer:
The first commit is required to make openmp work with boutcore

I am not sure the second commit is required right now, but setuptools likes to use the default linker, even if a different compiler is used. That can cause often issues - I stumbled across this issue when trying to pip install netcdf on marconi, where `CC` is set to `icc`, but `LDSHARED` was not set.


bendudson:
Is this still needed? Has anyone else tried this?

dschwoerer:
Yes, it is needed.

If preferred, I could only keep the first commit, but as it can be rather painful, I would like to keep the the comments, if the linking isn't working out of the box.

PR #1198, opened 2018-07-31
Coordinates location

johnomotani:
This PR allows Coordinates objects to be created at different locations. Follows the discussion in #1060.

The (default) CELL_CENTRE version is created from the input geometry, other locations are created by interpolating from CELL_CENTRE. The Coordinates objects are retrieved using `mesh->coordinates(location)` and derivatives/differential operators have been updated to use this, so staggered derivatives should now use metric coefficients and dx/dy at the correct locations.

Changes to add location to Field2D and implement interp_to for Field2D are cherry-picked from #1060.

### DIFF
--- a/src/invert/laplace/impls/cyclic/cyclic_laplace.cxx
+++ b/src/invert/laplace/impls/cyclic/cyclic_laplace.cxx
@@ -45,6 +45,10 @@
 
 LaplaceCyclic::LaplaceCyclic(Options *opt, const CELL_LOC loc)
     : Laplacian(opt, loc), Acoef(0.0), Ccoef(1.0), Dcoef(1.0) {
+  Acoef.setLocation(location);
+  Ccoef.setLocation(location);
+  Dcoef.setLocation(location);
+
   // Get options
 
   OPTION(opt, dst, false);
--- a/src/invert/laplace/impls/cyclic/cyclic_laplace.cxx
+++ b/src/invert/laplace/impls/cyclic/cyclic_laplace.cxx
@@ -244,6 +244,9 @@ const FieldPerp LaplaceCyclic::solve(const FieldPerp &rhs, const FieldPerp &x0) 
 const Field3D LaplaceCyclic::solve(const Field3D &rhs, const Field3D &x0) {
   TRACE("LaplaceCyclic::solve(Field3D, Field3D)");
 
+  ASSERT1(rhs.getLocation() == location);
+  ASSERT1(x0.getLocation() == location);
+
   Timer timer("invert");
 
   Mesh *mesh = rhs.getMesh();
--- a/src/mesh/coordinates.cxx
+++ b/src/mesh/coordinates.cxx
@@ -791,7 +791,7 @@ const Field3D Coordinates::Delp2(const Field3D &f, CELL_LOC outloc) {
       for (int jx = localmesh->xstart; jx <= localmesh->xend; jx++) {
         // Perform x derivative
 
-        laplace_tridag_coefs(jx, jy, jz, a, b, c);
+        laplace_tridag_coefs(jx, jy, jz, a, b, c, nullptr, nullptr, outloc);
 
         delft(jx, jz) = a * ft(jx - 1, jz) + b * ft(jx, jz) + c * ft(jx + 1, jz);
       }
--- a/include/invert_laplace.hxx
+++ b/include/invert_laplace.hxx
@@ -184,7 +184,8 @@ public:
 
   /// Coefficients in tridiagonal inversion
   void tridagCoefs(int jx, int jy, int jz, dcomplex &a, dcomplex &b, dcomplex &c,
-                   const Field2D *ccoef = nullptr, const Field2D *d = nullptr);
+                   const Field2D *ccoef = nullptr, const Field2D *d = nullptr,
+                   CELL_LOC loc = CELL_DEFAULT);
 
   /*!
    * Create a new Laplacian solver
@@ -212,7 +213,8 @@ protected:
   int outer_boundary_flags; ///< Flags to set outer boundary condition
 
   void tridagCoefs(int jx, int jy, BoutReal kwave, dcomplex &a, dcomplex &b, dcomplex &c,
-                   const Field2D *ccoef = nullptr, const Field2D *d = nullptr);
+                   const Field2D *ccoef = nullptr, const Field2D *d = nullptr,
+                   CELL_LOC loc = CELL_DEFAULT);
 
   void tridagMatrix(dcomplex **avec, dcomplex **bvec, dcomplex **cvec, dcomplex **bk,
                     int jy, int flags, int inner_boundary_flags, int outer_boundary_flags,
@@ -236,7 +238,8 @@ private:
 // These will be removed at some point
 
 void laplace_tridag_coefs(int jx, int jy, int jz, dcomplex &a, dcomplex &b, dcomplex &c,
-                          const Field2D *ccoef = nullptr, const Field2D *d = nullptr);
+                          const Field2D *ccoef = nullptr, const Field2D *d = nullptr,
+                          CELL_LOC loc = CELL_DEFAULT);
 
 int invert_laplace(const FieldPerp &b, FieldPerp &x, int flags, const Field2D *a,
                    const Field2D *c = nullptr, const Field2D *d = nullptr);
--- a/include/invert_laplace.hxx
+++ b/include/invert_laplace.hxx
@@ -111,35 +111,63 @@ public:
   /// Set coefficients for inversion. Re-builds matrices if necessary
   virtual void setCoefA(const Field2D &val) = 0;
   virtual void setCoefA(const Field3D &val) { setCoefA(DC(val)); }
-  virtual void setCoefA(BoutReal r) { Field2D f(r); setCoefA(f); }
+  virtual void setCoefA(BoutReal r) {
+    Field2D f(r);
+    f.setLocation(location);
+    setCoefA(f);
+  }
   
   virtual void setCoefC(const Field2D &val) = 0;
   virtual void setCoefC(const Field3D &val) { setCoefC(DC(val)); }
-  virtual void setCoefC(BoutReal r) { Field2D f(r); setCoefC(f); }
+  virtual void setCoefC(BoutReal r) {
+    Field2D f(r);
+    f.setLocation(location);
+    setCoefC(f);
+  }
   
   virtual void setCoefC1(const Field2D &UNUSED(val)) {
     throw BoutException("setCoefC1 is not implemented for this Laplacian solver");
   }
   virtual void setCoefC1(const Field3D &val) { setCoefC1(DC(val)); }
-  virtual void setCoefC1(BoutReal r) { Field2D f(r); setCoefC1(f); }
+  virtual void setCoefC1(BoutReal r) {
+    Field2D f(r);
+    f.setLocation(location);
+    setCoefC1(f);
+  }
   
   virtual void setCoefC2(const Field2D &UNUSED(val)) {
     throw BoutException("setCoefC2 is not implemented for this Laplacian solver");
   }
   virtual void setCoefC2(const Field3D &val) { setCoefC2(DC(val)); }
-  virtual void setCoefC2(BoutReal r) { Field2D f(r); setCoefC2(f); }
+  virtual void setCoefC2(BoutReal r) {
+    Field2D f(r);
+    f.setLocation(location);
+    setCoefC2(f);
+  }
   
   virtual void setCoefD(const Field2D &val) = 0;
   virtual void setCoefD(const Field3D &val) { setCoefD(DC(val)); }
-  virtual void setCoefD(BoutReal r) { Field2D f(r); setCoefD(f); }
+  virtual void setCoefD(BoutReal r) {
+    Field2D f(r);
+    f.setLocation(location);
+    setCoefD(f);
+  }
   
   virtual void setCoefEx(const Field2D &val) = 0;
   virtual void setCoefEx(const Field3D &val) { setCoefEx(DC(val)); }
-  virtual void setCoefEx(BoutReal r) { Field2D f(r); setCoefEx(f); }
+  virtual void setCoefEx(BoutReal r) {
+    Field2D f(r);
+    f.setLocation(location);
+    setCoefEx(f);
+  }
   
   virtual void setCoefEz(const Field2D &val) = 0;
   virtual void setCoefEz(const Field3D &val) { setCoefEz(DC(val)); }
-  virtual void setCoefEz(BoutReal r) { Field2D f(r); setCoefEz(f); }
+  virtual void setCoefEz(BoutReal r) {
+    Field2D f(r);
+    f.setLocation(location);
+    setCoefEz(f);
+  }
   
   virtual void setFlags(int f);
   virtual void setGlobalFlags(int f) { global_flags = f; }
--- a/src/invert/laplace/impls/cyclic/cyclic_laplace.hxx
+++ b/src/invert/laplace/impls/cyclic/cyclic_laplace.hxx
@@ -48,11 +48,20 @@ public:
   ~LaplaceCyclic();
   
   using Laplacian::setCoefA;
-  void setCoefA(const Field2D &val) override { Acoef = val; }
+  void setCoefA(const Field2D &val) override {
+    ASSERT1(val.getLocation() == location);
+    Acoef = val;
+  }
   using Laplacian::setCoefC;
-  void setCoefC(const Field2D &val) override { Ccoef = val; }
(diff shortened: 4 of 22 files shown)


include/bout/mesh.hxx
```diff
@@ -427,15 +428,43 @@ class Mesh {
   bool IncIntShear; ///< Include integrated shear (if shifting X)
 
   /// Coordinate system
-  Coordinates *coordinates() {
-    if (coords) { // True branch most common, returns immediately
-      return coords;
+  Coordinates *coordinates(const CELL_LOC location = CELL_DEFAULT) {
+    switch (location) { // True branch most common, returns immediately
+    case CELL_CENTRE: if (coords_centre) return coords_centre;
+    case CELL_XLOW: if (coords_xlow) return coords_xlow;
```
d7919: I think this could lead to bugs -- say the user requests XLOW but it is `nullptr` we won't hit the `return coords_xlow` but instead drop through to the next case, which means if `coords_ylow` is not `nullptr`, for example, then we'll end up returning that instead without the calling code having any means to tell.

include/bout/mesh.hxx
```diff
@@ -691,7 +720,7 @@ class Mesh {
   
   GridDataSource *source; ///< Source for grid data
   
-  Coordinates *coords;    ///< Coordinate system. Initialised to Null
+  Coordinates *coords_centre, *coords_xlow, *coords_ylow, *coords_zlow;    ///< Coordinate system. Initialised to Null
```
d7919: Will we ever want other locations? If so we might want to use a `map` that takes the location enum to return a coordinates instance, that should then be easier to extend without having to make changes in multiple places. I'm not sure if it would impact performance.

include/bout/mesh.hxx
```diff
@@ -730,7 +759,7 @@ class Mesh {
 
 private:
   /// Allocates a default Coordinates object
-  Coordinates *createDefaultCoordinates();
+  Coordinates *createDefaultCoordinates(const CELL_LOC location);
```
d7919: Is it worth setting a default location so existing calls still work and do the same thing?

include/difops.hxx
```diff
@@ -212,12 +212,12 @@ const Field2D Div_par_CtoL(const Field2D &var);
  * @param[in] kY  The diffusion coefficient
  * @param[in] f   The field whose gradient drives a flux
  */
-const Field2D Div_par_K_Grad_par(BoutReal kY, const Field2D &f);
-const Field3D Div_par_K_Grad_par(BoutReal kY, const Field3D &f);
-const Field2D Div_par_K_Grad_par(const Field2D &kY, const Field2D &f);
-const Field3D Div_par_K_Grad_par(const Field2D &kY, const Field3D &f);
-const Field3D Div_par_K_Grad_par(const Field3D &kY, const Field2D &f);
-const Field3D Div_par_K_Grad_par(const Field3D &kY, const Field3D &f);
+const Field2D Div_par_K_Grad_par(BoutReal kY, const Field2D &f, CELL_LOC outloc=CELL_DEFAULT);
```
d7919: I may be misunderstanding but I think these additions seem to be unrelated to the main purpose of this PR.

src/field/field2d.cxx
```diff
@@ -174,6 +175,32 @@ const IndexRange Field2D::region(REGION rgn) const {
   };
 }
 
+void Field2D::setLocation(CELL_LOC new_location) {
+  if (getMesh()->StaggerGrids) {
+    if (new_location == CELL_VSHIFT) {
+      throw BoutException(
+          "Field3D: CELL_VSHIFT cell location only makes sense for vectors");
```
d7919: Copy paste error 3D-->2D

include/bout/mesh.hxx
johnomotani: :+1: I think a `map` is a good idea

include/bout/mesh.hxx
johnomotani: will get rid of this by changing to use std::map

include/difops.hxx
johnomotani: agreed. Will split this PR: first bit to allow coordinates at different locations, second bit to update location handling of metric terms etc. in all the difops.

include/bout/mesh.hxx
johnomotani: It's a private method only used by Mesh::coordinates() so I don't think there should be any other existing calls.

johnomotani:
I've split up the PR now. This now contains just the changes to allow `Mesh::coordinates()` to take a location argument. Interpolating the metrics, etc. to different locations requires `Field2D` to allow non-`CELL_CENTRE` locations (first 4 commits) and `interp_to(Field2D)` (next 3 commits) to be implemented.

src/mesh/coordinates.cxx
```diff
@@ -170,6 +170,120 @@ Coordinates::Coordinates(Mesh *mesh)
...
+
+  // Copy nearest value into boundaries so that differential geometry terms can
+  // be interpolated if necessary
+  // Note: cannot use applyBoundary("neumann") here because applyBoundary()
+  // would try to create a new Coordinates object since we have not finished
+  // initializing yet, leading to an infinite recursion
+  for (auto bndry : localmesh->getBoundaries()) {
+    if (bndry->bx != 0) {
+      for(bndry->first(); !bndry->isDone(); bndry->next1d()) {
+        for (int i=0; i<localmesh->xstart; i++)
```
ZedThree: This just sets values at one end, do you need to set them at the other end too?

src/mesh/mesh.cxx
```diff
@@ -40,8 +40,8 @@ Mesh::~Mesh() {
     delete source;
   }
 
-  if (coords) {
-    delete coords;
+  for (const auto &it : coords_map) {
```
ZedThree: This shouldn't be `const`, as `it/coords_map` are being modified

src/mesh/coordinates.cxx
johnomotani: No, the other end is a different `bndry`, and bndry->bx/bndry-by change sign so that the index goes in the opposite direction. Maybe I should add some comments...

include/bout/mesh.hxx
```diff
@@ -691,7 +694,7 @@ class Mesh {
   
   GridDataSource *source; ///< Source for grid data
   
-  Coordinates *coords;    ///< Coordinate system. Initialised to Null
+  std::map<CELL_LOC, Coordinates*> coords_map; ///< Coordinate systems at different CELL_LOCs
```
d7919: Should we use a smart pointer here rather than a raw one?

src/mesh/interpolation.cxx
```diff
@@ -68,8 +68,11 @@ const Field3D interp_to(const Field3D &var, CELL_LOC loc, REGION region) {
     if (region != RGN_NOBNDRY) {
       // result is requested in some boundary region(s)
       result = var; // NOTE: This is just for boundaries. FIX!
+      result.allocate();
+    } else {
+      result.allocate();
+      invalidateGuards(result);
```
d7919: I don't think this `invalidateGuards` call is needed as it should be done in `allocate` if `result` doesn't already have a data block.

src/mesh/coordinates.cxx
```diff
@@ -170,6 +170,129 @@ Coordinates::Coordinates(Mesh *mesh)
...
+Field2D interpolate_and_Neumann(const Field2D &f, CELL_LOC location, bool diagonal) {
+  Mesh* localmesh = f.getMesh();
+  Field2D result = interp_to(f, location, RGN_NOBNDRY);
+  localmesh->communicate(result);
+
+  // Copy nearest value into boundaries so that differential geometry terms can
+  // be interpolated if necessary
+  // Note: cannot use applyBoundary("neumann") here because applyBoundary()
+  // would try to create a new Coordinates object since we have not finished
+  // initializing yet, leading to an infinite recursion
```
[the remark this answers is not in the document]
johnomotani: After a quick google, this seems like a good place to use an 'anonymous namespace'?
https://stackoverflow.com/questions/154469/unnamed-anonymous-namespaces-vs-static-functions
I've renamed the function interpolateAndNeumann and put it an an anonymous namespace.

src/mesh/coordinates.cxx
```diff
@@ -170,6 +170,129 @@ Coordinates::Coordinates(Mesh *mesh)
...
+      }
+    }
+  }
+
+  // Set corner guard cells to avoid exceptions
+  // Set diagonal elements to a large value, and off-diagonal to zero
+  if (diagonal) {
+    for (int i=0; i<localmesh->xstart; i++) {
+      for (int j=0; j<localmesh->ystart; j++) {
+        result(i, j) = 1.e4;
```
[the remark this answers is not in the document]
johnomotani: I think I had issues at the time when checkData was defaulting to RGN_ALL. Seems not to be a problem now, so I've set all the corner guard cells to BoutNaN.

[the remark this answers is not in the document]
johnomotani: Although now the code does call `invert3x3` with BoutNaNs as input, in a loop over all points
https://github.com/boutproject/BOUT-dev/blob/4547f8c1b264bf9820bec1c30c9fdc9a846193b1/src/mesh/coordinates.cxx#L553-L557
This 'works' because the condition for the error check in `invert3x3` will evaluate to `false` with NaNs (I guess)
https://github.com/boutproject/BOUT-dev/blob/4547f8c1b264bf9820bec1c30c9fdc9a846193b1/include/utils.hxx#L230-L236
but this seems a bit dirty... Would probably be better to loop over just grid cells and non-corner guard cells. But maybe that can wait until we have new `REGION` goodness #1245.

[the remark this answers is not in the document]
johnomotani: ...I think it turns out I didn't test this properly until I merged it into `set_coordinates_location` where the non-`CELL_CENTRE` coordinates actually get used: with NaN in the corner guard cells, the interpolated metrics fail the check that they are finite
https://github.com/boutproject/BOUT-dev/blob/99edc6ed68906ed0d7ee0164f0947a93db1d0d6e/src/mesh/coordinates.cxx#L249-L252
I think the easiest way to get around this is to use a region argument of `finite()` to skip the guard cells.

d7919:
Just a note that #1200 (also approved) is into this branch so it matters what order the PRs are merged -- probably want to do #1200 first and then this one. This will mean that this PR needs re-approving once #1200 is merged.

include/bout/operators_di.hxx
```diff
@@ -17,7 +17,7 @@
 /// @param i  The point where the result is calculated
 ///
 inline BoutReal DDX_C2(const Field3D &f, const DataIterator &i) {
-  return (f[i.xp()] - f[i.xm()])/(2.*mesh->coordinates()->dx[i]);
+  return (f[i.xp()] - f[i.xm()])/(2.*mesh->coordinates(f.getLocation())->dx[i]);
```
d7919: I know this might be giving a bit too much authority to Field?s but could we have a `Field?D.getCoordinates()` method that would return `fieldmesh->coordinates(location)`? This might make statements like this clearer as we'd replace `mesh->coordinates(f.getlocation())->dx[i]` with `f.getCoordinates()->dx[i]`. A further "bonus" is that it would mean we'd be using the field's mesh everywhere like this rather than the global one (i.e. strictly `mesh->coordinates(f.getlocation())->dx[i]`  should really be `f.getMesh()->coordinates(f.getlocation())->dx[i]`

bendudson: Yes, agree this would be good. It is adding a bit to Field, but would make the code more readable.

bendudson: Also in this case some way is needed to make this lookup faster; in this operator the coordinates are fetched for every point in the inner loop.

johnomotani: If the Field*D stored a coordinates pointer initialized set to `fieldmesh->coordinates(location)` instead of calling through fieldmesh, `f.getCoordinates()` could avoid the lookup. I guess it could then be inlined by the compiler as well?

bendudson: Good idea; it certainly should be a lot faster. Would just need to ensure that the coordinates pointer was updated when location was changed.

d7919: I'm happy to have a stab at implementing this unless anybody else has started?

d7919: Slightly more complicated than first thought -- has knock-on effects for tests etc. as now all staggering tests require a valid mesh/coordinates to exist which seems to be causing problems. Easy to get into an infinite loop with obvious design (creating coordinates triggers fields created which then tries to create coordinates etc.) but should be possible to address with a reasonable interface.

d7919: Design question : Should `getCoordinates()` will return the coordinates pointer at the correct location, should we also provide `getCoordinates(CELL_LOC loc)` that just returns `getMesh()->coordinates(loc)` as a convenience function? It would tidy some bits of code but might not be what we want.

d7919:
In boundary_standard - `BoundaryDivCurl::apply(Vector3D)` we get the location of the vector and use this to create the coordinates object. Vectors can have the CELL_VSHIFT location which buts each field at a different location so it's not clear what the correct location to use in constructing the coordinates is here. I would guess this should throw somewhere (maybe in the coordinates constructor for invalid locations, or locations not in a whitelist)?

include/bout/operators_di.hxx
d7919: #1266 should deal with most of these changes.

d7919:
Should we make `FieldPerp` able to be decentred as well?

src/invert/laplace/impls/cyclic/cyclic_laplace.cxx
```diff
@@ -90,7 +90,7 @@ const FieldPerp LaplaceCyclic::solve(const FieldPerp &rhs, const FieldPerp &x0)
   FieldPerp x(mesh); // Result
   x.allocate();
 
-  Coordinates *coord = mesh->coordinates();
+  Coordinates *coord = mesh->coordinates(location);
```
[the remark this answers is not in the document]
d7919: One thing I noticed is that we don't force/check that the coefficients are set at a consistent location so I think we might need a bit more work to make this a bit more robust?

Are you aware of any implementations that do cache location sensitive values in the constructor at the moment?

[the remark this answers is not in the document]
d7919: If FieldPerp had a variable location I guess this would help in these efforts -- can get location from passed fields and if needed check this is consistent with any precalculated data.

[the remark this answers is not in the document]
d7919: (I found `grep -i ">coordinates(" $(find ./{src,include} -name '*.?xx')` quite useful in finding any uses of coordinates, from which it's easy to pick out the few places where a location is not passed).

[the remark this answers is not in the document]
johnomotani: >Are you aware of any implementations that do cache location sensitive values in the constructor at the moment?

No, not at the moment.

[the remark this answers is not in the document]
johnomotani: > One thing I noticed is that we don't force/check that the coefficients are set at a consistent location so I think we might need a bit more work to make this a bit more robust?

That might be a reason for requiring the location in the constructor? If we know the location is set then we can check the location of the coefficients when they are passed in. Without having a location fixed from the constructor I can't think how we could guarantee a consistent location between calls to `setCoef*()` and `solve()`?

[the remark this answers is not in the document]
johnomotani: To remove dependence on the global `mesh`, should we (optionally) pass a mesh to the Laplacian constructor as well, then check that coefficients and rhs use the same mesh as the Laplacian?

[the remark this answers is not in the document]
johnomotani: > Without having a location fixed from the constructor I can't think how we could guarantee a consistent location between calls to setCoef*() and solve()?

I guess one possibility would be to set the location of the Laplacian object from the first coefficient or rhs passed in, and then check the others against that. I'm not in favour of that solution because it adds more code to go wrong (have to keep track of whether location has been set). I'd favour keeping the simple, explicit solution of setting the location of the Laplacian object when it's created as it is now.

johnomotani:
This PR was mostly merged with #1271. The remaining commits are adding location checking to Laplacian solvers, and getting the `laplace_tridag_coefs` at the correct location in `Delp2`. I think these are still worth merging.

d7919:
> This PR was mostly merged with #1271. The remaining commits are adding location checking to Laplacian solvers, and getting the `laplace_tridag_coefs` at the correct location in `Delp2`. I think these are still worth merging.

Yes, sorry I branched off just before these extra commits and forgot to merge them into the PR branch.

PR #1199, opened 2018-08-01
Don't use global mesh in BoundaryRegion

johnomotani:
Using the global mesh pointer can cause problems if there are multiple meshes, e.g. in the tests/MMS/derivatives3 test.

### DIFF
--- a/include/boundary_region.hxx
+++ b/include/boundary_region.hxx
@@ -21,9 +21,10 @@ enum BndryLoc {BNDRY_XIN=1,
 class BoundaryRegionBase {
 public:
   BoundaryRegionBase() = delete;
-  BoundaryRegionBase(Mesh* mesh) : localmesh(mesh) {}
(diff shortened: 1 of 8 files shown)


include/boundary_region.hxx
```diff
@@ -18,11 +20,14 @@ enum BndryLoc {BNDRY_XIN=1,
 
 class BoundaryRegionBase {
 public:
-  BoundaryRegionBase() {}
-  BoundaryRegionBase(const string &name) : label(name) {}
-  BoundaryRegionBase(const string &name, BndryLoc loc) : label(name), location(loc) {}
+  BoundaryRegionBase() = delete;
+  BoundaryRegionBase(Mesh* mesh) : localmesh(mesh) {}
```
ZedThree: This changes the interface! An alternative would be:
```cpp
BoundaryRegionBase(Mesh* localmesh = nullptr) : localmesh(localmesh ? localmesh : mesh) {}
```
to use the global `mesh` if one isn't provided.

PR #1200, opened 2018-08-01
Set coordinates location

johnomotani:
Uses location argument of `Mesh::coordinates(CELL_LOC location)` to get correctly staggered metrics and dx/dy in derivatives, differential operators and Laplace inversions.

### DIFF
--- a/include/bout/coordinates.hxx
+++ b/include/bout/coordinates.hxx
@@ -123,24 +123,24 @@ public:
   const Field3D Div_par(const Field3D &f, CELL_LOC outloc=CELL_DEFAULT, DIFF_METHOD method=DIFF_DEFAULT);
   
   // Second derivative along magnetic field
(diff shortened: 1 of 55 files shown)


bendudson:
Not sure why the changes to `wave-1d` and `wave-1d-y` MMS tests, but otherwise looks good.
We probably should have more tests with staggered variables to reduce changes of accidentally breaking all this new stuff.

johnomotani:
>Not sure why the changes to wave-1d and wave-1d-y MMS tests

The changes to remove `MXG=1` from wave-1d and `MYG=1` from wave-1d-y were necessary because they use staggered grids, so with the changes in this PR they need staggered metrics that are calculated by interpolation and interp_to needs 2 guard cells. Previously they had used staggered derivatives but never needed to call interp_to.

The changes to nx values in wave-1d are just to make the grid size powers of 2. They had 'missed' counting the guard cells previously.

PR #1202, opened 2018-08-07
Make Monitor::timestep settable

johnomotani:
Allows subclasses to overwrite the value of timestep in their constructors, which is useful for setting its value from options in BOUT.inp. When timestep was private its value could only be set in the initializer list, when the parent constructor Monitor::Monitor() was called, which is much more restrictive.

### DIFF
--- a/include/bout/monitor.hxx
+++ b/include/bout/monitor.hxx
@@ -50,9 +50,23 @@ public:
   virtual void cleanup(){};
 
 protected:
-  BoutReal timestep;
+  /// Get the currently set timestep for this monitor
+  BoutReal getTimestep() const { return timestep; }
+
+  /// Set the timestep for this Monitor
+  ///
+  /// Can only be called before the Monitor is added to a Solver
+  void setTimestep(BoutReal new_timestep) {
+    if (is_added) {
+      throw BoutException("Monitor::set_timestep - Error: Monitor has already"
+          "been added to a Solver, so timestep cannot be changed.");
+    }
+    timestep = new_timestep;
+  }
 
 private:
+  bool is_added = false; ///< Set to true when Monitor is added to a Solver
+  BoutReal timestep;
   int freq;
 };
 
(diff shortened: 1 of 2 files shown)


dschwoerer:
Now that `timestep` is not private any more - could you also add some documentation?
At least say that `timestep` must not be changed after the monitor is added to the solver?
As long as it is clear to the user that changing the `timestep` later results in undefined behaviour, I am happy for this to go in.

bendudson:
If doing this can lead to undefined behavior,  perhaps we should put some safety checks around it?  It may be a bit too Java,  but get and set functions which are protected would enable checks to be done.  An internal flag which is set when added to a solver, for example.

dschwoerer:
I think the value is mostly ignored, and only Monitor::freq is used - but some parts of the code might cache the value - but I think only the main timestep is stored in the solver implementation.
getter and setter might be useful - as the setter could even update the freq value, if it is set and the change is appropriate ...

johnomotani:
getter/setter sounds good to me. I think they could even be public rather than protected since they check that changing timestep is OK?

d7919:
I think in general we want to minimise the public interface to just what is absolutely required.

johnomotani:
Where the frequencies are updated when a monitor with a shorter timestep is added to the solver
https://github.com/boutproject/BOUT-dev/blob/9079d0422a457f3966acd75cb43a63bc9262ba8a/src/solver/solver.cxx#L651-L656
should freqDefault be updated too? Like
```
freqDefault *= multi
```
It looks to me as though at the moment monitors added in future would default to the fastest timestep (`mon->freq=1`) rather than the output timestep.

dschwoerer:
I think you are right - thanks for noticing :+1: 
Do you want me to fix my bug and add a test to catch this in the future?

johnomotani:
@dschwoerer :+1: sounds good!

PR #1208, opened 2018-08-16
Stencil small tweak to aid vectorisation

d7919:
Adds `BoutNaN` to provide a quiet_NaN compatible with `BoutReal`.

Makes the default stencil constructor header only and provides default values for stencil members directly. This small change has been seen to enable vectorisation in a number of test cases of the form of the code typically found in index_derivs (more to come on this in a future PR).

Note this defaults the stencil values to be quiet_NaN which is different to the current effective default of 0.

### DIFF
--- a/include/bout_types.hxx
+++ b/include/bout_types.hxx
@@ -25,9 +25,13 @@
 #include <vector>
 #include <string>
 #include <map>
+#include <limits>
 
 /// Size of real numbers
 typedef double BoutReal;
+
+/// Quiet NaN
+const BoutReal BoutNaN = std::numeric_limits<BoutReal>::quiet_NaN();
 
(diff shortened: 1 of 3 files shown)


d7919:
I wonder if we should think about marking all of the member functions (and non-members) as deprecated for the next release -- I think these may only be used in one or two places but bring a fair bit of code with them. Note also that the non-member `abs`, `max` and `min` routines combined with the stencil constructor taking a single BoutReal could cause problems due to implicit conversion. For example:

~~~~
BoutReal a = -1.;
auto result = abs(a);
~~~~

could potentially lead to `result` being a stencil (or a Field?D) I think.

bendudson:
Agree that stencils should be deprecated, then removed at some point.

PR #1209, opened 2018-08-17
New region block loop macro

ZedThree:
This makes uses of BLOCK_REGION_LOOP look more like traditional range-based/foreach loops:

```cpp
BLOCK_REGION_LOOP(index, region) {
    A[index] = B[index] + C[index]
}
```

This has no discernible effect on performance, but does make uses of the macro look a bit nicer.

It also provides a `BLOCK_REGION_LOOP_SECTION` which takes OpenMP pragmas as a third argument, to enable things like:

```cpp
BOUT_OMP(parallel)
BLOCK_REGION_LOOP_SECTION(index, region, for schedule(guided) nowait) {
  ...
}
```

I didn't include `for` by default in this macro -- but maybe we always want it there?

Now would also be a good time to change the name if we wanted, maybe `BOUT_REGION_LOOP`, or  `BOUT_FOREACH`?

### DIFF
--- a/include/bout/region.hxx
+++ b/include/bout/region.hxx
@@ -65,9 +65,8 @@
 /// Helper macros for iterating over a Region making use of the
 /// contiguous blocks of indices
 ///
-/// @param[in] region An already existing Region
 /// @param[in] index  The name of the index variable to use in the loop
-/// The rest of the arguments form the loop body.
+/// @param[in] region An already existing Region
 ///
 /// The following loops vectorise well when index is type int, needs testing
 /// with current approach where index is Ind2D or ind3D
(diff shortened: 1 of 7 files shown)


tests/unit/include/bout/test_region.cxx
```diff
@@ -242,7 +242,7 @@ TEST_F(RegionTest, regionLoopAllSerial) {
   auto region = mesh->getRegion3D("RGN_ALL");
 
   Field3D a = 0.0;
-  BLOCK_REGION_LOOP_SERIAL(region, i, a[i] = 1.0;);
+  BLOCK_REGION_LOOP_SERIAL(i, region) { a[i] = 1.0; }
```
d7919: Is there any benefit to also having a `BLOCK_REGION_LOOP_SECTION` version of these?

ZedThree: Yes, possibly -- `BLOCK_REGION_LOOP` is now implemented in terms of `BLOCK_REGION_LOOP_SECTION`, so in some respect it is already tested. But it wouldn't hurt to have an explicit test

ZedThree:
Thanks to @d7919 for spotting that the omp macros should fall back to the serial version if OpenMP isn't available.

include/bout/region.hxx
```diff
@@ -90,18 +90,21 @@
...
     for (auto index = block->first; index < block->second; ++index)
 #else
 // No OpenMP, so fall back to slightly more efficient serial form
-#define BLOCK_REGION_LOOP_SECTION(index, region, omp_pragmas)                            \
+#define BLOCK_REGION_LOOP_OMP(index, region, omp_pragmas)                                \
   BLOCK_REGION_LOOP_SERIAL(index, region)
 #endif
 
-#define BLOCK_REGION_LOOP(index, region) BLOCK_REGION_LOOP_SECTION(index, region, parallel for)
+#define BLOCK_REGION_LOOP(index, region) BLOCK_REGION_LOOP_OMP(index, region, parallel for)
```
d7919: I think this should have `schedule (guided)` as well.

ZedThree:
Now provides `BLOCK_REGION_LOOP_INNER` which has `for schedule(guided) nowait` as the omp directives. These should be nice sensible defaults for loops inside parallel regions.

`BLOCK_REGION_LOOP_SECTION` is now `BLOCK_REGION_LOOP_OMP` to be used generically for omp directives.

Suggestions for a shorter name for this family of macros?

- `BOUT_LOOP*`
- `BOUT_FOREACH*`
- `REGION_LOOP*`

`BOUT` ones might be nice as they are somewhat namespaced

include/bout/region.hxx
ZedThree: Thanks, fixed

d7919:
Happy to approve but would be good to fix the name in this PR before use gets more widespread.

ZedThree:
Any objections to `BOUT_FOR*`?

```cpp
BOUT_FOR(index, region) {
   A[index] = B[index] + C[index];
}

BOUT_FOR_SERIAL(index, region) {
   A[index] = B[index] + C[index];
}

BOUT_OMP(parallel)
BOUT_FOR_INNER(index, region) {
   A[index] = B[index] + C[index];
}

BOUT_FOR_OMP(index, region, parallel for firstprivate(foo)) {
   A[index] = B[index] + C[index];
}
```

d7919:
Sounds good to me. 

Should we add some documentation to describe the purpose of the different macros?

ZedThree:
Done and done

bendudson:
Well done everyone! This represents a surprisingly large amount of work and a big improvement over previous versions, both in performance and readability. Serious kudos all round.

PR #1210, opened 2018-08-17
Towards stencil deprecation

d7919:
Remove all member and non-member stencil functions (including all constructors) and convert stencil from class to struct (not required but as stencil purely public struct is probably more appropriate).

Requires a small change to `DDX_CWENO3` due to removal of `operator+` and `operator-` for stencil. This was the only location any of the stencil functions where used.


d7919:
The removed code was marked deprecated in the previous release (i.e. current master) through a statement in the header. 

I'd be surprised if anyone was using stencil routines in their own code, but it's possible that if they did they might not have noticed these were deprecated.

include/stencils.hxx
```diff
@@ -35,57 +32,9 @@
...
-  friend const stencil operator+(BoutReal lhs, const stencil &rhs);
-  friend const stencil operator-(BoutReal lhs, const stencil &rhs);
-  friend const stencil operator*(BoutReal lhs, const stencil &rhs);
-  friend const stencil operator/(BoutReal lhs, const stencil &rhs);
-  
-  BoutReal min() const; ///< Minimum value over stencil
-  BoutReal max() const; ///< Maximum value over stencil
-  const stencil abs() const; ///< Stencil with all values replaced by absolute values
+struct stencil {
+  ///< stencil 2 each side of the centre -- in effect means M?G > 2 is not supported
```
ZedThree: Very minor quibble, but the doxygen comment should just be `///` to refer to the next line.

Also reading this comment makes me this @JosephThomasParker is on to something with replacing `stencil` with a `vector`. We'd have to do something like `stencil[mesh->xstart]` to get the centre point, so does require some thinking about the interface

PR #1211, opened 2018-08-17
Add offsetting to Region and Ind2D/Ind3D

ZedThree:
- Replaces #1114 -- Basically that code moved into `SpecificInd`, with some support from `Region`

Thanks to @d7919 and @JosephThomasParker for benchmarking. #1208 was important for enabling vectorisation of loops with offsets.

With this, `Region` becomes pretty much a drop-in replacement for `DataIterator` that can be both parallelised with OpenMP and vectorised by the compiler. Near enough always guaranteed to be faster than `DataIterator`, and can get almost triple-nested loop performance.

This PR is not perfect:
- the `Region` ctors that take `RegionIndices` or `ContiguousBlocks` directly don't know about `ny/nz`. These ctors are not likely used in practice outside of tests, but it would be good to work out how to do them correctly
- There is currently no protection for making a `Region<Ind2D>` with `nz != 1` -- some compile-time magic would be best to ensure this
- No support for `FieldPerp` yet as in #1114, but that can come later
- It would be *nice* if `Region` were actually templated on `Field2D/Field3D` rather than `Ind2D/Ind3D`, but this seems to cause pain in other places (actually unrelated to this PR, but just mentioning it while I think about it)


ZedThree:
18bc402 is seemingly unrelated, but this PR triggered the bug somehow so it seems fitting to fix it here

d7919:
Could you add the performance test ?

examples/performance/iterator-offsets/iterator-offsets.cxx
```diff
@@ -0,0 +1,291 @@
...
+        for(int k=0;k<mesh->LocalNz;++k) {
+          result(i,j,k) = (a(i,j+1,k) - a(i,j-1,k));
+        }
+      }
+    }
+    );
+#endif
+
+  // Range based for DataIterator with indices
+  ITERATOR_TEST_BLOCK(
```
d7919: I think this (and several below) should be inside the above openmp preprocessor block.

examples/performance/iterator-offsets/iterator-offsets.cxx
```diff
@@ -0,0 +1,291 @@
...
+        s.c = a[i];
+        s.p = a[i.yp()];
+        s.pp = nan("");
+
+        result[i] = (s.p - s.m);
+        );
+    );
+#ifdef _OPENMP
+
+  ITERATOR_TEST_BLOCK(
```
d7919: I think the distinction with the `(inside)` case below has disappeared due to the removal of the `IndexOffset` class. I think one can be removed and the name updated.

examples/performance/iterator-offsets/iterator-offsets.cxx
```diff
@@ -0,0 +1,291 @@
...
+    "Region with stencil & raw ints (parallel section omp)",
+    const auto nz = mesh->LocalNz;
+    BOUT_OMP(parallel)
+    {
+      stencil s;
+      BLOCK_REGION_LOOP_PARALLEL_SECTION(
+        mesh->getRegion3D("RGN_NOY"), i,
+        s.mm = nan("");
+        s.m = a[i.ind - (nz)];
```
d7919: I guess we don't strictly have an `operator[](int)` function so this shouldn't be allowed but works due to implicit conversion?

ZedThree: Yes, we might want to mark `Ind3D(int)` as explicit to avoid implicit casting like this

ZedThree:
This a representative weak scaling:

ph_scaling_pr_constblocks_nowait_funcptr

Exact details differ between machines, etc., but broadly the following points hold:

1. `Region` does vectorise, even with offsetting built in to `SpecificInd`
2. `nowait` seems to be required at small grid sizes. We probably don't want to bake this into all the loop macros, but it's probably wise to have one that does include it
3. There's some nature of resource contention (memory bandwidth? cache size?) at large grid sizes that kills the performance, but note that this also applies to the nested loop
4. We probably can't do very much better than this form if we think the nested loop is close to ideal performance
5. The function pointer call kills the vectorisation, and replacing this should be one of the main goals of the version after 4.2

d7919:
Just to note this isn't quite right at the moment as the offsetting routines return a `SpecificInd` which has an ambiguous conversion to either `Ind2D` or `Ind3D`. A fix will come in another PR soon (hopefully).

ZedThree:
Might want to merge next again after #1209 is merged to use the new loop macro name in the iterator-offset tests

include/bout/region.hxx
```diff
@@ -157,9 +194,55 @@ public:
...
+    SpecificInd<N> new_ind{ind % n, ny, nz};
     return new_ind;
   }
+
+  /// Convenience functions for converting to (x, y, z)
+  int x() const { return (ind / nz) / ny; }
+  int y() const { return (ind / nz) % ny; }
+  int z() const { return (ind % nz); }
+
+  const inline SpecificInd xp(int dx = 1) const { return {ind + (dx * ny * nz), ny, nz}; }
```
d7919: I think all these need `<N>` adding?

ZedThree: Actually, I think it's the other way round -- we don't need to mention `<N>` anywhere inside the class definition because we don't actually use it. The whole class has been templated on `N`.

Likewise, I think we don't need to write `Region<T>` inside the `Region` class definition

ZedThree:
Ok, I think this is finally ready

include/bout/region.hxx
d7919: Ah ok, I thought the return types would require knowledge of what type to template on but I didn't test this was the case.

ZedThree:
Some important design questions that I hadn't thought of:

- We probably want to make the conversion from `int` to `Ind3D/2D` `explicit`, so that we can't index a `Field3D/2D` with an `int` -- any good reasons *not* to do this?
- Should `Region<Ind2D>` enforce `nz == 1`? For that matter, should `Ind2D` enforce `nz == 1`?
- Should `operator==` compare `ny, nz` for `Ind3D/2D`? If so, should `ny, nz` be public?

ZedThree:
Currently, checking `ny, nz` would be standing in for checking the meshes are the same. We would need to keep some reference to the mesh, but I think I'd like to make that a shared pointer or something. We can do that in the next version.

d7919:
Currently Region/Inds don't know about the mesh as I tried to make it a bit more general (and less interconnected) originally but I think eventually we'll probably end up wanting these objects to know about mesh

include/field2d.hxx
```diff
@@ -150,8 +150,8 @@ class Field2D : public Field, public FieldData {
   const BoutReal& operator[](const Ind2D &d) const {
     return data[d.ind];
   }
-  BoutReal& operator[](const Ind3D &d); 
-  const BoutReal& operator[](const Ind3D &d) const;
+  inline BoutReal& operator[](const Ind3D &d);
```
bendudson: I don't think the compiler can inline a function unless the body is available here in the header.

src/field/field2d.cxx
```diff
@@ -174,12 +174,13 @@ const IndexRange Field2D::region(REGION rgn) const {
   };
 }
 
-BoutReal& Field2D::operator[](const Ind3D &d) {
-  return data[fieldmesh->map3Dto2D(d)];
+// Not in header because we need to access fieldmesh
+inline BoutReal& Field2D::operator[](const Ind3D &d) {
```
bendudson: Isn't `fieldmesh` part of the class? Should be find in the header I would think.

ZedThree: `Mesh` is just forward declared in the Field2D header to avoid circular dependencies, but `mesh.hxx` is included (via `globals.hxx`) in the implementation file.

bendudson: Ah yes, tricky. Since it has to call a function `map3Dto2D` anyway perhaps it's not a problem that it won't be inlined.

include/field2d.hxx
ZedThree: Yes, it has to be "in the same translation unit", so included via this header :(

ZedThree: Not only was this not useful, it was actively harmful as it caused the linker to not see the implementation.

Also it turns out it wasn't being tested in the unit tests, so fixed that too

PR #1212, opened 2018-08-19
Matrix tensor size type

d7919:
Use `size_type` to define the type of size related variables in Matrix and Tensor

Currently set this to `int` whilst previous behaviour would be
equivalent to `unsigned int`. The motivation for changing to `int` is
that this allows `operator()` uses to vectorise when in loops with an
`int` index.

Changing loop indices to `unsigned int` (or `std::size_t`) doesn't
actually allow vectorisation even when `Matrix::size_type = std::size_t`
as `Array` currently also uses `int` as the indexing type. This seems to
be enough to prevent vectorisation in simple loops over `Matrix` elements.

### DIFF
--- a/include/utils.hxx
+++ b/include/utils.hxx
@@ -75,9 +75,13 @@ public:
   }
   
   inline T& operator()(size_type i1, size_type i2) {
+    ASSERT2(0<=i1 && i1<n1);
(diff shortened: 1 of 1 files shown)


ZedThree:
Long-term, this should use `Array<T>::size_type`, once that is defined and used consistently

PR #1220, opened 2018-08-24
Squashoutput

dschwoerer:
I want to extend this a bit further:
* add a `--delete` flag, that removes the non squashed files
* add an `--append` flag, to append to the squashed file, if the simulation is continued (probably slightly more complicated)

### DIFF
--- a/bin/bout-squashoutput
+++ b/bin/bout-squashoutput
@@ -1,0 +1,1 @@
+../tools/pylib/boutdata/squashoutput.py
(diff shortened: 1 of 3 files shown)


bendudson:
This is great. It may be better to put the script in the `bin/` directory, so it could then import the squashdata function and do the argument parsing there (see e.g. https://github.com/boutproject/BOUT-dev/blob/master/bin/bout-sonnet )

dschwoerer:
Thanks @johnomotani for providing this in the first place :+1:

Issue #1017, opened 2018-04-23
clang-format version dependent


dschwoerer:
I noticed that clang-format's output is version dependent.

Different between v4 and v5 can be seen here:
https://travis-ci.org/boutproject/BOUT-dev/jobs/369284888

If v6 and v7 are different again, than this is going to be a problem, if we want to have everything clang-auto formatted as currently planned by #538 

Is there a way to define our style more precisely to prevent this?
Should we expect a specific version of clang installed?

ZedThree:
That's a pain. We can try to tighten up the `.clang-format` config file which might help to some extent. I don't know how forwards compatible they are, which might be a sticking point.

We probably don't want tests to fail due to differences in formatting regardless.

dschwoerer:
I am avoiding the test-failures by removing any white space for comparison
Here is a branch with different formattings:
https://github.com/boutproject/BOUT-dev/commits/clang-format-diff
Here is e.g. the difference from v6 to v5:
https://github.com/boutproject/BOUT-dev/commit/b305c8c7475b49696c4239964c8fc214e0aa0dd1

ZedThree:
Oh dear, those are not great. I'll have a go at tightening up the format config file

dschwoerer:
Did you have a look?
I would remove the clang-format from #538 - although I thought it would be really nice to have an automatic way to format the code.

ZedThree:
Closing this: clang-format good, but tests relying on output of clang-format: bad. So we won't do that.

PR #1222, opened 2018-08-24
Add ind perp to region support

d7919:
Introduce a new index type for perpendicular fields -- not currently used but will be helpful for unifying the Field approaches.

Thanks to @JosephThomasParker for the hard work on initially doing this.

### DIFF
--- a/include/bout/mesh.hxx
+++ b/include/bout/mesh.hxx
@@ -673,7 +673,14 @@ class Mesh {
   /// Converts an Ind3D to an Ind2D using calculation
   Ind2D ind3Dto2D(const Ind3D &ind3D) { return {ind3D.ind / LocalNz, LocalNy, 1}; }
 
-  /// Converts an Ind3D to a raw int representing a 2D index using a lookup -- to be used with care
+  /// Converts an Ind3D to an IndPerp using calculation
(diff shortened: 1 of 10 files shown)


dschwoerer:
There are new failures:
https://omadesala.alfa1024.info/~bout-test/2018-08-24_22:34/
At least test-code-style is new ....

d7919:
The failure message was 

> ----- test-code-style -----
  Generating derivs.hxx
/bin/sh: clang-format: command not found
Formatting failed
../../../include/derivs.hxx changed upon recreation
Checking file out - to prevent further changes
-> Errors listed above

But this PR doesn't touch `dervis.hxx` -- is this failing just because clang-format isn't available?

dschwoerer:
Sorry, I should have check the details first.

This is part of the check to verify the file is generated:
`make -C ${autogen%/*} --no-print-directory`
But it seems that sometimes (not sure when and how) the checked out files have different time stamps, and then the generated file might be older - I don't think there is in general a way to avoid this.
That would probably require to switch to hash based rebuilds, rather then timestamp based ones, which make does not support, so that would make thinks rather more complicated - but avoid issues to accidentally require jinja2 in released versions ...

Installing clang-format would work around the issue, if it would be in the right version (#1017)

PR #1223, opened 2018-08-24
Add 'Laplace()' operator in boutcore

johnomotani:


dschwoerer:
Would you mind adding Parameters and Returns documentation?

I think otherwise the parameters will not be shown properly [in the documenation](https://omadesala.physics.dcu.ie/bout/user_docs/python_boutcore.html)

PR #1226, opened 2018-08-24
Fix interpolation boundaries

dschwoerer:
Keep workaround for lack of interpolation into boundary if converting to
fieldaligned
[Apply the hack for non-fieldaligned to fieldaligned](https://github.com/boutproject/BOUT-dev/compare/next...fix_interp_to?expand=1#diff-08115b2d2191f4b50484126770cfeb98L70)

Not sure this is the right thing to do, but at least fa and non-fa are this way consistent.

### DIFF
--- a/src/mesh/interpolation.cxx
+++ b/src/mesh/interpolation.cxx
@@ -146,7 +146,7 @@ const Field3D interp_to(const Field3D &var, CELL_LOC loc, REGION region) {
           // coordinates
 
           Field3D var_fa = fieldmesh->toFieldAligned(var);
-          Field3D result_fa;
+          Field3D result_fa = fieldmesh->toFieldAligned(result);
           result_fa.allocate();
           if (fieldmesh->ystart > 1) {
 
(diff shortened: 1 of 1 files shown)


src/mesh/interpolation.cxx
```diff
@@ -146,7 +146,10 @@ const Field3D interp_to(const Field3D &var, CELL_LOC loc, REGION region) {
           // coordinates
 
           Field3D var_fa = fieldmesh->toFieldAligned(var);
-          Field3D result_fa = fieldmesh->toFieldAligned(result);
+          Field3D result_fa(fieldmesh);
+          if (loc != RGN_NOBNDRY) {
```
dschwoerer: shouldn't it be `region` rather then `loc`?

johnomotani: yep, sorry. I must need a coffee...

src/mesh/interpolation.cxx
```diff
@@ -146,7 +146,10 @@ const Field3D interp_to(const Field3D &var, CELL_LOC loc, REGION region) {
           // coordinates
 
           Field3D var_fa = fieldmesh->toFieldAligned(var);
-          Field3D result_fa;
+          Field3D result_fa(fieldmesh);
+          if (loc != RGN_NOBNDRY) {
+            result = fieldmesh->toFieldAligned(result);
```
dschwoerer: and also `result_fa` rather then `result`?

johnomotani: :+1: thanks @dschwoerer

PR #1231, opened 2018-08-28
Fix constructors for BoundaryRegion classes

ZedThree:
A couple of recent PRs either changed the semantics of `BoundaryRegion` a little or partly clobbered other PRs

- Don't allow default ctor
- Pass `name` by value and move
- Make sure derived classes have default value for `passmesh`

### DIFF
--- a/include/boundary_region.hxx
+++ b/include/boundary_region.hxx
@@ -24,11 +24,11 @@ class BoundaryRegionBase {
 public:
 
   BoundaryRegionBase() = delete;
-  // default to global mesh object if none is passed as argument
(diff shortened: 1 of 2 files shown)


ZedThree:
No worries! :)

I'm not sure a test is possible, as the observable behaviour is exactly (or near enough) the same in both cases, it's just a optimisation. clang-tidy can spot some of these (which is how I did it in the first place, and notice the clobbering!), but I'm a little worried it might have some mutually exclusive patterns so I wouldn't want to fully automate it yet. And trying to do it without a compiler-based tool is asking for trouble.

PR #1236, opened 2018-08-28
Squash progress

dschwoerer:
This uses a secret, undocumented feature from zoidberg :grinning: 

The continue option is particularly useful if your squash job gets killed after running for an hour :-)

I could split them again, but I guess both options are useful.


tools/pylib/boutdata/squashoutput.py
```diff
@@ -70,21 +66,37 @@ def squashoutput(datadir=".", outputname="BOUT.dmp.nc", format="NETCDF4", tind=N
...
         datadirnew = tempfile.mkdtemp(dir=datadir)
         for f in glob.glob(datadir+"/BOUT.dmp.*.??"):
             if not quiet:
                 print("moving",f)
             shutil.move(f,datadirnew)
         oldfile=datadirnew+"/"+outputname
         datadir=datadirnew
+    if docontinue:
+        if append:
+            raise NotImplemented("append & docontinue: Case not handled")
```
ZedThree: Typo, should be `NotImplementedError` -- `NotImplemented` is a special value for returning from arithmetic operators

tools/pylib/boututils/datafile.py
```diff
@@ -628,6 +629,12 @@ def find_dim(dim):
             # Get new list of variables, and turn into a tuple
             dims = tuple(map(find_dim, dlist))
 
+            if name in self.handle.variables:
+                raise IOError("Variable %s already present",name)
```
ZedThree: This should be `ValueError` -- `IOError` means "something went wrong at the disk level"

tools/pylib/boututils/datafile.py
```diff
@@ -628,6 +629,12 @@ def find_dim(dim):
...
+            tmpname=name+"_bout_temp"
+            i=0
+            while tmpname in self.handle.variables:
+                tmpname=name+"_bout_temp_%d"%i
```
ZedThree: Were you meaning to increment `i` here? This looks like suspiciously like an infinite loop

tools/pylib/boutdata/squashoutput.py
```diff
@@ -99,9 +111,26 @@ def squashoutput(datadir=".", outputname="BOUT.dmp.nc", format="NETCDF4", tind=N
...
+        os.rmdir(datadirtmp)
+
     t_array_index = outputvars.index("t_array")
     outputvars.append(outputvars.pop(t_array_index))
 
+    if progress:
+        sizes_=outputs.sizes()
+        sizes={}
+        total=0
+        for var,size in sizes_.items():
```
ZedThree: Please could you put a comment here explaining why you're doing this? Or rename the variables to something more descriptive?

ZedThree:
Please could you start using a PEP8 tool for formatting? The lack of whitespace makes it difficult to read

dschwoerer:
Can I pep format the whole file I edited?
Then I can add an appropriate script to git, and this will never happen again ...

dschwoerer:
Thanks for all the feedback :+1: 
I think I have addressed everything.

dschwoerer:
I am not quite sure, as this makes the import much slower, importing `zoidberg` takes over a second, even on local disk, and it prints 3 warnings if matplotlib is not available. Is there a way to avoid importing everything of zoidberg? Maybe add a symlink to progress.py - that should speedup the import quit significantly:
```
$ time echo import zoidberg|python3
real    0m5.139s
user    0m1.232s
sys     0m0.496s
$ time (cd zoidberg ; echo import progress|python3)
real    0m0.038s
user    0m0.032s
sys     0m0.006s
```
I guess if import zoidberg would be faster, that would be great as well, but I don't know how to do that ...

dschwoerer:
The reccent commits are somewhat unrelated to the progress one.
But in order to avoid merge conflicts, I pushed directly to this branch.

Should I rebase and make this a separate PR?

The changes are:   Squash improvements

* Support appending and continue at the same time
* Simplifiy squash code
* Requires to know what the real length in time of a variable is
- implement getRealTlength() for netcdf
* Requires support to append to a variable in DataFile
* Makes the code more robost. The squashed file is only shortly moved,
and gets moved back before any slow operations starts.
* Filenames must be pure names, not paths. This was required to simplify
the code quite significantly. If writing to another disk is preffered,
using a simlink to the file that shall be created can be used as a
workaround, see e.g. the integrated test.

dschwoerer:
to be replaced by xarray ...

PR #1237, opened 2018-08-29
Sundials hybrid OpenMP-MPI

bendudson:
Adds an implementation of Sundials' NVector, based on nvector_parallel, with added OpenMP directives. 

Configure with:

    ./configure --with-sundials --enable-openmp --enable-sundials-openmp

Tested using MMS/time with nx=ny=1000. 

    1.000e+00         76               76       6.71e+00    53.1    0.0    0.0    1.2   45.7

Two odd things currently:

1. Little to no speedup observed. I suspect this is because I have BLAS/LAPACK installed, and many of the n_vector routines go to these if possible. Removing these calls from the nvector code may be needed to see a speedup. 
2. The timing reports only around 40-50% of time in the solver, though the RHS function consists of `ddt(f) = f;`. Copying data to/from fields is included in the RHS time, so perhaps this is it.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -87,6 +87,8 @@ AC_ARG_ENABLE(openmp,       [AS_HELP_STRING([--enable-openmp],
         [Enable building with OpenMP support])],,[enable_openmp=no])
 AC_ARG_ENABLE(pvode_openmp, [AS_HELP_STRING([--enable-pvode-openmp],
         [Enable building PVODE with OpenMP support])],,[enable_pvode_openmp=no])
+AC_ARG_ENABLE(sundials_openmp, [AS_HELP_STRING([--enable-sundials-openmp],
+        [Enable building OpenMP-MPI NVector for SUNDIALS])],,[enable_sundials_openmp=no])
 
 AC_ARG_VAR(EXTRA_INCS,[Extra compile flags])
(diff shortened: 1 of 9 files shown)


bendudson:
Perhaps an easier way to achieve this, without having to maintain our own nvector, would be to just build sundials with a BLAS library which supports OpenMP.

configure.ac
```diff
@@ -1011,6 +1017,13 @@ AS_IF([test "x$with_arkode" != "x" && test "x$with_arkode" != "xno"], [
     #include <arkode/arkode_bbdpre.h>
     extern int arkode_bbd_rhs(ARKODEINT, double, N_Vector, N_Vector, void *);
   ], [ARKBBDPrecInit(nullptr, 0, 0, 0, 0, 0, 0, arkode_bbd_rhs, nullptr);])
+  has_some_sundials="yes"
+])
+
+# Set a flag to enable compilation of hybrid OpenMP-MPI NVector
+AS_IF([test "x$has_some_sundials" != "no" && test "x$enable_sundials_openmp" != "x" && test "x$enable_sundials_openmp" != "xno"],[
```
ZedThree: We're in complete control of `$has_some_sundials`, so this could just be `$has_some_sundials == "no"`

ZedThree:
Closing for now, but leaving branch in case we want to return to this

PR #1238, opened 2018-08-29
More consistent/stringent pre/postconditions on Field* functions

ZedThree:
- Use `checkData(input)` rather than just `ASSERT(input.isAllocated())` to be more consistent with the arithmetic operators
- Make sure we set the y-index of `FieldPerp`s returned from functions
- Check that the y-index of a `FieldPerp` is set before using one

`checkData` is a more expensive check but it falls back to plain `isAllocated()` if `CHECK < 3`. Being consistent is a bit clearer too: we're more explicit about pre- and postconditions.

### DIFF
--- a/src/field/field3d.cxx
+++ b/src/field/field3d.cxx
@@ -380,8 +380,6 @@ Field3D & Field3D::operator=(const Field3D &rhs) {
 Field3D & Field3D::operator=(const Field2D &rhs) {
   TRACE("Field3D = Field2D");
   
-  ASSERT1(rhs.isAllocated());
-  
   /// Check that the data is valid
   checkData(rhs);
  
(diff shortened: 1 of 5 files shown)


ZedThree:
Yes, the `checkData` then gets done in the non-member functions (`FPERP_FPERP_OP_FIELD(op, ftype)`) so we don't want to do it twice (this is also what's done in the other field's arithmetic operators)

d7919:
Thanks, that makes sense!

src/field/field3d.cxx
```diff
@@ -401,6 +399,9 @@ Field3D & Field3D::operator=(const Field2D &rhs) {
 void Field3D::operator=(const FieldPerp &rhs) {
   ASSERT1(rhs.isAllocated());
```
bendudson: Can this be removed?

ZedThree: Yes, sorry, I meant to replace all the plain `isAllocated` checks with `checkData`

ZedThree:
Do we want to generally replace *all* `ASSERT1(isAllocated())` with `checkData`? At the default check level, they are equivalent, but automatically become stricter at higher check levels

PR #1239, opened 2018-08-29
Faster arakawa with bout for

d7919:
Slightly uglier versions of Arakawa brackets that are OMP parallelised and (mostly) vectorised. 

Had to split inner loop over z into three parts in order to get vectorisation.

Also adds a simple performance test for the bracket operator. This shows the new Arakawa methods are somewhere between 2-4x faster than the original method (that can still be accessed using BRACKET_ARAKAWA_OLD).

### DIFF
--- a/examples/performance/bracket/bracket.cxx
+++ b/examples/performance/bracket/bracket.cxx
@@ -1,0 +1,165 @@
+/*
+ * Testing performance of iterators over the mesh
+ *
+ */
+
+#include <bout.hxx>
+
+#include <chrono>
+#include <iomanip>
+#include <iostream>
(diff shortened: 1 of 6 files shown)


d7919:
2D,3D on 1 thread
bracket_2d_3d_nth_1

3D, 3D on 1 thread
bracket_3d_3d_nth_1


2D,3D on 8 threads
bracket_2d_3d_nth_8

3D,3D on 8 threads
bracket_3d_3d_nth_8

d7919:
Note ignore the `default` case in the previous plots -- these are actually "Simple".

PR #1240, opened 2018-08-29
Solver loadsave

bendudson:
New option `solver:loadsave_chunks` modifies how data is copied between BOUT++ Field objects and the 1D arrays used internally in the time integration solvers.

The old method (= `false`, the default) interleaves the fields, so that values which are spatially close together (and so likely to be strongly coupled) are close in the 1D arrays. This is used in some preconditioners, e.g. block diagonal, which assume nearby points are more likely to be coupled.

The new method (= `true`) puts each field into a contiguous block. This is faster, but now the coupling will not be near the diagonal. In many cases this may not matter.

Testing with `tests/integrated/time`, setting `nx = ny = 1000`. Old, default method:

    OMP_NUM_THREADS=1 ./time solver:loadsave_chunks=false
    1.000e+00         76               76       6.64e+00    53.4    0.0    0.0    1.2   45.4

New method:

    OMP_NUM_THREADS=1 ./time solver:loadsave_chunks=true
    1.000e+00         76               76       4.01e+00    41.6    0.0    0.0    2.1   56.3

New method with 4 threads. Only one field being evolved so no speedup:

    OMP_NUM_THREADS=4 ./time solver:loadsave_chunks=true
    1.000e+00         76               76       4.12e+00    42.3    0.0    0.0    2.0   55.7

Testing with `tests/integrated/test-interchange-instability`:

    OMP_NUM_THREADS=2 ./2fluid -d data_1 timestep=1e3 solver:loadsave_chunks=false
    1.000e+03        206       2.77e+00    48.5   41.0    0.0    0.2   10.3

New method:

    OMP_NUM_THREADS=2 ./2fluid -d data_1 timestep=1e3 solver:loadsave_chunks=true
    1.000e+03        206       2.70e+00    48.8   42.4    0.0    0.2    8.7

so seems to make a difference for some (1) cases, but not much.

### DIFF
--- a/include/bout/openmpwrap.hxx
+++ b/include/bout/openmpwrap.hxx
@@ -1,7 +1,22 @@
 /************************************************************************//**
- * \brief Openmp utilitiy wrappers
+ * \brief OpenMP utility wrappers
  * 
+ * A macro wrapper BOUT_OMP avoids noisy `unknown pragma` warnings when
+ * compiling without OpenMP support.
  * 
+ * Example
+ * -------
+ *
+ *    BOUT_OMP(parallel) {
+ *       ...
+ *    }
+ *
+ * becomes
+ *
+ *    #pragma omp parallel {
+ *       
+ *    } 
+ *
  **************************************************************************
  * Copyright 2017
  *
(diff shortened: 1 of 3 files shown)


d7919:
Would an "override" flag for `addRegion` be helpful to allow us to skip the check for an existing region?

ZedThree:
Closing for now, but leaving branch in case we want to return to this

PR #1241, opened 2018-08-29
Reduce memory usage of squashoutput

dschwoerer:
Can help to avoid out of memory errors


johnomotani:
I think it would be good to add also a `DataFile.sync()` method to the parent `DataFile` class and include a docstring there. This seems to be our documentation convention for `DataFile` methods.

PR #1246, opened 2018-08-31
Add -Wall -Wextra to default configure

ZedThree:
This PR fixes the remaining warnings from `-Wall -Wextra` and adds those flags to the default configure (with an option to disable them), along with `-Wnull-dereference -Wno-cast-function-type`.

I've turned off `-Wcast-function-type` because PETSc actually *requires* us to do bad things in order to use `MatFDColoringSetFunction`. We're otherwise not in the habit of casting function pointers, so this doesn't worry me too much.

I've not included `-Werror`, but maybe this should be turned on for development in `next` and turned off in `master`?

There's an [Autoconf macro](http://git.savannah.gnu.org/gitweb/?p=autoconf-archive.git;a=blob_plain;f=m4/ax_compiler_flags_cxxflags.m4) that adds a whole bunch of flags, but most are either not helpful or already included in `Wall/Wextra`. Additionally, there are a few C++ specific ones that cause a whole host of noise from third party C libraries like OpenMPI and PETSc, such as `-Wuseless-cast` and `-Wold-style-cast`. This PR also fixes those warnings, but doesn't turn them on.

There's also a GCC specific pragma in the unit tests in order to not warn about deprecated stuff we still want to test -- there are probably better ways of doing this.

### DIFF
--- a/src/fileio/impls/hdf5/h5_format.cxx
+++ b/src/fileio/impls/hdf5/h5_format.cxx
@@ -334,12 +334,15 @@ bool H5Format::addVar(const string &name, bool repeat, hid_t write_hdf5_type, in
     dataSet = H5Dopen(dataFile, name.c_str(), H5P_DEFAULT);
     if (dataSet < 0) {
       // Negative value indicates error, i.e. file does not exist, so create:
-      hsize_t init_size[3],init_size_local[3];
+      hsize_t init_size[3];
       if (parallel) {
-        init_size[0]=mesh->GlobalNx-2*mesh->xstart; init_size[1]=mesh->GlobalNy-2*mesh->ystart; init_size[2]=mesh->GlobalNz;
-      }
-      else {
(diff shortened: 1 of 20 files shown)


src/fileio/impls/hdf5/h5_format.cxx
```diff
@@ -666,7 +671,7 @@ bool H5Format::read_rec(void *data, hid_t hdf5_type, const char *name, int lx, i
 }
 
 bool H5Format::write_rec(int *data, const char *name, int lx, int ly, int lz) {
-  return write_rec(data, H5T_NATIVE_INT, H5T_NATIVE_INT, name, lx, ly, lz);
```
bendudson: Why has the number of arguments changed?

ZedThree: `write_hdf5_type` is used to initially create the variable, but that functionality has now been moved into `H5Format::addVar` so `write_hdf5_type` is no longer used here.

This is a private method so doesn't change the public API

PR #1247, opened 2018-08-31
Use BOUT_FOR everywhere and deprecate use of DataIterator

ZedThree:
- Deprecates Field* functions that use DI and related classes. In the version after 4.2, we'll be un-deprecating these but changing the return types to `Region<T>`, so hopefully everyone is already using `auto` everywhere! Accessing `x, y, z` also be functions, so there will be some slight breakage
- Uses the `BOUT_FOR` family of loop-macros everywhere possible. This should magically speed up large parts of the code (mostly calc), as well as enabling OpenMP in the same places

This is the last major piece of work for 4.2 -- after this, we're ready for the release candidate!

### DIFF
--- a/configure
+++ b/configure
@@ -786,8 +786,8 @@ enable_sigfpe
 enable_backtrace
 enable_shared
 enable_openmp
-enable_pvode_openmp
 with_openmp_schedule
+enable_pvode_openmp
 with_gcov
 enable_code_coverage
 with_hdf5
@@ -1444,8 +1444,6 @@ Optional Features:
   --enable-openmp         Enable building with OpenMP support
   --enable-pvode-openmp   Enable building PVODE with OpenMP support
   --disable-openmp        do not use OpenMP
-  --with-openmp-schedule=static/dynamic/guided/auto/runtime
-                          set OpenMP schedule (default: static)
   --enable-code-coverage  Whether to enable code coverage support
 
 Optional Packages:
@@ -1464,6 +1462,8 @@ Optional Packages:
   --with-mumps            Link with MUMPS library for direct matrix inversions
   --with-arkode           Use the SUNDIALS ARKODE solver
   --with-scorep           Enable support for scorep based instrumentation
+  --with-openmp-schedule=static
+                          Set OpenMP schedule (default: static)
   --with-gcov=GCOV        use given GCOV for coverage (GCOV=gcov).
   --with-hdf5=yes/no/PATH location of h5cc for serial HDF5 configuration
   --with-parallelhdf5=yes/no/PATH
@@ -2659,18 +2659,19 @@ else
   enable_openmp=no
 fi
 
+
+# Check whether --with-openmp_schedule was given.
+if test "${with_openmp_schedule+set}" = set; then :
+  withval=$with_openmp_schedule;
+else
+  with_openmp_schedule=static
+fi
+
 # Check whether --enable-pvode_openmp was given.
 if test "${enable_pvode_openmp+set}" = set; then :
   enableval=$enable_pvode_openmp;
 else
   enable_pvode_openmp=no
-fi
-
-# Check whether --openmp-schedule was given.
-if test "${with_openmp_schedule+set}" = set; then :
-  enableval=$with_openmp_schedule;
-else
-  with_openmp_schedule=static
 fi
 
 
@@ -5115,13 +5116,19 @@ $as_echo "$ac_cv_prog_cxx_openmp" >&6; }
       "none needed" | unsupported)
 	;; #(
       *)
-	OPENMP_CXXFLAGS="$ac_cv_prog_cxx_openmp -DOPENMP_SCHEDULE=$with_openmp_schedule" ;;
+	OPENMP_CXXFLAGS=$ac_cv_prog_cxx_openmp ;;
     esac
   fi
 
 
 HAS_OPENMP=$enable_openmp
-OPENMP_SCHEDULE=$with_openmp_schedule
+
+if test "x$enable_openmp" = "xyes"; then :
+
+  OPENMP_SCHEDULE=$with_openmp_schedule
+  OPENMP_CXXFLAGS="$OPENMP_CXXFLAGS -DOPENMP_SCHEDULE=$OPENMP_SCHEDULE"
+
+fi
 
 #############################################################
 # Code coverage using gcov
--- a/configure
+++ b/configure
@@ -787,6 +787,7 @@ enable_backtrace
 enable_shared
 enable_openmp
 enable_pvode_openmp
+with_openmp_schedule
 with_gcov
 enable_code_coverage
 with_hdf5
@@ -1443,6 +1444,8 @@ Optional Features:
   --enable-openmp         Enable building with OpenMP support
   --enable-pvode-openmp   Enable building PVODE with OpenMP support
   --disable-openmp        do not use OpenMP
+  --with-openmp-schedule=static/dynamic/guided/auto
+                          set OpenMP schedule
   --enable-code-coverage  Whether to enable code coverage support
 
 Optional Packages:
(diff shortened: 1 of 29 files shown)


src/field/field3d.cxx
```diff
@@ -1065,8 +1128,18 @@ void shiftZ(Field3D &var, int jx, int jy, double zangle) {
...
-    shiftZ(var, i.x, i.y, zangle);
+  const auto region_str = REGION_STRING(rgn);
+
+  // Only allow a whitelist of regions for now
+  ASSERT2(region_str == "RGN_ALL" || region_str == "RGN_NOBNDRY" ||
+          region_str == "RGN_NOX" || region_str == "RGN_NOY");
+
+  const Region<Ind2D> &region = var.getMesh()->getRegion2D(region_str);
+
+  // Could be OpenMP if shiftZ(Field3D, int, int, double) didn't throw
```
bendudson: Since an exception thrown would most likely just terminate the program, can we just allow it to happen. It makes debugging more difficult, but could be done by turning off OpenMP.

src/mesh/difops.cxx
```diff
@@ -307,115 +308,112 @@ const Field3D Vpar_Grad_par_LCtoC(const Field3D &v, const Field3D &f, REGION reg
...
+        vval.mm = v_fa[i.ymm()];
+        vval.m = v_fa[i.ym()];
+        vval.c = v_fa[i];
+        vval.p = v_fa[i.yp()];
+        vval.pp = v_fa[i.ypp()];
+
+        // Left side
+        result[i] = (vval.c >= 0.0) ? vval.c * fval.m : vval.c * fval.c;
+        // Right side
+        result[i] -= (vval.p >= 0.0) ? vval.p * fval.c : vval.p * fval.p;
```
bendudson: Doesn't look like the `.pp` and `.mm` components are used in this look. Leave as NaN?

src/mesh/index_derivs.cxx
```diff
@@ -1606,32 +1649,50 @@ const Field3D Mesh::indexD2DZ2(const Field3D &f, CELL_LOC outloc,
...
+          cv[jz] *= -kwave * kwave;
           if (shift)
             cv[jz] *= exp(0.5 * Im * (shift * kwave));
         }
+        for (int jz = kmax + 1; jz <= ncz / 2; jz++) {
+          cv[jz] = 0.0;
+        }
 
-        irfft(cv.begin(), ncz, result(jx, jy)); // Reverse FFT
+        irfft(cv.begin(), ncz, result(i.x(), i.y())); // Reverse FFT
```
bendudson: Can this indexing be made more efficient? Seems we're calculating the `x` and `y` indices, just to turn them back into a single index.

bendudson:
I think it would be good to add something to the manual on these `BOUT_FOR` macros and the support for OpenMP

src/field/field3d.cxx
```diff
@@ -857,21 +882,26 @@ BoutReal mean(const Field3D &f, bool allpe, REGION rgn) {
...
   }
 
   if(allpe) {
     // MPI reduce
-    BoutReal localresult[2] = {result[0], result[1]};
-    MPI_Allreduce(&localresult, &result, 2, MPI_DOUBLE, MPI_SUM, BoutComm::get());
+    BoutReal localresult = result;
+    MPI_Allreduce(&localresult, &result, 1, MPI_DOUBLE, MPI_SUM, BoutComm::get());
+    int localcount = count;
+    MPI_Allreduce(&localcount, &count, 1, MPI_INT, MPI_SUM, BoutComm::get());
```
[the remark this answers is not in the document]
d7919: Yes I didn't enjoy doubling this but the `OpenMP` reduction clause wouldn't work with the original two element array block (should be supported in new versions of `OpenMP`, >= 4.5 I think). We could of course use `int` in the reduction and then pack it into a two element `BoutReal` array to be communicated to avoid this. I'd guess that the global synchronisation imposed by the first call might not be such a big cost for the second call as procs will already be close to in sync, but this is just a guess.

src/mesh/difops.cxx
d7919: Good spot, I've removed these.

src/mesh/index_derivs.cxx
d7919: We're indexing a `Field3D` here as a 2D to return a pointer to the start of the z "slice". I think we could probably replace this with something like `&result(i, 0)`? This still involves some conversion from 2d to 3d but avoids getting the `x` and `y` first.

d7919:
The y offsetting does bounds check at sufficiently high check. Array has bounds checking which is why there isn't any in the Field [] indexing currently (out of bounds in X should always push you outside the array size).

bendudson:
Since regions can now be added, can this comment be removed?
https://github.com/boutproject/BOUT-dev/blob/1f53be7811d2e5f49993578b9d8796748bd55afc/include/bout/region.hxx#L662

d7919:
@JosephThomasParker is doing some timing checks to ensure this branch performs appropriately. I'd suggest we hold off on any merge until those results come in.

d7919:
I've removed the TODO but kept some of the commentary as it may be useful to know in the future.

manual/sphinx/developer_docs/data_types.rst
```diff
@@ -217,51 +217,201 @@ See the example ``examples/test-globalfield`` for more examples.
...
+be calculated::
+
+  Field3D f = ...;
+  Field3D g(0.0);
+  BOUT_FOR(i, f.getMesh()->getRegion3D("RGN_NOBNDRY")) {
+    g[i] = f[i.xp()] - f[i.xm()];
+  }
+
+The ``xp()`` function produces an offset of ``+1`` in ``X``, ``xm()``
+an offset of ``-1`` in the ``X`` direction. ``xpp()``
```
d7919: Actually these take an optional step size, which defaults to 1 (e.g. `xpp` is implemented as just `xp(2)`).

manual/sphinx/developer_docs/data_types.rst
```diff
@@ -217,51 +217,201 @@ See the example ``examples/test-globalfield`` for more examples.
...
+  BOUT_OMP(parallel for schedule(guided))
+  for (auto block = region.getBlocks().cbegin();
+       block < region.getBlocks().cend();
+       ++block)
+    for (auto index = block->first; index < block->second; ++index)
+
+The inner loop iterates over a contiguous range of indices, which
+enables it to be vectorised by GCC and Intel compilers.
+
+In order to both OpenMP parallelise, there must be enough blocks to
```
d7919: Probably don't need "both" here.

manual/sphinx/developer_docs/data_types.rst
```diff
@@ -217,51 +217,201 @@ See the example ``examples/test-globalfield`` for more examples.
...
+keep all threads busy. In order to vectorise, each of these blocks
+must be larger than the processor vector width, preferably several
+times larger. This can be tuned by setting the maximum block size, set
+in ``include/bout/region.hxx``::
+
+  #define MAXREGIONBLOCKSIZE 64
```
d7919: There's `mesh:maxregionblocksize` to change this at runtime -- the define just determines the default of this runtime parameter.

ZedThree:
I think this is now actually ready to go in.

David's caught me up on the meeting the other day -- OpenMP off by default still for now, but we should have a configure-time option for changing the default schedule. I'll do that in a separate PR

configure
```diff
@@ -2663,6 +2666,13 @@ else
   enable_pvode_openmp=no
 fi
 
+# Check whether --openmp-schedule was given.
+if test "${with_openmp_schedule+set}" = set; then :
+  enableval=$with_openmp_schedule;
```
d7919: Do we want to validate that the user has provided a valid option or wait for the compilation to fail?

I think there's some way to provide a default without having to do this sort of check, but I can't remember. @ZedThree may be able to comment.

configure
```diff
@@ -787,6 +787,7 @@ enable_backtrace
 enable_shared
```
d7919: I note you've changed `configure` but I think you actually need to change `configure.ac` and then regenerate `configure` from this -- @ZedThree what's the command to do this?

dschwoerer: `autoreconf`?
Changing `configure` directly will cause the changes to be lost the next time somebody runs autoreconf and adds the changes.

Should we add a test to make sure `configure` is generated from `configure.ac`?
Unfortunately the generated configure is slightly dependend on the version of autoreconf, making the test non-trivial ...

PR #1255, opened 2018-09-07
Morecommands

dschwoerer:
I would like to have more of the functions exposed as a simple script. As creating the script is rather repetitive work, I wrote a script that generates that, including the original documentation.
The script may not be nice, but it can be removed without any issues, if it is not working as expected.

I used that to create ` bin/bout-boutdata-restart-create` - which was the python function which I wanted to use ...

### DIFF
--- a/bin/bout-boutdata-restart-create
+++ b/bin/bout-boutdata-restart-create
@@ -1,0 +1,44 @@
+#!/usr/bin/env python3
+# PYTHON_ARGCOMPLETE_OK
+
+import argparse
+try:
+    import argcomplete
+except ImportError:
+    argcomplete=None
+from boutdata.restart import create
+
+parser = argparse.ArgumentParser("Create restart files from data (dmp) files.")
+parser.add_argument("path", type=str, default=...
+    ,help="Path to original restart files (default: \"data\")")
+parser.add_argument("--averagelast", type=int, default=...
+    ,help="Number of time points (counting from `final`, inclusive) to"
+     " average over (default is 1 i.e. just take last time-point)")
+parser.add_argument("--final", type=int, default=...
+    ,help="The last time point to use (default is last, -1)")
+parser.add_argument("--output", type=str, default=...
+    ,help="Path to write new restart files (default: current directory)")
+parser.add_argument("--informat", type=str, default=...
+    ,help="File extension of original files (default: \"nc\")")
+parser.add_argument("--outformat", type=str, default=...
+    ,help="File extension of new files (default: use the same as `informat`)")
+
+if argcomplete:
+    argcomplete.autocomplete(parser)
+
+args = parser.parse_args()
+
+# remove non-set optional arguments
+dic=args.__dict__
+
+if dic["output"] == ... and dic["path"] != ...:
+    dic["output"] = dic["path"]
+
+for k in dic:
+    if dic[k] == ...:
+        del dic[k]
+
+# Call the function create, using command line arguments
+create(**dic)
+
+
--- a/bin/bout-boutdata-restart-create
+++ b/bin/bout-boutdata-restart-create
@@ -6,7 +6,6 @@ try:
     import argcomplete
 except ImportError:
     argcomplete=None
-from boutdata.restart import create
 
 parser = argparse.ArgumentParser("Create restart files from data (dmp) files.")
 parser.add_argument("path", type=str, default=...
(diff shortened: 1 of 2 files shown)


ZedThree:
I think this is supposed to expose various functionality from the python library as an executable script?

If we're going to do this, maybe a more joined-up, comprehensive approach would be better? Rather than create a new script for each individual function, is there a way we expose more features through a unified interface?

For example, modules may have a `__main__.py` which can be executed via `python -m <module>`. Functions could be accessed via subcommands or similar.

bin/bout-boutdata-restart-create
```diff
@@ -0,0 +1,46 @@
...
+# PYTHON_ARGCOMPLETE_OK
+
+import argparse
+try:
+    import argcomplete
+except ImportError:
+    argcomplete=None
+
+parser = argparse.ArgumentParser("Create restart files from data (dmp) files.")
+parser.add_argument("path", type=str, default=...
```
ZedThree: The Ellipsis object, `...`, is for slicing. You should use `None` instead. Although these should really just use the default values from `restart.create`.

bin/bout-boutdata-restart-create
```diff
@@ -0,0 +1,46 @@
...
+    ,help="Path to original restart files (default: \"data\")")
```
ZedThree: Use the `default` argument to specify the default, rather than the help string. Also, don't escape quotes in nested quotes, just use the "other" one (single or double quotes)

bin/bout-boutdata-restart-create
```diff
@@ -0,0 +1,46 @@
...
+
+# remove non-set optional arguments
+dic=args.__dict__
+
+if dic["output"] == ... and dic["path"] != ...:
+    dic["output"] = dic["path"]
+
+for k in dic:
+    if dic[k] == ...:
+        del dic[k]
```
ZedThree: Set sensible defaults in the `parser.add_argument` and this isn't necessary. Similarly for the `dic = args.__dict__` above -- that just looks suspicious

johnomotani:
> For example, modules may have a __main__.py which can be executed via python -m <module>. Functions could be accessed via subcommands or similar.

On the other hand, don't we need executable scripts for the bash auto-completion functionality? I'd find that very very helpful.

ZedThree:
We've talked a little about pulling pylib out into a separate, stand-alone repo, and packaging it so it can be installed via PyPI. Certainly if we provided a `setup.py`, it would be pip-installable which would then generate executable scripts.

That method would also allow us to provide multiple entry points and executables.

bin/bout-boutdata-restart-create
dschwoerer: I found the defaults, which are stored in `create.__defaults__` - but they don't contain the name of the matching argument. Which way can I get them? I don't want to that manually ...

bin/bout-boutdata-restart-create
dschwoerer: But then I need to mix the quotes. And I need to read the string, checking which quotes are in there, and depending on that still need to escape ...
Is it worth all the trouble?

bin/bout-boutdata-restart-create
dschwoerer: Found it, will fix (link to remind myself) https://stackoverflow.com/questions/2677185/how-can-i-read-a-functions-signature-including-default-argument-values#2677263

dschwoerer:
Last time I used the `__main__`, [ben requested to put it in a separate file.](https://github.com/boutproject/BOUT-dev/pull/1220#issuecomment-415758325)
Further, providing several commands with argparse can be done, but that does not simplify the code.

I also thought having a separate repository would be useful ...
I am not sure how pypi works - can we get bash autocompletion?

bin/bout-boutdata-restart-create
dschwoerer: Resolved, I have found a way to extract the default value, and am now providing them.

bin/bout-boutdata-restart-create
dschwoerer: I am now checking what quotes are in the string, so here single quotes are used.

bin/bout-boutdata-restart-create
dschwoerer: I removed this part.
That changes the default output dir (which was previously the same as input), but as this is anyway the default of the function, I think this is fine.

PR #1256, opened 2018-09-07
Fix warnings with default configure

### LINKED ISSUE #1252: [next] build very noisy with gcc 8.1.1
Most are from `-Wnull-dereference`

See https://omadesala.alfa1024.info/~bout-test/2018-09-05_17:18/build.log

There are some others as well.

Good that `-Werror` was not included.

ZedThree:
Fixes #1252 

- Adds `getBase` and `isEnabled` as protected methods to `Output`, and makes `ConditionalOutput` a friend class. This simplifies ConditionalOutput a bit by removing the need for `base_is_cond` and `dynamic_cast`. Also adds explicit nullptr checks at at levels. Should be ok, as output is not a big performance hit
- Pulls the loop over the `region` in `checkData` that appears at `CHECK > 2` into a separate (detail) function. This avoids some really ugly workarounds to put `UNUSED` on `region` only at certain `CHECK` levels
- Uses a temporary reference in `BoutMesh::pack/unpack_data` to avoid a null dereference warning. I'm not really sure it "fixes" it so much as hides it from the compiler. It does have the advantage that it makes the code a bit more readable though.

The last remaining warning is due to this: https://github.com/boutproject/BOUT-dev/blob/master/src/mesh/boundary_standard.cxx#L88

Should we always do this check? It shouldn't cost anything (we're already doing a jump presumably).

### DIFF
--- a/include/output.hxx
+++ b/include/output.hxx
@@ -172,14 +172,14 @@ public:
   void print(const char *str, ...) override;
   void vprint(const char *str, va_list va) override {
     if (enabled) {
-      ASSERT0(base != nullptr);
+      ASSERT1(base != nullptr);
       base->vprint(str, va);
     }
   }
 
   /// Get the lowest-level Output object which is the base of this ConditionalOutput
   Output *getBase() override {
-    ASSERT0(base != nullptr);
+    ASSERT1(base != nullptr);
     return base->getBase();
   };
 
@@ -196,7 +196,7 @@ public:
 
   /// Check if output is enabled
   bool isEnabled() override {
-    ASSERT0(base != nullptr);
+    ASSERT1(base != nullptr);
     return enabled && base->isEnabled();
   };
 
--- a/include/output.hxx
+++ b/include/output.hxx
@@ -104,6 +104,11 @@ public:
 
   static Output *getInstance(); ///< Return pointer to instance
 
+protected:
+  friend class ConditionalOutput;
+  virtual Output *getBase() { return this; }
+  virtual bool isEnabled() { return true; }
+
 private:
   std::ofstream file;                 ///< Log file stream
   static const int BUFFER_LEN = 1024; ///< default length
@@ -129,7 +134,7 @@ public:
     if (enable)
       this->enable();
   };
-  bool isEnabled() { return false; }
+  bool isEnabled() override { return false; }
 };
 
 /// Layer on top of Output which passes through calls to write, print etc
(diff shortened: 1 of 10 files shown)


bshanahan:
12 as we we'd wsw2

include/output.hxx
```diff
@@ -140,21 +147,22 @@ public:
...
-      : base(base), enabled(base->enabled), base_is_cond(true) {};
+      : base(base), enabled(base->enabled) {};
 
   /// If enabled, writes a string using C printf formatting
   /// by calling base->vwrite
   /// This string is then sent to log file and stdout (on processor 0)
   void write(const char *str, ...) override;
   void vwrite(const char *str, va_list va) override {
     if (enabled) {
+      ASSERT0(base != nullptr);
```
dschwoerer: Should be ASSERT1 or ASSERT2, as this is never set to a nullptr.
In the Fields we avoid the checks, and in the case of the fields there is actually a fair chance that they have not allocated the data ...
I guess have more well defined definitions of what the levels mean would be nice.
I think level0 should be only some very basic checks, that are fast and also have rather nasty consequences - and nullptr dereference results in a _clean_ segfault ...

src/field/field2d.cxx
```diff
@@ -535,26 +535,36 @@ Field2D pow(BoutReal lhs, const Field2D &rhs, REGION rgn) {
...
-/// Check if the data is valid
-void checkData(const Field2D &f, REGION region) {
-  if(!f.isAllocated()) {
-    throw BoutException("Field2D: Operation on empty data\n");
-  }
-
+// Internal routine to avoid ugliness with interactions between CHECK
+// levels and UNUSED parameters
 #if CHECK > 2
+void checkData_is_finite_on_region(const Field2D &f, REGION region) {
```
bendudson: I'm not sure about this; possibly replaces one kind of ugliness with another. Pedantic point: probably should be `checkDataIsFiniteOnRegion`.

ZedThree: Good catch on the name, thanks!

Yes, I'm not very happy with it. I think the other alternatives were:

a)
```cpp
#if CHECK > 2
void checkData(const Field2d &f, REGION region)
#elif CHECK > 0
void checkData(const Field2d &f, REGION UNUSED(region))
#endif
{
```
b)
```cpp
void checkData(const Field2d &f, 
#if CHECK > 2
               REGION region)
#elif CHECK > 0
               REGION UNUSED(region))
#endif
{
```
Or of course, just don't warn on unused parameters (by default). It's not really helpful for end-users, but developers should really have it on.

dschwoerer: Maybe only not warn about unused parameters when compiled with checks?

PR #1261, opened 2018-09-12
Support for old install

dschwoerer:
CentOS ships with an older version of `install` `(install (GNU coreutils) 8.22)` where `-D` is differently defined:
```
  -D                  create all leading components of DEST except the last,
                        then copy SOURCE to DEST
```
whereas on more recent version, the definition is:
```
  -D                  create all leading components of DEST except the last,
                        or all components of --target-directory,
                        then copy SOURCE to DEST
```
As we require that the full path is created, this patch first creates the directories.
The `-D` flag is removed as the definition changed, and is not required anymore
The `-t` flag is not needed, and not listed in the various versions overview, thus I removed it, to avoid potential conflicts.

### DIFF
--- a/make.config.in
+++ b/make.config.in
@@ -154,22 +154,24 @@ install: libfast
 	$(PRE_INSTALL)     # Pre-install commands follow.
 
 	$(NORMAL_INSTALL)  # Normal commands follow.
-
-	$(INSTALL_DATA) -D include/*.hxx -t $(INSTALL_INCLUDE_PATH)
-	$(INSTALL_DATA) -D include/pvode/*.h -t $(INSTALL_INCLUDE_PATH)/pvode/
-	$(INSTALL_DATA) -D include/bout/*.hxx -t $(INSTALL_INCLUDE_PATH)/bout/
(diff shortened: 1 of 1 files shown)


bendudson:
I don't think this will work in general; `mkdir` only creates one directory, not several nested. If `DESTDIR` is `a/b/c` where `a` does not exist then it will usually fail.
I may be reading it wrong, but it looks from the descriptions as if `--target-directory` is empty then the meaning of `-D` does not change.

d7919:
`mkdir -p` should create child directories automatically, although I don't know if this is portable.

dschwoerer:
While it is called `$(MKDIR)`, it is actually set to mkdir -p, so creating several directories works with this version.

`-t` is the same as `--target-directory` - which we have been using.
I have tested a bit further, and the `-t` was needed, together with `-D` it created the directory. In older version this is not working. It is not documented in the man page that `-t directory` is allowed after the arguments, but it seems to work.

I slightly prefer the old version, as it is shorter, and adding/removing installed things does not require to change two parts ...

PR #1266, opened 2018-09-14
Coordinates location field metric

d7919:
Adds `Coordinates* fieldsCoordinate` to `Field` and provides `getCoordinates()` and `getCoordinates(CELL_LOC loc)` methods to get coordinates pointers. The first form gives the pointer for the coordinates object at the field's location. The second form gives the pointer at the requested location. The main advantage of introducing the second form is that it becomes easier to ensure that the coordinates objects are being returned from the same mesh that the field is on and hence this PR should also help with the push to remove references to the global mesh.

There are still some places where "raw" `mesh->coordinates(location)` calls are made (mostly in Laplacian routines) where the appropriate replacement was not immediately obvious.

### DIFF
--- a/include/bout/fv_ops.hxx
+++ b/include/bout/fv_ops.hxx
@@ -187,7 +187,7 @@ namespace FV {
     Field3D f = mesh->toFieldAligned(f_in);
     Field3D v = mesh->toFieldAligned(v_in);
 
-    Coordinates *coord = mesh->coordinates(f_in.getLocation());
+    Coordinates *coord = f_in.getCoordinates();
 
     Field3D result = 0.0;
     
@@ -343,7 +343,7 @@ namespace FV {
 
     CellEdges cellboundary;
     
-    Coordinates *coord = mesh->coordinates(n_in.getLocation());
+    Coordinates *coord = n_in.getCoordinates();
     
     if(v.covariant) {
       // Got a covariant vector instead
--- a/include/bout/operators_di.hxx
+++ b/include/bout/operators_di.hxx
@@ -17,7 +17,7 @@
 /// @param i  The point where the result is calculated
 ///
 inline BoutReal DDX_C2(const Field3D &f, const DataIterator &i) {
-  return (f[i.xp()] - f[i.xm()])/(2.*mesh->coordinates(f.getLocation())->dx[i]);
+  return (f[i.xp()] - f[i.xm()])/(2.*f.getCoordinates()->dx[i]);
 }
 
 /// \brief 2nd order central differencing in Y using yup/down fields
@@ -26,7 +26,7 @@ inline BoutReal DDX_C2(const Field3D &f, const DataIterator &i) {
 /// @param i  The point where the result is calculated
 /// 
 inline BoutReal DDY_C2(const Field3D &f, const DataIterator &i) {
-  return (f.yup()[i.yp()] - f.ydown()[i.ym()])/(2.*mesh->coordinates(f.getLocation())->dy[i]);
+  return (f.yup()[i.yp()] - f.ydown()[i.ym()])/(2.*f.getCoordinates()->dy[i]);
 }
 
 /// \brief 2nd order central differencing in Y assuming field aligned
@@ -35,7 +35,7 @@ inline BoutReal DDY_C2(const Field3D &f, const DataIterator &i) {
 /// @param i  The point where the result is calculated
 /// 
 inline BoutReal DDY_C2_FA(const Field3D &f, const DataIterator &i) {
-  return (f[i.yp()] - f[i.ym()])/(2.*mesh->coordinates(f.getLocation())->dy[i]);
+  return (f[i.yp()] - f[i.ym()])/(2.*f.getCoordinates()->dy[i]);
 }
 
 /// \brief 2nd order central differencing in Z 
@@ -44,7 +44,7 @@ inline BoutReal DDY_C2_FA(const Field3D &f, const DataIterator &i) {
 /// @param i  The point where the result is calculated
 /// 
 inline BoutReal DDZ_C2(const Field3D &f, const DataIterator &i) {
-  return (f[i.zp()] - f[i.zm()])/(2.*mesh->coordinates(f.getLocation())->dz);
+  return (f[i.zp()] - f[i.zm()])/(2.*f.getCoordinates()->dz);
 }
 
 
@@ -52,7 +52,7 @@ inline BoutReal DDZ_C2(const Field3D &f, const DataIterator &i) {
 /// d/dx( g^xx df/dx + g^xz df/dz) + d/dz( g^zz df/dz + g^xz df/dx)
 /// 
 BoutReal Delp2_C2(const Field3D &f, const DataIterator &i) {
-  Coordinates *metric = mesh->coordinates(f.getLocation());
+  Coordinates *metric = f.getCoordinates();
 
   //  o -- UZ -- o
   //  |          |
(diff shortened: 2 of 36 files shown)


d7919:
This should pretty much do what we want. There's one remaining issue that I'm not sure how to fix. The `fieldCoordinates` pointer is initialised to `nullptr` and this only gets set when `setLocation` is called with a new location. I've looked at setting the pointer appropriately in `setLocation` if it's `nullptr` but this causes problems (at least in the unit tests but probably more broadly) as we need to be able to instantiate fields in the default coordinates constructor -- these initially have no fieldCoordinates, which is fine, but towards the end of the constructor we do things like `abs` and `sqrt` on these fields, these routines in turn automatically call `setLocation` on the result and hence would try to do `fieldCoordinates = getMesh()->coordinates(CELL_CENTRE)`, which would lead to trying to construct the coordinates object again and this recursion would continue until a crash is achieved! 

I'm not sure how we can easily avoid this other than a substantial change to the coordinates creation process. If `fieldCoordinates` is a `nullptr` then the call to `getCoordinates` will fall back to `getMesh()->coordinates(getLocation())`, which is fine but might be slower than if we had a valid pointer. One possible solution would be to make `getCoordinates` store the `fieldCoordinates` pointer in this situation -- this should work OK as it means we can defer storing the pointer until we actually need the coordinates object to exist. Unfortunately it means `getCoordinates` can't be declared as const and hence would seemingly prevent the use of `getCoordinates` on fields passed into routines as const (e.g. all the derivatives). We could of course provide a const and non-const overload, with only the non-const one updating the pointer, but this might be unexpected behaviour (one overload does something the other doesn't)?

d7919:
Some the test jobs seem to fail, but not all of them. I don't seem to be able to produce this locally though so I'm a bit stumped.

bendudson:
Can Field have a constructor which takes a `Coordinate` pointer in addition to a `Mesh` pointer? This would allow the `Coordinate` to pass in `this`. Might that work?

d7919:
> Can Field have a constructor which takes a `Coordinate` pointer in addition to a `Mesh` pointer? This would allow the `Coordinate` to pass in `this`. Might that work?

This could probably work provided we copy `fieldCoordinates`. Digging a bit more I think the actual issue I'm seeing if something else:

> To reproduce need CHECK>0 and < 3 -- due to Field3DTest CreateCopyOnNullMesh. Essentially setLocation assumes the mesh is not nullptr as it tries to access members without checking it's nullptr -- really this is probably an issue with Field::getMesh which checks if fieldmesh is nullptr if it is then it returns a pointer to the global mesh without checking if global mesh is valid. Do we ever want to allow Fields to be created on a null mesh? If not then I would think we should check that mesh != nullptr in getMesh and if not then we should throw. That would mean this test (and others) are testing invalid behaviour and should be removed (it may also mean that some of the Field constructor code could be simplified).

d7919:
> Some of the test jobs seem to fail, but not all of them. I don't seem to be able to produce this locally though so I'm a bit stumped.. To reproduce need CHECK>0 and < 3 -- due to Field3DTest `CreateCopyOnNullMesh`. Essentially `setLocation` assumes the mesh is not `nullptr` as it tries to access members without checking it's `nullptr` -- really this is probably an issue with `Field::getMesh` which checks if `fieldmesh` is `nullptr` if it is then it returns a pointer to the global `mesh` _without_ checking if global `mesh` is valid. Do we ever want to allow Fields to be created on a null mesh? If not then I would think we should check that `mesh != nullptr` in `getMesh` and if not then we should throw. That would mean this test (and others) are testing invalid behaviour and should be removed (it may also mean that some of the Field constructor code could be simplified).

Removing these tests I still get fails when working with tests that construct fields using a copy constructor (currently the only constructor that explictly calls `setLocation`). Digging in with gdb it looks like this is because the unit tests use a `FakeMesh` object for which the `source` member is not set, and hence when we try to construct the default Coordinates object due to call to `setLocation` we end up dereferencing a null `source` pointer, leading to the error. This points to another potential unrelated issue/bug -- we never check that `source` is not a `nullptr` in `Mesh::get`, I don't think we make it easy to create a `Mesh` instance in which `source` is `nullptr` but it's clearly possible as this is what happens in the unit tests -- these `::get` routines should probably check `source != nullptr`, if it is `nullptr` then we have the option of throwing or acting like the requested variable wasn't found -- `return 1`. I think the later probably makes sense in `Mesh::get` and `Mesh::sourceHasVar`, whilst the former is sensible in `Mesh::readInts` (these seem to be the only places `Mesh` might use `source`).

d7919:
> Can Field have a constructor which takes a `Coordinate` pointer in addition to a `Mesh` pointer? This would allow the `Coordinate` constructor to pass in `this`. Might that work?

The downside of this approach is that it then makes it technically possible to produce a field that has a mesh pointer and a coordinates pointer that aren't consistent (e.g. coordinates has come from a different mesh).

johnomotani:
>> Can Field have a constructor which takes a Coordinate pointer in addition to a Mesh pointer? This would allow the Coordinate constructor to pass in this. Might that work?

> The downside of this approach is that it then makes it technically possible to produce a field that has a mesh pointer and a coordinates pointer that aren't consistent (e.g. coordinates has come from a different mesh).

The Coordinates object does know which mesh it belongs to though, so if we added a `getMesh()` method to Coordinates, we could check that they are consistent.

d7919:
> > > Can Field have a constructor which takes a Coordinate pointer in addition to a Mesh pointer? This would allow the Coordinate constructor to pass in this. Might that work?
> 
> > The downside of this approach is that it then makes it technically possible to produce a field that has a mesh pointer and a coordinates pointer that aren't consistent (e.g. coordinates has come from a different mesh).
> 
> The Coordinates object does know which mesh it belongs to though, so if we added a `getMesh()` method to Coordinates, we could check that they are consistent.

That's true.

I've run into one more issue with this -- I had the copy constructors doing `Field(f.fieldmesh, f.fieldCoordinates)` but this isn't possible for `Field3D::Field3D(Field2D &f)` as `fieldmesh` and `fieldCoordinates` are protected members of `Field`. The original code gets around this by using `f.getMesh()` instead of `f.fieldmesh`, but this won't work for the coordinates as `getCoordinates` has to ask the mesh for the default coordinates pointer if the field doesn't store one and hence potentially triggers the same recursive behaviour. We could instead make `getCoordinates` return `nullptr` if `fieldCoordinates` hasn't been set but I could see this leading to other issues!

d7919:
These changes are causing a lot of problems for the unit tests as the creation of the a field can now require a valid mesh that can create the default coordinates object. This is often not the case in most of the unit tests as they try to be as lightweight and isolated as possible.

The integrated tests all compile and run (altohg test-drift-instability-staggered currently fails one of the cases -- not clear if this is a bug or a bug fix.

johnomotani:
Maybe worth mentioning I usually see the first case of test-drift-instability-staggered fail on my desktop. It's always then passed on Travis though so I've not bothered to poke into why it fails.

d7919:
I've not thought through the full set of changes yet, but some of these issues might be avoided if we introduced a `copyLocation` method on fields for example. This could copy the location and the coordinates pointer from the passed field rather than trying to get the pointer from the mesh after storing the passed location.

d7919:
Yes I think the current setup is not going to work -- even being able to pass in the coordinates pointer in the default coordinates constructor doesn't help as it calls routines which take derivatives of fields and the derivative routines unconditionally call `setLocation` so we're always going to end up with this recursion if we try to ensure the coordinates pointer is set in the call to setLocation.

One workaround might be to provide a flag in the `mesh->coordinates` call that says don't try to make the coordinates object if not found.

d7919:
Replaced by #1271

PR #1271, opened 2018-09-19
Tidyup coordinates field fix

d7919:
Replaces #1266 

Adds Coordinates* fieldsCoordinate to Field and provides getCoordinates() and getCoordinates(CELL_LOC loc) methods to get coordinates pointers. The first form gives the pointer for the coordinates object at the field's location. The second form gives the pointer at the requested location. The main advantageAdds Coordinates* fieldsCoordinate to Field and provides getCoordinates() and getCoordinates(CELL_LOC loc) methods to get coordinates pointers. The first form gives the pointer for the coordinates object at the field's location. The second form gives the pointer at the requested location. The main advantage of introducing the second form is that it becomes easier to ensure that the coordinates objects are being returned from the same mesh that the field is on and hence this PR should also help with the push to remove references to the global mesh.

There are still some places where "raw" mesh->coordinates(location) calls are made (mostly in Laplacian routines) where the appropriate replacement was not immediately obvious. being returned from the same mesh that the field is on and hence this PR should also help with the push to remove references to the global mesh.

There are still some places where "raw" mesh->coordinates(location) calls are made (mostly in Laplacian routines) where the appropriate replacement was not immediately obvious.

Note this also brings in the recent changes to next.

### DIFF
--- a/include/field.hxx
+++ b/include/field.hxx
@@ -72,11 +72,11 @@ class Field {
   }
 
 #ifdef TRACK
-  std::string getName() const { return name; }
-  void setName(std::string s) { name = s; }
+  DEPRECATED(std::string getName() const) { return name; }
+  DEPRECATED(void setName(std::string s)) { name = s; }
 #else
-  std::string getName() const { return ""; }
-  void setName(std::string UNUSED(s)) {}
+  DEPRECATED(std::string getName()) const { return ""; }
+  DEPRECATED(void setName(std::string UNUSED(s))) {}
 #endif
   std::string name;
 
--- a/include/field.hxx
+++ b/include/field.hxx
@@ -132,7 +132,7 @@ class Field {
   virtual const IndexRange region(REGION rgn) const = 0;
  protected:
   Mesh * fieldmesh;
-  Coordinates * fieldCoordinates;
+  mutable Coordinates * fieldCoordinates = nullptr;
   /// Supplies an error method. Currently just prints and exits, but
   /// should do something more cunning...
   DEPRECATED(void error(const char *s, ...) const);
--- a/include/field.hxx
+++ b/include/field.hxx
@@ -110,6 +110,10 @@ class Field {
   /// Returns a pointer to the coordinates object at this field's
   /// location from the mesh this field is on.
   virtual Coordinates *getCoordinates() const;
+
+  /// Returns a pointer to the coordinates object at the requested
+  /// location from the mesh this field is on.
+  virtual Coordinates *getCoordinates(CELL_LOC loc) const;
   
   /*!
    * Return the number of nx points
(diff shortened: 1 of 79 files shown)


src/field/field.cxx
```diff
@@ -34,21 +34,42 @@
...
 #endif
 }
 
-Field::Field(Mesh *localmesh) : fieldmesh(localmesh) {
+Field::Field(Mesh *localmesh) : fieldmesh(localmesh), fieldCoordinates(nullptr) {
   if (fieldmesh == nullptr) {
     fieldmesh = mesh;
   }
+
+/// Note we would like to do `fieldCoordinates = getCoordinates();` here but can't
```
dschwoerer: Isn't that an implementation detail, and thus only have `//` comments?

src/field/field.cxx
```diff
@@ -34,21 +34,42 @@
...
 }
 
+Coordinates *Field::getCoordinates() const {
+  if (fieldCoordinates) {
+    return fieldCoordinates;    
+  } else {
+    /// Note we don't set fieldCoordinates here as the routine would become
+    /// non-const limiting our ability to call it in places like derivatives
+    /// where fields are passed as const.
+    fieldCoordinates = getMesh()->coordinates(getLocation());
```
dschwoerer: I am very confused, doesn't the comment is opposite of what is done?
That's why fieldCoordinates is mutable, so it can be changed in const context ...

src/field/field.cxx
d7919: Thanks, you're right.

src/field/field.cxx
d7919: Thanks, this should be removed (there due to history).

d7919:
I've changed the base to next as this includes a merge-commit from next. Hence this PR incorporates #1198 as well, should #1198 be merged first I think this PR should appear much smaller.

include/field.hxx
```diff
@@ -122,6 +128,7 @@ class Field {
   virtual const IndexRange region(REGION rgn) const = 0;
  protected:
   Mesh * fieldmesh;
+  Coordinates * fieldCoordinates;
```
johnomotani: Should this be snake_case: `field_coordinates`? Or possibly `fieldcoordinates` to be like `fieldmesh`?

src/field/field.cxx
```diff
@@ -17,7 +17,7 @@
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  * GNU Lesser General Public License for more details.
- *
+ *civil
```
johnomotani: typo?

src/field/field2d.cxx
```diff
@@ -78,6 +78,8 @@ Field2D::Field2D(const Field2D& f) : Field(f.fieldmesh), // The mesh containing
   checkData(f);
 #endif
 
+  location = f.location;
```
johnomotani: This is repeated at line 94

PR #1273, opened 2018-09-20
Add some syntactic sugar to Options

bendudson:
In order to read an integer from a section, the Options interface is:

```
int value;
Options::getRoot()->getSection("mysection")->get("myvalue", value, 42);
```

This commit adds:
* `root()` which is like `getRoot()` but returns a reference
* `[]` operator which is equivalent to `getSection`
* a templated `get` function which returns the value rather than
  putting the value into a given reference.

The above example becomes:

```
int value = Options::root()["mysection"].get("myvalue", 42);
```

### DIFF
--- a/include/options.hxx
+++ b/include/options.hxx
@@ -119,7 +119,11 @@ public:
 
   /// Get a pointer to the only root instance (singleton)
   static Options* getRoot();
-
+  
+  /// Alternative to getRoot.
+  /// Gets a reference to the only root instance
+  static Options& root() { return *getRoot(); }
+  
   /*!
    * Free all memory
    */
@@ -157,6 +161,13 @@ public:
   void get(const string &key, bool &val, bool def);
   void get(const string &key, string &val, const string &def);
 
+  template<typename T>
+  T get(const string &key, T def) {
+    T value;
+    get(key, value, def);
+    return value;
+  }
+  
   // This is a temporary replacement for 4-argument get
   // and will be removed in the next major release
   template<typename T, typename U>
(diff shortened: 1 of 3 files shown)


d7919:
Looks nice. I guess people might try to do something like `int value = Options::root()["mysection"]["myvalue"]` if they see this like a nested dictionary so just have to make sure it is clearly documented (as I think this won't raise an error, just create an empty section)? 

Is it possible to nest `[]`, so something like `Options::root()["Section"]["subSection"].get("var",42)`?

I'm not sure if this looks any nicer, but, do we need get or could we do something like `Options::root()["Section"]["subSection"]("var",42)`?

bendudson:
I think `int value = Options::root()["mysection"]["myvalue"]` would be a compile-time error when converting from `Options&` to `int`. Good point though this should be a test.

Yes, arbitrary nesting for accessing sub-sub-sections etc is supported.

I did try that `operator()` implementation too. It is quite nice, but if you need/want to specify the return type because it's incorrectly inferred from the default value then it becomes less nice:
```
Options::root()["Section"]["subSection"].operator()<BoutReal>("var",42);

Options::root()["Section"]["subSection"].get<BoutReal>("var",42)
```
Perhaps both could be supported.

I have a more drastic rewrite in progress: https://github.com/boutproject/BOUT-dev/tree/options-more-bling
This allows `int value = Options::root()["mysection"]["myvalue"]`, and `Options::root()["mysection"]["myvalue"] = 3`. It's backward compatible, but currently fails because `restart` is used as both a section and a value.

bendudson:
Replaced by PR #1277

PR #1277, opened 2018-09-21
Pour some sugar on Options

bendudson:
This is a backward-compatible change to the `Options` class. An `Options` object can now have a value and/or children; casting and assignment operators are used to provide a reasonably nice interface. The old interface is supported by wrapping the new interface, so could be removed at some point.

This is an alternative to PR #1273

Replaces 
```
int value;
Options::getRoot()->getSection("mysection")->get("myvalue", value, 42);
```
with
```
int value = Options::root()["mysection"]["myvalue"];
```
which will throw an exception if the option is not found, or to use a default:
```
int value = Options::root()["mysection"]["myvalue"].withDefault(42);
```
or to read as a specified type:
```
auto value = Options::root()["mysection"]["myvalue"].as<BoutReal>();
```
Setting values can be done using:
```
Options::root()["mysection"]["myvalue"] = 3.14;
```
which will use an empty `source`. To specify the source use `assign`:
```
Options::root()["mysection"]["myvalue"].assign(3.14, "source");
```
This will throw if overwriting a value from the same source. To force an assignment:
```
Options::root()["mysection"]["myvalue"].force(3.14, "source");
```

### DIFF
--- a/include/options.hxx
+++ b/include/options.hxx
@@ -310,13 +310,12 @@ public:
   Options *parent_instance;
   string full_name; // full path name for logging only
 
-  /// An Option object can be a section (node), a value (leaf), or empty (undecided)
-  enum class OptionType { empty, section, value };
+  /// An Option object can be a section and/or a value, or neither (empty)
 
-  OptionType type = OptionType::empty; ///< The type of this Option object
-  
-  std::string name; ///< If a section then this is the section name
+  bool is_section = false; ///< Is this Options object a section?
   std::map<string, Options*> children; ///< If a section then has children
+
+  bool is_value = false; ///< Is this Options object a value?
   OptionValue value; ///< If a value
   
   void _set(const string &val, const string &source, bool force);
--- a/include/options.hxx
+++ b/include/options.hxx
@@ -61,25 +61,36 @@ using std::string;
  * which can be used as a map.
  *
  *     Options options;
+ *     
+ *     // Set values
  *     options["key"] = 1.0;
  *
+ *     // Get values. Throws BoutException if not found
  *     int val = options["key"]; // Sets val to 1 
  *
- * It can be useful to record where values came from by adding a source label:
- * 
- *     options["key"].setTo(1.0, "code"); // Sets a value with source "code"
+ *     // Return as specified type. Throws BoutException if not found
+ *     BoutReal var = options["key"].as<BoutReal>();
  *
- * Equivalent interfaces are:
+ *     // A default value can be used if key is not found
+ *     BoutReal value = options["pi"].withDefault(3.14);
+ *    
+ *     // Assign value with source label. Throws if already has a value from same source
+ *     options["newkey"].assign(1.0, "some source");
+ *
+ *     // Force assign a new value
+ *     options["newkey"].force(2.0, "some source");
+ *
+ * A legacy interface is also supported:
  * 
- *     options.set("key", 1.0, "code"); // Sets a key
+ *     options.set("key", 1.0, "code"); // Sets a key from source "code"
  *
  *     int val;
- *     options.get("key", val, 0.0); // Sets val to 1.0
+ *     options.get("key", val, 0); // Sets val to 1, default to 0 if not found
  *
  * If a variable has not been set then the default value is used
  *
  *     int other;
- *     options.get("otherkey", other, 2.0); // Sets other to 2.0 because "otherkey" not found
+ *     options.get("otherkey", other, 2.0); // Sets other to 2 because "otherkey" not found
  *
  * Internally, all values are stored as strings, so conversion is performed silently:
  *
@@ -98,17 +109,26 @@ using std::string;
  * 
  *     Options &section = options["section"];
  * 
- * or
- *     Options *section = options.getSection("section");
+ * which can be nested:
+ *
+ *     options["section"]["subsection"]["value"] = 3;
  *
  * This always succeeds; if the section does not exist then it is created.
+ *
+ * The legacy interface uses pointers:
+ *
+ *     Options *section = options.getSection("section");
+ *
+ * e.g.
+ *     options->getSection("section")->getSection("subsection")->set("value", 3);
+ * 
  * Options also know about their parents:
  *
  *     Options &parent = section.parent();
  *     
  * or
  * 
- *     section->getParent() == &options // Pointer to options object
+ *     Options *parent = section->getParent();
  *
  * Root options object
  * -------------------
@@ -148,7 +168,13 @@ public:
   static void cleanup();
 
   /// Get a sub-section or value
-  /// Note: Using this makes this object a section
+  ///
+  /// Example:
+  ///
+  /// Options parent;
+  /// auto child  = parent["child"];
+  ///
+  /// parent is now a section.
   Options& operator[](const string &name);
 
   /// Assignment from any type T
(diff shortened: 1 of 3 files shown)


bendudson:
Damn... at the moment that's not supported, but it should be easy to add. It's a matter of making `withDefault` a template, rather than just an overloaded function. At the moment you could do something like `withDefault( static_cast<BoutReal>(2.0) )` but the choice of types is limited (`bool`, `int`, `BoutReal` or `string`). I will have a go

d7919:
Sorry!

src/sys/options.cxx
```diff
@@ -10,286 +10,303 @@
...
                             "'%s' - old value was '%s'.",
-                            source.c_str(), opt.value.c_str(), oldopt.value.c_str());
+                            source.c_str(), val.c_str(), value.value.c_str());
       }
     }
   }
-  options[key_l] = opt;
+  
+  value.value = val;
+  value.source = source;
```
ZedThree: We're making a copy here, so we could take `source` by value and move it

include/options.hxx
```diff
@@ -191,21 +370,57 @@ public:
...
-  std::map<string, OptionValue> options;
-  std::map<string, Options*> sections;
+  /// An Option object can be a section and/or a value, or neither (empty)
 
-  void _set(const string &key, const string &val, const string &source, bool force);
+  bool is_section = false; ///< Is this Options object a section?
+  std::map<string, Options*> children; ///< If a section then has children
+
+  bool is_value = false; ///< Is this Options object a value?
+  OptionValue value; ///< If a value
```
ZedThree: Does it make sense to give this a default empty value (`OptionValue value{}`)? We would need to give `OptionValue::used` a default value as well (presumably `false`).

include/options.hxx
```diff
@@ -111,45 +155,178 @@ public:
...
-    set(key,t,source,true);
+  /// Get a sub-section or value
+  ///
+  /// Example:
+  ///
+  /// Options parent;
+  /// auto child  = parent["child"];
+  ///
+  /// parent is now a section.
+  Options& operator[](const string &name);
```
ZedThree: Do we need a `const` version as well? If implemented, it would need to `throw` if a section/value didn't exist, instead of creating a new subsection. We'd probably also want a `const` version of `getSection` as well in that case.

---

A possible downside to this implementation is that this:

```cpp
int value = options["sub-section"];
```

is a runtime error, rather than a compile-time error, but I'm not sure there's much to be done about that.

ZedThree:
This is currently failing due to `test-code-style` complaining about `Options::operator=<BoutReal>(const BoutReal &inputvalue)` passing a `const BoutReal` by reference. Unfortunately, this is a specialisation of the `operator=(const T &inputvalue)` template, so it has to be exactly `const BoutReal&`.

I'm not really sure not passing this test is a "failure" -- maybe we should drop it?

bendudson:
Quick question: Should we try using `std::optional` to replace the `is_section` and `is_value` logic. Not essential, and just an internal implementation detail, but might be nice. It's in C++17, but some header-only implementations for C++11:
https://github.com/TartanLlama/optional
or
https://github.com/martinmoene/optional-lite

bendudson:
Currently I've just changed `operator=` to take argument by value, so BoutReal specialisation is possible; will this do terrible things?

ZedThree:
That's not completely terrible, as it reads like we are taking our own copy, which we kind of are. But returning a reference to it is a little problematic. Although, should it return a `T`, rather than an `Options`?

ZedThree:
There's a fun bug here that Clang catches but GCC doesn't. Basically in

```cpp
options["foo"]
```

the conversion operator means the compiler can't distinguish
between the builtin

```cpp
operator[](long, const char*)   // as in the weird `1["foo"]` vs `"foo"[1]`
```

and our

```cpp
operator[](Options, const char*)  // `this` is now convertible to `std::ptrdiff_t`
```

See https://stackoverflow.com/a/8917084/2043465 for a more in-depth
explanation.

Simple fix is to overload `operator[]` to take `const char*`:
```diff
   Options& operator[](const string &name);
+  // Prevent ambiguous overload resolution due to operator T() below
+  Options& operator[](const char *name) { return (*this)[std::string(name)]; }
```

bendudson:
Note: This (making member functions templates) doesn't seem to affect `boutcore`; setting and getting options seems to work as in `next`. On a related note, some unit tests for the python modules would be useful.

manual/sphinx/user_docs/bout_options.rst
```diff
@@ -386,25 +386,98 @@ Implementation
...
+
+    options["mysection"]["myswitch"] = true;
+
+To get options, they can be assigned to a variable::
+
+    int nout = options["nout"];
+
+If the option is not found then a ``BoutException`` will be thrown. A
+default value can be given, which will be used if the option has not
+been set::
```
ZedThree: Should we mention this value is cached (although not if `options` is `const`)?

manual/sphinx/user_docs/bout_options.rst
```diff
@@ -386,25 +386,98 @@ Implementation
...
+
+It is common for BOUT++ models to read in many settings which have the
+same variable name as option setting (e.g. "nout" here). A convenient
+macro reads options into an already-defined variable::
+
+    int nout;
+    OPTION(options, nout, 1);
+
+where the first argument is a section, second argument is the variable
+whose name will also be used as the option string.
```
ZedThree: The third argument is used as a default value

include/options.hxx
```diff
@@ -105,76 +150,255 @@ using std::string;
 class Options {
 public:
   /// Constructor. This is called to create the root object
-  Options();
+  Options() : parent_instance(nullptr) {}
```
ZedThree: `parent_instance` has a default value, so we can skip setting it in the ctor. But it doesn't do any harm

include/options.hxx
```diff
@@ -105,76 +150,255 @@ using std::string;
...
+  /// Assign a value to the option.
+  /// This will throw an exception if already has a value
+  ///
+  /// Example:
+  ///
+  /// Options option;
+  /// option["test"].assign(42, "some source");
+  ///
+  template<typename T>
+  void assign(T val, const std::string &source="") {
```
ZedThree: I think we could take `source` by value here, as `_set` is making a copy anyway

include/options.hxx
```diff
@@ -105,76 +150,255 @@ using std::string;
...
-  void set(const string &key, const int &val, const string &source="", bool force=false);
-  void set(const string &key, BoutReal val, const string &source="", bool force=false);
-  void set(const string &key, const bool &val, const string &source="", bool force=false);
-  void set(const string &key, const char * &val, const string &source="", bool force=false) {
-    _set(key,val,source,force);
+  
+  /// Force to a value
+  /// Overwrites any existing setting
+  template<typename T>
+  void force(T val, const string &source = "") {
```
ZedThree: Likewise, `source` could be a copy

tests/unit/sys/test_options.cxx
```diff
@@ -368,3 +371,108 @@ TEST_F(OptionsTest, SetSameOptionTwice) {
...
+  ASSERT_TRUE(options["sub-section"]["int_key"].isSet());
+  
+  int value = options["sub-section"]["int_key"].withDefault(99);
+  EXPECT_EQ(value, 42);
+}
+
+TEST_F(OptionsTest, NewIsSetDefault) {
+  Options options;
+  ASSERT_FALSE(options.isSet());
+  int value = options.withDefault(42);
```
ZedThree: `value` isn't used -- could be dropped

PR #1278, opened 2018-09-21
Add location checking for generated field ops involving Field2Ds

d7919:
Field2D now has a location so we should check that this is consistent with the other fields that it interacts with. This currently only adds checks in the generated field operations but can be expanded to more places.

Unfortunately (or fortunately) this doesn't pass the tests at the moment as it appears some of the tests use inconsistent field locations.

### DIFF
--- a/src/field/gen_fieldops.jinja
+++ b/src/field/gen_fieldops.jinja
@@ -1,6 +1,6 @@
 // Provide the C++ wrapper for {{operator_name}} of {{lhs}} and {{rhs}}
 {{out}} operator{{operator}}(const {{lhs.passByReference}}, const {{rhs.passByReference}}) {
-  {% if lhs == rhs == "Field3D" %}
+  {% if ((lhs in ["Field3D", "Field2D"]) and (rhs in ["Field3D", "Field2D"])) %}
 #if CHECK > 0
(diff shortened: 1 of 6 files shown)


d7919:
Also makes sure Field2D results get their location set as well.

d7919:
It seems like this reimplements 31b55f57cf8a1af54b968203b787fcd48eea58bb by @dschwoerer which was missed from the coordinates_location alternative to #1060

d7919:
Changes in #1284 may be enough to fix this.

d7919:
Two most recent commits plus #1284 should be enough to get this branch working and passing the tests.

d7919:
Some fixes now included to ensure fields are returned with a correct location when using shortcuts (i.e. for DDZ on Field2D, DDX when nx=1 etc.).

The tweak in 4b37404 may want to be a little more widespread to ensure coordinates operations are done consistently.

d7919:
Should the tweak 4b37404 actually use `outloc` rather than `f.getLocation()`? In fact I suspect it might be that we want to check `f.getCoordinates() == this` or similar? (`outloc` unlikely correct option as it defaults to `CELL_DEFAULT` which is not a valid field location).

d7919:
#1286 demonstrates that this branch should work once 1284 is in next.

PR #1285, opened 2018-09-25
Add MACRO_FOR_EACH for variadic macros

bendudson:
Rather than needing a separate macro for each number of arguments e.g. `SOLVE_FOR(a)`, `SOLVE_FOR2(a,b)`, `SOLVE_FOR3(a,b,c)` etc. this provides a way for a macro to expand a variable number of arguments. Currently limited to 10, but could be expanded quite easily if needed.

The definition of `SOLVE_FOR` becomes:

```
#define SOLVE_FOR1(var) solver->add(var, #var);

#define SOLVE_FOR(...)  \
  MACRO_FOR_EACH(SOLVE_FOR1, __VA_ARGS__)
```

For now the other macros are kept for backward compatibility.

Modest amounts of black magic are needed, but as far as I know all standard, no compiler extensions.

### DIFF
--- a/include/bout/macro_for_each.hxx
+++ b/include/bout/macro_for_each.hxx
@@ -55,7 +55,13 @@
 ///
 ///    test(a) test(b) test(c)
 ///
-/// Note that no ; is inserted after each expansion
+/// Notes:
+///
+///   - No semicolon is inserted after each expansion
+///   - No braces are put around the expansion. These
+///     should usually be added in the top-level macro
+///     to avoid surprising results.
+/// 
 #define MACRO_FOR_EACH(mac, ...)                                        \
   _GET_FOR_EACH_EXPANSION(__VA_ARGS__,                                  \
                           _me_10, _me_9, _me_8, _me_7, _me_6, _me_5,    \
@@ -74,7 +80,13 @@
 ///
 ///    test(a); test(b); test(c);
 ///
-/// Note that a ; is inserted after each expansion
+/// Notes:
+///
+///   - A ; is inserted after each expansion
+///   - No braces are put around the expansion. These
+///     should usually be added in the top-level macro
+///     to avoid surprising results.
+///
 #define MACRO_FOR_EACH_FN(fn, ...)                                      \
   _GET_FOR_EACH_EXPANSION(__VA_ARGS__,                                  \
                           _fe_10, _fe_9, _fe_8, _fe_7, _fe_6, _fe_5,    \
--- a/include/bout/macro_for_each.hxx
+++ b/include/bout/macro_for_each.hxx
@@ -1,0 +1,84 @@
+
+#ifndef __MACRO_FOR_EACH_H__
+#define __MACRO_FOR_EACH_H__
+
+// Provides a macro MACRO_FOR_EACH which applies a
+// macro to each argument in a VA_ARGS list
+// 
+// Based on this blog post by Daniel Hardman:
+//    https://codecraft.co/2014/11/25/variadic-macros-tricks/
+// This answer on StackOverflow:
+//    https://stackoverflow.com/questions/11761703/overloading-macro-on-number-of-arguments/11763277
+// See also: 
+//    https://github.com/cormacc/va_args_iterators
+//
+
+
+/// _me_x set of macros expand a number of arguments without ';' between them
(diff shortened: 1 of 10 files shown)


include/bout/macro_for_each.hxx
```diff
@@ -0,0 +1,84 @@
...
+/// of the following arguments.
+/// Currently supports up to 10 arguments.
+///
+/// Example:
+///
+///    MACRO_FOR_EACH(test, a, b, c)
+///
+/// expands to
+///
+///    test(a) test(b) test(c)
```
d7919: I guess this form is most useful when `test` is actually another macro that itself includes `;`? Are there any other use cases?

bendudson: The main use for this (and the reason I went down this rabbit hole) was to declare lists of variables. In SD1D the reactions are integrated over the cells using Simpson's rule, so each rate is calculated three times, for the left, centre and right side of each cell:
https://github.com/boutproject/SD1D/blob/master/sd1d.cxx#L971
I'd like to add many more reactions, and don't want to have to repeat each one three times. I've improved this a bit:
https://github.com/boutproject/SD1D/blob/new-multifluid/reaction_elastic.cxx#L57
but there's still a lot of boilerplate code which could (perhaps) be removed. What I'd like is to be able to write:

```
INTEGRATE(i, weight, Ni, Nn, Ti, Tn, Vi, Vn) {
    BoutReal R = a0 * Ni * Nn * Cs0 * sqrt((16. / PI) * Ti) * Nnorm / Omega_ci;
    Fel[i] += weight * (Vi - Vn) * R;
    Eel[i] += (3. / 2) * weight * (Ti - Tn) * R;
}
```
which would expand to something like:
```
for (auto i : region(RGN_NOBNDRY)) {   /// << Maybe have region as argument
    QuadValue Ti(Ti, i), Ni(Ni, i), Vi(Vi, i), ... /// << Here MACRO_FOR_EACH
    for (int j : QR.indices) {   /// << Some kind of gensym for index j
        weight = QR[j];
        BoutReal Ti = Ti[j], Ni = Ni[j], ... /// << MACRO_FOR_EACH
        // User code
    }
}
```
so inside the expression the Field2D or Field3D variables in the arguments to `INTEGRATE` are declared as `BoutReal`s, shadowing original variables. I'm not sure if this would be great or terrible. Perhaps both.

bendudson: For the kind of (mis-)feature this enables, see the updated SD1D reaction rate integration code: https://github.com/boutproject/SD1D/blob/new-multifluid/reaction_elastic.cxx#L47

I will admit that the language abuse needed is quite appalling:
https://github.com/boutproject/SD1D/blob/new-multifluid/reaction.hxx#L117

This uses a trick from https://stackoverflow.com/questions/866012/is-there-a-way-to-define-variables-of-two-different-types-in-a-for-loop-initiali in which `for` loops are used to define variables in a new scope. This enables two nested loops with a bunch of variables introduced in both to be wrapped up into what looks like a single loop, but is actually 19 nested loops and two `if...else` clauses.

include/bout/physicsmodel.hxx
```diff
@@ -325,7 +325,7 @@ private:
   }
 
 /// Macro to replace solver->add, passing variable name
-#define SOLVE_FOR(var) solver->add(var, #var)
+#define SOLVE_FOR1(var) solver->add(var, #var);
 #define SOLVE_FOR2(var1, var2) { \
```
d7919: Do you want to deprecate these specific count versions (for count >1) or do you expect we will keep these?

tests/unit/include/bout/test_macro_for_each.cxx
```diff
@@ -0,0 +1,86 @@
...
+}
+
+// No braces are put around the expansion
+// This is needed in some cases, but can
+// lead to unexpected behaviour.
+TEST(MacroForEachTest, NoBraces) {
+  int a = 0;
+
+  // Only the first expansion is disabled
+  if(false)
```
d7919: Nice!

ZedThree: Maybe a stupid question, my brain is not fully in gear -- if this conditional gets optimised away, that's still proof that the macro is working, right?

PR #1287, opened 2018-09-27
Remove mergeYupYdown()

johnomotani:
`Field3D::mergeYupYdown()` was used to set yup/ydown to point to the field itself. Was used by ParallelTransformIdentity. It is unnecessary because there are field-aligned versions of all the operators, which can be used when hasYupYdown()==false.

Removing the method simplifies several conditionals, which can now be just 'var.hasYupYdown()' where previously they also had to check 'var.yup() != var' and 'var.ydown() != var'.

Also rename 'splitYupYdown()' to 'createYupYdown()' since it no longer makes sense to contrast 'split' with 'merge'.

### DIFF
--- a/include/bout/paralleltransform.hxx
+++ b/include/bout/paralleltransform.hxx
@@ -56,10 +56,9 @@ public:
 class ParallelTransformIdentity : public ParallelTransform {
 public:
   /*!
-   * Merges the yup and ydown() fields of f, so that
-   * f.yup() = f.ydown() = f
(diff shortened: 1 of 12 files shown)


ZedThree:
This is backwards incompatible, so probably not going to get in just yet, but plausibly soon.

d7919:
Presumably if we wish to remove things in the next version we should deprecate them in v4.2 so you may want to explore deprecating this now so that you're able to remove for v4.3?

johnomotani:
Closing: see discussion in #1321.

PR #1290, opened 2018-09-28
Use templates in index_derivs.cxx

johnomotani:


johnomotani:
Oops, created this before writing the description. Sorry!

Since location was implemented fully for `Field2D`, the `Field2D` and `Field3D` versions of derivatives in index_derivs.cxx are basically identical. This PR uses templates to remove this duplication.

The set of changes in af8a133 seem like a straightforward win to me: they don't require any changes elsewhere in the code.

Templating some of the y-derivatives (4f61165) required `Field2D` to have yup/ydown methods. `Field2D::yup()` and `Field2D::ydown()` just return a reference to the field itself. This seems sensible for ShiftedMetric, but possibly not for FCI? I guess FCI probably shouldn't use Field2D much anyway, although `test-fci-slab` does take derivatives of `Field2D` metric components in `geometry()`; I don't know if these will be correct?

The changes to Field2D caused a bit of complication. I had to move the call to `geometry()` out of the Coordinates constructor, because Field2D derivatives now need a ParallelTransform. Calling `Coordinates()` then `setParallelTransform()` then `geometry()` in `Mesh::coordinates()` where the `Coordinates` objects are created gets around this though.
The Field2D derivatives do need #1287 to get into the right branch though: #1287 removes the conditionals like `((&v.yup() != &v) || (&v.ydown() != &v))` which is needed since Field2D returns itself for yup and ydown, so any derivative with a Field2D would fall through to the to/fromFieldAligned branch otherwise.

Finally, in 68fa467 I found some template goodness that lets templated functions like `indexVDDX` specialize to mixed Field2D/Field3D arguments, like `indexVDDX(Field2D, Field3D)` without converting `Field2D`->`Field3D` and with the return type being `Field2D` if both arguments are `Field2D` and `Field3D` otherwise. This uses the template `promoteField<T1,T2>` defined in `field3d.hxx` to get the right return type.

johnomotani:
Just to note: this PR will need re-jigging for changes in #1302, #1306 before we can merge.

johnomotani:
#1384 makes this PR redundant.

PR #1291, opened 2018-09-28
Change Solver::add to take std::string

bendudson:
`Solver::add` and `Solver::constrain` previously took variable names as `const char*`. This made generating variable names quite clunky e.g. `(d.name+"_x").c_str()`. Now changed to `const std::string&`.

Some reformatting, particularly of `SlepcSolver`, and adding `override` to `SlepcSolver` definitions.

### DIFF
--- a/include/bout/solver.hxx
+++ b/include/bout/solver.hxx
@@ -214,10 +214,10 @@ class Solver {
    * Add a variable to be solved. This must be done
    * in the initialisation stage, before the simulation starts.
    */ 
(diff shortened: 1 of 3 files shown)


ZedThree:
These all "consume" `name` by copying into `d.name`, so we could instead take `name` by value. This has two nice features:

1. taking a copy in the function signature lets the compiler generate the most efficient code (not much of a concern here, admittedly)
2. the signature now documents that the lifetime of the argument, i.e. don't need to worry about possibly dangling references

PR #1293, opened 2018-10-03
Use input field/vector location for CELL_DEFAULT in vecops

ZedThree:
Some operations used `CELL_DEFAULT` to mean `CELL_CENTRE`, others used it to mean use the input field/vector location, others didn't handle it at all. Now all vector operations use it to mean use the input field/vector location.


src/field/vecops.cxx
```diff
@@ -34,69 +34,85 @@
...
-const Vector3D Grad(const Field3D &f, 
-                    CELL_LOC outloc_x, CELL_LOC outloc_y, CELL_LOC outloc_z) {
+const Vector3D Grad(const Field3D &f, CELL_LOC outloc_x, CELL_LOC outloc_y,
+                    CELL_LOC outloc_z) {
   Vector3D result(f.getMesh());
 
   TRACE("Grad( Field3D )");
 
-  if(outloc_x == CELL_DEFAULT)
+  if (outloc_x == CELL_DEFAULT) {
```
d7919: I think this might be dangerous, currently assume that either all components are at the same location or all are staggered in direction of component. Here we're allowing some to be set from input field and others to set from argument.

d7919: I realise you're not changing this but thought I should flag!

src/field/vecops.cxx
```diff
@@ -34,69 +34,85 @@
...
 
-  Coordinates* metric_x = mesh->coordinates(outloc_x);
-  Coordinates* metric_z = mesh->coordinates(outloc_z);
+  Coordinates *metric_x = mesh->coordinates(outloc_x);
+  Coordinates *metric_z = mesh->coordinates(outloc_z);
 
-  result.x = DDX(f, outloc_x) - metric_x->g_12*DDY(f, outloc_x) / SQ(metric_x->J * metric_x->Bxy);
+  result.x = DDX(f, outloc_x) -
+             metric_x->g_12 * DDY(f, outloc_x) / SQ(metric_x->J * metric_x->Bxy);
   result.y = 0.0;
```
d7919: Note that `result.y` doesn't get a location set meaning it's centred unlike the other two fields. This feels odd.

src/field/vecops.cxx
```diff
@@ -110,6 +126,10 @@ const Field2D Div(const Vector2D &v, CELL_LOC outloc) {
   Mesh *localmesh = v.x.getMesh();
   Field2D result(localmesh);
 
+  if (outloc == CELL_DEFAULT) {
+    outloc = v.getLocation();
```
d7919: Note this could be a vector only location (VSHIFT) which should be invalid for getting/setting a field location.

src/field/vecops.cxx
ZedThree: Ah, I see! Yes, maybe we just shouldn't expose this as part of the public API

d7919: Yup, maybe with an ASSERT for further safety?

src/field/vecops.cxx
ZedThree: Good point -- is the "correct" output location well defined in that case? Maybe `CELL_CENTRE`?

src/field/vecops.cxx
ZedThree: True. Should it use `outloc_y` similarly to the other two locations?

ZedThree:
Sounds like `CELL_VSHIFT` may be generally a little bit broken, or at least suspicious.

This PR is the minimal set of changes to make vecops not completely broken. This is blocking for v4.2, as elm-pb won't run without it, so I'd like to get this in and try and fix vecops properly in either 4.2.1 or 4.3

src/field/vecops.cxx
```diff
@@ -34,69 +34,85 @@
...
-const Vector2D Grad(const Field2D &f, CELL_LOC UNUSED(outloc)) {
+const Vector2D Grad(const Field2D &f, CELL_LOC outloc) {
   Vector2D result(f.getMesh());
 
   TRACE("Grad( Field2D )");
   
-  result.x = DDX(f);
-  result.y = DDY(f);
-  result.z = DDZ(f);
+  if (outloc == CELL_DEFAULT) {
```
d7919: Can we just pass `CELL_DEFAULT` through to the derivative operators? Ideally the interpretation of `CELL_DEFAULT` can be minimised to just the lowest level where it's actually needed.

src/field/vecops.cxx
```diff
@@ -34,69 +34,85 @@
...
   Vector3D result(f.getMesh());
 
   TRACE("Grad( Field3D )");
 
-  if(outloc_x == CELL_DEFAULT)
+  if (outloc_x == CELL_DEFAULT) {
     outloc_x = f.getLocation();
-  if(outloc_y == CELL_DEFAULT)
+  }
+  if (outloc_y == CELL_DEFAULT) {
```
d7919: As above, can we just pass this through?

src/field/vecops.cxx
```diff
@@ -34,69 +34,85 @@
...
 }
 
 const Vector3D Grad(const Field3D &f, CELL_LOC outloc) {
-  if(outloc == CELL_VSHIFT)
+  if (outloc == CELL_VSHIFT) {
     return Grad(f, CELL_XLOW, CELL_YLOW, CELL_ZLOW);
-  
+  }
+
+  if (outloc == CELL_DEFAULT) {
```
d7919: As with other can we just pass this through without a check?

src/field/vecops.cxx
d7919: Yes I think that's correct.

src/field/vecops.cxx
d7919: I'm not sure, @bendudson might have a better idea of what this should do in this case.

d7919: Also, as above we might be able to just pass this through to the derivative routines to decide what to do. We can then take the location of the returned result to decide which coordinates location we want.

johnomotani: I think it has to be `CELL_CENTRE` :+1:

ZedThree:
`CELL_DEFAULT` means "do the right thing, if unambiguous; otherwise, throw", where "do the right thing" covers two input fields at the same location or exactly one staggered and one cell centred. This should go into the documentation for `CELL_DEFAULT`

d7919:
A further issue here -- routines like `Grad` which return a vector don't set the vector's location (particularly tricky for the three location argument versions). This can also impact later routines which try to get the input vector location as the vector location may no longer be set appropriately.

d7919:
#1308 (into this branch) attempts to fix some of these issues.

src/field/vecops.cxx
ZedThree: This is ok now. `CELL_DEFAULT` is passed down to `DD?`, and is then used to set the location of the result to the location of `result.x`

src/field/vecops.cxx
ZedThree: Now handled by `Vector3D::setLocation`

ZedThree:
I'm happy with this. @johnomotani you're basically in charge of staggering code at this point, do you want to sign off on this?

include/vecops.hxx
```diff
@@ -82,23 +104,29 @@ const Field3D Div(const Vector3D &v, const Field3D &f);
...
-const Vector3D Curl(const Vector3D &v, 
-                    CELL_LOC outloc_x, CELL_LOC outloc_y, CELL_LOC outloc_z);
+const Vector3D DEPRECATED(Curl(const Vector3D &v, 
+			       CELL_LOC outloc_x, CELL_LOC outloc_y, CELL_LOC outloc_z));
 
 // Upwinding routines
 
 /// Advection of a scalar field \p f by a velocity vector \p v
+///
+/// Both vectors must be at the same location, which cannot be CELL_VSHIFT
```
johnomotani: This should say 'The vector and the field must...'.

src/field/vecops.cxx
```diff
@@ -120,28 +160,31 @@ const Field2D Div(const Vector2D &v, CELL_LOC outloc) {
...
   vcn.toContravariant();
-  
-  result = DDX(metric->J*vcn.x, outloc);
-  result += DDY(metric->J*vcn.y, outloc);
-  result += DDZ(metric->J*vcn.z, outloc);
+
+  result = DDX(metric->J * vcn.x, outloc);
+  result += DDY(metric->J * vcn.y, outloc);
+  result += DDZ(metric->J * vcn.z, outloc);
   result /= metric->J;
```
johnomotani: We need to set the location of J here (note we do allow `v.getLocation==CELL_VSHIFT`). I think we need:
```
  result = DDX(vcn.x.getCoordinates()->J * vcn.x, outloc);
  result += DDY(vcn.y.getCoordinates()->J * vcn.y, outloc);
  result += DDZ(vcn.z.getCoordinates()->J * vcn.z, outloc);
```

src/field/vecops.cxx
```diff
@@ -154,53 +197,60 @@ const Field3D Div(const Vector3D &v, CELL_LOC outloc) {
...
   Vector2D vcn = v;
   vcn.toContravariant();
 
   Field2D result(localmesh);
-  result = FDDX(metric->J*vcn.x, f, outloc);
-  result += FDDY(metric->J*vcn.y, f, outloc);
-  result += FDDZ(metric->J*vcn.z, f, outloc);
+  result = FDDX(metric->J * vcn.x, f, outloc);
+  result += FDDY(metric->J * vcn.y, f, outloc);
+  result += FDDZ(metric->J * vcn.z, f, outloc);
```
johnomotani: locations of `J` again.

src/field/vecops.cxx
```diff
@@ -154,53 +197,60 @@ const Field3D Div(const Vector3D &v, CELL_LOC outloc) {
...
   vcn.toContravariant();
-  
-  result = FDDX(metric->J*vcn.x, f, outloc, method);
-  result += FDDY(metric->J*vcn.y, f, outloc, method);
-  result += FDDZ(metric->J*vcn.z, f, outloc, method);
+
+  Field3D result(localmesh);
+  result = FDDX(metric->J * vcn.x, f, outloc, method);
+  result += FDDY(metric->J * vcn.y, f, outloc, method);
+  result += FDDZ(metric->J * vcn.z, f, outloc, method);
```
johnomotani: and again

src/field/vecops.cxx
```diff
@@ -212,70 +262,89 @@ const Vector2D Curl(const Vector2D &v, CELL_LOC outloc) {
...
-  result.x = (DDY(vco.z, outloc_x) - DDZ(vco.y, outloc_x))/localmesh->coordinates(outloc_x)->J;
-  result.y = (DDZ(vco.x, outloc_y) - DDX(vco.z, outloc_y))/localmesh->coordinates(outloc_y)->J;
-  result.z = (DDX(vco.y, outloc_z) - DDY(vco.x, outloc_z))/metric_z->J;
+  result.x = (DDY(vco.z, outloc) - DDZ(vco.y, outloc)) / metric->J;
+  result.y = (DDZ(vco.x, outloc) - DDX(vco.z, outloc)) / metric->J;
+  result.z = (DDX(vco.y, outloc) - DDY(vco.x, outloc)) / metric->J;
 
   // Coordinate torsion
-  result.z -= metric_z->ShiftTorsion*vco.z / metric_z->J;
+  result.z -= metric->ShiftTorsion * vco.z / metric->J;
```
johnomotani: I think we have to have `vco.z.getLocation()==outloc` here, because otherwise one of the derivatives above would throw, but it's not asserted in this function. I think we should either assert than `v.getLocation()==outloc` or put `interp_to(vco.z, outloc)` here.

src/field/vecops.cxx
ZedThree: I am coming around to thinking that if `outloc` is `CELL_DEFAULT` and `v` is at `CELL_VSHIFT`, then the result location should be `CELL_CENTRE`, as that's the only location that we currently support, at least for scalar-result operations.

On the other hand, for vector-result operations, you might expect the result to be at the same location as the input, rather than `CENTRE`, which is not possible for `VSHIFT` and in that case it is definitely better to fail.

Is it better to be consistent across vector operations, and require the user to be explicit? Or better to do "the right thing"?

src/field/vecops.cxx
ZedThree: I'm not sure I follow -- which combinations of input and output locations do we support here? Is it just the usual one staggered and one not? If so, then one of the derivatives here should catch problems. If they have to be at the same location, we should not support `outloc` at all.

src/field/vecops.cxx
johnomotani: @ZedThree I don't have a strong opinion: I suspect there is nobody using CELL_VSHIFT anyway. For what it's worth, my first thought was that `Div` on a CELL_VSHIFT vector would default to a result at CELL_CENTRE...

ZedThree:
@johnomotani That should be everything fixed now!

PR #1302, opened 2018-10-08
Derivatives bugfixes

### LINKED ISSUE #1294: Geometry/topology fields at poloidal boundary
I've run into a bug when interpolating metrics, etc. near the y-boundary of closed field line regions. I think I know how to fix it; writing down here for comments and so I don't forget any parts:

* The integrated shear is not continuous when crossing from poloidal angle theta=2pi -> theta=0. Therefore the metric components, etc., are not necessarily continuous across the boundary.
* To get the correct metric components when we interpolate to a `CELL_YLOW` grid the guard cells of the `CELL_CENTRE` metric components need to contain the values corresponding to extending the metric components continously from the grid cells.
* When reading geometry from input file options, this is initially done 'correctly' -- 'y' is negative in the lower boundary guard cells and >2pi in the upper boundary guard cells; but then the guard cells are overwritten by a call to `communicate`.
* When reading the grid from a file, the y-guard cells are not included, so the communication is needed to fill the guard cells, but does not give the 'correct' values

I think this bug also causes a problem when y-derivatives of the metric are taken, e.g. to calculate G1/G2/G3 https://github.com/boutproject/BOUT-dev/blob/8389878169b4e4824c2d486dad7d9d65fa0db8c5/src/mesh/coordinates.cxx#L313-L315

Steps to fix the bug:

* Add an option to skip communication when reading variables from the grid to `Mesh::get(Field2D)` (`Mesh::get(Field3D)` already has this option)
* When loading geometrical quantities, skip communications so that the y-boundary guard cells have correct values
    * for backward compatibility, do communicate if a grid file does not include y-boundary guard cells
* Update hypnotoad to calculate and output the y-boundary guard cell values (I'd be grateful if someone would be willing to update hypnotoad...). This is the enhancement requested in #24.

johnomotani:
This is a collection of small bugfixes, mostly to do with staggered grids. I found them while writing the test in https://github.com/boutproject/BOUT-dev/tree/test-difops, but I'm trying to (at least partially) fix #1294 before that branch is ready to merge, so maybe it's useful to review these in the meantime.

### DIFF
--- a/src/mesh/coordinates.cxx
+++ b/src/mesh/coordinates.cxx
@@ -679,41 +679,40 @@ const Field3D Coordinates::Vpar_Grad_par(const Field3D &v, const Field3D &f, CEL
 const Field2D Coordinates::Div_par(const Field2D &f, CELL_LOC outloc,
                                    DIFF_METHOD method) {
   TRACE("Coordinates::Div_par( Field2D )");
-  // Perversely the relevant coordinates object might not be `this` if
-  // f isn't at the same location.
-  if (f.getLocation() != location)
-    return f.getCoordinates()->Div_par(f, outloc, method);
 
-  return Bxy * Grad_par(f / Bxy, outloc, method);
+  // Need Bxy at location of f, which might be different from location of this
+  // Coordinates object
+  Field2D Bxy_floc = f.getCoordinates()->Bxy;
+
+  return Bxy * Grad_par(f / Bxy_floc, outloc, method);
 }
 
 const Field3D Coordinates::Div_par(const Field3D &f, CELL_LOC outloc,
                                    DIFF_METHOD method) {
   TRACE("Coordinates::Div_par( Field3D )");
   
-  // Perversely the relevant coordinates object might not be `this` if
-  // f isn't at the same location.
-  if (f.getLocation() != location)
-    return f.getCoordinates()->Div_par(f, outloc, method);
+  // Need Bxy at location of f, which might be different from location of this
+  // Coordinates object
+  Field2D Bxy_floc = f.getCoordinates()->Bxy;
 
   if (f.hasYupYdown()) {
     // Need to modify yup and ydown fields
-    Field3D f_B = f / Bxy;
+    Field3D f_B = f / Bxy_floc;
     if (&f.yup() == &f) {
       // Identity, yup and ydown point to same field
       f_B.mergeYupYdown();
     } else {
       // Distinct fields
       f_B.splitYupYdown();
-      f_B.yup() = f.yup() / Bxy;
-      f_B.ydown() = f.ydown() / Bxy;
+      f_B.yup() = f.yup() / Bxy_floc;
+      f_B.ydown() = f.ydown() / Bxy_floc;
     }
     return Bxy * Grad_par(f_B, outloc, method);
   }
 
   // No yup/ydown fields. The Grad_par operator will
   // shift to field aligned coordinates
-  return Bxy * Grad_par(f / Bxy, outloc, method);
+  return Bxy * Grad_par(f / Bxy_floc, outloc, method);
 }
 
 /////////////////////////////////////////////////////////
--- a/src/mesh/difops.cxx
+++ b/src/mesh/difops.cxx
@@ -239,7 +239,7 @@ const Field3D Div_par(const Field3D &f, const Field3D &v) {
 
 const Field3D Div_par_flux(const Field3D &v, const Field3D &f, CELL_LOC outloc, DIFF_METHOD method) {
   Coordinates *metric = v.getCoordinates(outloc);
-  return metric->Bxy*FDDY(v, f/metric->Bxy, outloc, method)/sqrt(metric->g_22);
+  return metric->Bxy*FDDY(v, f/f.getCoordinates()->Bxy, outloc, method)/sqrt(metric->g_22);
 }
 
 const Field3D Div_par_flux(const Field3D &v, const Field3D &f, DIFF_METHOD method, CELL_LOC outloc) {
--- a/src/mesh/difops.cxx
+++ b/src/mesh/difops.cxx
@@ -239,7 +239,7 @@ const Field3D Div_par(const Field3D &f, const Field3D &v) {
 
 const Field3D Div_par_flux(const Field3D &v, const Field3D &f, CELL_LOC outloc, DIFF_METHOD method) {
   Coordinates *metric = v.getCoordinates(outloc);
-  return -metric->Bxy*FDDY(v, f/metric->Bxy, outloc, method)/sqrt(metric->g_22);
+  return metric->Bxy*FDDY(v, f/metric->Bxy, outloc, method)/sqrt(metric->g_22);
 }
 
 const Field3D Div_par_flux(const Field3D &v, const Field3D &f, DIFF_METHOD method, CELL_LOC outloc) {
--- a/src/mesh/difops.cxx
+++ b/src/mesh/difops.cxx
@@ -556,19 +556,19 @@ const Field3D Div_par_K_Grad_par(BoutReal kY, const Field3D &f, const CELL_LOC o
 }
 
 const Field2D Div_par_K_Grad_par(const Field2D &kY, const Field2D &f, const CELL_LOC outloc) {
-  return kY*Grad2_par2(f, outloc) + Div_par(kY, outloc)*Grad_par(f, outloc);
+  return interp_to(kY, outloc)*Grad2_par2(f, outloc) + Div_par(kY, outloc)*Grad_par(f, outloc);
 }
 
 const Field3D Div_par_K_Grad_par(const Field2D &kY, const Field3D &f, const CELL_LOC outloc) {
-  return kY*Grad2_par2(f, outloc) + Div_par(kY, outloc)*Grad_par(f, outloc);
+  return interp_to(kY, outloc)*Grad2_par2(f, outloc) + Div_par(kY, outloc)*Grad_par(f, outloc);
 }
 
 const Field3D Div_par_K_Grad_par(const Field3D &kY, const Field2D &f, const CELL_LOC outloc) {
-  return kY*Grad2_par2(f, outloc) + Div_par(kY, outloc)*Grad_par(f, outloc);
+  return interp_to(kY, outloc)*Grad2_par2(f, outloc) + Div_par(kY, outloc)*Grad_par(f, outloc);
 }
 
 const Field3D Div_par_K_Grad_par(const Field3D &kY, const Field3D &f, const CELL_LOC outloc) {
-  return kY*Grad2_par2(f, outloc) + Div_par(kY, outloc)*Grad_par(f, outloc);
+  return interp_to(kY, outloc)*Grad2_par2(f, outloc) + Div_par(kY, outloc)*Grad_par(f, outloc);
 }
 
 /*******************************************************************************
--- a/src/mesh/difops.cxx
+++ b/src/mesh/difops.cxx
@@ -789,25 +789,28 @@ const Field3D b0xGrad_dot_Grad(const Field3D &phi, const Field3D &A, CELL_LOC ou
 /*!
  * Calculate location of result
  */
-CELL_LOC bracket_location(const CELL_LOC &f_loc, const CELL_LOC &g_loc, const CELL_LOC &outloc) {
-  if(!mesh->StaggerGrids)
-    return CELL_CENTRE;
+// use anonymous namespace so this function is only available in this file
+namespace {
+  CELL_LOC bracket_location(const CELL_LOC &f_loc, const CELL_LOC &g_loc, const CELL_LOC &outloc, Mesh* localmesh=mesh) {
+    if(!localmesh->StaggerGrids)
+      return CELL_CENTRE;
 
-  if(outloc == CELL_DEFAULT){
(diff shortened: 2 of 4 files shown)


src/mesh/coordinates.cxx
```diff
@@ -679,41 +679,40 @@ const Field3D Coordinates::Vpar_Grad_par(const Field3D &v, const Field3D &f, CEL
...
-  // f isn't at the same location.
-  if (f.getLocation() != location)
-    return f.getCoordinates()->Div_par(f, outloc, method);
 
-  return Bxy * Grad_par(f / Bxy, outloc, method);
+  // Need Bxy at location of f, which might be different from location of this
+  // Coordinates object
+  Field2D Bxy_floc = f.getCoordinates()->Bxy;
+
+  return Bxy * Grad_par(f / Bxy_floc, outloc, method);
```
d7919: Implicit assumption that `location == outloc` (or at least that `location ==` location of return from `Grad_par`), should we assert this (maybe level 1?) to make it explicit?

src/mesh/coordinates.cxx
```diff
@@ -679,41 +679,40 @@ const Field3D Coordinates::Vpar_Grad_par(const Field3D &v, const Field3D &f, CEL
...
+      f_B.ydown() = f.ydown() / Bxy_floc;
     }
     return Bxy * Grad_par(f_B, outloc, method);
   }
 
   // No yup/ydown fields. The Grad_par operator will
   // shift to field aligned coordinates
-  return Bxy * Grad_par(f / Bxy, outloc, method);
+  return Bxy * Grad_par(f / Bxy_floc, outloc, method);
 }
```
d7919: Note previous comment about implicit assumption here.

src/mesh/difops.cxx
```diff
@@ -239,7 +239,7 @@ const Field3D Div_par(const Field3D &f, const Field3D &v) {
 
 const Field3D Div_par_flux(const Field3D &v, const Field3D &f, CELL_LOC outloc, DIFF_METHOD method) {
   Coordinates *metric = v.getCoordinates(outloc);
-  return -metric->Bxy*FDDY(v, f/metric->Bxy, outloc, method)/sqrt(metric->g_22);
+  return metric->Bxy*FDDY(v, f/f.getCoordinates()->Bxy, outloc, method)/sqrt(metric->g_22);
```
d7919: There's a sign change here -- is that intentional?

src/sys/derivs.cxx
```diff
@@ -340,9 +340,9 @@ const Field3D D2DYDZ(const Field3D &f, CELL_LOC outloc, DIFF_METHOD method, REGI
       for(int k=0;k<f.getMesh()->LocalNz;k++) {
         int kp = (k+1) % (f.getMesh()->LocalNz);
         int km = (k-1+f.getMesh()->LocalNz) % (f.getMesh()->LocalNz);
-        result(i,j,k) = 0.25*( +(f(i,j+1,kp) - f(i,j-1,kp))/(coords->dy(i,j+1))
-                               -(f(i,j+1,km) - f(i,j-1,km))/(coords->dy(i,j-1)) )
-                    / coords->dz;
+        result(i,j,k) = 0.25*( +(f(i,j+1,kp) - f(i,j-1,kp))
+                               -(f(i,j+1,km) - f(i,j-1,km)) )
+                    / coords->dy(i,j) / coords->dz;
```
bendudson: Now assuming that dy is the same at j+1 and j-1

src/mesh/index_derivs.cxx
```diff
@@ -2635,7 +2635,7 @@ const Field2D Mesh::indexFDDX(const Field2D &v, const Field2D &f, CELL_LOC outlo
   if ((method == DIFF_SPLIT) || ((method == DIFF_DEFAULT) && (fFDDX == nullptr))) {
```
d7919: I know you've not changed this but we only ever pass `method = DIFF_DEFAULT` so why do we allow/perform the option `DIFF_SPLIT` version?

src/mesh/difops.cxx
johnomotani: Yes, it's intentional. I think at some point this sign error was covering the one fixed in 41d05ad89252993a271aa411265bf31857a77e90. The sign here is correct (I think) and consistent with both the fixed U1 method, and the other methods (at least I've tested C2).

src/mesh/index_derivs.cxx
```diff
@@ -2767,7 +2767,7 @@ const Field3D Mesh::indexFDDX(const Field3D &v, const Field3D &f, CELL_LOC outlo
 
   if (this->xstart > 1) {
     // Two or more guard cells
-    if (StaggerGrids) {
+    if (StaggerGrids && vloc != diffloc) {
```
d7919: This line seems to be equivalent to that on line 2736 above -- I think this means we should never be able to hit the throw just below (as we'd already have thrown). Think the main practical application of this is that the nested `if` below only needs to care about one of `vloc` and `diffloc` (although having both may help make things clear).

src/mesh/index_derivs.cxx
```diff
@@ -2837,7 +2837,7 @@ const Field3D Mesh::indexFDDX(const Field3D &v, const Field3D &f, CELL_LOC outlo
     }
   } else if (this->xstart == 1) {
```
d7919: Again not something you've changed, just something I've noticed -- I think this only needs to be else if we assume `xstart>0` is asserted somewhere.

src/mesh/index_derivs.cxx
```diff
@@ -2924,7 +2924,7 @@ const Field2D Mesh::indexFDDY(const Field2D &v, const Field2D &f, CELL_LOC outlo
   if ((method == DIFF_SPLIT) || ((method == DIFF_DEFAULT) && (fFDDY == nullptr))) {
     // Split into an upwind and a central differencing part
     // d/dx(v*f) = v*d/dx(f) + f*d/dx(v)
-    return indexVDDY(v, f, outloc, DIFF_DEFAULT) + f * indexDDY(v);
+    return indexVDDY(v, f, outloc, DIFF_DEFAULT) + interp_to(f, outloc) * indexDDY(v);
```
d7919: Do we need to assert `v.getLocation()== outloc` here or should we pass `outloc` into the last `indexDDY` call?

src/sys/derivs.cxx
johnomotani: Possibly I don't understand the original intention, but I don't think this assumption is being made: a non-uniform dy correction is not needed here because `dy` is constant in z (and `dz` is constant). The old version looks to me like DDY(f)[z+1] is uses dy(y+1) while DDY(f)[z-1] uses dy(y-1): I think they should both use dy(y) because dy is constant in z, and that this is what the new version is doing. When I test with https://github.com/boutproject/BOUT-dev/tree/test-difops, the old version does not converge, while the new one converges at 2nd order.

src/mesh/index_derivs.cxx
johnomotani: I think it's possible for a user to call the `FDDX` function with a `method=DIFF_SPLIT` argument that would be passed through and have an effect here.

src/mesh/index_derivs.cxx
johnomotani: I don't think xstart>0 is actually required. It should be possible to run 2d simulations with no guard cells in the 3rd direction. Then this test is needed so that an exception is thrown if the user tries to take a derivative in the 3rd direction.

src/mesh/index_derivs.cxx
johnomotani: that's a bug, should be passing outloc :+1:

src/mesh/coordinates.cxx
johnomotani: :+1: I've added checks through all the Coordinates derivatives/difops now.

src/mesh/index_derivs.cxx
ZedThree: `diffloc` is changed to `inloc` at those lines though? When is `diffloc` ever not `inloc`?

What combinations of locations were causing problems?

src/mesh/index_derivs.cxx
ZedThree: Currently this does throw if `xstart == 0` (the final `else` clause), so I agree we should move that to an `ASSERT` near the top.

src/mesh/index_derivs.cxx
```diff
@@ -1881,7 +1881,7 @@ const Field3D Mesh::indexVDDX(const Field3D &v, const Field3D &f, CELL_LOC outlo
     Mesh::flux_func func = sfVDDX;
     DiffLookup *table = UpwindTable;
 
-    if (vloc == CELL_XLOW) {
+    if ((vloc == CELL_XLOW) && (inloc == CELL_CENTRE)) {
```
ZedThree: I think this change means that the following snippet:

```cpp
    if ((vloc == CELL_XLOW) && (inloc == CELL_CENTRE)) {
      // V staggered w.r.t. variable
      func = sfVDDX;
      table = UpwindStagTable;
      diffloc = CELL_CENTRE;
    } else if ((vloc == CELL_CENTRE) && (inloc == CELL_XLOW)) {
      // Shifted
      func = sfVDDX;
      table = UpwindStagTable;
      diffloc = CELL_XLOW;
    } else {
 ```

could be rewritten as:

```cpp
    if (((vloc == CELL_XLOW) && (inloc == CELL_CENTRE))
        || ((vloc == CELL_CENTRE) && (inloc == CELL_XLOW))){
      // V staggered w.r.t. variable
      func = sfVDDX;
      table = UpwindStagTable;
    } else {
```

As now `diffloc` is explicitly the same as `inloc`, which is set at 1868. Unless I am missing something?

The same applies to the `if` statements at 2272, 2538, etc.

src/mesh/index_derivs.cxx
d7919: > It's not equivalent: at line 2736 diffloc=inloc, but diffloc can be changed at 2742 or 2747. I actually added this because in testing combinations of vloc/inloc/outloc that should throw, some did not without this condition.

But I think the possible changes at 2742 and 2747 force `diffloc = inloc` according to the logic in the `if`s

src/mesh/index_derivs.cxx
d7919: I know in some routines there's an "early exit" check that says, e.g. `if nz == 1 {return 0;}`, maybe we that here.

src/sys/derivs.cxx
d7919: Is it possible to merge ` / coords->dy(i,j) / coords->dz` into ` / (coords->dy(i,j) * coords->dz)` for clarity?

src/mesh/index_derivs.cxx
johnomotani: There were quite a few of these. I've cut them down in 380d891, as well as getting rid of all the 'diffloc's.

src/mesh/index_derivs.cxx
```diff
@@ -2305,87 +2191,15 @@ const Field3D Mesh::indexVDDY(const Field3D &v, const Field3D &f, CELL_LOC outlo
...
-              vval.mm = vval.m;
-              vval.m = vval.c;
-            }
-            // Shifted in one direction -> shift in another
-            // Could produce warning
-          }
-          result[i] = func(vval, fval);
-        }
-      }
-    } else if (vUseUpDown) {
```
ZedThree: What happened to this branch?

src/mesh/index_derivs.cxx
```diff
@@ -677,11 +681,11 @@ const Field2D Mesh::applyXdiff(const Field2D &var, Mesh::deriv_func func,
           s.p = var[i.xp()];
           s.pp = var[i.xpp()];
 
-          if ((location == CELL_CENTRE) && (loc == CELL_XLOW)) {
+          if (outloc == CELL_XLOW) {
             // Producing a stencil centred around a lower X value
             s.pp = s.p;
             s.p = s.c;
-          } else if (location == CELL_XLOW) {
+          } else if (inloc == CELL_XLOW) {
```
ZedThree: Is there another possibility at this point? If not, this could lose the `if` statement.

johnomotani: :+1: that one can go

src/mesh/index_derivs.cxx
johnomotani: Sorry, should have put this in a separate commit with an explanation: it's not correct to mix fields with yup/ydown and fields that are shifted to field-aligned, because the physical locations of their grid points are different. So I just got rid of them rather than updating the diffloc, etc. in those branches, but I should have been not quite so lazy.

johnomotani:
@d7919 asked on one of the commits whether the ASSERT0s I've used to check locations should be ASSERT1s. I used ASSERT0 to match the previous behaviour, which was mostly to throw an exception for unhandled locations, and so that the test I'm writing can check that all the operators throw for all permutations of inloc/outloc(/vloc) that aren't implemented, so that at the end of the test (when running in a long version) all permutations have been tested and either converge or throw. The test could behave differently at different check levels though, I think that's sensible, so I'll change them to ASSERT1 unless someone has a different preference?

johnomotani:
I'll clang-format that last commit. Do you want me to do the whole file while I'm at it, or just the new bits I've been changing?

ZedThree:
Just the bits you've changed is fine

PR #1303, opened 2018-10-08
Squash test

dschwoerer:
Before I wanted to use squashoutput to append data to some of my longer running simulations, I thought I write a small test :-)
The found bugs (in squashoutput) are now fixed.

The test currently fails in about 2s - but normally should pass in around 4s, so could be enabled by default.

Note that the test would currently still fail.
I havent been using `Field2D` - so I might be missing a magic flag in `BOUT.inp` or it is a bug in `Field2D`.
The outer most guard cells have a zero more than the guard cells towards the domain:
```
  0, 0, 0, 0, 0, 0, 0, 0,
  0, 0, 0, 0, 0, 0, 0, 0,
  0, 0, 20, 20, 20, 20, 0, 0,
  0, 0, 20, 20, 20, 20, 0, 0,
  0, 0, 20, 20, 20, 20, 0, 0,
  0, 0, 20, 20, 20, 20, 0, 0,
  0, 0, -20, -20, -20, -20, 0, 0,
  0, 0, 0, -60, -60, -60, 0, 0 ;
```
Not sure why ...

### DIFF
--- a/tests/integrated/test-squash/runtest
+++ b/tests/integrated/test-squash/runtest
@@ -55,6 +55,15 @@ run("../../../bin/bout-squashoutput -qdcal 9 data --outputname ../para.nc")
 
 verify("20.nc", "para.nc")
 
+# Parallel and in two pieces without dump_on_restart
+run("rm -f para.nc")
+run("mpirun -np 4 ./squash -q -q -q")
+run("../../../bin/bout-squashoutput -qdcl 9 data --outputname ../para.nc")
+run("mpirun -np 4 ./squash -q -q -q restart dump_on_restart=false")
+run("../../../bin/bout-squashoutput -qdcal 9 data --outputname ../para.nc")
+
+verify("20.nc", "para.nc")
+
 # Sequential test
 run("rm -f seq.nc")
 run("./squash -q -q -q nout=20")
--- a/tests/integrated/test-squash/runtest
+++ b/tests/integrated/test-squash/runtest
@@ -1,0 +1,63 @@
+#!/usr/bin/env python3
+
+import netCDF4 as nc
+import itertools as it
+import os
+
+
+def run(cmd):
+    s = os.system(cmd)
+    if s:
+        raise RuntimeError("%s exited with non-zero %d value" % (cmd, s))
+
+
+def verify(f1, f2):
+    org = nc.Dataset(f1)
+    para = nc.Dataset(f2)
+    for v in org.variables:
+        if org[v].shape != para[v].shape:
+            print(org[v])
+            print(para[v])
+            raise RuntimeError("shape mismatch in ", v)
+        if v in ["MXSUB","MYSUB","NXPE","NYPE","iteration"]:
+            continue
+        if (org[v][:] != para[v][:]).any():
+            a=org[v]
+            b=para[v]
+            print(a)
+            print(b)
+            lst=[range(x) for x in org[v].shape]
+            for i in it.product(*lst):
+                if a[i] != b[i]:
+                    print(i,":",a[i],"!=",b[i])
+            # There are differences, but I think only guard cells are affected
+            raise RuntimeError("data mismatch in ", v)
+
+run("make")
+
+# Run once to get normal data
+run("./squash -q -q -q nout=20")
+run("mv data/BOUT.dmp.0.nc 20.nc")
+
+# Parallel test
+run("rm -f para.nc")
+run("mpirun -np 4 ./squash -q -q -q nout=20")
+run("../../../bin/bout-squashoutput -qdcl 9 data --outputname ../para.nc")
+
+verify("20.nc", "para.nc")
+
+# Parallel and in two pieces
+run("rm -f para.nc")
+run("mpirun -np 4 ./squash -q -q -q")
+run("../../../bin/bout-squashoutput -qdcl 9 data --outputname ../para.nc")
+run("mpirun -np 4 ./squash -q -q -q restart")
+run("../../../bin/bout-squashoutput -qdcal 9 data --outputname ../para.nc")
+
+verify("20.nc", "para.nc")
+
+# Sequential test
+run("rm -f seq.nc")
+run("./squash -q -q -q nout=20")
+run("../../../bin/bout-squashoutput -qdcl 9 data --outputname ../seq.nc")
+
+verify("20.nc", "seq.nc")
--- a/tests/integrated/test-squash/runtest
+++ b/tests/integrated/test-squash/runtest
@@ -1,8 +1,7 @@
 #!/usr/bin/env python3
 
 from boututils.datafile import DataFile
-import itertools as it
-import os
+import itertools
 import time
 import numpy as np
 from boututils.run_wrapper import launch_safe, shell_safe
@@ -10,17 +9,20 @@ from boututils.run_wrapper import launch_safe, shell_safe
 #requires: all_tests
 #requires: netcdf
 
-class timer(object):
 
+class timer(object):
+    """Context manager for printing how long a command took
+
+    """
     def __init__(self, msg):
         self.msg = msg
 
     def __enter__(self):
         self.start = time.time()
 
-    def __exit__(self, a, b, c):
+    def __exit__(self, exc_type, exc_value, traceback):
         end = time.time()
-        print("%12.8f %s" % (end - self.start, self.msg))
+        print("{:12.8f}s {}".format(end - self.start, self.msg))
 
 
 def timed_shell_safe(cmd, *args, **kwargs):
@@ -40,6 +42,9 @@ def timed_launch_safe(cmd, *args, **kwargs):
 
 
 def verify(f1, f2):
+    """Verifies that two BOUT++ files are identical
+
+    """
     with timer("verify %s %s" % (f1, f2)):
         d1 = DataFile(f1)
         d2 = DataFile(f2)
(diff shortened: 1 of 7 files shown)


ZedThree:
Tests are good, thanks! 4s is quite long though, any chance you could reduce `nout` and the grid sizes to get this as fast as possible?

tests/integrated/test-squash/runtest
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env python3
+
+import netCDF4 as nc
```
ZedThree: This is tied to `netCDF4` -- is it possible to just use `Datafile`?

tests/integrated/test-squash/runtest
```diff
@@ -0,0 +1,72 @@
...
+import itertools as it
+import os
+
+
+def run(cmd):
```
ZedThree: `boututils` includes `launch_safe`, which this looks to be a reimplementation of. Is it possible to just use `launch_safe`?

tests/integrated/test-squash/runtest
dschwoerer: It anway requires netcdf, the way it is implemented now. Without it becomes far more complicated, as I need to detect the correct file extension etc.
Without that extra code, Datafile could be used, but I see no advantage ...
I prefer netcdf, as it does not load everything immediately.
If you feel strongly about that change, I can change it though.

tests/integrated/test-squash/runtest
dschwoerer: Yes, but is only 4 lines of code - thus basically as long as the other version. And for launch_safe I always need to check where to import it from, which is far more effort, without any gain, as far as I can see. Also, it make it slower. It also makes it harder to add timing instructions to the calls.
Again, if you feel strongly about this, I can change it, but I feel to see the advantage.

tests/integrated/test-squash/runtest
```diff
@@ -0,0 +1,72 @@
...
+# Parallel test
+run("rm -f para.nc")
+run("mpirun -np 4 ./squash -q -q -q nout=20")
+run("../../../bin/bout-squashoutput -qdcl 9 data --outputname ../para.nc")
+
+verify("20.nc", "para.nc")
+
+# Parallel and in two pieces
+run("rm -f para.nc")
+run("mpirun -np 4 ./squash -q -q -q")
```
d7919: Does this have to use exactly four processors or could this be an option (e.g. for travis it would be nice to limit to two processors if possible)?

tests/integrated/test-squash/runtest
ZedThree: Sorry, I was not very clear, I meant the `netCDF4` python module, which is not always available everywhere and can be a pain to install on certain systems (e.g. see the problems we had getting it working on Travis!). `Datafile` takes care of this, supporting several NetCDF python modules. We should be using it instead of "bare" NetCDF modules like this.

Also, the lazy-loading of NetCDF files is because they are backed by HDF5 which handles that, so if reading HDF5 files via `Datafile` loads everything immediately, that's our bug.

This test should probably also be marked as requiring NetCDF at any rate.

tests/integrated/test-squash/runtest
ZedThree: It's four lines of code that need to be read and checked to see what they do by everyone reading this, versus something that's well integrated into our python library that already has a bunch of checks, documentation, handling for various systems, etc.

It took me 10 seconds to find what module `launch_safe` is in, I'm sure it would've taken you less time -- please reuse stuff rather than reinventing the wheel.

tests/integrated/test-squash/runtest
```diff
@@ -0,0 +1,72 @@
...
+
+def run(cmd):
+    s = os.system(cmd)
+    if s:
+        raise RuntimeError("%s exited with non-zero %d value" % (cmd, s))
+
+
+def verify(f1, f2):
+    org = nc.Dataset(f1)
+    para = nc.Dataset(f2)
```
ZedThree: Please could you rename these? I assume "org" means "original", but I have no clue what "para" means

tests/integrated/test-squash/runtest
```diff
@@ -0,0 +1,72 @@
...
+    for v in org.variables:
+        if org[v].shape != para[v].shape:
+            print(org[v])
+            print(para[v])
+            raise RuntimeError("shape mismatch in ", v)
```
ZedThree: Put the bad values straight into the exception, rather than printing them

tests/integrated/test-squash/runtest
```diff
@@ -0,0 +1,72 @@
...
+        if v in ["MXSUB", "MYSUB", "NXPE", "NYPE", "iteration"]:
+            continue
+        if (org[v][:] != para[v][:]).any():
```
ZedThree: Use `np.allclose` for comparing arrays of floats

tests/integrated/test-squash/runtest
dschwoerer: how do I get newlines?
Now the error looks terrible:
```
RuntimeError: ('data mismatch in ', 'f2', '(3, 0, 1):-3.0!=-3.0000000000000004\n(3, 0, 2):-3.0!=-3.0000000000000004\n(3, 1, 1):3.0!=3.0000000000000004\n(3, 1, 2):3.0!=3.0000000000000004\n(3, 2, 1):3.0!=3.0000000000000004\n(3, 2, 2):3.0!=3.0000000000000004\n(3, 3, 1):-3.0!=-3.0000000000000004\n(3, 3, 2):-3.0!=-3.0000000000000004\n', BoutArray([[[ 0.,  0.,  0.,  0.],
            [ 0.,  0.,  0.,  0.],
            [ 0.,  0.,  0.,  0.],
            [ 0.,  0.,  0.,  0.]],

           [[ 0., -1., -1.,  0.],
            [ 0.,  1.,  1.,  0.],
            [ 0.,  1.,  1.,  0.],
            [ 0., -1., -1.,  0.]],

           [[ 0., -2., -2.,  0.],
            [ 0.,  2.,  2.,  0.],
            [ 0.,  2.,  2.,  0.],
            [ 0., -2., -2.,  0.]],

           [[ 0., -3., -3.,  0.],
            [ 0.,  3.,  3.,  0.],
            [ 0.,  3.,  3.,  0.],
            [ 0., -3., -3.,  0.]],

           [[ 0., -4., -4.,  0.],
            [ 0.,  4.,  4.,  0.],
            [ 0.,  4.,  4.,  0.],
            [ 0., -4., -4.,  0.]]]), BoutArray([[[ 0.,  0.,  0.,  0.],
            [ 0.,  0.,  0.,  0.],
            [ 0.,  0.,  0.,  0.],
            [ 0.,  0.,  0.,  0.]],

           [[ 0., -1., -1.,  0.],
            [ 0.,  1.,  1.,  0.],
            [ 0.,  1.,  1.,  0.],
            [ 0., -1., -1.,  0.]],

           [[ 0., -2., -2.,  0.],
            [ 0.,  2.,  2.,  0.],
            [ 0.,  2.,  2.,  0.],
            [ 0., -2., -2.,  0.]],

           [[ 0., -3., -3.,  0.],
            [ 0.,  3.,  3.,  0.],
            [ 0.,  3.,  3.,  0.],
            [ 0., -3., -3.,  0.]],

           [[ 0., -4., -4.,  0.],
            [ 0.,  4.,  4.,  0.],
            [ 0.,  4.,  4.,  0.],
            [ 0., -4., -4.,  0.]]]))
```
Its not obvious anymore that boundary cells are affected ...

tests/integrated/test-squash/runtest
dschwoerer: was parallel, but has no meaning any more. I switched to `d1` and `d2`.

tests/integrated/test-squash/runtest
dschwoerer: I thought 4 would be nice to get domain decomposition in both direction, but it does not really matter, as it currently relies on collect, which should have it's own test ...

tests/integrated/test-squash/runtest
dschwoerer: launch_safe is also not what I want :-1: 
But as code is read more often then written, I will change ...

tests/integrated/test-squash/runtest
```diff
@@ -1,72 +1,90 @@
...
+                continue
+            if not np.allclose(d1[v], d2[v]):
+                # continue
+                a = d1[v]
+                b = d2[v]
+                err = ""
+                lst = [range(x) for x in d1[v].shape]
+                for i in it.product(*lst):
+                    if a[i] != b[i]:
+                        err += str(i) + ":" + str(a[i]) + "!=" + str( b[i]) + """
```
ZedThree: This is probably clearer as:

```python
err += "{}: {} != {}\n".format(i, a[i], b[i])
```

d7919:
This is into next, should it be v4.2?

ZedThree:
Tests pass for me, even for Field2D.

I've used `launch_safe`, rather than `shell_safe` for actually running the physics model, as this allows setting the number of OpenMP threads.

It's still too slow to be in by default though, I think.

dschwoerer:
Ah, I fixed the test by accident.
Using only 1 guard cell in y direction fixes the issue. One cell in the 2nd guard cell is not set, so having only 1 guard cells avoids the issue.
I am still curious as to why that happens ...

The test could be speed-up by not testing the shell interface of squash, but rather the python interface directly - however I implemented it this way, as I want to use the shell interface ...

ZedThree:
Testing the wrapper seems reasonable. It could be sped up by just making the netcdf files directly, instead of spinning up BOUT++ and running a physics model. On the other hand, the slow bit might be writing to disk and this will always be slow.

I seem to recall there being some issues with setting the outer-most guard cell, but I forget exactly what.

PR #1304, opened 2018-10-09
Add fmod function for inputs, and a test in test-initial

bshanahan:
This PR adds a function "fmod" for input fields, which is a modulo operator for floating point numbers.


ZedThree:
Looks good, please can you add this to the table of functions in the documentation?

PR #1305, opened 2018-10-09
Version 4.2 release candidate

ZedThree:
This is the release candidate for version 4.2.0. New bugfixes should go straight into this branch, new features should go into next. Once we're happy, this will go into master, and we'll merge master into next.

The only difference in this PR from current next is f02880b: bumping the version number, and a merge of master which was missing.

The release date and DOI in the CITATION.cff file need to be updated just before release.

I've run `test_suite --all`, which is a very nice addition, letting us run all the integrated tests in one go! All of them passed first time, which is even better. Some of the MMS tests are still failing, but these are things like WENO3, which I think is just tricky to MMS. elm-pb also doesn't pass MMS (and needs #1293 to run), but that is also tricky to MMS.

I think the following open PRs need to go into this: #1302 #1293 #1303 

---

This is an enormous release, with about 360 merged PRs, and almost 2,500 commits! I've been trying to summarise them all in a changelog (which also needs to go in!), and there's a lot of good stuff that's gone in in the last year.

This PR doesn't need to be reviewed as such, it's just open to collect any bugfixes that need to go in before the release.


d7919:
Net ~27k lines is a large increase to the full code!

d7919:
#1302 has already been merged into next (sorry) -- is there anything in next that we don't want in 4.2 or can we just merge next in here to fetch these last few bits?

PR #1310, opened 2018-10-10
Rename Mesh::coordinates -> Mesh::getCoordinates; deprecate old name

ZedThree:
While we are polishing things for the release candidate -- renames `Mesh::coordinates` to `Mesh::getCoordinates` to be more in line with our naming convention. The old name is deprecated.

I used rtags to rename the symbol, so it got all instances of `Mesh::coordinates` and not just things named `->coordinates`. Some of these could potentially use the fields version, but I couldn't be bothered to go through and check

---

There's a couple of other quick changes I'd like to make to `Mesh::getCoordinates`:

1. change the `coords_map.insert(...)` to `coords_map.emplace(location, createDefaultCoordinates(location));`. I'm not 100% sure `emplace` should be used here. At least, `make_pair` could be used to remove the type name

2. Move the throw to the top as a precondition, rather than burying it in the middle of the conditional. Should it always be on, or at `CHECK > 1/2/3`?

### DIFF
--- a/examples/bout_runners_example/diffusion_3D.cxx
+++ b/examples/bout_runners_example/diffusion_3D.cxx
@@ -60,8 +60,8 @@ int physics_init(bool restarting) {
         // The boundary lies (1/2)*dx away from the last point As there
         // are 2 boundaries there will effectively add one more line
         // segment in the domain. Hence
-        mesh->coordinates()->dx = Lx/(internal_x_points);
-        mesh->coordinates()->dy = Ly/(internal_y_points);
+        mesh->getCoordinates()->dx = Lx/(internal_x_points);
+        mesh->getCoordinates()->dy = Ly/(internal_y_points);
     }
     // ************************************************************************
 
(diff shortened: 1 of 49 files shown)


d7919:
I'm sure I've already changed the coordinates routine to use asserts so not sure why it looks like I haven't!

d7919:
> I'm sure I've already changed the coordinates routine to use asserts so not sure why it looks like I haven't!

Ah yes, in #1293 (not yet merged) https://github.com/boutproject/BOUT-dev/blob/vecops-consistent-cell-default/include/bout/mesh.hxx#L433

d7919:
As this touches a lot of files we might want to hold this until all other changes are into the rc?

d7919:
> 2\. Move the throw to the top as a precondition, rather than burying it in the middle of the conditional. Should it always be on, or at `CHECK > 1/2/3`?

I don't think it should always be on as we have a large number of these calls throughout the code.

ZedThree:
Yes, probably hold off on merging this till the other PRs are in.

ZedThree:
This is good to go in now -- all other PRs have been resolved. Last commit just fixed the conflicts

PR #1312, opened 2018-10-11
Throw error messages for non-python collect routines

ZedThree:
Someone nice who knows IDL, Matlab/Octave and/or Mathematica could maybe actually fix these routines and submit pull requests :)

Does the minimal thing for #394 -- tells the user it's broken


ZedThree:
If I knew who to request to review this, I would ask them to properly fix these routines! I don't think I could have made them worse, at any rate. I will use my admin-privileges to merge this at the end of the day.

bendudson:
I think for now this is the Right Thing to do; it's not worth spending a lot of time fixing these routines if no-one is using them, or even trying to use them. The first person to ask gets the job of fixing them ;)

umansky:
What is broken there? If I just comment out the message line where the routine is stopped then my IDL post-processing goes just fine.

PR #1313, opened 2018-10-11
Provide `getRegion` method on Fields

d7919:
Used to return the requested Region from fieldmesh.


include/field2d.hxx
```diff
@@ -154,6 +154,10 @@ class Field2D : public Field, public FieldData {
    */
   const IndexRange DEPRECATED(region(REGION rgn)) const override;
 
+  /// Return a Region<Ind2D> reference to use to iterate over this field
+  const Region<Ind2D>& getRegion(REGION region) const;
```
ZedThree: Could these be marked `inline`?

PR #1315, opened 2018-10-11
Test difops

johnomotani:
There are quite a few commits here, but it does add a test and a bugfix, so worth considering for v4.2?

The test is an MMS test using boutcore for all the differential operators and derivatives, with all combinations of inloc/outloc, and vloc if necessary. It checks that all the operators either converge at 2nd order (a couple use U1, so first order for those) or throw an exception (if CHECK>0). The full test is pretty long (oddly, testing for Exceptions via cython is slower than most of the successful tests), so there's a short version for the test-suite that only tests the difops (not the derivatives), skips the checks for exceptions and skips operators that require 3d grid refinement. Arguably, the short version is still too slow, but if so we need to decide what to take out (maybe operators that few people use, or some unlikely locations like ZLOW/ZLOW).

Summary of changes:
* c04647a - 3288cca make interfaces for Field2D and Field3D versions of operators more consistent. This made them easier to add to boutcore.
* 721d251 - 095cfca are additions to boutcore: a Field2D class; a bunch of extra operators (many of these are just need a name adding to code-generation loops, after the changes in the previous bullet); a `boutcore.bout_CHECK` variable to access the value of `CHECK`; some extra keywords to allow differential operators to raise C++ exceptions as Python exceptions.
* c7205a4 and 2bcfe64 add testing of non-uniform dx/dy to the MMS/derivatives3 test.
* f515601 adds a Python class that provides symbolic differential operators and analytic tokamak-like geometry. It started as a copy of mms.py, but I changed the normalizations and added extra stuff. I'd be in favour of replacing mms.py with mms_alternate.py, but wasn't sure if anyone/anything depended on mms.py so didn't want to just change it.
* 2f407e6 - 473cc54 are the new MMS test of difops and derivatives. They fail until the next set of commits is added...
* 0b06173 - 23c0d1b change how the metric, etc. is loaded: extrapolate into guard cells instead Neumann boundary conditions, fixes glitches in G1/G2/G3 at the boundaries; don't communicate metrics (unless no guard cells are available, as when loading from a grid file), because metric components don't have to be periodic in poloidal angle (can depend on integrated shear) - then when the YLOW versions are interpolated, you get large errors near the poloidal branch-cut if the guard cells are filled by communicating. Since the 'fix' does not fix loading from grid files, it is not ideal, but at least lets tests check the rest of the code-base as long as they load the grid from options.
* 3568a7b is trivial tidying up, removing an unnecessary interp_to in Grad2_par2.

### DIFF
--- a/include/field_data.hxx
+++ b/include/field_data.hxx
@@ -62,12 +62,20 @@ class FieldVisitor;
 */
 class FieldData {
 public:
-  FieldData();
+  FieldData(Mesh* m);
   virtual ~FieldData();
 
   // Visitor pattern support
   virtual void accept(FieldVisitor &v) = 0;
   
+  virtual Mesh * getDataMesh() const{
+    if (fielddatamesh){
+      return fielddatamesh;
+    } else {
+      return mesh;
+    }
+  }
+
   // Defines interface which must be implemented
   virtual bool isReal() const = 0; ///< Returns true if field consists of BoutReal values
   virtual bool is3D() const = 0;   ///< True if variable is 3D
@@ -92,6 +100,7 @@ public:
   FieldGeneratorPtr getBndryGenerator(BndryLoc location);
 
 protected:
+  Mesh* fielddatamesh;
   vector<BoundaryOp *> bndry_op; ///< Boundary conditions
   bool boundaryIsCopy;           ///< True if bndry_op is a copy
   bool boundaryIsSet;            ///< Set to true when setBoundary called
--- a/src/field/field2d.cxx
+++ b/src/field/field2d.cxx
@@ -45,7 +45,8 @@
 
 #include <bout/assert.hxx>
 
-Field2D::Field2D(Mesh *localmesh) : Field(localmesh), deriv(nullptr) {
+Field2D::Field2D(Mesh *localmesh) :
+  Field(localmesh), FieldData(localmesh), deriv(nullptr) {
 
   boundaryIsSet = false;
 
@@ -66,6 +67,7 @@ Field2D::Field2D(Mesh *localmesh) : Field(localmesh), deriv(nullptr) {
 }
 
 Field2D::Field2D(const Field2D& f) : Field(f.fieldmesh), // The mesh containing array sizes
+                                     FieldData(f.fieldmesh),
                                      data(f.data), // This handles references to the data array
                                      deriv(nullptr) {
   TRACE("Field2D(Field2D&)");
@@ -95,7 +97,9 @@ Field2D::Field2D(const Field2D& f) : Field(f.fieldmesh), // The mesh containing 
   boundaryIsSet = false;
 }
 
-Field2D::Field2D(BoutReal val, Mesh *localmesh) : Field(localmesh), deriv(nullptr) {
+Field2D::Field2D(BoutReal val, Mesh *localmesh) :
+    Field(localmesh), FieldData(localmesh), deriv(nullptr) {
+
   boundaryIsSet = false;
 
   nx = fieldmesh->LocalNx;
--- a/src/field/field3d.cxx
+++ b/src/field/field3d.cxx
@@ -45,8 +45,8 @@
 
 /// Constructor
 Field3D::Field3D(Mesh *localmesh)
-    : Field(localmesh), background(nullptr), deriv(nullptr), yup_field(nullptr),
-      ydown_field(nullptr) {
+    : Field(localmesh), FieldData(localmesh), background(nullptr),
+      deriv(nullptr), yup_field(nullptr), ydown_field(nullptr) {
 #ifdef TRACK
   name = "<F3D>";
 #endif
@@ -71,6 +71,7 @@ Field3D::Field3D(Mesh *localmesh)
 /// later)
 Field3D::Field3D(const Field3D &f)
     : Field(f.fieldmesh),                // The mesh containing array sizes
+      FieldData(f.fieldmesh),
       background(nullptr), data(f.data), // This handles references to the data array
       deriv(nullptr), yup_field(nullptr), ydown_field(nullptr) {
 
@@ -100,8 +101,8 @@ Field3D::Field3D(const Field3D &f)
 }
 
 Field3D::Field3D(const Field2D &f)
-    : Field(f.getMesh()), background(nullptr), deriv(nullptr), yup_field(nullptr),
-      ydown_field(nullptr) {
+    : Field(f.getMesh()), FieldData(f.getMesh()), background(nullptr),
+      deriv(nullptr), yup_field(nullptr), ydown_field(nullptr) {
 
   TRACE("Field3D: Copy constructor from Field2D");
 
@@ -118,8 +119,8 @@ Field3D::Field3D(const Field2D &f)
 }
 
 Field3D::Field3D(const BoutReal val, Mesh *localmesh)
-    : Field(localmesh), background(nullptr), deriv(nullptr), yup_field(nullptr),
-      ydown_field(nullptr) {
+    : Field(localmesh), FieldData(localmesh), background(nullptr),
+      deriv(nullptr), yup_field(nullptr), ydown_field(nullptr) {
 
   TRACE("Field3D: Copy constructor from value");
 
--- a/src/field/field_data.cxx
+++ b/src/field/field_data.cxx
@@ -6,8 +6,11 @@
 #include <field_factory.hxx>
 #include "unused.hxx"
 
-FieldData::FieldData() : boundaryIsCopy(false), boundaryIsSet(true) {
-  
+FieldData::FieldData(Mesh* m) :
+    fielddatamesh(m), boundaryIsCopy(false), boundaryIsSet(true) {
+  if (fielddatamesh == nullptr) {
+    fielddatamesh = mesh;
+  }
 }
 
 FieldData::~FieldData() {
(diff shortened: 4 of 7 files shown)


src/mesh/coordinates.cxx
```diff
@@ -648,11 +703,11 @@ const Field2D Coordinates::DDZ(const Field2D &f, CELL_LOC loc,
 // Parallel gradient
 
 const Field2D Coordinates::Grad_par(const Field2D &var, CELL_LOC outloc,
-                                    DIFF_METHOD UNUSED(method)) {
+                                    DIFF_METHOD method) {
   TRACE("Coordinates::Grad_par( Field2D )");
   ASSERT1(location == outloc || outloc == CELL_DEFAULT);
```
d7919: You didn't change this here but I'm not sure this assert is correct -- I think it should be (location == outloc || (outloc == CELL_DEFAULT && location == var.getLocation())`?

src/mesh/coordinates.cxx
```diff
@@ -669,9 +724,9 @@ const Field3D Coordinates::Grad_par(const Field3D &var, CELL_LOC outloc,
 
 const Field2D Coordinates::Vpar_Grad_par(const Field2D &v, const Field2D &f,
                                          CELL_LOC outloc,
-                                         DIFF_METHOD UNUSED(method)) {
+                                         DIFF_METHOD method) {
   ASSERT1(location == outloc || outloc == CELL_DEFAULT);
```
d7919: and here.

src/mesh/coordinates.cxx
```diff
@@ -728,38 +783,33 @@ const Field3D Coordinates::Div_par(const Field3D &f, CELL_LOC outloc,
 // second parallel derivative (b dot Grad)(b dot Grad)
 // Note: For parallel Laplacian use Laplace_par
 
-const Field2D Coordinates::Grad2_par2(const Field2D &f, CELL_LOC outloc) {
+const Field2D Coordinates::Grad2_par2(const Field2D &f, CELL_LOC outloc, DIFF_METHOD method) {
   TRACE("Coordinates::Grad2_par2( Field2D )");
   ASSERT1(location == outloc || outloc == CELL_DEFAULT);
```
d7919: and here.

src/mesh/coordinates.cxx
```diff
@@ -728,38 +783,33 @@ const Field3D Coordinates::Div_par(const Field3D &f, CELL_LOC outloc,
...
-  Field2D result = DDY(1. / sg, outloc) * DDY(f, outloc) / sg + D2DY2(f, outloc) / g_22;
+  Field2D result = DDY(1. / sg, outloc, method) * DDY(f, outloc, method) / sg + D2DY2(f, outloc, method) / g_22;
 
   return result;
 }
 
-const Field3D Coordinates::Grad2_par2(const Field3D &f, CELL_LOC outloc) {
+const Field3D Coordinates::Grad2_par2(const Field3D &f, CELL_LOC outloc, DIFF_METHOD method) {
   TRACE("Coordinates::Grad2_par2( Field3D )");
   ASSERT1(location == outloc || outloc == CELL_DEFAULT);
```
d7919: and here..

src/mesh/coordinates.cxx
```diff
@@ -728,38 +783,33 @@ const Field3D Coordinates::Div_par(const Field3D &f, CELL_LOC outloc,
...
 
   sg = sqrt(g_22);
-  sg = DDY(1. / sg) / sg;
+  sg = DDY(1. / sg, outloc, method) / sg;
 
   if (outloc == CELL_DEFAULT) {
     outloc = f.getLocation();
   }
 
-  if (sg.getLocation() != outloc) {
```
d7919: I don't think we assert that this case can't arise so are we allowed to remove this?

src/mesh/coordinates.cxx
```diff
@@ -728,38 +783,33 @@ const Field3D Coordinates::Div_par(const Field3D &f, CELL_LOC outloc,
...
-    localmesh->communicate(sg);
-    sg = interp_to(sg, outloc);
-  }
-
-  result = ::DDY(f, outloc);
+  result = ::DDY(f, outloc, method);
 
-  r2 = D2DY2(f, outloc) / interp_to(g_22, outloc);
+  r2 = D2DY2(f, outloc, method) / g_22;
```
d7919: I don't think we know outloc is exactly location so I'm not sure we can remove the `interp_to`.

src/mesh/difops.cxx
```diff
@@ -167,15 +167,15 @@ const Field3D Grad_parP(const Field3D &apar, const Field3D &f) {
 * vparallel times the parallel derivative along unperturbed B-field
 *******************************************************************************/
 
-const Field2D Vpar_Grad_par(const Field2D &v, const Field2D &f, const CELL_LOC outloc) {
-  return f.getCoordinates(outloc)->Vpar_Grad_par(v, f, outloc);
+const Field2D Vpar_Grad_par(const Field2D &v, const Field2D &f, const CELL_LOC outloc, const DIFF_METHOD method) {
```
d7919: I'm not sure why `test-code-style` doesn't catch this (and it's not new code) but usually we don't allow (for style reasons) the `const` qualifier on things that are passed by value. I've probably just misremembered/interpreted the style rules.

src/mesh/difops.cxx
```diff
@@ -784,7 +789,7 @@ const Field2D bracket(const Field2D &f, const Field2D &g, BRACKET_METHOD method,
     result = 0.0;
   }else {
     // Use full expression with all terms
-    result = b0xGrad_dot_Grad(f, g) / f.getCoordinates(result_loc)->Bxy;
+    result = b0xGrad_dot_Grad(f, g, outloc) / f.getCoordinates(result_loc)->Bxy;
```
d7919: presumably outloc and result_loc must be the same here?

src/mesh/coordinates.cxx
```diff
@@ -28,23 +28,30 @@ Coordinates::Coordinates(Mesh *mesh)
...
 
-  if (mesh->get(dx, "dx")) {
+  // If the source does not include y-boundary guard cells then communications
+  // are needed to fill them. This provides backwards compatibility even though
+  // the communicated guard cell values may not be correct, because the
+  // integrated shear is not necessarily continuous when poloidal angle goes
+  // from 2pi->0
+  bool need_comms = !mesh->sourceHasYGuards();
+
+  if (localmesh->get(dx, "dx", 0., need_comms)) {
```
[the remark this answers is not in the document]
johnomotani: think i'm missing your point... need_comms is a local variable, the functionality is in `sourceHasYGuards()` which is a member of Mesh.

src/mesh/coordinates.cxx
johnomotani: could re-order the handling of CELL_DEFAULT like we had in the index_derivs functions to make it more obvious, but I think location==outloc always so it's OK without the interp_to.

src/mesh/difops.cxx
johnomotani: I thought it looked unusual, but didn't know why... I'll make a commit that gets rid of those const's.

src/mesh/difops.cxx
johnomotani: Currently that's what `bracket_location` does. I think it's another leftover from trying to interpolate the result of derivatives. We could tidy up by getting rid of it and just asserting that all the locations are the same, since I don't think we support anything else.

src/mesh/coordinates.cxx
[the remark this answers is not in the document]
d7919: If `sourceHasGuards` is all that is required then couldn't we put the call to this inside `mesh::get` so that `mesh::get` does the right thing?

src/mesh/difops.cxx
d7919: If we don't need it then I'd be in favour of removing it. My real thought here was that if we know that `outloc==result_loc` has to be true so we shouldn't need both symbols.

src/mesh/coordinates.cxx
```diff
@@ -28,23 +28,30 @@ Coordinates::Coordinates(Mesh *mesh)
...
     output_warn.write("\tWARNING: differencing quantity 'dx' not found. Set to 1.0\n");
     dx = 1.0;
   }
 
-  if (mesh->periodicX) {
-    mesh->communicate(dx);
+  if (localmesh->periodicX) {
+    localmesh->communicate(dx);
```
ZedThree: We're potentially communicating `dx` above, but also doing so here. I appreciate it's for a different reason, but it looks a bit odd

src/mesh/coordinates.cxx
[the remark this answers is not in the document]
johnomotani: I was originally going to load zShift in Coordinates (instead of only in ShiftedMetric), and pass 'false' for that, but I think there might be a better solution, and in any case is better to add the extra argument only when it's actually needed, so I'll put the call inside Mesh::get as you suggest.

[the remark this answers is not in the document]
d7919: This should all only be done once (or a few times) during initialisation so even if we pay a few extra communications that aren't strictly needed (as we overwrite guards etc.) then I think that is ok if it helps make the code cleaner (e.g. not having to repeat the default argument everywhere).

ZedThree:
This is very big and complicated, so it's not ideal for it to go straight into 4.2 right now. Especially given the changes to grid loading, etc.

We're also starting to put more of a hard dependency on the Python API, and if that is the case, we need to make a concerted effort to really polish that. For example, it needs extensive tests to verify everything is wrapped correctly, more documentation on the individual build steps and files, less reliance on bash scripts to parse C++, etc.

Let's not rush to get this in now, but spend a bit of time sorting out some of the details, like the grid sources, and get it in 4.3

PR #1319, opened 2018-10-12
Boundary conditions fixes

johnomotani:
Fixes 2 bugs:
* calling applyBoundary() on a field with location `CELL_ZLOW` did nothing, because the conditionals missed CELL_ZLOW.
* Fields created on meshes that aren't the global mesh could be wrong because the methods in boundary_standard.cxx all used the global mesh.


src/mesh/boundary_standard.cxx
```diff
@@ -271,6 +271,8 @@ void BoundaryDirichlet::apply(Field2D &f,BoutReal t) {
           }	
         }
       }
+    } else {
+      throw BoutException("Unhandled staggering in BoundaryDirichlet.");
```
ZedThree: Could we bring all these `throw`s to the top of the functions as explicit preconditions? I guess with `ASSERT1`, but `ASSERT0` might also be ok?

src/mesh/boundary_standard.cxx
```diff
@@ -23,88 +23,92 @@
     lead to an out of bounds access error later but we add it here to provide a
     more explanatory message.
  */
-void verifyNumPoints(BoundaryRegion *region, int ptsRequired) {
-  TRACE("Verifying number of points available for BC");
+namespace {
+  void verifyNumPoints(BoundaryRegion *region, int ptsRequired) {
```
ZedThree: I think you've just added this to an anonymous namespace and indented it? Github is not very helpful, even with whitespace changes hidden

johnomotani: yep, that's all I did.

src/mesh/boundary_standard.cxx
johnomotani: I re-ordered the if-else statements. Mega whitespace changes: github does a terrible job with the diff, but if you pull the branch `git -w HEAD^^` does a much better job of ignoring the whitespace changes...

src/field/field_data.cxx
```diff
@@ -6,8 +6,11 @@
 #include <field_factory.hxx>
 #include "unused.hxx"
 
-FieldData::FieldData() : boundaryIsCopy(false), boundaryIsSet(true) {
-  
+FieldData::FieldData(Mesh* m) :
+    fielddatamesh(m), boundaryIsCopy(false), boundaryIsSet(true) {
```
d7919: I think this could be `fielddatamesh(m? m : mesh)` to avoid the need for the logic inside the function.

src/field/field_data.cxx
```diff
@@ -24,17 +27,17 @@ void FieldData::setBoundary(const string &name) {
   
   output_info << "Setting boundary for variable " << name << endl;
   /// Loop over the mesh boundary regions
-  for(const auto& reg : mesh->getBoundaries()) {
+  for(const auto& reg : getDataMesh()->getBoundaries()) {
```
d7919: I think the constructor ensures that fielddatamesh can never be `nullptr` so it might be ok to just reference `fielddatamesh` here rather than asking `getDataMesh` to return it?

src/field/vector2d.cxx
```diff
@@ -148,6 +149,8 @@ Vector2D* Vector2D::timeDeriv() {
 /////////////////// ASSIGNMENT ////////////////////
 
 Vector2D & Vector2D::operator=(const Vector2D &rhs) {
+  fielddatamesh = rhs.fielddatamesh;
```
d7919: I don't recall seeing similar statements for operator= in the fields, should there be?

src/mesh/boundary_standard.cxx
d7919: (and replaced mesh with localmesh!)

src/mesh/boundary_standard.cxx
```diff
@@ -121,7 +125,13 @@ void BoundaryDirichlet::apply(Field2D &f){
 void BoundaryDirichlet::apply(Field2D &f,BoutReal t) {
   // Set (at 2nd order) the value at the mid-point between the guard cell and the grid cell to be val
   // N.B. Only first guard cells (closest to the grid) should ever be used
-  
+
+  Mesh* localmesh = f.getMesh();
+
+  // Check for staggered grids
+  CELL_LOC loc = f.getLocation();
+  ASSERT1(localmesh->StaggerGrids || loc == CELL_CENTRE);
```
d7919: I think this can never fail -- it shouldn't be possible to set a location other than CELL_CENTRE if staggered grids are disabled. I guess it doesn't hurt much to have a check here, but our units tests should (hopefully) tell us if this condition is ever accidentally relaxed.

src/field/field_data.cxx
ZedThree: I'd keep the explicit `nullptr` check though:

`fielddatamesh(m == nullptr ? m : mesh)`

src/mesh/boundary_standard.cxx
ZedThree: True, I don't think this is an enlightening precondition, but as you say, it doesn't hurt.

src/field/field_data.cxx
ZedThree: I've fixed this, and also moved it into the header -- the ctor body is now empty

src/field/field_data.cxx
johnomotani: I introduced this because of errors with the MMS/diffusion test: if a Field3D is declared in global scope, then it's constructor is called before the global `mesh` is allocated/initialized. Then fielddatamesh will be `nullptr` unless `getDataMesh()` is called to replace `nullptr` with the value of `mesh` once `mesh` has been initialized. So i think `getDataMesh` is necessary.

johnomotani: PS is it better to have the check in `getDataMesh()` be `if (fielddatamesh != nullptr)` instead of `if (fielddatamesh)`?

src/field/vector2d.cxx
johnomotani: I think there should be... I will add.

@ZedThree could you push the changes you just made please?

ZedThree:
I had forgotten how horrifying the boundary conditions code is. Slightly regretting saying anything now...

@johnomotani I believe you've done the following in each `BoundaryCondition::apply`:

- Turned this conditional into an `ASSERT` precondition:
```cpp
  CELL_LOC loc = f.getLocation();	
  if(mesh->StaggerGrids && loc != CELL_CENTRE) {
```
- Unindented the rest of the function

There are now three top-level branches in each `::apply`:
1. `CELL_XLOW`
2. `CELL_YLOW`
3. everything else, i.e. `CELL_CENTRE` and `CELL_ZLOW`

Is that correct?

src/field/vector2d.cxx
johnomotani: While looking at `Field3D::operator=(const Field2D &rhs)`, I have a question; should we set `fieldmesh = rhs.getMesh();`, or `ASSERT1(fieldmesh == rhs.getMesh());` in this operator?

src/field/field_data.cxx
ZedThree: I'm ok with this, but stuff declared at global scope is a bit dubious

src/field/vector2d.cxx
ZedThree: Done

johnomotani:
@ZedThree yes, exactly.

src/field/field2d.cxx
```diff
@@ -249,6 +249,7 @@ Field2D &Field2D::operator=(const Field2D &rhs) {
 
   // Copy the data and data sizes
   fieldmesh = rhs.fieldmesh;
+  fielddatamesh = rhs.fielddatamesh;
```
d7919: Do we need to check that fieldmesh/fielddatamesh are identical? If so why do we need two variables?

ZedThree: They definitely should be the same. We should `ASSERT1` in `Field2D` and `Field3D`.

There are two because there's one in each of the two base classes. I think this points towards our class hierarchy being a little weird

d7919: I was wondering if this might be an opportunity to address the class structure? We're currently introducing more opportunities to form an inconsistent state (albeit with appropriate checks to try to guard against this). All vectors and fields apart from fieldperp have fielddata and all fields have Field. I'd guess FieldPerp should also have fielddata (and if not why not?) and if so then it looks like `FieldData` is the generic stuff to all field like objects (which would presumably include getting the single relevant mesh) whilst `Field` provides the things fields have that vectors don't (coordinates, anything else?).

d7919: Or could Field inherit from FieldData?

src/field/vector2d.cxx
d7919: I think to be consistent with other places this should probably be an assert. If it is an assert we should ensure we assert for `Field3D::operator=(const Field3D &rhs)` etc. as well to be consistent.

src/field/vector3d.cxx
```diff
@@ -160,6 +163,8 @@ Vector3D & Vector3D::operator=(const Vector3D &rhs) {
 }
 
 Vector3D & Vector3D::operator=(const Vector2D &rhs) {
+  fielddatamesh = rhs.x.getMesh();
```
d7919: Should this be `getDataMesh()` for consistency? (I know they should return the same thing)

src/mesh/boundary_standard.cxx
```diff
@@ -23,88 +23,92 @@
     lead to an out of bounds access error later but we add it here to provide a
     more explanatory message.
  */
-void verifyNumPoints(BoundaryRegion *region, int ptsRequired) {
-  TRACE("Verifying number of points available for BC");
+namespace {
```
d7919: Are all the changes in this file separate from the FieldData gaining a mesh pointer?

src/mesh/boundary_standard.cxx
```diff
@@ -318,178 +325,169 @@ void BoundaryDirichlet::apply(Field3D &f,BoutReal t) {
...
+  if( loc == CELL_XLOW ) {
+    // X boundary, and field is shifted in X
+
+    if(bndry->bx > 0) {
+      // Outer x boundary
+
+      for(; !bndry->isDone(); bndry->next1d()) {
+        BoutReal xnorm = 0.5*(   localmesh->GlobalX(bndry->x)
+            + localmesh->GlobalX(bndry->x - bndry->bx) );
+        BoutReal ynorm = localmesh->GlobalY(bndry->y);
```
d7919: Should we introduce `Mesh::GlobalZ` for symmetry?

src/mesh/boundary_standard.cxx
```diff
@@ -318,178 +325,169 @@ void BoundaryDirichlet::apply(Field3D &f,BoutReal t) {
...
+            int xi = bndry->x + i*bndry->bx;
+            int yi = bndry->y ;
+
+            f(xi, yi, zk) = 2*f(xi - bndry->bx, yi , zk) - f(xi - 2*bndry->bx, yi, zk);
+          }
+        }
+      }
+    }
+  } else {
+    // CELL_CENTRE or CELL_ZLOW
```
d7919: Do we not care about the Z location in the `CELL_ZLOW` case for 3D fields?

src/mesh/boundary_standard.cxx
johnomotani: I guess the changes are separate in the sense that they could all be made without adding a mesh pointer to FieldData, but the mesh pointer in FieldData is required for the boundary conditions to work correctly when the fieldmesh is not the same as the global mesh.
* in verifyNumPoints we need to get localmesh from the `BoundaryRegion *region`; `BoundaryRegion*`s are created in a method of `FieldData`; `FieldData` had to gain a mesh pointer so that `BoundaryRegion*`s aren't always created with the global `mesh`.
* the `BoundaryRegion* bndry` of the `BoundaryOp`s could previously be wrong if the global and local `Mesh` are different (e.g. wrong number of points, or wrong number of guard cells, etc.).

src/mesh/boundary_standard.cxx
johnomotani: Going through all the boundary operators again will be painful... There's a lot of cut-and-paste code in boundary_standard.cxx, we could probably cut it down massively by having a single set of `BoundaryOp::apply()` methods implemented for the `BoundaryOp` class, and just having the single point stencils, one staggered and one unstaggered, for each derived class.

That would change quite a lot of code (on the other hand, I've touched most of the lines in boundary_standard.cxx already, at least for whitespace) so maybe you'd rather put off until 4.3. Also, since probably nobody is using CELL_ZLOW it doesn't matter too much if it stays broken for a little while longer (at least it's less broken now, constant value boundary conditions work).

I don't like leaving things broken though, so I'll try to make a fixed version, but I'll put it in a new PR.

ZedThree: I think we're going to put this PR off till 4.3 -- there's a few questions about `Field`/`FieldData` and the classes that do and don't inherit from them that we really need to resolve, and that will involve some possibly major structural changes.

All the boundary stuff needs a thorough look through and rethink as well. Can we use `Region`, or a subclass instead of `BoundaryRegion`? Is there a way we can template the boundary conditions on simple kernels and stencils? @d7919 has been doing something like this for derivatives, so maybe we can extend that work to here as well.

bendudson:
Replaced by #1334

PR #1322, opened 2018-10-14
Check TwistShift==true in ShiftedMetric

johnomotani:
The ShiftedMetric class requires TwistShift==true to get yup/ydown fields correct at the branch cut where poloidal angle and zShift jump.

Add a getter method to return the value of TwistShift from the Mesh, and use it in ShiftedMetric to throw an exception if TwishShift==false.

This check isn't strictly necessary if a simulation doesn't have a closed field line region, but I think it's simpler to just check always and make a few people possibly doing SOL-only simulations set TwistShift=true. Anyway, hopefully for 4.3 we can change so TwistShift doesn't need to be switched on for ShiftedMetric (see comments on #1294), and then the default (TwistShift=false) will be correct.


ZedThree:
This has also prompted lots of discussion! I'm wary of adding more things to the `Mesh` interface, especially as it might tie us up later. There's several other ways we could implement twist-shift, and one of them is bound to be better than how we currently do it. Moving twist-shift into `ParallelTransform` for example.

I'm also worried this check might be overzealous. It might also catch out simulations of linear devices for example. `test-yupydown` certainly gets caught out (this would need fixing at the very least!).

A possible alternative might be something like this:

```cpp
  bool twistshift = Options::root()["TwistShift"].withDefault(false);
  bool shift_without_twist = Options::root()["ShiftWithoutTwist"].withDefault(false);
  if (!twistshift and !shift_without_twist) {
    throw BoutException("ShiftedMetric usually requires the option TwistShift=true\n"
                        "    Use ShiftWithouttwist=true to use ShiftedMetric without TwistShift");
  }
```

This gives a nice warning message, along with an option for suppressing the warning (well, an exception still, to make sure people see it). Thanks to @bendudson for the suggestion!

It's a bit hacky, as it's directly getting the `TwistShift` input option without going through `Mesh`, and adds a new option that we might drop later.

Yet another option is to do nothing yet. Also problematic!

johnomotani:
I think we could get away without the second option? In cases like linear devices, TwistShift does nothing anyway (they have no closed field line region), so it should be safe to just set it to true.

I'd be happy with checking the global option instead of adding a method to Mesh though. They're both a bit hacky, but as you say going without changing Mesh makes it easier to update later without worrying about changing interfaces :+1:

johnomotani:
Ah I see, `TwistShift=true` requires `ShiftAngle` to be set somewhere. We do need a second option to avoid that...

PR #1331, opened 2018-10-19
Remove more deprecated derivs boundaries

d7919:
Follows on from #1326


ZedThree:
Not sure why the push test is failing, I can't reproduce. The PR tests succeeds though, so that should be sufficient.

I am a bit worried there's some UB hidden somewhere though, and we've just plastered over it.

PR #1334, opened 2018-10-21
BoundaryOp refactor

johnomotani:
Simplify boundary_standard.cxx:
* Most boundary conditions now use BoundaryOp::apply, which handles looping and staggering. The derived classes BoundaryNeumann, etc. mostly provide just methods to apply the boundary condition at a point, e.g.
https://github.com/boutproject/BOUT-dev/blob/cf7aa24b0ee2870623cb398fb22dd81b34b7b0a8/src/mesh/boundary_standard.cxx#L415-L445
BoundaryDirichlet still has its own apply() method because it handled 2nd (and 3rd, 4th, etc.) guard cells in a non-standard way, setting them equal to the boundary value expression (evaluated at the guard cell's location); I have left this as-is. BoundaryNeumann_NonOrthogonal, BoundaryZeroLaplace, BoundaryZeroLaplace2 and BoundaryConstLaplace also implement overriding apply methods.
* Remove BoundaryOpBase class. This is replaced by using templates to implement several methods for both BoundaryOp and BoundaryOpPar with a minimum of code duplication. Not totally sure this form is better/clearer. It does avoid several `static_cast`s. Could undo this, but would take a little digging because I didn't do a good job of implementing this refactor in separate commits.

Fixes a few bugs: setting boundary conditions for fields at CELL_ZLOW (previously this case fell though and the guard cells were not changed); use the fieldmesh of the field, not the global mesh, in all boundary conditions; correct one of the coefficients in `BoundaryDirichlet_4thOrder` (now renamed to `BoundaryDirichlet_O5`; implement BoundaryNeumann_O4 for staggered grids.

Add `GlobalZ` method to Mesh for consistency with `GlobalX`, `GlobalY`.

Add a couple of higher order extrapolating boundary conditions, `free_o4` and `free_o5`.

Change `BoundaryRobin` to take account of grid spacing, so now the derivative coefficient passed in (`b`) does not have to include dx or dy. Effectively derivative in boundary condition is now DDX or DDY instead of indexDDX or indexDDY.

Test the order of accuracy of Dirichlet, Neumann and free boundary conditions in MMS/derivatives3.

Add new boundary condition `dirichlet_smooth`, which suppresses grid-scale oscillations at the boundary enough to be used in `wave-1d`/`wave-1d-y` MMS tests. It has the same accuracy as the standard `dirichlet` condition, but uses two grid points to set the boundary condition. Setting the guard cell g by extrapolating from the boundary value, v, and first grid cell, f1 would give 'g1 = 2*v-f1'; setting from the second grid cell would give 'g2 = 4/3*v-1/3*f2'; this bc takes the average of the two, setting 'g = 4/3*v-1/2*f1-1/6*f2'.

Test both `dirichlet_smooth` and `neumann` boundary conditions for `StaggerGrids=true` or `false` in `MMS/wave-1d` and `MMS/wave-1d-y`.

Test both `dirichlet` and `neumann` boundary conditions on both boundaries in `MMS/diffusion`.

### DIFF
--- a/include/boundary_factory.hxx
+++ b/include/boundary_factory.hxx
@@ -14,6 +14,12 @@ class BoundaryFactory;
 using std::string;
 using std::map;
 
+template<typename T>
+using BoundaryOpRegion = typename std::conditional<std::is_same<T, BoundaryOpPar>::value, BoundaryRegionPar, BoundaryRegion>::type;
+
+template<typename T>
+using BoundaryRegionOp = typename std::conditional<std::is_same<T, BoundaryRegionPar>::value, BoundaryOpPar, BoundaryOp>::type;
+
 /// Create BoundaryOp objects on demand
 /*!
  * This implements a simple string parser, used to match boundary condition
@@ -55,7 +61,7 @@ using std::map;
  * Subsequent calls to create() or createFromOptions() can make use
  * of the boundary type "myboundary".
  *
- * BoundaryOpBase *bndry = bf->create("myboundary()", new BoundaryRegionXOut("xout", 0, 10, localmesh));
+ * BoundaryOp *bndry = bf->create("myboundary()", new BoundaryRegionXOut("xout", 0, 10, localmesh));
  * 
  * where the region is defined in boundary_region.hxx
  * 
@@ -69,12 +75,16 @@ class BoundaryFactory {
   static void cleanup(); ///< Frees all memory
 
   /// Create a boundary operation object
-  BoundaryOpBase* create(const string &name, BoundaryRegionBase *region);
-  BoundaryOpBase* create(const char* name, BoundaryRegionBase *region);
+  template<typename T>
+  BoundaryRegionOp<T>* create(const string &name, T* region);
+  template<typename T>
+  BoundaryRegionOp<T>* create(const char* name, T* region);
 
   /// Create a boundary object using the options file
-  BoundaryOpBase* createFromOptions(const string &varname, BoundaryRegionBase *region);
-  BoundaryOpBase* createFromOptions(const char* varname, BoundaryRegionBase *region);
+  template<typename T>
+  BoundaryRegionOp<T>* createFromOptions(const string &varname, T* region);
+  template<typename T>
+  BoundaryRegionOp<T>* createFromOptions(const char* varname, T* region);
 
   /*!
    * Add available boundary conditions and modifiers
@@ -119,14 +129,18 @@ class BoundaryFactory {
   // map<string, BoundaryModifier*> par_modmap;
 
   // Functions to look up operations and modifiers
-  BoundaryOp* findBoundaryOp(const string &s);
+  // Standard or parallel boundary conditions
+  template<typename T>
+  T* findBoundaryOp(const string &s);
+  // Boundary modifiers
   BoundaryModifier* findBoundaryMod(const string &s);
-  // Parallel boundary conditions
-  BoundaryOpPar* findBoundaryOpPar(const string &s);
-  // To be implemented...
-  // BoundaryModifier* findBoundaryMod(const string &s);
 
 };
+
+template<>
+BoundaryOp* BoundaryFactory::findBoundaryOp<BoundaryOp>(const string &s);
+template<>
+BoundaryOpPar* BoundaryFactory::findBoundaryOp<BoundaryOpPar>(const string &s);
 
 #endif // __BNDRY_FACTORY_H__
 
--- a/include/boundary_op.hxx
+++ b/include/boundary_op.hxx
@@ -18,16 +18,29 @@ class BoundaryModifier;
 using std::string;
 using std::list;
 
-class BoundaryOpBase {
+/// An operation on a boundary
+class BoundaryOp {
 public:
-  BoundaryOpBase() {}
-  virtual ~BoundaryOpBase() {}
+  BoundaryOp()  : bndry(nullptr), apply_to_ddt(false), gen(nullptr) {}
+  BoundaryOp(BoundaryRegion *region)
+    : bndry(region), apply_to_ddt(false), gen(nullptr) {}
+  BoundaryOp(BoundaryRegion *region, std::shared_ptr<FieldGenerator> g)
+    : bndry(region), apply_to_ddt(false), gen(std::move(g)) {}
+  virtual ~BoundaryOp() {}
+
+  // Note: All methods must implement clone, except for modifiers (see below)
+  virtual BoundaryOp* clone(BoundaryRegion *UNUSED(region), const list<string> &UNUSED(args)) {
+    ASSERT1(false); // this implementation should never get called
+    return nullptr;
+  }
 
   /// Apply a boundary condition on field f
-  virtual void apply(Field2D &f) = 0;
-  virtual void apply(Field2D &f,BoutReal UNUSED(t)){return apply(f);}//JMAD
-  virtual void apply(Field3D &f) = 0;
-  virtual void apply(Field3D &f,BoutReal UNUSED(t)){return apply(f);}//JMAD
+  virtual void apply(Field2D &f,BoutReal t = 0.) {
+    applyTemplate<Field2D>(f, t);
+  }
+  virtual void apply(Field3D &f,BoutReal t = 0.) {
+    applyTemplate<Field3D>(f, t);
+  }
 
   virtual void apply(Vector2D &f) {
     apply(f.x);
@@ -40,30 +53,10 @@ public:
     apply(f.y);
     apply(f.z);
   }
-};
-
-/// An operation on a boundary
-class BoundaryOp : public BoundaryOpBase {
-public:
-  BoundaryOp() {
-    bndry = nullptr;
-    apply_to_ddt = false;
-  }
-  BoundaryOp(BoundaryRegion *region) {bndry = region; apply_to_ddt=false;}
-  ~BoundaryOp() override {}
-
-  // Note: All methods must implement clone, except for modifiers (see below)
-  virtual BoundaryOp* clone(BoundaryRegion *UNUSED(region), const list<string> &UNUSED(args)) {
-    return nullptr;
-  }
 
   /// Apply a boundary condition on ddt(f)
-  virtual void apply_ddt(Field2D &f) {
-    apply(ddt(f));
-  }
-  virtual void apply_ddt(Field3D &f) {
-    apply(ddt(f));
-  }
+  virtual void apply_ddt(Field2D &f);
+  virtual void apply_ddt(Field3D &f);
   virtual void apply_ddt(Vector2D &f) {
     apply(ddt(f));
   }
@@ -73,6 +66,41 @@ public:
 
   BoundaryRegion *bndry;
   bool apply_to_ddt; // True if this boundary condition should be applied on the time derivatives, false if it should be applied to the field values
+
+protected:
+  std::shared_ptr<FieldGenerator> gen; // Generator
+
+  // Apply boundary condition at a point
+  virtual void applyAtPoint(Field2D &UNUSED(f), BoutReal UNUSED(val), int
+      UNUSED(x), int UNUSED(bx), int UNUSED(y), int UNUSED(by), int UNUSED(z),
+      Coordinates* UNUSED(metric)) {
+    ASSERT1(false);
+  }
+  virtual void applyAtPoint(Field3D &UNUSED(f), BoutReal UNUSED(val), int
+      UNUSED(x), int UNUSED(bx), int UNUSED(y), int UNUSED(by), int UNUSED(z),
+      Coordinates* UNUSED(metric)) {
+    ASSERT1(false);
+  }
+
+  // Apply to staggered grid
+  virtual void applyAtPointStaggered(Field2D &UNUSED(f), BoutReal UNUSED(val),
+      int UNUSED(x), int UNUSED(bx), int UNUSED(y), int UNUSED(by), int
+      UNUSED(z), Coordinates* UNUSED(metric)) {
+    ASSERT1(false);
+  }
+  virtual void applyAtPointStaggered(Field3D &UNUSED(f), BoutReal UNUSED(val),
+      int UNUSED(x), int UNUSED(bx), int UNUSED(y), int UNUSED(by), int
+      UNUSED(z), Coordinates* UNUSED(metric)) {
+    ASSERT1(false);
+  }
+
+  // extrapolate to further guard cells
+  virtual void extrapFurther(Field2D &f, int x, int bx, int y, int by, int z);
+  virtual void extrapFurther(Field3D &f, int x, int bx, int y, int by, int z);
+
+private:
+  template<typename T>
+  void applyTemplate(T &f, BoutReal t);
 };
 
 class BoundaryModifier : public BoundaryOp {
@@ -80,6 +108,7 @@ public:
   BoundaryModifier() : op(nullptr) {}
   BoundaryModifier(BoundaryOp *operation) : BoundaryOp(operation->bndry), op(operation) {}
   virtual BoundaryOp* cloneMod(BoundaryOp *op, const list<string> &args) = 0;
+  virtual BoundaryOpPar* cloneMod(BoundaryOpPar *UNUSED(op), const list<string> &UNUSED(args)) { throw BoutException("BoundaryModifier should not be called on a BoundaryOpPar."); }
 protected:
   BoundaryOp *op;
 };
--- a/include/boundary_region.hxx
+++ b/include/boundary_region.hxx
@@ -20,40 +20,15 @@ enum BndryLoc {BNDRY_XIN=1,
                BNDRY_PAR_FWD=16,   // Don't include parallel boundaries
                BNDRY_PAR_BKWD=32};
 
-class BoundaryRegionBase {
-public:
-
-  BoundaryRegionBase() = delete;
-  BoundaryRegionBase(std::string name, Mesh *passmesh = nullptr)
-      : localmesh(passmesh ? passmesh : mesh), label(std::move(name)) {}
-  BoundaryRegionBase(std::string name, BndryLoc loc, Mesh *passmesh = nullptr)
-      : localmesh(passmesh ? passmesh : mesh), label(std::move(name)), location(loc) {}
-
-  virtual ~BoundaryRegionBase() {}
-
-  Mesh* localmesh; ///< Mesh does this boundary region belongs to
-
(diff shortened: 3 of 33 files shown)


johnomotani:
This should replace #1319.

include/boundary_op.hxx
```diff
@@ -73,13 +66,49 @@ public:
...
   bool apply_to_ddt; // True if this boundary condition should be applied on the time derivatives, false if it should be applied to the field values
+
+protected:
+  std::shared_ptr<FieldGenerator> gen; // Generator
+
+  // Apply boundary condition at a point
+  virtual void applyAtPoint(Field2D &UNUSED(f), BoutReal UNUSED(val), int
+      UNUSED(x), int UNUSED(bx), int UNUSED(y), int UNUSED(by), int UNUSED(z),
+      Coordinates* UNUSED(metric)) {
+    ASSERT1(false);
```
d7919: This might lead to a confused user as the error message won't say what is wrong just that the assert failed so a throw might be better. 

I think if you want it to be an error to not override these routines then you can make these be pure virtual, e.g. 

~~~~
virtual void applyAtPoint(Field2D &f, BoutReal val, int x, int bx, int y, int by, int z,
      Coordinates* metric) = 0;
~~~~

and it's then a compile time error.

I notice you pass in the metric here, but you're also passing a field -- could we just get the metric from the field?

d7919: This might lead to a confused user as the error message won't say what is wrong just that the assert failed so a throw might be better. 

I think if you want it to be an error to not override these routines then you can make these be pure virtual, e.g. 

~~~~
virtual void applyAtPoint(Field2D &f, BoutReal val, int x, int bx, int y, int by, int z,
      Coordinates* metric) = 0;
~~~~

and it's then a compile time error.

I notice you pass in the metric here, but you're also passing a field -- could we just get the metric from the field?

d7919: This might lead to a confused user as the error message won't say what is wrong just that the assert failed so a throw might be better. 

I think if you want it to be an error to not override these routines then you can make these be pure virtual, e.g. 

~~~~
virtual void applyAtPoint(Field2D &f, BoutReal val, int x, int bx, int y, int by, int z,
      Coordinates* metric) = 0;
~~~~

and it's then a compile time error.

I notice you pass in the metric here, but you're also passing a field -- could we just get the metric from the field?

d7919: This might lead to a confused user as the error message won't say what is wrong just that the assert failed so a throw might be better. 

I think if you want it to be an error to not override these routines then you can make these be pure virtual, e.g. 

~~~~
virtual void applyAtPoint(Field2D &f, BoutReal val, int x, int bx, int y, int by, int z,
      Coordinates* metric) = 0;
~~~~

and it's then a compile time error.

I notice you pass in the metric here, but you're also passing a field -- could we just get the metric from the field?

d7919: This might lead to a confused user as the error message won't say what is wrong just that the assert failed so a throw might be better. 

I think if you want it to be an error to not override these routines then you can make these be pure virtual, e.g. 

~~~~
virtual void applyAtPoint(Field2D &f, BoutReal val, int x, int bx, int y, int by, int z,
      Coordinates* metric) = 0;
~~~~

and it's then a compile time error.

I notice you pass in the metric here, but you're also passing a field -- could we just get the metric from the field?

include/boundary_standard.hxx
```diff
@@ -10,107 +10,91 @@
...
+  using BoundaryOp::BoundaryOp; // inherit BoundaryOp constructors
   BoundaryOp* clone(BoundaryRegion *region, const list<string> &args) override;
 
   using BoundaryOp::apply;
-  void apply(Field2D &f) override;
-  void apply(Field2D &f,BoutReal t) override;
-  void apply(Field3D &f) override;
-  void apply(Field3D &f,BoutReal t) override;
+  void apply(Field2D &f,BoutReal t = 0.) override {
+    applyTemplate<Field2D>(f, t);
```
d7919: You probably don't actually need to specify `Field2D` or `Field3D` here as the compiler should be able to infer the type, although being explicit certainly doesn't hurt.

include/boundary_op.hxx
d7919: This might lead to a confused user as the error message won't say what is wrong just that the assert failed so a throw might be better. 

I think if you want it to be an error to not override these routines then you can make these be pure virtual, e.g. 

~~~~
virtual void applyAtPoint(Field2D &f, BoutReal val, int x, int bx, int y, int by, int z,
      Coordinates* metric) = 0;
~~~~

and it's then a compile time error.

I notice you pass in the metric here, but you're also passing a field -- could we just get the metric from the field?

johnomotani:
> Does it take only a uniform dx/dy into account, or also a non-uniform dx/dy?

@dschwoerer no, the Robin bc doesn't account for variation of dx/dy, just uses the value dx/dy at the boundary grid point. Sorry, I hadn't looked at #1179 before now.

include/boundary_op.hxx
d7919: This might lead to a confused user as the error message won't say what is wrong just that the assert failed so a throw might be better.

I think if you want it to be an error to not override these routines then you can make these be pure virtual, e.g.

~~~~
virtual void applyAtPoint(Field2D &f, BoutReal val, int x, int bx, int y, int by, int z,
      Coordinates* metric) = 0;
~~~~

and it's then a compile time error.

I notice you pass in the metric here, but you're also passing a field -- could we just get the metric from the field?

johnomotani: I implemented like this because subclasses don't have to implement these methods if they override `apply()`. I didn't want to have to put some sort of null version of these four methods in subclasses that don't need them, so had to provide a default implementation. Having at least a helpful error message is a good call.

Another possibilty would be to take `applyAtPoint`, `applyAtPointStaggered` and `extrapPoint` and put them in a subclass like `BoundaryOpWithApply` that inherits from `BoundaryOp` and implements the generic `apply` method. Then the actual implementations could inherit from `BoundaryOp` if they provide `apply` or from `BoundaryOpWithApply` and provide the low-level methods.

I think the `BoundaryOpWithApply` subclass could be a case for using the CRTP design https://en.wikipedia.org/wiki/Curiously_recurring_template_pattern; CRTP is extra code and (to me at least) adds work for the reader to understand, but might have a performance benefit: [from the wikipedia article] _"However, if base class member functions use CRTP for all member function calls, the overridden functions in the derived class will be selected at compile time. This effectively emulates the virtual function call system at compile time without the costs in size or function call overhead (VTBL structures, and method lookups, multiple-inheritance VTBL machinery) at the disadvantage of not being able to make this choice at runtime."_

Any preferences for:
1. Leaving all the methods in `BoundaryOp`
2. Providing a default `apply` method in an intermediate subclass `BoundaryOpWithApply` so that implementations either subclass `BoundaryOp` and implement `apply` or subclass `BoundaryOpWithApply` and provide `applyAtPoint`, etc.
3. Use CRTP to (if I understand right) avoid cost of virtual methods in 2.
4. Something else?

d7919: This might lead to a confused user as the error message won't say what is wrong just that the assert failed so a throw might be better.

I think if you want it to be an error to not override these routines then you can make these be pure virtual, e.g.

~~~~
virtual void applyAtPoint(Field2D &f, BoutReal val, int x, int bx, int y, int by, int z,
      Coordinates* metric) = 0;
~~~~

and it's then a compile time error.

I notice you pass in the metric here, but you're also passing a field -- could we just get the metric from the field?

d7919: This might lead to a confused user as the error message won't say what is wrong just that the assert failed so a throw might be better.

I think if you want it to be an error to not override these routines then you can make these be pure virtual, e.g.

~~~~
virtual void applyAtPoint(Field2D &f, BoutReal val, int x, int bx, int y, int by, int z,
      Coordinates* metric) = 0;
~~~~

and it's then a compile time error.

I notice you pass in the metric here, but you're also passing a field -- could we just get the metric from the field?

d7919:
Sorry for the duplicate comments!

d7919:
Sorry for the duplicate comments!

include/boundary_op.hxx
johnomotani: > I notice you pass in the metric here, but you're also passing a field -- could we just get the metric from the field?

Good point. Not sure if there are performance implications of having the f.getCoordinates() call on the inside of the loop? Passing `Coordinates* metric` as an argument will be gone in an update anyway, because it turned out to be convenient to calculate a grid spacing outside and pass that instead.

src/mesh/boundary_standard.cxx
```diff
@@ -23,473 +21,168 @@
...
+        ptsAvail = ptsAvailGlobal;
+        gridType = "global";
+      }else{
+        ptsAvail = ptsAvailLocal;
+        gridType = "local";
+      }
+
+      break;
+    }
+#if CHECK > 2 //Only fail on Unrecognised boundary for extreme checking
```
ZedThree: I know this isn't part of this PR, but I forget the rationale for turning this on only at high checking. The check is free, so what situation would we not want to check this?

src/mesh/boundary_standard.cxx
```diff
@@ -23,473 +21,168 @@
...
+
+#endif
   }
-#if CHECK > 2 //Only fail on Unrecognised boundary for extreme checking
-  default : {
-    throw BoutException("Unrecognised boundary region (%s) for verifyNumPoints.",region->location);
+
+  // 2nd order extrapolation to a point
+  template<typename T>
+  void extrap2nd(T &f, int x, int bx, int y, int by, int z) {
```
ZedThree: See comment about about `extrap`->`extrapolate`

src/mesh/boundary_standard.cxx
```diff
@@ -525,454 +218,238 @@ void BoundaryDirichlet::apply(Field3D &f,BoutReal t) {
...
-  Field2D *dt = f.timeDeriv();
-  for(bndry->first(); !bndry->isDone(); bndry->next())
-    (*dt)(bndry->x,bndry->y) = 0.; // Set time derivative to zero
-}
+            f(xi, yi, zk) = 2*f(xi - bndry->bx, yi , zk) - f(xi- 2*bndry->bx, yi , zk);
+          }
+        }
+      }
+    }
+    if (bndry->bx < 0){
```
ZedThree: While we're fixing things here, this should be `else if` -- `bndry.bx` can't be both `< 0` and `> 0`

src/mesh/boundary_standard.cxx
```diff
@@ -525,454 +218,238 @@ void BoundaryDirichlet::apply(Field3D &f,BoutReal t) {
...
+          for(int i=0;i<bndry->width;i++) {
+            int xi = bndry->x + i*bndry->bx;
+            int yi = bndry->y ;
 
+            f(xi, yi, zk) = 2*f(xi - bndry->bx, yi , zk) - f(xi- 2*bndry->bx, yi , zk);
+          }
+        }
+      }
+    }
+    if(bndry->by !=0){
```
ZedThree: Note that this branch can be taken in addition to the `bndry->bx` branches above -- is this correct?
The `bndry` iterator is not reset, it just carries on, and it's not clear to me at least that this would do the correct thing.

None of the current `BoundaryRegion`s has both `bx` and `by` `!= 0`, so what do we actually assume are valid values? Is it one or the other? Or can we really have "diagonal" boundaries? If not, then:

 - this should be an `else if`
 - the `BoundaryRegion` ctor should enforce exactly one non-zero `bx/by`

include/boundary_region.hxx
```diff
@@ -20,23 +20,30 @@ enum BndryLoc {BNDRY_XIN=1,
...
+  BoundaryRegion() = delete;
+  BoundaryRegion(std::string name, BndryLoc loc, Mesh *passmesh = nullptr)
       : localmesh(passmesh ? passmesh : mesh), label(std::move(name)), location(loc) {}
+  BoundaryRegion(std::string name, int xd, int yd, Mesh *passmesh = nullptr)
+      : bx(xd), by(yd), width(2), localmesh(passmesh ? passmesh : mesh), label(std::move(name)) {}
+  virtual ~BoundaryRegion() {}
 
-  virtual ~BoundaryRegionBase() {}
+  int x,y; ///< Indices of the point in the boundary
+  int bx, by; ///< Direction of the boundary [x+dx][y+dy] is going outwards
```
ZedThree: These should be `const`

include/boundary_region.hxx
```diff
@@ -20,23 +20,30 @@ enum BndryLoc {BNDRY_XIN=1,
...
+
+  int width; ///< Width of the boundary
```
ZedThree: `const`

include/boundary_factory.hxx
```diff
@@ -14,6 +14,12 @@ class BoundaryFactory;
 using std::string;
 using std::map;
 
+template<typename T>
+using BoundaryOpRegion = typename std::conditional<std::is_same<T, BoundaryOpPar>::value, BoundaryRegionPar, BoundaryRegion>::type;
+
+template<typename T>
+using BoundaryRegionOp = typename std::conditional<std::is_same<T, BoundaryRegionPar>::value, BoundaryOpPar, BoundaryOp>::type;
```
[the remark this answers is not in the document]
johnomotani: * `BoundaryOpRegion` is either a `BoundaryRegion` or `BoundaryRegionPar` depending on whether `T` is a `BoundaryOp` or `BoundaryOpPar`.
* `BoundaryRegionOp` is a `BoundaryOp/Par` depending on `T` being `BoundaryRegion/Par`.

Suggestions for more intelligent names welcome :+1:

src/mesh/boundary_standard.cxx
johnomotani: I think it only makes sense for a BoundaryRegion to have one of `bx` and `by` non-zero, not both. Will add a check to the constructor...

include/boundary_region.hxx
johnomotani: We can't easily make `width` a `const` because BoundaryWidth needs to change it.

I don't see a way around it: the closest to a workaround I can think of would be for BoundaryWidth to make a new BoundaryRegion for itself with a different width, but BoundaryWidth::clone doesn't know the sub-class of BoundaryRegion it has as it's bndry member, so we'd need a factory class for BoundaryRegions...

BoundaryWidth might be useful. I can at least think of one scenario: if you want to evolve fields on the y-boundaries, you need evolving `CELL_YLOW` fields at both boundaries. That's fine at ystart because the `mesh->ystart` point is at the boundary. But at yend, you'd need to evolve the `mesh->yend+1` point, but you can't. Instead, you could use the BoundaryWidth modifier to set boundary conditions on the other fields with a width of 3, allowing you to move the boundary in one point. I admit it's a bit unlikely because if I actually wanted to do something like this, I'd probably have to implement all the sheath boundary conditions by hand, but it's enough for me to think BoundaryWidth is worth keeping.

src/mesh/boundary_standard.cxx
johnomotani: I'll take out this `#if CHECK > 2` for now. It's easy to remove the commit if anyone remembers why it was there.

include/boundary_region.hxx
bendudson: Would another possibility be to make the width an input to the boundary constructor? As currently we can specify the value of the Dirichlet/Neumann boundary, we could add width as an optional argument. Support for keywords in the input file would be nice...

ZedThree: Ah, I didn't realise there was a `BoundaryWidth` modifier. That does complicate things a little bit.

It's possible most of the `BoundaryRegion`s could be replaced with some kind of wrapper around `Region`.

johnomotani: Could be nice to do as @bendudson suggests and make the width an argument to the boundary condition input. I've now replaced most of the method bodies for Boundary*::clone with two template functions, one that gets an expression from arguments and one that ignores arguments. So now there's only one function that needs changing to add a width argument. Would be easy to add 'width' now, but it seems unlikely to be used, so more sensible to wait for possible keyword arguments from options (?) rather than adding a positional argument that we have to keep supporting.

bendudson: I've added keywords in commit 115f86750f (branch `boundary-keywords`), but no testing yet. I based it on `next`, but could rebase it and merge into this branch if it's useful.

johnomotani: I've now merged #1346 and implemented a 'width' keyword. Then in the next commit made several member variables (including `BoundaryRegion::width`) `const`, and deprecated the BoundaryWidth modifier (throws an exception with a message to use the 'width' keyword).

johnomotani:
@bendudson what is the reason for having the BoundaryWidth modifier? It's being trickier than I'd anticipated to update; understanding the purpose would help make sure I don't break anything...

bendudson:
@johnomotani I'm a bit unsure what the purpose was of the width modifier, and since it doesn't seem to be used in any of the examples, it can probably be removed. At least I wouldn't worry too much about trying to preserve it. I think it was either for:
- Operators like C4
- Staggered cells, where in the simple implementation the boundary needs to apply to an extra point. This extra point is now handled properly (?) by the boundary condition operators.

johnomotani:
There was a comment in the manual about `BoundaryWidth` not being thread-safe because it changed the `width` of the `BoundaryRegion` while applying the `BoundaryOp` and then changed it back. That is no longer true, but are the boundary conditions thread-safe in general? I don't have a good understanding of what thread-safety means, but it looks to me like they are not, because they use shared `BoundaryRegion` objects as iterators, and `BoundaryRegion::first()`, `BoundaryRegion::next1d()`, etc. change the state of the shared object so if different threads were trying to apply boundary conditions to different fields at the same time, they might interfere with each other. Is that right? I've put a comment about the boundary conditions not being thread-safe in the manual, but someone who understands the concept better could probably write much more clearly...
See https://github.com/boutproject/BOUT-dev/blob/ebda6f316a184e419876a573a4bdd58a44b425aa/manual/sphinx/user_docs/boundary_options.rst#limitations

dschwoerer:
If you are not sure, could you not mention that at all?

Having wrong documentation is IMO way more confusing then no documentation.
I don't think BOUT++ generally makes any sort of guarantees in that
regard, thus saying one thing isn't thread safe implies the rest is -
which might be the case, but I don't know.

johnomotani:
> @johnomotani do you still intend to follow up on this?

No. I don't remember now what exactly the purpose was, so don't know if in principle it would be worth updating and merging. In any case I don't have any time to work on it now, so I'm not going to do anything. Anyone is welcome to take it over if they think it's useful, although at this point it might be easier to start over from scratch!

PR #1337, opened 2018-10-25
A couple of small bugfixes for Zoidberg

ZedThree:
Calling `FieldTracer.follow_field_lines` would fail at `if len(x_values) < 1000:` if `x_values` is scalar -- despite being a numpy array, it doesn't have a `len()`. Instead, use `np.atleast_1d` to make sure inputs are both numpy arrays and also at least 1D.

Also silences some odeint warnings in the straight stellarator test.

Lastly, runs pytest on Travis -- we only have python tests for zoidberg at the mo, but we should actually run them!

### DIFF
--- a/.pip_install_for_travis.sh
+++ b/.pip_install_for_travis.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 export PATH=${HOME}/.local/bin:${PATH}
 pip3 install --user --upgrade pip setuptools
+pip3 install --user --upgrade pip scipy numpy
 
 for package in $@
 do
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,7 +53,7 @@ matrix:
       - *default_env
       - CONFIGURE_OPTIONS='--enable-shared'
       - SCRIPT_FLAGS="-uim -t python -t shared"
-      - PIP_PACKAGES='netcdf4 cython'
+      - PIP_PACKAGES='netcdf4 cython sympy'
   - env:
       - *default_env
       - CONFIGURE_OPTIONS='--enable-openmp'
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ addons:
     - hdf5-tools
     - python3
     - python3-pip
+    - python3-pytest
     - python3-numpy
     - python3-scipy
     - lcov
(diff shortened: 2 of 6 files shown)


d7919:
How closely integrated with the rest of the bout lib is zoidberg -- should this have its own top level repository?

bendudson:
Zoidberg uses the bout python libraries for things like EQDSK file reading, but is otherwise quite independent. It could possibly be made separate, if that's the direction we want to go in. It would make sense if the core python routines were one package, Zoidberg another, boutcore a third etc. Then they might be a `pip install` away...

bendudson:
Ah, no `sympy` on Travis?

ZedThree:
Looks like it's only available for python 2. Also we need scipy > 0.14 (although requirements.txt says >1.0) and the system version is 0.13.

I'll try installing them through pip

ZedThree:
This also adds the python checks in same travis section as the unit tests. That may or may not be a good idea. We don't really need to run them in every configuration.

If we moved all the python stuff out into separate repos, they could be tested completely independently. Although some things like collect would need to be tied to the C++ library version as well some how.

d7919:
I think splitting things out into separate repositories might be quite nice. We can then seek to have them pip install-able and provide appropriate requirements files to ensure they pull in appropriate versions of the dependencies.

d7919:
(alternative would be to use git submodules or similar so cloning bout pulls these extra repositories in automatically).

ZedThree:
We'll need at least `collect` for the integrated tests, so probably submodules?

d7919:
True, although if it was pip available we can just add it to the list of python modules to install.

dschwoerer:
On 10/25/2018 05:13 PM, Ben Dudson wrote:
> Zoidberg uses the bout python libraries for things like EQDSK file
> reading, but is otherwise quite independent. It could possibly be made
> separate, if that's the direction we want to go in. It would make sense
> if the core python routines were one package, Zoidberg another, boutcore
> a third etc. Then they might be a |pip install| away...

I think that may not make as much sense for boutcore, as it requires
bout, and thus all it's dependencies, so I am not sure whether that is
beneficial to install with pip - as pip is not a general package
manager, and is generally not able to pull in (non-python) dependencies.
I am more in favor of packaging that via more capable package managers,
as rpm or apt.

Further, it makes some sense to run boutcore tests with different
configurations on travis, and when the BOUT++ library changes.

For the other python libraries, it makes more sense to separate them,
although `boututils` and `boutdata` could be merged.

TomNicholas:
@dschwoerer I might be wrong but boutcore would still fit perfectly well into the conda approach right? Conda is what is used by the pydata ecosystem (including numpy, scipy, xarray etc.). That's not mutually exclusive with `pip install`-ing `boutdata` and `boututils` either

PR #1338, opened 2018-10-26
Boutexception backtrace tidy

ZedThree:
A couple of changes to `BoutException`:

- Mixing the backtrace and the message causes problems with tests. Backtrace functionality is still present, but moved into `BoutException::getBacktrace`. `BOUTMAIN` now calls that instead of `what()` so that physics models still print a backtrace if it is available.

- Clear the MsgStack in BoutException destructor: If an exception is thrown while a TRACE is active, we won't clear up the msg_stack. We also won't know how many messages to pop, so just clear everything. This should make the unit tests less confusing if something goes wrong

Also improves the error message in `SolverFactory::createSolver` which needs the first change.


ZedThree:
Related, this looks pretty cool: https://github.com/bombela/backward-cpp

dschwoerer:
Note that this makes debugging in boutcore much harder, as Cython calls `.what()` to get the error message, and now that does not contain the backtrace.

The third commit does not require the other ones. We only catch BoutException, so we can add an getter for `e.message`.

In my opinion the main question we should discuss is, do we want the exception message to contain the information if BOUT++ is called from other codes, or do we only want to contain the backtrace only when the full BOUT++ framework is used to run a simulation?

Imho I think the backtrace is often very important. As an example you can use the third commit. The changed exception message does not explain on it self what was tried to achieve, and what failed. I thought first that a file did not exist. I am not saying we should make that message more clear. There are many messages like that. The backtrace gives that information for very little cost (makes the message much longer if not needed) - but if we would extend all messages to include that, then that would mean much more work, and still long messages. Probably not as long, but also not as much information, as how we got there can never be included. A simplification in unit tests does not justify decreased usability.

Thus I am in favor of not merging this PR as long as `what()` does not contain the backtrace.

ZedThree:
The backtrace is sometimes useful, but it is very noisy and can obscure the actual exception message. Even without the backtrace, `what()` is currently more noisy than it needs to be:

```
terminate called after throwing an instance of 'BoutException'
  what():  ====== Exception thrown ======
Some error message
====== Back trace ======
Stacktrace not enabled.

Aborted (core dumped)
```

and with `CHECK = 0`:

```
terminate called after throwing an instance of 'BoutException'
  what():  ====== Exception thrown ======
Some error message
Enable checking (configure with --enable-check or set flag -DCHECK > 1) to get a trace
Stacktrace not enabled.

Aborted (core dumped)
```

Even harder to see what actual error message! With this PR:

```
terminate called after throwing an instance of 'BoutException'
  what():  Some error message
Aborted (core dumped)
```

Also, for Cython, you can specify a conversion routine which could get the backtrace information

JosephThomasParker:
Putting a bad value for a solver in the orzag-tang example, I get the same backtrace whether or not checking is turned on:
```	Option solver:type = spt (data/BOUT.inp)
Error encountered
Error when trying to create a Solver: Could not find 'spt'
====== Exception path ======
[bt] #1 ./mhd() [0x407686]
main at /home/joseph/code/bout/bout-elm/examples/orszag-tang/mhd.cxx:122 (discriminator 24)
[bt] #2 /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf0) [0x7fa18a705830]
__libc_start_main at ??:?
[bt] #3 ./mhd() [0x41efd9]
_start at ??:?
```
This only points to the call `BOUTMAIN(MHD);` in the physics model, and back-traces no further.

ZedThree:
I've moved the backtrace generation back into the constructor.

I've also reversed the output of `BoutException::getBacktrace` so it now looks like:

```
Error encountered
====== Exception path ======
[bt] #6 ./mhd() [0x412c9a]
_start at /home/abuild/rpmbuild/BUILD/glibc-2.27/csu/../sysdeps/x86_64/start.S:122
[bt] #5 /lib64/libc.so.6(__libc_start_main+0xeb) [0x7f106975cfeb]
__libc_start_main at ??:?
[bt] #4 ./mhd() [0x41306d]
main at /home/peter/Codes/BOUT-dev/examples/orszag-tang/mhd.cxx:122
[bt] #3 ./mhd() [0x4aff5e]
Solver::create(Options*) at /home/peter/Codes/BOUT-dev/src/solver/solver.cxx:816
[bt] #2 ./mhd() [0x51e9ee]
SolverFactory::createSolver(Options*) at /home/peter/Codes/BOUT-dev/src/solver/solverfactory.cxx:58
[bt] #1 ./mhd() [0x4b786d]
SolverFactory::createSolver(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&, Options*) at /home/peter/Codes/BOUT-dev/src/solver/../../include/bout/solverfactory.hxx:37 (discriminator 2)
====== Back trace ======

Error when trying to create a Solver: Could not find 'spt'

--------------------------------------------------------------------------
MPI_ABORT was invoked on rank 0 in communicator MPI COMMUNICATOR 3 DUP FROM 0
with errorcode 1.

NOTE: invoking MPI_ABORT causes Open MPI to kill all MPI processes.
You may or may not see output from other processes, depending on
exactly when Open MPI kills them.
--------------------------------------------------------------------------
```

This puts the more important stuff at the bottom, where hopefully it's a bit easier to read.

SO provides an example of the Python conversion function: https://stackoverflow.com/a/49245784/2043465
I'll have a stab at implementing it tomorrow

ZedThree:
I finally got around to implementing a basic wrapper for `BoutException` in boutcore so that the backtrace is preserved.

It might be nicer to implement a custom type, `BoutError`, and raise one of those instead of a `RuntimeError` but I didn't manage to get that working.

PR #1341, opened 2018-10-29
Fix FV::Div_Par_K_Grad_Par

johnomotani:
Previously, fup/fdown and Kup/Kdown fields were defined but not used. The inputs fin and Kin were copied to f and K before checking hasYupYdown(), but since yup_field and ydown_field are not copied in the Field3D assignment operator, the yup/ydown branch could never be used.

Second commit makes K/Kup/Kdown/f/fup/fdown into Field3D& references to avoid copies. Is this better or worse style?

### DIFF
--- a/src/mesh/fv_ops.cxx
+++ b/src/mesh/fv_ops.cxx
@@ -122,31 +122,16 @@ namespace FV {
     Mesh *mesh = Kin.getMesh();
     Field3D result(0.0, mesh);
 
-    // K and f fields in yup and ydown directions
-    Field3D Kup(mesh), Kdown(mesh);
-    Field3D fup(mesh), fdown(mesh);
-    Field3D f(mesh);
-    Field3D K(mesh);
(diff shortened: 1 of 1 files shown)


ZedThree:
`const&` is good, it automatically extends the lifetime of the temporary to the lifetime of the reference.

I might even use `const auto&` here

src/mesh/fv_ops.cxx
```diff
@@ -168,20 +156,20 @@ namespace FV {
...
+        BoutReal gradient = 2.*(f[i] - fdown[iym]) / (coord->dy[i] + coord->dy[iym]);
         
         BoutReal flux = c * J * gradient / g_22;
         
         result[i] -= flux / (coord->dy[i] * coord->J[i]);
       }
     }
     
-    if (!(K.hasYupYdown() && f.hasYupYdown())) {
+    if (!(Kin.hasYupYdown() && fin.hasYupYdown())) {
```
bendudson: Should this just use `if (!use_yup_ydown)` ?

PR #1342, opened 2018-10-29
Calculate yup/ydown on demand

johnomotani:
Call calcYUpDown wherever yup/ydown fields are required (it now does nothing if yup/ydown are already present and valid), instead of in Mesh::communicate().

* makes `yup_field` and `ydown_field` members of Field3D `mutable`
* adds `mutable bool has_yup_ydown` to Field3D, so that in `Solver::load_vars()` we can mark the yup/ydown fields as invalid without deleting them
* new `Field3D& ynextMutable(int dir) const` which can be used to get non-`const` versions of the mutable `yup_field` and `ydown_field` from a `const Field3D`
* `calcYUpDown()` methods take a `const Field3D` as argument

### DIFF
--- a/include/bout/mesh.hxx
+++ b/include/bout/mesh.hxx
@@ -611,7 +611,7 @@ class Mesh {
 
   /// Calculate yup/ydown fields in case they can be computed for an
   /// intermediate variable without communicating
-  void calcYUpDown(Field3D &f) {
+  void calcYUpDown(const Field3D &f) {
     getParallelTransform().calcYUpDown(f);
   }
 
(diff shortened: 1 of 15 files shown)


ZedThree:
I don't think this is the correct thing to do. `calcYupYdown` is changing the logical state of a field, therefore it should not be done on a `const Field3D`.

I think this would only be acceptable if the parallel slices were *completely* hidden from the user, i.e. they were an implementation detail.

Having a `bool hasValidParallelSlices` is not a bad idea, but we'd have to be very rigorous about keeping it correct. All the arithmetic-assignment operators would need to mark it as invalid for instance.

johnomotani:
Think @ZedThree is right. For example if a user wanted to force a function to use the field-aligned version instead of yup/ydown while using ShiftedMetric everywhere else, this implementation would prevent them from doing so because the `const Field3D` would be modified :-1:

Closing this, but I'll take over the `bool hasValidParallelSlices` idea to #1176.

Issue #1336, opened 2018-10-25
Expanding number of yup/ydown fields


ZedThree:
A possible implementation would be to use `std::vectors<Field3D>` -- note storing actual values rather than pointers/references/shared_ptr etc.

`field3d.hxx`
```diff
- Field3D* yup_field, ydown_field;
+ std::vector<Field3D> yup_fields, ydown_fields;
```
`Field3D::yup/ydown()` could then look something like this:
```cpp
  Field3D &yup(std::vector<Field3D>::size_type index = 0) {
    if (yup_fields.empty()) {
      return *this;
    }
    ASSERT2(index < yup_fields.size()); // Check for communicate
    return yup_fields[index];
  }
```
If we haven't set any yup/ydown fields, then the vector is empty, in which case we can return the field itself.

`splitYupYdown` would `emplace` fields into the vectors, up to the number of y-guards.
`mergeYupYdown` would just `clear` the vectors.
The field destructor wouldn't need to do anything! The vector would get cleaned up automatically.

`Field3D::hasYupYdown` would just `return !yup_fields.empty() and !ydown_fields.empty()`. Although, perhaps confusingly, `hasYupYdown` returning `false` doesn't mean that you can't do `f.yup()` and get a result -- but it might not be correct if you need to do a parallel transform.

The identity parallel transform could copy `this` into the yup/ydowns, which wouldn't be too bad due to COW.

---

One issue is that currently we have three possible states for `yup`: `nullptr`, `this` and a unique field. This proposal would change this to just two states, changing the behaviour of `mergeYupYdown` and the meaning of `hasYupYdown`:

`test_field3d.cxx`:
```diff
TEST_F(Field3DTest, MergeYupYDown) {
  Field3D field;

  field = 0.;

  EXPECT_FALSE(field.hasYupYdown());

  field.mergeYupYdown();

- EXPECT_TRUE(field.hasYupYdown());
+ EXPECT_FALSE(field.hasYupYdown());
```

I've implemented this proposal and this is the only change to a test that needs doing.

There's also a small issue with working with unallocated fields -- copying one throws at CHECK > 2. This causes problems if the vector of yups needs to be reallocated. One solution is to simply `yup_fields.reserve(yguards)` when a field is created; this is a memory pessimisation though, especially as it might ripple through the vector of yups...

ZedThree:
Looks like #1345 is pretty much the agreed-on design

PR #1345, opened 2018-10-29
Add multiple parallel slices

ZedThree:
<s>Nothing uses the extra parallel slices yet -- waiting for the derivatives work to be finished</s>.

See #1336 for some discussion


include/field3d.hxx
```diff
@@ -226,30 +228,42 @@ class Field3D : public Field, public FieldData {
...
+    return !yup_fields.empty() and !ydown_fields.empty();
   }
 
   /// Return reference to yup field
-  Field3D& yup() { 
-    ASSERT2(yup_field != nullptr); // Check for communicate
-    return *yup_field; 
+  Field3D &yup(std::vector<Field3D>::size_type index = 0) {
+    if (yup_fields.empty()) {
+      return *this;
```
johnomotani: Do we want to return `*this` rather than throwing an exception? Might cause confusion if `ShiftedMetric` is being used, but the user calls `yup()` before `calcYUpDown()`.

Actually, maybe this needs a `bool hasValidYUpDown` type check. If `mergeYupYdown()` has been called and `hasValidYUpDown=true` has been set, we want to return `*this`, but otherwise it would be an error.

src/field/field3d.cxx
```diff
@@ -181,54 +169,53 @@ Field3D* Field3D::timeDeriv() {
...
-  case -1:
-    return ydown();
-  default:
-    throw BoutException("Field3D: Call to ynext with strange direction %d. Only +/-1 currently supported", dir);
+const Field3D& Field3D::ynext(int dir) const {
+  // Asked for more than yguards
+  if (std::abs(dir) > fieldmesh->ystart) {
+    throw BoutException(
+        "Field3D: Call to ynext with %d which is more than number of yguards (%d)", dir,
+        fieldmesh->ystart);
```
johnomotani: Should we check only for `CHECK > 0`?

include/field3d.hxx
```diff
@@ -226,30 +228,42 @@ class Field3D : public Field, public FieldData {
   
   /// Check if this field has yup and ydown fields
   bool hasYupYdown() const {
-    return (yup_field != nullptr) && (ydown_field != nullptr);
+    return !yup_fields.empty() and !ydown_fields.empty();
```
johnomotani: If we added a `bool hasValidYUpDown`, could be `return hasValidYUpDown`, since if `mergeYupYdown()` has been called, then calling `yup()`, etc., should be allowed so this should return true even if `yup_fields` is empty.

I'd been deleting yup/ydown fields on #1176, but now I've changed that to calling `setHasValidYUpDown(false)`. 4a7cbcfc9b80db86b6454552cb003569d7ffa0d4 is meant to have all the places that yup/ydown should be marked as invalid.

include/field3d.hxx
ZedThree: Yes, this was initially an optimisation for the identity transform, but just copying the field into the parallel slice vectors should be sufficient -- COW should prevent this being too costly.

I'm currently thinking we can have a `bool hasValidParallelSlices` that the non-`const` methods set to `false`, with a public `setParallelSlicesValid` (or some less horrible name) that sets it to `true`. The parallel transforms would call that, as well as possibly advanced users if they what they're doing.

ZedThree:
I think this is the right direction to go in, but unfortunately, it changes both the semantics and behaviour in a backwards-incompatible fashion.

`Field3d::mergeYupYdown()` no longer makes `f.yup()` **identified with** `f`, it clears all the parallel slices. I could change it so that it instead essentially does the identity transform, but it would still mean that:

```
f.mergeYupYdown();
&f.yup() == &f;
```
would no longer hold. Is this acceptable?

---

This PR needs to make the non-trivial parallel transforms actually calculate multiple parallel slices for it to be useful. I'm working on writing some sanity/unit tests for the shifted metric so that I can change that happily.

johnomotani:
:+1: I think the backwards-incompatibility is acceptable; anyone who has physics-model code that would notice the change is surely following `next` development closely enough not to be confused!

ZedThree:
I think this is about ready to go in. I've not added support for multiple parallel slices to FCI yet. Requires some thinking about.

I'm also pondering deprecating the current `yup/ydown`, etc. methods and replacing them with:

```cpp
// Call mesh->getParallelTransform().calcYupYdown()
void createParallelSlices();
// Delete parallel slices; replace mergeYupYdown
void clearParallelSlices();
// Replace hasYupYdown
bool hasParallelSlices() const;
// Replace yup/ydown/ynext
Field3D& getParallelSlice(int dir);
```

bshanahan:
I think the suggested change is a good idea. It makes things a bit more
logical and flexible.

I'm absolutely no expert but I can imagine that this could be slightly
slower (to call getparallelslice 2-4 times per derivative function).

johnomotani:
> this could be slightly slower (to call getparallelslice 2-4 times per derivative function).

@bshanahan it's OK: if you set `MYG=1` in the options (or mesh) so there's only one guard cell in the y-direction then it'll only calculate 2 parallel slices instead of 4.

include/bout/index_derivs_interface.hxx
```diff
@@ -207,14 +207,13 @@ template <typename T>
 T DDY(const T& f, CELL_LOC outloc = CELL_DEFAULT, const std::string& method = "DEFAULT",
       REGION region = RGN_NOBNDRY) {
   AUTO_TRACE();
-  if (std::is_base_of<Field3D, T>::value && f.hasYupYdown()
-      && ((&f.yup() != &f) || (&f.ydown() != &f))) {
+  if (std::is_base_of<Field3D, T>::value && f.hasYupYdown()) {
```
d7919: This hasn't changed, but this condition means that Field2D always goes through the "aligned" branch. It should actually equally be able to go through the orthogonal branch without any issue now (the previous case was that `f.yup() != &f` would always be false for `Field2D` so it would always go through the second branch anyway, so the Field3D check was just a compile-time way of ensuring this). I'm not sure which is preferred/more efficient etc. but might be worth experimenting with.

Secondly I think you proposed providing alternative names rather than yup/ydown -- is it worth using the new name in the function here (i.e. `f.hasParallelSlices()` rather than `f.hasYupYdown()`), or have these renames not happened yet?

include/bout/paralleltransform.hxx
```diff
@@ -125,7 +125,7 @@ public:
   }
 
   /// A 3D array, implemented as nested vectors
-  using arr3Dvec = std::vector<std::vector<std::vector<dcomplex>>>;
+  using arr3Dvec = std::vector<std::vector<Array<dcomplex>>>;
```
d7919: Would #1505 help with this? What's the motivation for just changing the inner to an Array rather than the other nested levels?

include/bout/paralleltransform.hxx
```diff
@@ -137,6 +137,20 @@ private:
...
   arr3Dvec fromAlignedPhs;
 
+  /// Helper POD for parallel slice phase shifts
+  struct ParallelSlicePhase {
+    arr3Dvec phase_shift;
+    int y_offset;
+  };
+
+  /// Cache of phase shifts for the parallel slices. Slices are stored
+  /// in the following order:
```
d7919: Presumably the order shouldn't really matter as the struct carries the index with it etc. ? I don't suppose it really matters but we might (potentially but probably not) get better memory access etc. if this was ordered as `{-n, ...-1, 1, ...., n}`

include/bout/paralleltransform.hxx
```diff
@@ -137,6 +137,20 @@ private:
...
+  ///     {+1, ..., +n, -1, ..., -n}
+  /// slice[i] stores offset i+1
+  /// slice[2*i + 1] stores offset -(i+1)
+  /// where i goes from 0 to (n-1), with n the number of y guard cells
+  std::vector<ParallelSlicePhase> parallel_slice_phases;
+
   arr3Dvec yupPhs; ///< Cache of phase shifts for calculating yup fields
```
d7919: Should these two phases go as we're now using `parallel_slice_phases`? Also I know this is meant for internal use but `parallel_slice_phases` doesn't match our usual naming convention, is that ok?

include/field3d.hxx
```diff
@@ -228,30 +230,30 @@ class Field3D : public Field, public FieldData {
...
   }
 
   /// Return reference to yup field
-  Field3D& yup() { 
-    ASSERT2(yup_field != nullptr); // Check for communicate
-    return *yup_field; 
+  Field3D &yup(std::vector<Field3D>::size_type index = 0) {
+    ASSERT2(index < yup_fields.size()); // Check for communicate
```
d7919: I don't think this comment is correct (or if it is it's not super clear what it really means)?

include/field3d.hxx
```diff
@@ -228,30 +230,30 @@ class Field3D : public Field, public FieldData {
...
   /// Return const reference to ydown field
-  const Field3D& ydown() const { 
-    ASSERT2(ydown_field != nullptr);
-    return *ydown_field; 
+  const Field3D &ydown(std::vector<Field3D>::size_type index = 0) const {
+    ASSERT2(index < ydown_fields.size());
+    return ydown_fields[index];
   }
 
   /// Return yup if dir=+1, and ydown if dir=-1
```
d7919: This comment need to be generalised I think.

src/field/field3d.cxx
```diff
@@ -150,54 +135,53 @@ Field3D* Field3D::timeDeriv() {
...
-  if((yup_field != this) && (yup_field != nullptr))
+  if (!yup_fields.empty()) {
     return;
+  }
 
-  // yup_field and ydown_field null
-  yup_field = new Field3D(fieldmesh);
-  ydown_field = new Field3D(fieldmesh);
+  for (int i = 0; i < fieldmesh->ystart; ++i) {
+    yup_fields.emplace_back(fieldmesh);
```
d7919: We're passing a mesh here to the `emplace_back` method of a `std::vector<Field3D>` -- what does this do? Does it work out how to call one of the `Field3D` constructors, what if we want to specify other parameters (e.g. location in the future)?

src/field/field3d.cxx
```diff
@@ -150,54 +135,53 @@ Field3D* Field3D::timeDeriv() {
 void Field3D::splitYupYdown() {
   TRACE("Field3D::splitYupYdown");
   
-  if((yup_field != this) && (yup_field != nullptr))
+  if (!yup_fields.empty()) {
```
d7919: Should we also check `ydown_fields`?

src/field/field3d.cxx
d7919: Yes I think this should be disabled at low/0 check.

src/mesh/parallel/identity.cxx
```diff
@@ -0,0 +1,11 @@
+#include "bout/paralleltransform.hxx"
```
d7919: I think we should really treat `identity` as any other parallel transform rather than including the class definition in the general `paralleltransform.hxx`.  See 4135a74 of #1464 and related commits for an example of what this looks like. I just mention as adding this file is a part of those changes so it might be worth doing that split now.

src/mesh/parallel/identity.cxx
```diff
@@ -0,0 +1,11 @@
...
+#include "bout/mesh.hxx"
+
+void ParallelTransformIdentity::calcYUpDown(Field3D& f) {
```
d7919: I think the other transforms provide a Field2D form as well which does nothing (although I might be confused with to/from field aligned).

include/bout/index_derivs_interface.hxx
ZedThree: It would be nice to be able to treat them the same!

Renames haven't happened, I was thinking of doing them in another PR in case there was any discussion on the names, but happy to just do them here (names can always be changed before the next release)

include/bout/paralleltransform.hxx
ZedThree: Yes, likely. I only *needed* the inner one to be an `Array`, and hadn't considered the others.

include/bout/paralleltransform.hxx
ZedThree: It should only matter for the initialisation, but is otherwise arbitrary. The current scheme was just easy to implement.

I suspect memory access is not the bottleneck here, but I don't have any data to support that.

include/bout/paralleltransform.hxx
ZedThree: Yes, I had just left them in while I was doing the initial testing. Yes, `camelCase` is the usual here, will change

include/field3d.hxx
ZedThree: Oops, I edited the code without touching the comment, a classic mistake!

src/field/field3d.cxx
ZedThree: Yes, `emplace_back` takes `class... Args` and `forward`s them to the member type ctor:
https://en.cppreference.com/w/cpp/container/vector/emplace_back

Should be trivial to add other arguments

src/field/field3d.cxx
ZedThree: Possibly? They are private and should always be in sync, so I think we would have problems if only one was empty.

johnomotani: We could add an `ASSERT1(ydown_fields.empty())` just to check?

ZedThree:
Now working with FCI. I've replaced the existing FCI test with a much faster one that tests both C2 and C4!

Not a great deal more to do on this. I think it would be nice if the parallel transforms had an option to do fewer parallel slices than `MYG`, but that could always go in later.

There's also an issue of parallel boundary conditions... I'm tempted to leave that till the BC refactor though, on account of boundaries are horrible.

src/field/field3d.cxx
ZedThree: Yes, that's sensible. Probably in `mergeYupYdown` too

src/mesh/parallel/identity.cxx
ZedThree: It looks like that is proposed in #1432

d7919: It looks like I was thinking of the field aligned shifts in `Mesh` https://github.com/boutproject/BOUT-dev/blob/next/include/bout/mesh.hxx#L686

include/interpolation.hxx
```diff
@@ -118,15 +118,7 @@ const T interp_to(const T& var, CELL_LOC loc, REGION region = RGN_ALL) {
...
-        if (var.hasYupYdown() && ((&var.yup() != &var) || (&var.ydown() != &var))) {
-          // Field "var" has distinct yup and ydown fields which
-          // will be used to calculate a derivative along
-          // the magnetic field
-          throw BoutException(
-              "At the moment, fields with yup/ydown cannot use interp_to.\n"
-              "If we implement a 3-point stencil for interpolate or double-up\n"
-              "/double-down fields, then we can use this case.");
-
+        if (var.hasYupYdown()) {
```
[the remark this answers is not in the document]
d7919: Would one way to do this universally be to make sure `calcYupYdown` doesn't actually populate the yup/down etc. fields if `location/=CELL_CENTRE`?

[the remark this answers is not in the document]
d7919: Actually I guess it's an issue for x staggering as well

include/stencils.hxx
```diff
@@ -46,49 +46,73 @@ void inline populateStencil(stencil &s, const FieldType& f, const typename Field
...
+      } else {
+        s.pp = f[i.template plus<2, direction>()];
+      }
+    }
     break;
   case(STAGGER::C2L):
-    if (nGuard == 2) s.mm = f[i.template minus<2, direction>()];
+    if (nGuard == 2) {
+      if (direction == DIRECTION::YOrthogonal) {
+        s.mm = f.ynext(-2)[i.template minus<2, direction>()];
```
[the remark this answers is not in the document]
d7919: This presumably is an issue in next as well where we just use one guard cell? I guess we might want (and I think I recall this has been suggested) that ShiftedMetric throws if stagger grids is enabled for the time being?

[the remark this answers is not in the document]
johnomotani: I proposed to throw from `ShiftedMetric` if staggered grids were enabled before, but now I'd rather not. Soon `ShiftedMetric` should move into `Coordinates` and get a location; then it should work with staggering in x- and z-directions should work, I think. Also soon, I'm hoping to be able to use `ShiftedMetric::toFieldAligned` to get the inputs to staggered y-derivatives (might need to add some option/flag to turn off `calcYUpDown()` for staggered simulations, but I think that'd be the only change needed and would avoid having to introduce a whole new `ParallelTransform`).

src/mesh/parallel/shiftedmetric.cxx
```diff
@@ -187,6 +184,72 @@ void ShiftedMetric::shiftZ(const BoutReal* in, const std::vector<dcomplex>& phs,
...
+  std::vector<Field3D> results{};
+
+  for (auto& phase : phases) {
+    // In C++17 std::vector::emplace_back returns a reference, which
+    // would be very useful here!
+
+    // FIXME: initialisation to -1 to avoid checkData choking on the
+    // uninitialised regions in assignment into the Field parallel
+    // slices in calcYUpDown
+    results.emplace_back(-1.0, &mesh);
```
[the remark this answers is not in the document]
ZedThree: Agreed, I think we're over-checking, and sometimes we want to be able to copy invalid fields!

tools/pylib/zoidberg/test_zoidberg.py
```diff
@@ -1,69 +1,68 @@
...
+    offsets = chain(range(1, nslice + 1), range(-1, -(nslice + 1), -1))
+    field_line_maps = ["xt_prime", "zt_prime"]
+
+    for field_line_map, offset in product(field_line_maps, offsets):
+        var = zoidberg.parallel_slice_field_name(field_line_map, offset)
+        print("Current field: ", var)
+        assert var in maps
+
+        # Each map should have the same shape as the grid
+        maps[var].shape == (nx, ny, nz)
```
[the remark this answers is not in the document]
ZedThree: Yes, good spot, thanks!

include/field3d.hxx
johnomotani: :+1: am I right in thinking that the `std::vector` will not deallocate when we call `clear`, so we're not actually losing anything by deleting them?

src/mesh/parallel/shiftedmetric.cxx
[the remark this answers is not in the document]
bendudson: Is this a change for another PR?

src/mesh/parallel/shiftedmetric.cxx
```diff
@@ -187,6 +184,72 @@ void ShiftedMetric::shiftZ(const BoutReal* in, const std::vector<dcomplex>& phs,
...
+    auto& current_result = results.back();
+    // FIXME: uncomment the following after fixing Field3D::operator=
+    // current_result.allocate();
```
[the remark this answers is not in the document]
bendudson: I guess this can be uncommented now

tests/MMS/spatial/fci/mms.py
```diff
@@ -26,22 +25,11 @@
 Bpx = Bp + (x-0.5)*Lx * Bpprime  # Note: x in range [0,1]
 B = sqrt(Bpx**2 + Bt**2)
 
-def FCI_Grad_par(f):
-    return ( Bt * diff(f, y)*2.*pi/Ly + Bpx * diff(f, z)*2.*pi/Lz ) / B
+def FCI_ddy(f):
+    return ( Bt * diff(f, y)*2.*pi/Ly + Bpx * diff(f, z)*2.*pi/Lz )
```
bendudson: Not sure about this: Dividing by B normalises the Bt and Bpx components, otherwise the result depends on the magnitude of B, not just the direction.

tools/pylib/zoidberg/test_zoidberg.py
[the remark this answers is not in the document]
bendudson: resolved?

include/field3d.hxx
ZedThree: `std::vector::clear` erases its contents, so the values will be lost but it does retain its capacity though, so next time we create the parallel slices we won't need to allocate any new memory, if that's what you meant.

src/mesh/parallel/shiftedmetric.cxx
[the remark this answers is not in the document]
ZedThree: Yes, this is waiting for #1533 to go in

tests/MMS/spatial/fci/mms.py
ZedThree: This was originally because I'd changed the MMS test to `DDY`, but it turns out that doesn't really make much sense for FCI, and `Grad_par` should always be used instead. `Grad_par` does require this `B` here and also in `g_22`

I've done the easy thing and added a simple calculation of the field line length to zoidberg, but it could really do with doing it properly. This is a bit tricky without breaking lots of things!

include/interpolation.hxx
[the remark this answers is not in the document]
ZedThree: Think I finally grok all this. Given that the parallel direction is only `y` if we're field-aligned, it's not possible to do any nature of interpolation involving `y` unless we are actually field-aligned.

So everywhere we try to do a stagger in `y`, we should check we're doing it with something field-aligned.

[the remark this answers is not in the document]
johnomotani: @ZedThree yes, that's a good summary :+1:

include/stencils.hxx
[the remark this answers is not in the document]
ZedThree: After some discussion with @d7919 we think we should be able to do staggered y-derivatives, but the result will be field-aligned, and need `fromFieldAligned` calling on it. We might even be able to do this for FCI as it might just be an interpolation.

There's more work to be done there though, and it might be easier once the y-tagging goes in so I'm going to leave it for another PR.

ZedThree:
Ok, think this is finally, finally done!

PR #1346, opened 2018-10-31
Boundary keywords and bugfix

bendudson:
Not sure if this should go into `next` or `BoundaryOp-refactor` first... Contains a new feature (keywords for boundary operations), and a fix for a previously existing bug which was found while adding unit tests.

### DIFF
--- a/tests/unit/mesh/test_boundary_factory.cxx
+++ b/tests/unit/mesh/test_boundary_factory.cxx
@@ -1,0 +1,92 @@
+#include "gtest/gtest.h"
+
+#include "boundary_factory.hxx"
+#include "boundary_region.hxx"
+
+#include "test_extras.hxx"
+
+/// Global mesh
+extern Mesh *mesh;
+
+class TestBoundary : public BoundaryOp {
+public:
+  BoundaryOp *clone(BoundaryRegion *UNUSED(region), const std::list<std::string> &args,
+                    const std::map<std::string, std::string> &keywords) override {
+    auto *testboundary = new TestBoundary();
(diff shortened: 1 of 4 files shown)


johnomotani:
I don't have a strong opinion. I'm merging into BoundaryOp-refactor at the moment, but I'd lean to merging this PR into next already since it has a bugfix and a test.

d7919:
Should the bug fix strictly go into master? If it's lived in the code for a long time then I guess it's probably not causing any problems.

tests/unit/mesh/test_boundary_factory.cxx
```diff
@@ -0,0 +1,92 @@
...
+
+  // Check no brackets
+  
+  auto *boundary = fac->create("testboundary", &region);
+
+  EXPECT_TRUE( boundary != nullptr );
+  
+  EXPECT_EQ(typeid(*boundary), typeid(TestBoundary));
+
+  delete boundary;
```
ZedThree: Please could you split this into multiple tests? The first few lines could be a fixture

bendudson: Sorry, me being lazy. Is there a good way to ensure that the boundary factory singleton is deleted?

ZedThree: `BoundaryFactory::cleanup` looks like it does the job, although if you put the fixture code into `static void SetUpTestCase` (note spelling!), it'll only get called once for all the tests. I think that's ok in this case, as none of the tests modify the factory

PR #1350, opened 2018-11-02
Support BoutReal and file-level attributes for netCDF and HDF5 files

johnomotani:
Add support for BoutReal attributes (in addition to string and int), and for adding attributes to the top-level of the data file, not just to individual variables.

This supports interaction of physics models with xarray, for example storing normalization parameters as attributes of the file.

### DIFF
--- a/src/fileio/datafile.cxx
+++ b/src/fileio/datafile.cxx
@@ -1064,6 +1064,35 @@ void Datafile::setAttribute(const string &varname, const string &attrname, int v
   }
 }
 
+void Datafile::setAttribute(const string &varname, const string &attrname, BoutReal value) {
+
+  TRACE("Datafile::setAttribute(string, string, BoutReal)");
+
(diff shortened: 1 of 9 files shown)


src/fileio/datafile.cxx
```diff
@@ -1064,6 +1064,35 @@ void Datafile::setAttribute(const string &varname, const string &attrname, int v
...
+  Timer timer("io");
+
+  if(!file)
+    throw BoutException("Datafile::write: File is not valid!");
+
+  if(openclose && (flushFrequencyCounter % flushFrequency == 0)) {
+    // Open the file
+    int MYPE;
+    MPI_Comm_rank(BoutComm::get(), &MYPE);
+    if(!file->openw(filename, MYPE, appending))
```
ZedThree: I think you could replace `MYPE` with `BoutComm::rank()`

PR #1365, opened 2018-11-09
Fix submodules for releases

ZedThree:
GitHub Releases don't include submodules, and it's not possible to use `git submodule update` unless you're in a git repo. So we first need to make a git repo, then parse the .gitmodules file and add the
submodules from that. Finally, we delete the temporary git repo

Unfortunately, this will check out the latest commit on master for the submodules rather than the specific commit that we have in the real repo. I don't think there's a nice way round that, except for storing the commit hash in a file somewhere, and that's a little gross.

I believe GitHub releases allow you to upload your own binary files, so we could make the tarball ourselves.

The useful taken from: https://stackoverflow.com/questions/11258737/restore-git-submodules-from-gitmodules/11258810#11258810


bin/bout-submodule-helper
```diff
@@ -0,0 +1,71 @@
...
+if $(git rev-parse --is-inside-work-tree > /dev/null 2>&1); then
+    # Good news! We're in a git repo. We can do the sensible thing
+    git submodule update --init --recursive
+    exit 0
+else
+    echo "Not in a git repo: creating temporary repo in order to clone submodules"
+    # Better hope README.md exists... If anything goes wrong here, we
+    # need to delete the git repo we've made, or it won't be possible
+    # to rerun this script.
+    git init >/dev/null || (rm -rf ./.git; exit 1)
```
dschwoerer: Could we make sure .git does not exist?

Further, you could write an exit handler, to delete .git - that way you don't need to keep it everywhere ...

ZedThree:
We could avoid all this by uploading our own tarball to the release, but I'm not sure if that replaces GitHub's version

dschwoerer:
I tried, and it does not seem to replace it:
https://github.com/dschwoerer/unheatmap/releases
Maybe another name helps, but if we give it something descriptive or mention this in the description, it should still be possible to find the correct file.

I think that would be much cleaner, as it retains the correct version, and also works in an offline environment.

ZedThree:
I think the sensible thing is to upload our own tarballs and explicitly state that we don't support the GitHub ones.

I'll add a `make tarball` target instead of this horribleness

PR #1366, opened 2018-11-13
Move the definition of `__thefunc__` and add AUTO_TRACE

d7919:
Moved definition of `__thefunc__` into msg_stack and adds `AUTO_TRACE` macro that works like `TRACE` but automatically generates the recorded string using `__thefunc__`.

### DIFF
--- a/include/msg_stack.hxx
+++ b/include/msg_stack.hxx
@@ -39,6 +39,17 @@ class MsgStack;
 /// The maximum length (in chars) of messages, not including terminating '0'
 #define MSG_MAX_SIZE 127
 
+/// The __PRETTY_FUNCTION__ variable is defined by GCC (and some other families) but is not a part 
+/// of the standard. The __func__ variable *is* a part of the c++11 standard so we'd like to fall back
+/// to this if possible. However as these are variables/constants and not macros we can't just
+/// check if __PRETTY_FUNCITON__ is defined or not. Instead we need to say if we support this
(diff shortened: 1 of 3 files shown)


d7919:
Originally `__thefunc__` was define in `scorepwrapper.hxx` but it seems more appropriate to separate this from the profiling header.

dschwoerer:
Is there a reason `AUTO_TRACE` ignores arguments?
If I add arguments, I would expect them to be included, thus not getting an error seems unexpected.

include/msg_stack.hxx
```diff
@@ -172,4 +184,6 @@ private:
 #define TRACE(...)
 #endif
 
+#define AUTO_TRACE(...) TRACE("%s", __thefunc__)
```
ZedThree: Please could you add a doxygen comment?

d7919:
@dschwoerer and @ZedThree  I think I've now addressed your comments.

include/msg_stack.hxx
```diff
@@ -172,4 +184,23 @@ private:
...
+ *
+ * Example
+ * -------
+ *
+ * {
+ *   AUTO_TRACE();
+ *
+ * } // Scope ends, message popped
+ */
+#define AUTO_TRACE()("%s", __thefunc__)
```
bendudson: I think this is missing a `TRACE` macro call

PR #1367, opened 2018-11-13
Set location in LaplaceCyclic::solve() and DC()

johnomotani:
Previously, location of the result was not set in `LaplaceCyclic::solve(Field3D,Field3D)` or `DC(Field3D)`.

This is a bug-fix - should it be merged to `master` as well?


d7919:
@johnomotani I think there were a couple of extra fixes that made it into master than weren't in this PR? If so we should probably include those in next as well. (Sorry I think this was my fault, I hadn't noticed #1368 was into master and mirroring this PR).

johnomotani:
It's a bit messy (sorry, my fault for merging this PR too quickly) but the extra changes in #1368 made it into next in #1371. Still, I think we'll want to merge master->next soonish to bring the fixes from #1372 (when/if that's merged) and #1374 into next.

PR #1368, opened 2018-11-13
Set location in LaplaceCyclic::solve() and DC()

johnomotani:
Merge bugfix from #1367 to master also.

### DIFF
--- a/src/invert/laplace/impls/cyclic/cyclic_laplace.cxx
+++ b/src/invert/laplace/impls/cyclic/cyclic_laplace.cxx
@@ -258,7 +258,7 @@ const Field3D LaplaceCyclic::solve(const Field3D &rhs, const Field3D &x0) {
   x.allocate();
   x.setLocation(location);
 
-  Coordinates *coord = mesh->getCoordinates(location);
+  Coordinates *coord = rhs.getCoordinates();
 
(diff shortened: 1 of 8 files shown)


src/invert/laplace/impls/cyclic/cyclic_laplace.cxx
```diff
@@ -256,8 +256,9 @@ const Field3D LaplaceCyclic::solve(const Field3D &rhs, const Field3D &x0) {
   Mesh *mesh = rhs.getMesh();
   Field3D x(mesh); // Result
   x.allocate();
+  x.setLocation(location);
 
-  Coordinates *coord = mesh->getCoordinates();
+  Coordinates *coord = mesh->getCoordinates(location);
```
d7919: A minor change but `rhs->getCoordinates()` might be preferred here.

d7919:
Are the other Laplacian implementations setting the location correctly?

johnomotani:
> Are the other Laplacian implementations setting the location correctly?

Good point. Sorry I didn't check properly the first time around. There were indeed a few more places that location needed setting (see 7965a62).

I guess this kind of issue is telling us we need more tests... sorry I don't have time to design them at the moment!

PR #1372, opened 2018-11-14
Use Mesh::addCoordinates(location) to initialize Coordinates

### LINKED ISSUE #1369: getCoordinates() can hang
When `Mesh::getCoordinates(location)` is called it may need to calculate a new `Coordinates` object. This requires interpolation, derivatives and _communication_. However, we do not guarantee that `getCoordinates()` is called on all processors at the same time. I ran into this issue when it is called from the initialization of boundary conditions, where different processors call different combinations of boundary conditions (or no boundary conditions).

The solution to this is to create all the `Coordinates` objects on staggered grids during initialization. I've implemented this for other reasons, d3c66c7, but it's mixed up with a lot of other changes that should probably be split into small chunks. I'll try and figure out tomorrow what a sensible work-plan is for introducing the changes I've been thinking about...

johnomotani:
Instead of creating Coordinates objects when Mesh::getCoordinates is called, require the user to explicitly request them before using staggered fields. Either with REQUEST_LOCATION(location1, ...) macro, which adds coordinates to the global mesh pointer, or by calling Mesh::addCoordinates(location) for each location required.

Fixes #1369.

Staggered grid section of manual and `examples/staggered_grid` also updated.

I switched to storing the `Coordinate`s in `std::unique_ptr` rather than `std::shared_ptr` because we want the `Mesh` object to be the only owner of the `Coordinate`s. The change might also offer some small performance advantage.

### DIFF
--- a/include/bout/mesh.hxx
+++ b/include/bout/mesh.hxx
@@ -434,20 +434,25 @@ class Mesh {
     ASSERT1(location != CELL_DEFAULT);
     ASSERT1(location != CELL_VSHIFT);
 
-    if (coords_map.count(location)) { // True branch most common, returns immediately
-      return coords_map[location].get();
-    } else {
-      // No coordinate system set. Create default
-      // Note that this can't be allocated here due to incomplete type
-      // (circular dependency between Mesh and Coordinates)
-      coords_map.emplace(location, createDefaultCoordinates(location));
-      return coords_map[location].get();
+#if CHECK > 0
+    if (!coords_map.count(location)) {
+      throw BoutException("Error: Coordinates for %s have not been added to "
+          "this Mesh. You should use REQUEST_STAGGER(location) before "
+          "initializing fields staggered to 'CELL_LOC location'.",
+          CELL_LOC_STRING(location).c_str());
     }
+#endif
+    return coords_map.at(location).get();
   }
 
   Coordinates *DEPRECATED(coordinates(const CELL_LOC location = CELL_CENTRE)) {
     return getCoordinates(location);
   }
+
+  /// Add Coordinates object at a certain location.
+  /// If replace_coords is set to true, reset the object in coords_map if it
+  /// already exists, otherwise add a new one
+  void addCoordinates(const CELL_LOC location, bool replace_coords = false);
 
   // First derivatives in index space
   // Implemented in src/mesh/index_derivs.hxx
@@ -712,7 +717,7 @@ class Mesh {
   
   GridDataSource *source; ///< Source for grid data
   
-  std::map<CELL_LOC, std::shared_ptr<Coordinates> > coords_map; ///< Coordinate systems at different CELL_LOCs
+  std::map<CELL_LOC, std::unique_ptr<Coordinates> > coords_map; ///< Coordinate systems at different CELL_LOCs
 
   Options *options; ///< Mesh options section
   
@@ -751,7 +756,7 @@ class Mesh {
 
 private:
   /// Allocates default Coordinates objects
-  std::shared_ptr<Coordinates> createDefaultCoordinates(const CELL_LOC location);
+  std::unique_ptr<Coordinates> createDefaultCoordinates(const CELL_LOC location);
 
   //Internal region related information
   std::map<std::string, Region<Ind3D>> regionMap3D;
--- a/include/bout/mesh.hxx
+++ b/include/bout/mesh.hxx
@@ -755,9 +755,6 @@ class Mesh {
                            REGION region = RGN_NOBNDRY);
 
 private:
-  /// Allocates default Coordinates objects
-  std::unique_ptr<Coordinates> createDefaultCoordinates(const CELL_LOC location);
-
   //Internal region related information
   std::map<std::string, Region<Ind3D>> regionMap3D;
   std::map<std::string, Region<Ind2D>> regionMap2D;
--- a/include/bout/mesh.hxx
+++ b/include/bout/mesh.hxx
@@ -717,7 +717,7 @@ class Mesh {
   
   GridDataSource *source; ///< Source for grid data
   
-  std::map<CELL_LOC, std::unique_ptr<Coordinates> > coords_map; ///< Coordinate systems at different CELL_LOCs
+  std::map<CELL_LOC, std::shared_ptr<Coordinates> > coords_map; ///< Coordinate systems at different CELL_LOCs
 
   Options *options; ///< Mesh options section
   
--- a/include/bout/mesh.hxx
+++ b/include/bout/mesh.hxx
@@ -437,8 +437,8 @@ class Mesh {
 #if CHECK > 0
     if (!coords_map.count(location)) {
       throw BoutException("Error: Coordinates for %s have not been added to "
-          "this Mesh. You should use REQUEST_STAGGER(location) before "
-          "initializing fields staggered to 'CELL_LOC location'.",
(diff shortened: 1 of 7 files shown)


johnomotani:
If we really wanted to avoid changing user code, we could call `addCoordinates()` from `Field3D::setLocation()`, since `addCoordinates()` does nothing if the `Coordinates` for that location have already been added. But potentially we call `setLocation()` quite a lot, so I'm inclined to stay with the REQUEST_LOCATION design and keep the (mostly redundant) calls and checks out of `setLocation()`. Any preferences?

d7919:
I'm not sure if this should be going into master as it seems to suggest users need to change how they do things, which I'd have thought should only change with a non-bug-fix release. On the other hand it is a route to avoiding a known bug. Any thoughts @bendudson and @ZedThree  ?

d7919:
One possibility is to actually go with the brute force method (generate all possible coordinates during startup -- downside memory consumption) as a temporary bug fix for this release and then use the approach in this PR for next as a way of properly side-stepping the issue for the next release?

johnomotani:
Agree the brute force as a temporary fix for master is a good idea. Would it be OK to put REQUEST_LOCATION into master as well (although it won't be required to call it) so that code can be compatible with both master and next? [is there a word for the opposite of deprecation? :wink: ]

johnomotani:
I think this is ready now. As @d7919 suggested, this is a temporary work-around to fix the bug without requiring any changes to user code, before #1392 introduces the long-term implementation in `next` where users will call `REQUEST_LOCATION(location)` for each staggered location they need. `REQUEST_LOCATION` is already present in this PR for forward-compatibility, but does nothing because all the locations are already added to `coords_map`.

I found a couple of small bugs while writing the unit tests for #1392, so I've included the fixes here to go into `master`, bdb56d8 and 8795aca.

johnomotani:
I've taken out the `REQUEST_LOCATION` macro. As a bugfix this PR should 'just work', so I think it's ready to go in?

include/bout/mesh.hxx
```diff
@@ -712,7 +717,7 @@ class Mesh {
   
   GridDataSource *source; ///< Source for grid data
   
-  std::map<CELL_LOC, std::shared_ptr<Coordinates> > coords_map; ///< Coordinate systems at different CELL_LOCs
+  std::map<CELL_LOC, std::unique_ptr<Coordinates> > coords_map; ///< Coordinate systems at different CELL_LOCs
```
ZedThree: I'm not sure about this change to `unique_ptr` -- is there a reason for that?

include/bout/mesh.hxx
```diff
@@ -751,7 +756,7 @@ class Mesh {
 
 private:
   /// Allocates default Coordinates objects
-  std::shared_ptr<Coordinates> createDefaultCoordinates(const CELL_LOC location);
+  std::unique_ptr<Coordinates> createDefaultCoordinates(const CELL_LOC location);
```
ZedThree: The implementation of this was deleted so this should go too

src/mesh/impls/bout/boutmesh.cxx
```diff
@@ -839,6 +839,26 @@ int BoutMesh::load() {
...
+    if (LocalNx >= 4) {
+      addCoordinates(CELL_XLOW);
+    }
+    if (LocalNy >= 4) {
+      addCoordinates(CELL_YLOW);
+    }
+    // Can always add ZLOW Coordinates, since z-interpolation on Field2D is a
+    // null operation
+    addCoordinates(CELL_ZLOW);
+  }
```
ZedThree: This logic is in `Coordinates::geometry` too -- does it need to be here?

include/bout/mesh.hxx
johnomotani: @ZedThree We were discussing in the thread on #1392 a while ago. I think the conclusion was that `unique_ptr` makes sense because the `Mesh` should always keep ownership of the `Coordinates` objects, but that we should (eventually?) return references rather than pointers. I'm happy to go with whatever you think best on this though; should I change it back?

src/mesh/impls/bout/boutmesh.cxx
johnomotani: I think it does need to be here. This code ensures the `Coordinates` are created at each location initially. The code in `Coordinates::geometry()` replaces (and therefore re-calculates) the staggered `Coordinates` objects after the `CELL_CENTRE` `Coordinates` is changed by `geometry()`. The idea is that #1392 will update things so that here we only create the `CELL_CENTRE` `Coordinates` (i.e. lines 845-860 are deleted), the user has to call `Mesh::addCoordinates(location)` for any needed staggered location, and if the user calls `geometry()` it has to be before the staggered `Coordinates` are initialized.

Maybe there is a neater way, but I don't think we want to call `Coordinates::geometry()` here because it would duplicate work already done when `geometryNoRecalculate` is called as part of the constructor in `addCoordinates(CELL_CENTRE)`.

include/bout/mesh.hxx
johnomotani: Also, even if `unique_ptr` is the thing to do, it is not a bugfix, so should be in #1392 only, not in this PR. I'll take it out of this one.

ZedThree:
After quite a bit of discussion, we think this probably isn't the best way forward. The relationship between `Mesh` and `Coordinates` is already a bit too complicated.

Instead, let's just not calculate the Christoffel symbols in `geometry` -- they are the real cause of the bug. We'll move the calculation out into a separate method, which the vector functions can call. This would be a breaking change, but we think it's unlikely to actually affect anyone.

Doing this would neatly sidestep the root cause of the bug you were seeing, and also have other nice side-effects as it would pave the way for removing them entirely (or at least out to a separate structure).

include/bout/mesh.hxx
ZedThree: I think we can even go further and just store values. It was originally stored as a pointer because we have to make the `Mesh` before making a `Coordinates` so it had to be initialised to `nullptr`. But as we're now storing multiple instances in a `map`, we can initialise the map to empty and hand out pointers for now, and references later.

ZedThree:
I think there was also a bug fix for the non-uniform corrections in here -- @johnomotani do you want to put them in another PR?

johnomotani:
I'd agree moving the calculation of the Christoffel symbols could be a good idea, and happy to have separate things for the vector operations, but I'm concerned about `G1`, `G2` and `G3`. If I understand right, the idea is to make `geometry()` not need to call `communicate()`.  `G1`, `G2` and `G3` are needed by `Delp2`, `Laplace` and Laplacian inversion code. We'd need to make sure they are initialized, and if not in `geometry()` then somewhere else. But I guess what you wrote above about vector functions would apply here too; an initialization function that's called from a differential operator would be allowed to use `communicate()` because all processors would have to call the operator.

Actually, I think `geometry()` is not the only issue. The `Coordinates::Coordinates(Mesh*,CELL_LOC,const Coordinates*)` constructor also needs to communicate after interpolating the metric components, otherwise we wouldn't be able to take derivatives of the metric components later. The original cause of the bug was when constructing a new `Coordinates` object inside `getCoordinates()` when `getCoordinates()` wasn't called by all processors simultaneously (because `getCoordinates(CELL_YLOW)` was first called in a boundary condition).

PS I'll put the bugfix to non-uniform corrections in a separate PR anyway. The fix to `BoutMesh` in 8795aca is included in #1470 already.

PR #1373, opened 2018-11-14
Print command line options to output_info

johnomotani:
Print the command line executable+arguments (contents of argv) to `output_info`. I think it would be nice to keep a record of how the simulation was run, to help ease of reproducibility.


bendudson:
Agree this is a good idea. The BOUT.settings file also contains the settings actually used,  including any changes made on the command line.

PR #1377, opened 2018-11-15
Skip special entries from input file

### LINKED ISSUE #602: BOUT.settings contains datadir
`BOUT.settings` includes `datadir` and if it is used as input file from another location, this will cause the `datadir` to be overwritten. Instead of the (correct) version from the command line, the (wrong) version from the datafile is used.

Expected behaviour:
* ignore `datadir` from `BOUT.inp` (the input file)
(alternatively: don't write `datadir`)

dschwoerer:
Resolves #602

### DIFF
--- a/src/sys/options/options_ini.cxx
+++ b/src/sys/options/options_ini.cxx
@@ -113,8 +113,17 @@ void OptionINI::read(Options *options, const string &filename) {
         string key, value;
         // Get a key = value pair
         parse(buffer, key, value);
-        // Add this to the current section
-        section->set(key, value, filename);
+	// check for some special entries from the root section
+	if (section == options &&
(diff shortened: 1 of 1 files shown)


bendudson:
Thanks @dschwoerer for fixing, glad you've found a solution. I can't help feeling though that this is a bit of a sticking plaster on a more fundamental problem. I would be ok with merging this for now, but would feel happier with a solution which didn't involve special cases like this.

dschwoerer:
I think it is not a particular elegant solution, but it seems rather simple.
As long as there is only a short list of special cases, I think it is not worth complicating the rest of the code base ...

ZedThree:
Replaced by #1382

PR #1382, opened 2018-11-16
Don't clobber command line short options when reading input file

ZedThree:
Alternative to #1377 

Instead of the input file possibly taking precedence over the command line short options, `datadir`/`optionfile`/`settingsfile` are now either the command line option or the default and the input file is ignored.

This might not actually be the behaviour we want either. I think we want command line > input file > default, and that's a little trickier as we need either the command line option or the default in order to read the input file!

```cpp
    if (Options::root().isSet("datadir") and not still_default_datadir) {
      Options::root()["datadir"] = data_dir;
    }
 ```

where `still_default_datadir` is initially `true` and `false` if we use `-d datadir`. I think that's the right conditional?

Also I've spotted we have `DEFAULT_DIR = "data"` in `bout++.cxx` but use `"data"` literally as a default in various places.

### DIFF
--- a/src/bout++.cxx
+++ b/src/bout++.cxx
@@ -261,11 +261,6 @@ int BoutInitialise(int &argc, char **&argv) {
     throw BoutException(_("DataDir \"%s\" does not exist or is not accessible\n"),data_dir);
   }
   
-  // Set options
-  Options::getRoot()->set("datadir", string(data_dir));
-  Options::getRoot()->set("optionfile", string(opt_file));
-  Options::getRoot()->set("settingsfile", string(set_file));
-
   // Set the command-line arguments
   SlepcLib::setArgs(argc, argv); // SLEPc initialisation
   PetscLib::setArgs(argc, argv); // PETSc initialisation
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by dschwoerer** [comment]
Note that this makes the output more noisy:
```
        Option datadir = test (test2/BOUT.settings) overwritten with:
                datadir = test2 ()
        Option optionfile = bout.inp (test2/BOUT.settings) overwritten with:
                optionfile = BOUT.settings ()
        Option settingsfile = bout.settings (test2/BOUT.settings) overwritten with:
                settingsfile = testsettings ()
```

Besides that, it is nice as it does not need any special handling of that case. :+1:

**Comment by ZedThree** [comment]
I think the correct conditional might be even worse than I though:

```cpp
    if (not Options::root().isSet("datadir")
        or (Options::root().isSet("datadir") and not still_default_datadir)) {
      Options::root()["datadir"].force(data_dir);
    }
```

- If "datadir" wasn't set in the input file, we want to set the default value.
- If it was set in the input file and not on the command line, we don't want to overwrite it.
- If it was set in the input file and also on the command line, we want to overwrite it.

It's a bit unwieldy, but I think does the right thing

dschwoerer:
I think `datadir` and `inputfile` should never ever be read from the inputfile.
Otherwise, they can be self-contradicting, and the only clean way forward would be restarting with the new `datadir` and `inputfile` set (e.g. by commandline)
The case about `settingsfile` is imo not so clear - but I think for consistency, the should not be read.
I think `logfile` is not part of the `Options` - but otherwise it should be excluded as well, as it already has been opened, thus it may already have overwritten data, and also to get the full log, the already written log would need to be copied (or the simulation restarted)

Setting datadir and input file in the input file is futile - ether it was found, and they are not needed, or they are not, in which case they don't help. The settingsfile and logfile might however be useful, but imho it is fine to only allow setting them via the commandline, for consistency and simplicity.

ZedThree:
Yes, that probably makes sense, I'm happy to go with that

PR #1383, opened 2018-11-16
Absorb used options to keep options tree clean

dschwoerer:
The code is ugly, but it annoyed me that there is always a `d = True` etc in the `BOUT.settings` ...

### DIFF
--- a/src/bout++.cxx
+++ b/src/bout++.cxx
@@ -193,6 +193,7 @@ int BoutInitialise(int &argc, char **&argv) {
     }
   }
   bool color_output = false; // Will be set true if -c is in the options
+  // Remove arguments that are used here
   char ** argvcleaned = (char**) malloc(sizeof(char *)*(argc+1));
   int icleaned=0;
   argvcleaned[icleaned]=argv[0];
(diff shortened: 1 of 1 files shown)


src/bout++.cxx
```diff
@@ -193,6 +193,9 @@ int BoutInitialise(int &argc, char **&argv) {
     }
   }
   bool color_output = false; // Will be set true if -c is in the options
+  char ** argvcleaned = (char**) malloc(sizeof(char *)*(argc+1));
```
ZedThree: Please don't use `malloc`! This should just be a `std::vector`

dschwoerer: I wasn't sure as other parts of the code might assume it is a plain pointer, and thus call free or remalloc - which I though is not safe if the memory has not been allocated directly by malloc.

ZedThree:
I think this is supposed to remove the command line options we use in `BoutInitialise` from `argv`? Please could you drop in a couple of comments to that effect. It's not clear what "cleaning" means in this context.

Also, I'm not sure the `output`s are necessary. This is something that we are doing behind the scenes.

dschwoerer:
Oh sorry, outputs should of course be removed :+1:

src/bout++.cxx
ZedThree: I think that's undefined behaviour.

This might be cleaner if you use a `std::vector<std::string>` to store the original `argv` for later when we write it out, and then just set `argv[i-1][0] = argv[i][0] = 0;` (to set the flag and its argument to empty string).

dschwoerer:
Replaced by #1400

PR #1384, opened 2018-11-19
Overhaul derivatives for vectorisation and user registration

d7919:
Overhaul of index derivatives. Designed to provide the following main advantages: 

1. Possible for user registration of derivative methods.
2. Inner loop of derivative methods vectorise were possible and can be openmp parallelised.
3. Reduction of duplicate code.
4. More consistent treatment of staggering in all directions and for both field types.

Also provides the following minor changes:

1. More robust treatment of enums in boutcore processing.
2. `uppercase` routine.
3. Adds a direction enum class, a staggering enum class and a derivative class (standard, upwind, flux etc.) enum class.
4. A couple of minor helper routines have been added to mesh to provide methods to request the number of points/guard cells in a particular direction.
5. Performance examples for measuring the cost of `DDX`, `DDY` and `DDZ` with various supported methods.
6. Provide an overload for top level derivative methods that takes a `std::string` instead of a `DIFF_METHOD` enum. This is required to allow for user registered routines. The `std::string` methods should be preferred and the `DIFF_METHOD` overloads may become deprecated.
7. Provides a templated `minus` and `plus` offsetting routine for `SpecificInd` which provides a simple way to offset in a requested direction with compile time information that works nicely with routines templated on the `DIRECTION` enum class.
8. Adds a `fourth` option to `mesh:dd?`, currently `C2` is the only valid predefined option.

Still to do:

1. Documentation. (main user documentation added, no developer docs added yet but hopefully source documentation covers this).
2. More testing. (never ending task! Basic unit testing for derivative store and template combination code added)

Main implementation details:

1. To achieve vectorisation for the "inner loop" we had to remove the function pointer indirection, 
meaning we have to effectively create an instance of the loop for every possible method and with all possible combinations of staggering, direction etc. This is achieving using some templating routines (in `include/bout/template_combinations.hxx`) which can do "something" for a cartesian product of options.
2. The "something" in 1. is to create derivative instances and register them in a singleton derivative store (one for each field type). This store can register routines and can return these at runtime given user options etc.
3. All the `indexDD?` routines now just call a single kernel (`indexStandardDerivative`) that gets the requested method and applies it in the appropriate direction. All the `indexVDD?` and `indexFDD?` routines now just call a single kernel (`indexFlowDerivative`) that gets the requested method and applies it.

The performance examples suggest that most derivative methods now vectorise, giving performance comparable to direct hard-coded nested loop. Whilst this is around a factor 3-4 improvement over the current implementation in next this is not reflected in physics models due to the other costs involved in a full model.

Compilation of the library is likely to be slower due to the template combination expansion that must be done at compile time.


include/bout/index_derivs.hxx
```diff
@@ -0,0 +1,789 @@
...
+public:
+  template <DIRECTION direction, STAGGER stagger, int nGuards, typename T>
+  void standard(const T& var, T& result, REGION region) const {
+    TRACE("%s", __thefunc__);
+    ASSERT2(meta.derivType == DERIV::Standard || meta.derivType == DERIV::StandardSecond
+            || meta.derivType == DERIV::StandardFourth)
+    ASSERT2(var.getMesh()->template getNguard<direction>() >= nGuards);
+
+    BOUT_FOR(i, var.getRegion(region)) {
+      result[i] = apply(populateStencil<direction, stagger, nGuards>(var, i));
```
bendudson: This is awesome! So glad it's got to the point where the whole derivatives mess (as it was) boils down to something this "simple".

include/bout/index_derivs.hxx
```diff
@@ -0,0 +1,789 @@
...
+/// Some helper defines for now that allow us to wrap up enums
+/// and the specific methods.
+#define e(family, value) enumWrapper<family, family::value>
+
+/////////////////////////////////////////////////////////////////////////////////
+/// Here's an example of registering a couple of DerivativeType methods
+/// at once for no staggering
+/////////////////////////////////////////////////////////////////////////////////
+
+// Could use Ben's magic macro for thing here to register multiple routines at once
```
bendudson: Could do, but probably shouldn't. There is already a near-explosive concentration of magic in this part of the code, so probably a wise choice to keep it written out. Maybe once thaum levels have decayed a little we can use it here.

include/bout/deriv_store.hxx
```diff
@@ -0,0 +1,415 @@
...
+  static DerivativeStore& getInstance() {
+    static DerivativeStore instance;
+    return instance;
+  }
+
+  /// Returns a vector of all registered method names for the
+  /// specified derivative type, direction and stagger.
+  std::vector<std::string> getAvailableMethods(DERIV derivType, DIRECTION direction,
+                                               STAGGER stagger = STAGGER::None) const {
+    TRACE("%s", __thefunc__);
```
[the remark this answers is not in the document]
d7919: Yes, now `AUTO_TRACE` is in we could use that here (or try to provide our own message).
On a related note I think we might want to increase the maximum allowed size of a single message in the stack as it's currently limited to 127 chars, which is small enough that some of the templated routines get truncated before the template specialisation is given.

ZedThree:
Some possible things spotted by codacy: https://app.codacy.com/app/bout-project/BOUT-dev/pullRequest?prid=2583567#newIssuesView

d7919:
> Some possible things spotted by codacy: https://app.codacy.com/app/bout-project/BOUT-dev/pullRequest?prid=2583567#newIssuesView

Thanks, I've hopefully addressed these now.

src/sys/utils.cxx
```diff
@@ -61,6 +61,14 @@ const std::string lowercase(const std::string &str) {
   return strlow;
 }
 
+// Convert a string to upper case
+const std::string uppercase(const std::string& str) {
+  std::string strlow(str);
```
johnomotani: Would make more sense to call this `strup` or `strupper`?

d7919: It would -- lazy copy and paste by me there! I'll fix it.

d7919:
I've now added initial unit tests for the derivative store and the template code for producing combinations. I've yet to add tests for the index derivative methods and I'm not sure if it makes sense to include these as unit tests or not?

johnomotani:
> I've yet to add tests for the index derivative methods and I'm not sure if it makes sense to include these as unit tests or not?

At some point an updated version of #1315 should MMS test all the derivative methods, including checks that exceptions are thrown for not-allowed combinations of inloc/outloc. (Although this depends on resolving the issues @ZedThree raised with relying on boutcore for tests...)

include/bout/deriv_store.hxx
```diff
@@ -0,0 +1,486 @@
...
+    flux.clear();
+    registeredMethods.clear();
+  }
+
+private:
+  // Make the constructor private so we can't make instances outside
+  // of the struct
+  DerivativeStore(){};
+  // No copy constructor allowed
+  DerivativeStore(const DerivativeStore& junk) = delete;
```
ZedThree: These should both be `public` and the default ctor `= delete`

include/bout/deriv_store.hxx
```diff
@@ -0,0 +1,486 @@
...
+  /// Provides a routine to produce a unique key given information about the
+  /// specific type
+  /// required. This is templated so requires compile-time information. Need to
+  /// also
+  /// supply
+  /// a non-templated version to account for run-time choices
+  /// Note : We could include the derivType in the key -- this would allow us to
+  /// store
+  /// all methods with the same function interface in the same map, which might
+  /// be nice.
```
ZedThree: Emacs `M-q` :)

include/bout/deriv_store.hxx
d7919: If I do this then the line that defines the static instance (`static DerivativeStore instance;`) will not work or am I missing something?

ZedThree: Oops, sorry, brain not in gear! I think the deleted copy ctor should be public though

ZedThree:
Documentation built here: https://bout-dev.readthedocs.io/en/more_templating_difops_templatemagic/

include/bout/index_derivs.hxx
```diff
@@ -0,0 +1,826 @@
...
+            theFunc);
+      } else {
+        const auto theFunc = std::bind(
+            // Method to store in function
+            &Method::template upwindOrFlux<Direction::value, Stagger::value, 2,
+                                           FieldType>,
+            // Arguments -- first is hidden this of type-bound, others are placeholders
+            // for input field, output field, region
+            Method{}, _1, _2, _3, _4);
+        derivativeRegister.template registerDerivative<Direction, Stagger, Method>(
```
ZedThree: Would it make sense for `registterDerivative` to take `Direction`, `Stagger`, and `Method` as arguments, rather than template parameters? Then the function could be called without the need to specify `template`

d7919: There should be an overload already that would allow that but I think it would make the call slightly uglier potentially (if I remember correctly, which I might not).

ZedThree: I think you can just change this overload:

```cpp
  template <typename Direction, typename Stagger, typename Method>
  void registerDerivative(standardFunc func) {
    AUTO_TRACE();
    registerDerivative(func, Method{}.meta.derivType, Direction{}.lookup(),
                       Stagger{}.lookup(), Method{}.meta.key);
  };
```

to take `Direction`, `Stagger` and `Method` by argument

d7919: The `Direction`, `Stagger` and `Method` types here are wrapped up values/enums/types provided by the template combination code so we should be able to use type deduction with the templated routine to give 

~~~~
 /// Templated versions of the above registration routines.                                                                                                                                                       
  template <typename Direction, typename Stagger, typename Method>
  void registerDerivative(standardFunc func, Direction direction, Stagger stagger, Method method) {
    AUTO_TRACE();
    registerDerivative(func, method.meta.derivType, direction.lookup(),
                       stagger.lookup(), method.meta.key);
  };
~~~~

The call then looks like

~~~~
        derivativeRegister.registerDerivative(
            theFunc, Direction{}, Stagger{}, Method{});
~~~~

include/bout/deriv_store.hxx
```diff
@@ -0,0 +1,484 @@
...
+    }
+  }
+
+  /// The following stores what actual method to use when DIFF_DEFAULT
+  /// is passed. The key is determined using the getKey routine here,
+  /// where the name we pass is determined by the type of method (standard,
+  /// upwind etc.). Note for now we'll always use STAGGER::None as we
+  /// currently assume the default method is independent of staggering --
+  /// it might be useful to relax this assumption!
+  storageType<std::size_t, std::string> defaultMethods;
```
ZedThree: Should this be private?

d7919: Yes it should! I think this was public because of a different original design.

ZedThree:
Should we namespace some of the new things? The template combo stuff might go in one, though I've not got a name to mind. Possibly `bout::utils`?

include/bout/deriv_store.hxx
```diff
@@ -0,0 +1,484 @@
...
+    AUTO_TRACE();
+    const auto realName =
+        nameLookup(name,
+                   defaultMethods.at(
+                       getKey(direction, stagger, DERIV_STRING(DERIV::StandardSecond))));
+    const auto key = getKey(direction, stagger, realName);
+    const auto resultOfFind = standardSecond.find(key);
+    if (resultOfFind != standardSecond.end())
+      return resultOfFind->second;
+    throw BoutException("Couldn't find requested method %s in map for "
```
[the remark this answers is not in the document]
d7919: I put it this way as I expected the true condition to be most common but can reverse.

include/bout/mesh.hxx
```diff
@@ -756,5 +800,201 @@ private:
...
+    tmp.setLocation(outloc);
+    return tmp;
+  }
+
+  // Lookup the method
+  auto& derivativeStore = DerivativeStore<T>::getInstance();
+  typename DerivativeStore<T>::standardFunc derivativeMethod;
+
+  if (order == 1) {
+    derivativeMethod = derivativeStore.getStandardDerivative(method, direction, stagger);
```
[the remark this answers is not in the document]
d7919: As it is indeed known at compile time there should be no branching cost, just developer cost of reading these lines! Moving order into the store stuff would break some of the symmetry with respect to the other derivative types I think.

include/bout/mesh.hxx
```diff
@@ -756,5 +800,201 @@ private:
...
+  if (outloc == CELL_DEFAULT)
+    outloc = inloc;
+  const STAGGER stagger = getStagger(inloc, outloc, allowedStaggerLoc);
+
+  // Check for early exit
+  const int nPoint = getNpoints<direction>();
+
+  if (nPoint == 1) {
+    auto tmp = T(0., f.getMesh());
+    tmp.setLocation(outloc);
```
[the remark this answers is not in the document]
d7919: Agreed.

include/bout/region.hxx
```diff
@@ -226,6 +226,46 @@ public:
...
+    case(DIRECTION::X):
+      return xp(dd);
+    case(DIRECTION::Y):
+    case(DIRECTION::YAligned):
+    case(DIRECTION::YOrthogonal):
+      return yp(dd);
+    case(DIRECTION::Z):
+      return zp(dd);
+#if CHECK > 2      
+    default: throw BoutException("Invalid direction passed to SpecificInd.plus<int, direction>");
```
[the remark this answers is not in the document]
d7919: We could possibly make it compile time with a static assert at the top. I'm not sure if it's possible to ever hit this statement as we currently handle each member of the direction enum class, I think this mostly is here to help remind us if we add a new direction that we need some handling here.

include/bout/index_derivs_interface.hxx
```diff
@@ -0,0 +1,178 @@
...
+    auto tmp = T(0., localmesh);
+    tmp.setLocation(outloc);
+    return tmp;
+  }
+
+  // Lookup the method
+  auto& derivativeStore = DerivativeStore<T>::getInstance();
+  typename DerivativeStore<T>::standardFunc derivativeMethod;
+
+  derivativeMethod =
```
ZedThree: Could this just be `auto derivativeMethod = `, skipping the declaration above?

d7919: Yes it could now that we've removed the branches!

Issue #1361, opened 2018-11-08
Checking in Field constructors


ZedThree:
Currently, there's lots of checking via calls to `checkData` all over the place, and I wonder if it's possible to remove some of them. Take the following example:

```cpp
Field3D f, a{0.0}, b{a};

f = (a + 1) * b;
```
There are seven calls to `checkData(Field3D)` (one in the copy ctor for `b`, two in the addition, three in the multiplication, one in the assignment) and two to `checkData(BoutReal)` (one in the initialisation of `a`, one in addition).

I would like to remove the `checkData(Field3D)` from the copy-constructor and -assignment operator. This would allow e.g. making them both `noexcept`. We've got much better at pre- and post-conditions in Field operations so I think it would be ok to relax them in copy construction/assignment.

A potential downside would be not catching the following situations:

```cpp
Field3D a, b{a};
// or
void foo(Field3D f);
Field3D c;
foo(c);
```

I think the first case is probably too rare to worry about and would be caught as soon you tried to use `b`.
The second case is likely also very rare -- we normally don't pass Fields by value, and this could be solved by calling `checkData(f)` immediately inside `foo` anyway (which we should be doing!).

See e.g. #914 for a related issue.

d7919:
I would certainly support removing the checks on copying fields. I think it should be valid to create a copy of an empty field -- it's just an error to then try to read/write data from this without allocating first.

ZedThree:
Removed in #1533

PR #1385, opened 2018-11-19
Field* constructor tidy

ZedThree:
1. Implement `swap` and move constructors for `Field2D/3D`
2. Default initialises a few things that we were always setting to the same value (useful to allow default initialisation needed for the move ctors)
     - Note: I've changed the default value of `Field::boundaryIsSet` because this was `true` when the `Field2D/3D` constructors were immediately setting it to `false`.
     - This also means a couple of ctors can just be `= default`
3. Lots of `noexcept` on `Array`

I'm going to remove `checkData` from the copy constructors, and possibly `Field*::operator=`, but I would like to hear some opposing view points to that (see #1361)


include/bout/array.hxx
```diff
@@ -161,17 +161,19 @@ public:
...
   /*!
    * Return size of the array. Zero if the array is empty.
    */
-  int size() const {
+  int size() const noexcept {
     if(!ptr)
       return 0;
 #ifdef BOUT_ARRAY_WITH_VALARRAY
+    // Note: std::valarray::size is not noexcept, so Array::size
+    // shouldn't be either if we're using valarrays
```
d7919: But we're not doing that here? Should we wrap the signature in the `ifdef` as well to allow this?

include/field3d.hxx
```diff
@@ -443,22 +445,43 @@ class Field3D : public Field, public FieldData {
...
+
+  /// Array sizes (from fieldmesh). These are valid only if fieldmesh is not null
+  int nx{-1}, ny{-1}, nz{-1};
 
-  int nx, ny, nz;  ///< Array sizes (from fieldmesh). These are valid only if fieldmesh is not null
-  
   /// Internal data array. Handles allocation/freeing of memory
   Array<BoutReal> data;
 
   CELL_LOC location = CELL_CENTRE; ///< Location of the variable in the cell
```
d7919: You've changed other existing defaults from `=` to `{}`, should you do it here as well for consistency?

PR #1386, opened 2018-11-19
Use python unittest in runtests

ZedThree:
(Note this includes #1382 as one of the tests needs that, please ignore those commits)

This replaces the bash script `test-command-args/runtest` with a python version. It has several advantages:

- runs all the tests even if one of them fails (the bash version stops at the first failure)
- easier to add new tests
- tests always clean up after themselves

It's no slower than the bash version, even accounting for the python startup time. It's still compatible with our `test_suite`.

It uses the python standard library `unittest` module so nothing else needs to be installed, but it could be made even nicer by using `pytest` (which we would likely need to check people had during `configure`).

I'm not expecting this to go in, but I thought it was interesting. We might consider rewriting more of our test `runtest` scripts using `unittest` or `pytest`. This has some downsides, but the upsides might outweigh them

### DIFF
--- a/src/bout++.cxx
+++ b/src/bout++.cxx
@@ -261,11 +261,6 @@ int BoutInitialise(int &argc, char **&argv) {
     throw BoutException(_("DataDir \"%s\" does not exist or is not accessible\n"),data_dir);
   }
   
-  // Set options
-  Options::getRoot()->set("datadir", string(data_dir));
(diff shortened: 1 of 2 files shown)


ZedThree:
Actually, I think this should go in. It might be interesting to use python unit testing facilities more widely, but this test in particular needs fixing up -- it doesn't clean up after itself properly, and it's awkward to add more tests cleanly

PR #1358, opened 2018-11-07
Remove last DataIterator references/uses

ZedThree:
Fixes #1354 #1355 

This does remove the experimental outer loop stuff. We might want to update that for the new `Region` stuff instead (or keep it in an experimental branch?)

PR #1388, opened 2018-11-19
Fix examples so most compile

d7919:
A few examples still don't compile due to features missing from the current library. 

Intend to produce a separate PR that provides an integrated test that just checks we can compile all the examples.

### DIFF
--- a/examples/2Dturbulence_multigrid/esel.cxx
+++ b/examples/2Dturbulence_multigrid/esel.cxx
@@ -69,7 +69,7 @@ protected:
       return 1;
     }
 
-    Coordinates *coord = mesh->coordinates();
+    Coordinates *coord = mesh->getCoordinates();
     
     // generate coordinate system 
     // coord->J = R0/B0 ;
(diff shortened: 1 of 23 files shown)


d7919:
Examples that still don't compile are:

1. blob2d-outerloop -- uses dataiterator linked routines (proposed for removal in #1358)
2. non-local_1d -- uses removed `bindex` struct.
3. uedge-benchmark -- uses removed `bindex` struct.

d7919:
I should note this includes a bug fix in commit a45bc40 which I can pull out if required. This may also need to go into master.

d7919:
Should we fix non-local_1d before merging this?

bendudson:
non-local_1d might need quite a bit more work to tidy up. I have a version of the nonlocal heat transport code from this example in SD1D which works with the latest BOUT++ version: https://github.com/boutproject/SD1D/tree/nonlocal
It might be easiest to just put the SD1D version into BOUT++ examples, or remove non-local_1d from examples.

ZedThree:
+1 for removing non-local_1d, at least for now.

PR #1389, opened 2018-11-19
Adding simple integrated test to compile all example files

d7919:
This test attempts to run `make` in any directory within `examples` that contains a makefile and reports those for which the make command fails.

This is currently too expensive to be enabled by default and hence requires `all_tests`. It would be a good idea to run this on release candidates.

### DIFF
--- a/tests/integrated/test-compile-examples/README.md
+++ b/tests/integrated/test-compile-examples/README.md
@@ -1,0 +1,6 @@
+test-compile-examples
+=====================
+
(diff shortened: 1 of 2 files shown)


d7919:
It took a bout 90s to run this test on a local machine so I think this is too slow to run by default/in travis but is probably useful for developers changing APIs, removing files etc.

d7919:
This test will currently fail. PR #1388 helps fix most of the cases that fail but three remain as noted in that PR.

PR #1394, opened 2018-11-20
Migrate include files

d7919:
This is an example/demonstration PR that shows the sorts of widespread changes that would be required to move all top level includes under `include/bout`.

As can be seen this touches most files. I don't propose this goes in now (and I plan to close this PR after a day or so) but thought it might be helpful to keep this in mind (perhaps in time for the next major release?).

This change set can be produced using the following and the changes to boutcore in ac1e4f2 and 926cb1d

~~~~
#!/bin/bash                                                                                                                                           
                                                                                                                                                      
#A simple script for migrating header files from `include` to `include/bout`                                                                          
                                                                                                                                                      
#First step -- git mv existing headers and create redirection, deprecated header                                                                      
                                                                                                                                                      
for i in $(ls include/*.hxx)                                                                                                                          
do                                                                                                                                                    
    # Work out names                                                                                                                                  
    ff=$(basename ${i})                                                                                                                               
    newLoc=include/bout/${ff}                                                                                                                         
    newInc=$(echo ${newLoc} | awk -F"/" '{print $2"/"$3}')                                                                                            
    # Actually move header and create replacement that redirects                                                                                      
    ( git mv ${i} ${newLoc} && echo -e '#pragma message "This header has moved to '${newInc}', use of the header in the old location is deprecated."\\
n#include <'${newInc}'>' > ${i} )                                                                                                                     
done                                                                                                                                                  
                                                                                                                                                      
#Second step fix existing include statements                                                                                                          
for i in $(ls include/*.hxx)                                                                                                                          
do                                                                                                                                                    
    ff=$(basename ${i})                                                                                                                               
    #Modify the directories in find to also change examples, tests etc.                                                                               
    sed --in-place 's|\#include *\(["<]\)'"${ff}"'|\#include \1bout/'"${ff}"'|g' $(find ./{src,include} -name '*.?xx')                                
done 
~~~~


dschwoerer:
Duplicate of #538 ?

d7919:
Ah thanks, I didn't realise/remember this was included in the 4to5 script. I'm intending closing this PR soon anyway and just using it to demonstrate what this change looks like and to prompt discussion.

PR #1395, opened 2018-11-21
Pass zShift to ShiftedMetric constructor

ZedThree:
Moved the logic for getting zShift from the grid file, etc., into `Mesh::setParallelTransform`. `ShiftedMetric` ctor now explicitly takes zShift, which technically breaks backwards compatibility, but if you're calling the ctor yourself, you probably know what you're doing? 
We could keep backwards compatibility by moving all the ctor logic into a separate `init` method or something (can't just call the other ctor as we `zShift` first)

This allows greater separation between ShiftedMetric and Mesh which should make it easier to test ShiftedMetric

### DIFF
--- a/include/bout/paralleltransform.hxx
+++ b/include/bout/paralleltransform.hxx
@@ -93,7 +93,7 @@ public:
 class ShiftedMetric : public ParallelTransform {
 public:
   ShiftedMetric() = delete;
-  ShiftedMetric(Mesh &mesh);
+  ShiftedMetric(Mesh &mesh, Field2D zShift);
   
   /*!
    * Calculates the yup() and ydown() fields of f
--- a/src/mesh/mesh.cxx
+++ b/src/mesh/mesh.cxx
@@ -292,24 +292,44 @@ void Mesh::setParallelTransform() {
 
   // Convert to lower case for comparison
   ptstr = lowercase(ptstr);
-    
-  if(ptstr == "identity") {
+
+  if (ptstr == "identity") {
     // Identity method i.e. no transform needed
     transform = std::unique_ptr<ParallelTransform>(new ParallelTransformIdentity());
-      
-  }else if(ptstr == "shifted") {
-    // Shifted metric method
-    transform = std::unique_ptr<ParallelTransform>(new ShiftedMetric(*this));
-      
-  }else if(ptstr == "fci") {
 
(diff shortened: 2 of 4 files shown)


johnomotani:
I've been wondering if `zShift` should be moved from `ShiftedMetric` to `Coordinates` to make getting staggered-grid versions easier? When I was writing `ShiftToFieldAligned` in #1177 (before `Coordinates` at different locations) I had to have 3 different `zShift` variables to store the interpolated versions as well as the `CELL_CENTRE` one.

This would be an alternative way to keep backward compatibility, since you wouldn't need to pass zShift to the constructor at all, just fetch it from `localmesh->getCoordinates()->zShift`.

johnomotani:
@ZedThree I think of `zShift` as some information about the coordinate system (e.g. it gives the toroidal position of points with `paralleltransform=identity`) so maybe belongs in `Coordinates`. Also, if/when I add `ShiftToFieldAligned` (assuming it stays as a separate class) then it would need some copied logic for loading `zShift` (and the logic is not totally trivial if we add/subtract ShiftAngle in order to get avoid one set of FFTs in the branch-cut guard cells see #1294).

I think the struct+map thing sounds good, but is not enough to make `ShiftedMetric` useful for y-staggered grids: I'd leave `ShiftedMetric` as is*, and leave it to `ShiftToFieldAligned` to handle the staggered grid cases (I'll use the design you suggest above when I get around to updating #1177). 

--
*i.e. not compatible with staggering, we could allow for example staggering in the x-direction by having a staggered `zShift`, etc. but it doesn't seem worth implementing that just for the sake of it

johnomotani:
> What if we moved all the internal variables into a struct, ShiftedInternals, and then had a map<CELL_LOC, ShiftedInternals>.

Another option might be to make `ParallelTransform`s have a location and keep a `map<CELL_LOC, ParallelTransform>` in the `Mesh`, like we now do for `Coordinates`. We could rename `Mesh::addCoordinates` to `Mesh::addLocation` and initialize both `Coordinates` and `ParallelTransform` in the same method.

ZedThree:
We really need to sit down and properly design how all these systems should interact, and what belongs to what. @d7919 has suggested that the whole parallel transform needs to be moved into `Coordinates`, which I agree with. This might then go at least some way to taking care of the staggered grids problem.

---

Closing this as #1413 achieves the same thing, but is backwards compatible

PR #1396, opened 2018-11-21
Add implementation of make_unique

ZedThree:
This is essentially the same implementation that is in C++14. 

It's guarded the macro `__cpp_lib_make_unique`: most useful compilers have this if they define `std::make_unique`. I've also added `using std::make_unique` guarded by the same macro where I've used `make_unique`. Thus, if you are using a compiler that has `std::make_unique` *and* defines that macro, `std::make_unique` is called. In all other cases, our `make_unique` is called.

This way it should be painless to upgrade to C++14 when we do (for 5.0 hopefully!)

### DIFF
--- a/src/fileio/formatfactory.cxx
+++ b/src/fileio/formatfactory.cxx
@@ -14,6 +14,10 @@
 #include <output.hxx>
 #include <string.h>
 
+#ifdef __cpp_lib_make_unique
+using std::make_unique;
+#endif
+
 FormatFactory *FormatFactory::instance = nullptr;
 
 FormatFactory* FormatFactory::getInstance() {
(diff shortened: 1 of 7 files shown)


src/fileio/formatfactory.cxx
```diff
@@ -14,6 +14,10 @@
 #include <output.hxx>
 #include <string.h>
 
+#ifdef __cpp_lib_make_unique
```
bendudson: Should this code go into `utils.hxx` so that `make_unique` is always defined? It does mean putting `using std::make_unique` in a header file though... 
Is this a good time to introduce a namespace, so `bout::make_unique` is always defined, importing `std::make_unique` if needed? 
(happy for other suggestions for namespace)

ZedThree: I was um-ing and ah-ing about that myself! It might be the cleanest thing to do though.

I'm happy with `bout` as the namespace, possibly even `bout::utils`

PR #1398, opened 2018-11-22
Fix setCoefs in InvertPar

arkabokshi:
Field3D and BoutReal override implementations of setCoefs C, D, E in
invert_parderiv.hxx were wrong (called setCoefB everywhere).

This was not picked up by the test because it does not set coefs C, D, E and only uses the Field2D versions.


ZedThree:
Would you be able to expand the test so this doesn't happen again?

arkabokshi:
Have updated the integrated test with C, D, E coefficients. But this still does not test BoutReal and Field3D coefs. John says that a unit test might be more appropriate but we do not know how to go about it.

PR #1399, opened 2018-11-22
More fixes for invparderiv

d7919:
Collection of bug fixes for the parallel inversion code.


d7919:
Another inconsistency -- the serial version will throw if there are open surfaces whilst the parallel cyclic version doesn't.

bendudson:
The serial version was just an initial proof of concept, used for quick tests when developing the more complicated parallel version. Since the parallel version also works in serial, it may be easiest to just delete the serial version.

d7919:
> The serial version was just an initial proof of concept, used for quick tests when developing the more complicated parallel version. Since the parallel version also works in serial, it may be easiest to just delete the serial version.

Agreed!

d7919:
This code could possibly be further improved but I think the bug fix (if it's right) shouldn't wait on this.

PR #1400, opened 2018-11-22
Two small improvements to command line options

### LINKED ISSUE #1378: Running a program with `--help` tries to write to BOUT.settings
This is pretty surprising! Should we actually immediately call `BoutFinalise` and `exit`? That doesn't feel right either, but `./foo --help` should really not touch any files

ZedThree:
- Exit nicely for `--help` (fixes #1378)
- Remove consumed options from argv (alternative to #1383)

### DIFF
--- a/src/bout++.cxx
+++ b/src/bout++.cxx
@@ -64,7 +64,7 @@ const char DEFAULT_LOG[] = "BOUT.log";
 
 #include <strings.h>
 #include <string>
-#include <list>
+#include <vector>
 using std::string;
 #include <sys/types.h>
 #include <sys/stat.h>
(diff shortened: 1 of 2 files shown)


src/bout++.cxx
```diff
@@ -189,60 +183,96 @@ int BoutInitialise(int &argc, char **&argv) {
...
         return 1;
       }
-      i++;
-      data_dir = argv[i];
-      
+
+      data_dir = argv[++i];
+
+      original_argv.push_back(argv[i - 1]);
+      original_argv.push_back(argv[i]);
```
dschwoerer: Why duplicate the push_back code, and not take a full copy initially?
Also, in case this is not an option we are processing here, it does not get copied, as far as I can see.
Is this intentional?

PR #1402, opened 2018-11-23
"Dumb" constructor for Coordinates and Vector tests

ZedThree:
- Add a constructor for `Coordinates` that takes the more useful members which enables using it in the unit tests
- Add a bunch of tests for the Vector-Vector operations (dot and cross products) that need the metric

The Vector tests are basically just sanity checks. They're also really regression tests as I didn't take the time to calculate the correct answer myself and just used the existing results!

### DIFF
--- a/include/bout/coordinates.hxx
+++ b/include/bout/coordinates.hxx
@@ -53,6 +53,12 @@ public:
   /// Constructor interpolating from another Coordinates object
   Coordinates(Mesh *mesh, const CELL_LOC loc, const Coordinates* coords_in);
   
+  /// Standard constructor from input
+  Coordinates(Mesh* mesh, Field2D dx, Field2D dy, BoutReal dz, Field2D J, Field2D Bxy,
+              Field2D g11, Field2D g22, Field2D g33, Field2D g12, Field2D g13,
+              Field2D g23, Field2D g_11, Field2D g_22, Field2D g_33, Field2D g_12,
+              Field2D g_13, Field2D g_23);
+
   ~Coordinates() {}
   
   /*!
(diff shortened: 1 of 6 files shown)


include/bout/coordinates.hxx
```diff
@@ -53,6 +53,12 @@ public:
   /// Constructor interpolating from another Coordinates object
   Coordinates(Mesh *mesh, const CELL_LOC loc, const Coordinates* coords_in);
   
+  /// Standard constructor from input
```
d7919: Should we make this private and make google test (or a helper) a friend such that users don't start trying to use this?

d7919: (or actually test something derived from coordinates that just provides this constructor and no other functionality?)

ZedThree: I'd rather not go with the first, just so as to avoid making googletest a hard dependency. The second might be a reasonable approach though.

But do we need to hide it? It's not actively harmful -- one has to not only make a `Coordinates` object but also tell the `Mesh` about it somehow, and we don't currently have anything for that. I had to add `setCoordinates` to `FakeMesh` to actually use this in the tests

ZedThree: That's reasonable. I'll try using a derived class in the tests

ZedThree:
The latest version allows all fields that are read from `mesh->get()` in the usual constructor to be passed in, along with a flag that defaults to `true` for immediately calling `geometry`. This ensures that the object will be completely constructed, although it's still possible to pass in inconsistent values for e.g. the co- and contravariant metric elements and J/Bxy.

After some discussion with @d7919, `Coordinates` is probably due a much larger rewrite. There are a number of invariants that are not properly maintained. An interface that maintains these invariants would be better, although it would mean properly encapsulating most of the members. Several different constructors could be provided as well, to ensure everything gets set consistently.

Issue #1213, opened 2018-08-21
Delete comm_group?


ZedThree:
I don't think this is used anywhere any more. Delete?

[Source](https://github.com/boutproject/BOUT-dev/blob/master/src/sys/comm_group.cxx)
[Header](https://github.com/boutproject/BOUT-dev/blob/master/include/comm_group.hxx)

bendudson:
Deprecate for now? Then delete next major version.

ZedThree:
Sounds good

d7919:
Closed by #1406

PR #1406, opened 2018-11-23
Remove comm_group as not used anywhere

bendudson:
Provides an interface for MPI gather/scatter operations, but too low level to be really useful. If this functionality is needed again, I would prefer a better API.

Fixes issue #1213


d7919:
I think you need to remove this from `src/sys/makefile`. It's also mentioned in the developer docs so probably want to remove this.

manual/sphinx/developer_docs/code_layout.rst
```diff
@@ -345,8 +345,6 @@ The current source code files are:
 
 - sys
 
-   - :doc:`boutcomm.cxx<../_breathe_autogen/file/boutcomm_8cxx>`
```
ZedThree: This is `boutcomm.cxx` rather than `comm_group.cxx`

d7919: Sorry, there is a line referring to comm_group in there. I'll fix, thanks for spotting!

PR #1409, opened 2018-11-27
Fix MAYBE_UNUSED for some compilers

### LINKED ISSUE #1407: MAYBE_UNUSED causes compiling failure
In the definition of `MAYBE_UNUSED`, if `__has_cpp_attribute` is defined but `__has_cpp_attribute(maybe_unused)` is not, then `MAYBE_UNUSED` is not defined and causes compililation failures.
https://github.com/boutproject/BOUT-dev/blob/13fda6d569486bfda4c100c8547391ade2a15ee8/include/unused.hxx#L43-L51

ZedThree:
Some compilers may have `__has_cpp_attribute` but not
`maybe_unused`. Make sure MAYBE_UNUSED always gets defined to
something

Fixes #1407

### DIFF
--- a/include/unused.hxx
+++ b/include/unused.hxx
@@ -47,7 +47,7 @@
 #endif
 #ifndef MAYBE_UNUSED
 #if defined(__GNUC__)
-# define MAYBE_UNUSED(x) x __attribute__((unused))
+# define MAYBE_UNUSED(x) [[gnu::unused]] x
 #else
 # define MAYBE_UNUSED(x) x
 #endif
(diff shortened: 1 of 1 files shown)


include/unused.hxx
```diff
@@ -44,10 +44,13 @@
 #if __has_cpp_attribute(maybe_unused)
 # define MAYBE_UNUSED(x) [[maybe_unused]] x
 #endif
-#elif defined(__GNUC__)
+#endif
+#ifndef MAYBE_UNUSED
+#if defined(__GNUC__)
 # define MAYBE_UNUSED(x) x __attribute__((unused))
```
d7919: This could be `[[gnu::unused]] x` to make it look more like the maybe_unused branch above? This might also allow all the `MAYBE_UNUSED` calls to not take an argument (not sure if this is desirable).

Issue #1414, opened 2018-11-29
Possible issue with DST/DST_rev?


ZedThree:
Given a simple 8-point `f = sin(x)`, x in `[0, 2pi]`, `DST(f)` gives:
```
f = [0, 0.707107, 1, 0.707107, 1.22465e-16, -0.707107, -1, -0.707107, ]
DST(f) = [(0, 0), (0.226083, 0), (0.944508, 0), (-0.20323, 0), (0.105938, 0), (-0.0593545, 0), (0.0272555, 0), (0, 0), ]
```
while SciPy gives for the same `f`:
```
array([ 1.27275880e+00,  7.39103740e+00, -2.42182027e+00,  1.57009246e-16,
       -1.07899223e+00,  4.73683948e-07, -8.50431216e-01,  2.22044605e-16])
```
Note that for our DST, the peak is at index 2, rather than 1, which is surprising.

For a DST, I would expect to see `fftw_plan_r2r_1d` with `FFTW_RODFT10`, but we just use `fftw_plan_dft_r2c_1d` -- is this right?

Note that this is both with and without OpenMP, so it's not a thread issue.

d7919:
I'd vote for removing the DST code and reimplementing when we properly deal with z-boundaries.

bendudson:
hold on... it is quite useful at the moment for some things. I think the issue is related to reserving one point in Z for the boundaries. It's not done properly yet, but is useful for trying things out. Perhaps just a health warning on it for now, since DST does assume that boundaries are imposed when they are not, at least in the main code. Code using these DST routines apply their own boundaries in Z e.g https://github.com/bendudson/isothermal-4field/blob/master/isothermal-4field.cxx#L256

d7919:
Ah ok, I thought it was more experimental. I think a health warning might be useful as it's likely to be reserved for advanced users at the moment.

PR #1415, opened 2018-11-29
Add a couple of sanity tests for ffts

ZedThree:
Also exposes the `fft_init` function to make the tests silent.

Note I originally wrote tests for `DST`/`DST_rev` as well, but see #1414 for a possible bug there


include/fft.hxx
```diff
@@ -78,4 +78,28 @@ void DST(const BoutReal *in, int length, dcomplex *out);
...
+/// Should the FFT functions find and use an optimised plan?
+void fft_init(bool fft_measure);
+/// Should the FFT functions find and use an optimised plan?
+///
+/// If \p options is not nullptr, it should contain a bool called
+/// "fftw_measure". If it is nullptr, use the global `Options` root
+void fft_init(Options* options = nullptr);
+
+/// Returns the fft of a real signal \p in using fftw_forward
+Array<dcomplex> rfft(const Array<BoutReal>& in);
```
d7919: Can we / should we move all fft routines into this namespace now and make the non-namespaced variants just call the namespaced version?

tests/unit/invert/test_fft.cxx
```diff
@@ -0,0 +1,149 @@
...
+TEST(FFTTest, rfft) {
+
+  // Make sure fft functions are quiet by setting fft_measure to false
+  bout::fft::fft_init(false);
+
+  constexpr int size{8};
+
+  // Calculate sin(x) + cos(2x) on [0, 2pi]
+  Array<BoutReal> indices(size);
+  std::iota(indices.begin(), indices.end(), 0);
```
d7919: Should this really be `0.0` so that `std::iota` generates reals rather than ints that get converted?

d7919: Also a comment to note that after this call `indices` contains  {0,1,2...,size-1} might help.

tests/unit/invert/test_fft.cxx
```diff
@@ -0,0 +1,149 @@
...
+
+  Array<BoutReal> input(size);
+  std::transform(indices.begin(), indices.end(), input.begin(),
+                 [](BoutReal i) -> BoutReal {
+                   return std::sin(i * TWOPI / size) + std::cos(2. * i * TWOPI / size);
+                 });
+
+  Array<dcomplex> output(size);
```
d7919: Shouldn't we expect a smaller output than the input (e.g. size/2)? It might be good to not overallocate as that way we can catch attempts to go beyond our expected bounds better.

tests/unit/invert/test_fft.cxx
```diff
@@ -0,0 +1,149 @@
...
+  std::iota(indices.begin(), indices.end(), 0);
+
+  Array<BoutReal> input(size);
+  std::transform(indices.begin(), indices.end(), input.begin(),
+                 [](BoutReal i) -> BoutReal {
+                   return std::sin(i * TWOPI / size) + std::cos(2. * i * TWOPI / size);
+                 });
+
+  // Compute forward real FFT
+  auto output = bout::fft::rfft(input);
```
d7919: Is it worth checking the length (and type?) of output here is as we would expect?

tests/unit/invert/test_fft.cxx
```diff
@@ -0,0 +1,149 @@
...
+  std::fill(expected_imag.begin(), expected_imag.end(), 0.);
+  expected_imag[1] = -0.5;
+
+  for (int i = 0; i < size; ++i) {
+    EXPECT_NEAR(real(output[i]), expected_real[i], BoutRealTolerance);
+    EXPECT_NEAR(imag(output[i]), expected_imag[i], BoutRealTolerance);
+  }
+}
+
+TEST(FFTTest, irfftWithArray) {
```
d7919: Is it worth having a test that shows that the "round-trip" is a no-op?

include/fft.hxx
```diff
@@ -78,4 +86,39 @@ void DST(const BoutReal *in, int length, dcomplex *out);
  */
 void DST_rev(dcomplex *in, int length, BoutReal *out);
 
+/// Should the FFT functions find and use an optimised plan?
+void fft_init(bool fft_measure);
+/// Should the FFT functions find and use an optimised plan?
```
bendudson: duplicate line 89

src/invert/fft_fftw.cxx
```diff
@@ -457,7 +464,29 @@ void DST_rev(dcomplex *in, int length, BoutReal *out) {
...
-#else
-  throw BoutException("This instance of BOUT++ has been compiled without fftw support.");
 #endif
 }
+
+Array<dcomplex> rfft(const Array<BoutReal>& in) {
+  ASSERT1(!in.empty());
+
+  int size{in.size()};
+  Array<dcomplex> out{(size / 2) + 1};
```
bendudson: Does this handle the case where `size` is odd?

PR #1416, opened 2018-12-04
Move index derivs out of mesh

d7919:
Makes index derivatives free functions, mesh versions just call these.

Using namespace `bout::derivative::index` to store the index derivative routines.


bendudson:
Anyone see what's going on with NetCDF attributes in Travis? This `test-io` test passes for me, but seems to now be failing here.

d7919:
Yes it also fails if you rerun a case that had previously passed on travis. My uninformed guess would be that the netcdf lib or environment has changed.

d7919:
@bendudson it seems the pip install of the netcdf4 module is failing https://travis-ci.org/boutproject/BOUT-dev/jobs/463318511#L897 and this means we end up failing back onto the scipy module which I'm guessing doesn't handle attributes (or in a way we don't deal with).

d7919:
To be explicit it seems to that there is an upgrade to the cftime module from 1.0.2.1 to 1.0.3 and we're unable to pip install the new version currently.

ZedThree:
Sounds like there'll be a bugfix release soon for cftime. We might have to specify a particular version of netcdf4 in the mean time

d7919:
This now also contains a small tweak to the integrated tests (stopCheck) to avoid an unreproducible error on travis.

PR #1418, opened 2018-12-05
Attempt to provide attribute support to datafile when using scipy bac…

d7919:
…kend

Currently untested as I can't seem to be able to create a file with attributes that scipy is happy to read (even using scipy to generate the file!)


ZedThree:
Dropped support for scipy backend in #1524

PR #1419, opened 2018-12-06
Workaround for gcc ICE to do with binding class bound member with

d7919:
transient instance of class

Previous code worked ok for v7.3.1 but failed for some versions of
v7.3.0 (and passed for others!)


bshanahan:
works for me!

PR #1421, opened 2018-12-06
Fix failure of collect() with integer slice argument

johnomotani:
Collect may fail if integer argument is given, e.g. for 'yind'. Fix this by checking for 'numpy.integer' as well as 'int'.

I guess this should go into `master` as it's a bug fix?


d7919:
This is into master so will get into next in the next bug-fix merge I guess.

PR #1423, opened 2018-12-07
Start to introduce zstart and zend

d7919:
Starts to introduce zstart and zend to mesh. These will be required to introduce guard cells in the z-direction. As a first step these are introduced and the size of the guard cells forced to 0 with no parallelisation. The next step is to replace loops over `0, mesh->LocalNz` with `zstart,zend`. This second stage is likely to touch a lot of files so we may wish to introduce the zstart/end first and then slowly migrate loops as separate work.


src/mesh/impls/bout/boutmesh.cxx
```diff
@@ -415,6 +425,12 @@ int BoutMesh::load() {
         _("\tERROR: Cannot split %d Y points equally between %d processors\n"), MY, NYPE);
   }
 
+  MZSUB = MZ / NZPE;
+  if ((MY % NYPE) != 0) {
```
ZedThree: This should presumably be `MZ % NZPE`?

src/mesh/impls/bout/boutmesh.cxx
```diff
@@ -2541,6 +2560,7 @@ void BoutMesh::outputVars(Datafile &file) {
   file.add(MYG, "MYG", false);
   file.add(nx, "nx", false);
   file.add(ny, "ny", false);
+  file.add(LocalNz, "nz", false);
```
ZedThree: Should we add `nz` instead of `LocalNz` here? Or is that for a separate PR?

Also, should we add `MZG` and `MZSUB`? Or also for future PR?

ZedThree:
I've fixed the conflicts btw, happy to also fix my above comments before pushing

ZedThree:
@d7919 Ok with my last four commits?

ZedThree:
For reference, he ok'd them, out-of-band

PR #1424, opened 2018-12-09
Adding toString specialisations

bendudson:
- Handle bool, converting to "true" or "false"
 - Extend precision of BoutReal conversion (update unit test)

Added more unit tests

### DIFF
--- a/include/utils.hxx
+++ b/include/utils.hxx
@@ -371,21 +371,42 @@ inline void checkData(BoutReal UNUSED(f)){};
  */ 
 char* copy_string(const char* s);
 
-/*!
- * Convert a value to a string
- * by writing to a stringstream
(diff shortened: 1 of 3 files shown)


include/utils.hxx
```diff
@@ -371,21 +371,37 @@ inline void checkData(BoutReal UNUSED(f)){};
...
- * by writing to a stringstream
- */
+
+/// Convert a value to a string
+/// by writing to a stringstream
 template <class T>
-const std::string toString(const T& val) {
+std::string toString(const T& val) {
   std::stringstream ss;
   ss << val;
```
ZedThree: `ss << std::boolalpha << val` might remove the need to specialise for `bool`

PR #1425, opened 2018-12-10
Options use variant type (towards a DataFile replacement)

bendudson:
This PR aims to extend the `Options` tree structure to handle all input and output data, including grid files, dump and restart files. This is done by using `variant` types as the storage inside Options, rather than a string as done before. 

## Storage

A `variant` type is a type-safe union, which can hold a set of different types, keeping track of which type is stored at a given time. It was added to `std` in C++17, but is available in standalone header file implementations e.g.  https://github.com/mpark/variant
Some helper routines are defined in `include/bout/sys/variant.hxx`, which imports the `variant` class into the `bout::utils` namespace.

`Options` can now store `bool`, `int`, `BoutReal`, `std::string`, `Field2D`, `Field3D`, `Array<BoutReal>`, `Matrix<BoutReal>`, and `Tensor<BoutReal>` values, and converts between them when possible. An advantage for current use is that putting any of these types into an `Option` then getting it out again is exact, with no writing or reading strings.

## Conversion between types

Conversions e.g.
```
Option option = 32;
std::string str = option; // Converts to "32"
```
or using FieldFactory, passing a Mesh pointer to get a Field
```
Option option = " 1 + sin(y)"
auto value = option.as<Field3D>(mesh);
```
Also added an `attributes` map for each `Option`, which currently stores the `source`, but could be used to store other attributes. That is also a `variant`, so currently attributes can be `bool`, `int`, `BoutReal`, or `std::string`.

If a conversion cannot be done, it should either fail at compile time, or throw a `BoutException`.

Unit tests are in `tests/unit/sys/test_options.cxx` and `tests/unit/sys/test_options_fields.cxx` (for the ones involving a Mesh and fields).

## Options tree manipulation

The options tree nodes can now be copied and joined onto each other:
```
Options option1, option2;
option1["key1"] = 42;
option2["key2"] = option1;
// Now have option2["key2"]["key1"] = 42
```
This assignment makes a copy recursively, since `Options` store their children by value. This means that after the assignment `option1` and `option2["key2"]` point to different `Options` objects. This minimises opportunities for unexpected non-local modifications.

## Reading and writing to NetCDF
Examples are in integrated test `tests/integrated/test-options-netcdf/`

To write the current ``Options` tree (e.g. from BOUT.inp) to a NetCDF file:
```
OptionsNetCDF("settings.nc").write(Options::root());
```
and to read it in again:
```
Options data = OptionsNetCDF("settings.nc").read();
```
Fields can also be stored and written:
```
Options fields;
fields["f2d"] = Field2D(1.0);
fields["f3d"] = Field3D(2.0);
OptionsNetCDF("fields.nc").write(fields);
```
This should allow the input settings and evolving variables to be combined into a single tree (see above on joining trees) and written to the output dump or restart files.

Reading fields is a bit more tricky. Currently 1D data is read as an `Array<BoutReal>`, 2D as `Matrix<BoutReal>` and 3D as `Tensor<BoutReal>`. These can be extracted directly from the `Options` tree, or converted to a `Field` if given a `Mesh`:
```
Options fields_in = OptionsNetCDF("fields.nc").read();
Field2D f2d = fields_in["f2d"].as<Field2D>(mesh);
Field3D f3d = fields_in["f3d"].as<Field3D>(mesh);
```
Currently this only works if the data in the `Matrix` or `Tensor` is the same size as the `Field`. In the case of grid files, the fields only need a part of the global values. Some kind of mapping from the global index to local index is needed, probably defined by `Mesh`. For now it should be possible to be compatible with the current system, so that all quantities from the grid file are accessed through `Mesh::get`.

## Time dependence

When writing the ouput dump files, some variables should have a time dimension added, and then be added to each time they are written. This has been implemented using an attribute: If variables in the `Options` tree have an attribute `"time_dimension"` then that is used as the name of the time dimension in the output file. This allows multiple time dimensions e.g. high frequency diagnostics and low frequency outputs, to exist in the same file.
```
Options data;
data["scalar"] = 1.0;
data["scalar"].attributes["time_dimension"] = "t";
  
data["field"] = Field3D(2.0);
data["field"].attributes["time_dimension"] = "t";
  
OptionsNetCDF("time.nc").write(data);
  
// Update time-dependent values. This can be done without `force` if the time_dimension
// attribute is set
data["scalar"] = 2.0;
data["field"] = Field3D(3.0);
  
// Append data to file
OptionsNetCDF("time.nc", OptionsNetCDF::FileMode::append).write(data);
```
Some issues:
- Currently all variables in the `Options` tree are written when passed to `OptionsNetCDF::write`. This means that the variables with different time dimensions should be stored in different `Options` trees, so they can be written at different times. One possibility is to have an optional argument to `write`, so that only variables with one specified time dimension were updated.


src/sys/options.cxx
```diff
@@ -131,115 +93,162 @@ template <> std::string Options::as<std::string>() const {
...
-    val = false;
-    output_info << _("\tOption ") << full_name << " = false";
+  
+  bool result;
+  
+  if (bout::utils::holds_alternative<bool>(value)) {
+    result = bout::utils::get<bool>(value);
+  
+  } else if(bout::utils::holds_alternative<std::string>(value)) {
+    std::string strvalue = bout::utils::get<std::string>(value);
```
ZedThree: Good place for `auto` -- type is named on right

src/sys/options.cxx
```diff
@@ -131,115 +93,162 @@ template <> std::string Options::as<std::string>() const {
...
+  
+    char c = static_cast<char>(toupper((strvalue)[0]));
```
ZedThree: Also `auto`

include/utils.hxx
```diff
@@ -222,10 +223,11 @@ public:
   void ensureUnique() {
     data.ensureUnique();
   }
-  
+
+  /// Underlying 1D storage array
+  Array<T> data;
```
d7919: This looks like it could be dangerous, although I can see several advantages to doing this. 

Would making Fields a friend of these classes be one way to avoid exposing this more broadly?

The "risk" might be worth it considering some of the benefits.

bendudson: Agree this could be dangerous. Would having a function `getData` to return the Array be better? That would allow Fields to be constructed from Matrix/Tensor, without giving too much access to the internals. It does still rely on implementation detail though.

For reading in grid files, I'd like to be able to select a region of a `Tensor` and copy it into a `Field3D`, based on a global offset. I don't think indexing can use `Ind3D` since the array size may be larger than a field and a different shape. Is there an efficient way to convert from `Ind3D` to (x,y,z), add an offset to each index, then use that to index a Tensor?

d7919: Another way to do it would be something like:

~~~~
Region<Ind3D> theTensorSubset(xstartSection, xendSection......);
auto tensorIndices = theTensorSubset.getIndices();
auto fieldIndices = field.getRegion("RGN_ALL").getIndices();
ASSERT1(tensorIndices.size() == fieldIndices.size();
int count = 0;
BOUT_FOR_SERIAL(i, field.getRegion("RGN_ALL")) {
theField[i] = theTensor[tensorIndices[count]];
count++;
}
~~~~

This doesn't use a

include/external/mpark/config.hpp
```diff
@@ -0,0 +1,79 @@
+// MPark.Variant
```
d7919: Should/could we treat this (and related) as an external sub-module rather the distributing with BOUT++?

d7919: I think the advantage is that it might make it easier to get updates should there be bug fixes etc.  rather than manually merging changes (and it perhaps makes it clearer that we shouldn't touch these files, although that's indicated by being in externals) but the downside is that an initial checkout of BOUT++ will still need a network connection at the configure stage.

ZedThree: I'd prefer to stick it as a submodule under `externalpackages/`

bendudson:
I think this is ready to go in, if it is a direction people would like to go in. It is not yet ready to replace DataFile, but is a step towards that, and can already do useful things. The main questions are:
1. Are we ok with a hard dependency on an external header file library?
2. How should the header be included? As a submodule, or in the code?
3. Where should the header go? It is needed to compile BOUT++ code, including physics models, so would need to be installed into the `include` path somewhere.

d7919:
I'm happy for this to be a hard dependency as it should be trivial to obtain and I'd prefer inclusion as a submodule to make it easy to keep things up to date whilst maintaining a fixed version with each commit.

The following is probably quite a horrible way of making the header(s) in `externalpackages/variant` available. In `include/external/variant` we can have something like

~~~~
#define THE_VARIANT_LOCATION VARIANT_LOCATION_SET_IN_BOUT_FLAGS
#include THE_VARIANT_LOCATION
~~~~

(https://stackoverflow.com/a/42858714 suggests this should be valid but I've not tested). So for example if we had `BOUT_FLAGS+=-DVARIANT_LOCATION_SET_IN_BOUT_FLAGS=${BOUT_TOP}/externalpackages/variant/variant.hxx` this would point to the correct place. If we install bout then we just have to make sure the macro is updated to point to the appropriate location. Alternatively I guess we can just copy/link the headers into `include` at configure time.

ZedThree:
I'm ok with a hard dependency, and having it as a submodule has a bunch of advantages, including being able to easily upgrade it.

For installation, I would propose just copying the relevant files into `include`.

include/bout/sys/variant.hxx
```diff
@@ -0,0 +1,155 @@
...
+
+namespace details {
+
+/// Helper class for casting between two types
+/// The general case is that casting can't be done
+///
+/// Note: This must be a class/struct because partial
+/// specialisation is not allowed for functions
+template <typename Target, typename Source, bool>
+struct _StaticCastOrThrow {
```
ZedThree: I don't think we're allowed to name things with a single underscore followed by an uppercase letter.

Given this is in `bout::details::`, maybe just drop the underscore?

include/options.hxx
```diff
@@ -158,13 +161,76 @@ public:
...
+  class AttributeType : public bout::utils::variant<bool, int, BoutReal, std::string> {
+  public:
+    using Base = bout::utils::variant<bool, int, BoutReal, std::string>;
+
+    /// Constructor
+    AttributeType() {}
+    /// Copy constructor
+    AttributeType(const AttributeType& other) : Base(other) {}
+    /// Move constructor
+    AttributeType(AttributeType&& other) : Base(other) {}
```
ZedThree: I think this should this be `Base(std::move(other))` as `other` becomes an l-value in the initialiser list

include/options.hxx
```diff
@@ -158,13 +161,76 @@ public:
   /// @param[in] sectionName   Name of the section, including path from the root
   Options(Options *parent_instance, std::string full_name)
       : parent_instance(parent_instance), full_name(std::move(full_name)){};
-  
+
+  /// Copy constructor
+  Options(const Options& other);
```
ZedThree: Should we also implement the move ctor/assignment operator?

ZedThree:
There are three tests that print something to screen. They might need one of the `output`s disabling for the duration of the test?
```
[ RUN      ] OptionsFieldTest.RetrieveStringfromField3D
	Option  = <Field3D> ()
[ RUN      ] OptionsFieldTest.RetrieveStringfromField2D
	Option  = <Field2D> ()
[ RUN      ] OptionsFieldTest.RetrieveField3DfromString
	Option paralleltransform = identity (default)
[----------] 11 tests from OptionsFieldTest (2 ms total)
```

ZedThree:
Thanks @bendudson ! :tada: Progress towards replacing `DataFile`!

PR #1432, opened 2018-12-10
Add region arguments for toFieldAligned/fromFieldAligned

johnomotani:
Add region arguments (where it is valid to give `RGN_NOX` or `RGN_NOBNDRY`) so functions like `toFieldAligned`/`fromFieldAligned` can be used both when y-guard cells have been set and when they have not. e.g. when using `fromFieldAligned` on the results of interpolation or derivatives, y-guard cells cannot be included, but when using `toFieldAligned` on the inputs to interpolation or derivatives, y-guard cells must be included.

Pass `RGN_NOBNDRY` to `fromFieldAligned` in various places where the guard cells aren't valid.

Pulled out of #1176; is independent of other stuff in that PR, which now probably won't be merged.


src/mesh/parallel/shiftedmetric.cxx
```diff
@@ -177,18 +176,24 @@ void ShiftedMetric::shiftZ(const BoutReal *in, const std::vector<dcomplex> &phs,
 }
 
 //Old approach retained so we can still specify a general zShift
-const Field3D ShiftedMetric::shiftZ(const Field3D &f, const Field2D &zangle) {
+const Field3D ShiftedMetric::shiftZ(const Field3D &f, const Field2D &zangle, const REGION region) {
   ASSERT1(&mesh == f.getMesh());
+  ASSERT1(region == RGN_NOX || region == RGN_NOBNDRY); // Never calculate x-guard cells here
```
bendudson: Not sure this is necessary; it may be inefficient to perform shifting in the X guard cells, but shouldn't cause a problem. There are some cases where it may be useful e.g. calculating `D2DXDY`

johnomotani: I think the problem I was originally trying to avoid was when shifting in the corner guard cells. I remember getting 'uninitialized value' warnings from Valgrind that I wanted to protect against. They must have been coming from interpolated `zShift` fields though, so probably the right solution is to make sure all points in `zShift`s get initialized to something even when checking is turned off, instead of trying to force the corners to not be used here.

I'll get rid of this check :+1:

src/mesh/parallel/shiftedmetric.cxx
```diff
@@ -129,30 +129,28 @@ void ShiftedMetric::calcYUpDown(Field3D &f) {
...
 
   Field3D result(&mesh);
   result.allocate();
-  
-  for(int jx=0;jx<mesh.LocalNx;jx++) {
-    for(int jy=0;jy<mesh.LocalNy;jy++) {
-      shiftZ(f(jx,jy), phs[jx][jy], result(jx,jy));
-    }
+
+  for (const auto &i : mesh.getRegion2D(REGION_STRING(region))) {
```
d7919: Could this use the `BOUT_FOR`(`_SERIAL` )macro rather than the range based loop?

src/mesh/parallel/shiftedmetric.cxx
```diff
@@ -177,18 +175,23 @@ void ShiftedMetric::shiftZ(const BoutReal *in, const std::vector<dcomplex> &phs,
...
-    for(int jy=0;jy<mesh.LocalNy;jy++) {
-      shiftZ(f(jx,jy), mesh.LocalNz, zangle(jx,jy), result(jx,jy));
-    }
+  // We only use methods in ShiftedMetric to get fields for parallel operations
+  // like interp_to or DDY.
+  // Therefore we don't need x-guard cells, so do not set them.
+  // (Note valgrind complains about corner guard cells if we try to loop over
+  // the whole grid, because zShift is not initialized in the corner guard
+  // cells.)
+  for(const auto &i : mesh.getRegion2D(REGION_STRING(region))) {
```
d7919: BOUT_FOR?

src/mesh/parallel/shiftedmetric.cxx
```diff
@@ -73,28 +73,24 @@ ShiftedMetric::ShiftedMetric(Mesh &m) : mesh(m), zShift(&m) {
...
-    for(int jy=0;jy<mesh.LocalNy;jy++){
-      for(int jz=0;jz<nmodes;jz++) {
-  	BoutReal kwave=jz*2.0*PI/zlength; // wave number is 1/[rad]
-  	fromAlignedPhs[jx][jy][jz] = dcomplex(cos(kwave*zShift(jx,jy)) , -sin(kwave*zShift(jx,jy)));
-  	toAlignedPhs[jx][jy][jz] =   dcomplex(cos(kwave*zShift(jx,jy)) ,  sin(kwave*zShift(jx,jy)));
-      }
+  BOUT_FOR(i, zShift.getRegion("RGN_ALL")) {
+    for(int jz=0;jz<nmodes;jz++) {
+      BoutReal kwave=jz*2.0*PI/zlength; // wave number is 1/[rad]
+      fromAlignedPhs[i.x()][i.y()][jz] = dcomplex(cos(kwave*zShift(i.x(),i.y())) , -sin(kwave*zShift(i.x(),i.y())));
```
d7919: I think you can just index `zShift` using `i` (i.e. `zShift[i]`) here.

src/mesh/parallel/shiftedmetric.cxx
```diff
@@ -73,28 +73,24 @@ ShiftedMetric::ShiftedMetric(Mesh &m) : mesh(m), zShift(&m) {
...
   //Yup/Ydown phases -- note we don't shift in the boundaries/guards
-  for(int jx=0;jx<mesh.LocalNx;jx++){
-    for(int jy=mesh.ystart;jy<=mesh.yend;jy++){
-      BoutReal yupShift = zShift(jx,jy) - zShift(jx,jy+1);
-      BoutReal ydownShift = zShift(jx,jy) - zShift(jx,jy-1);
-      
-      for(int jz=0;jz<nmodes;jz++) {
-  	BoutReal kwave=jz*2.0*PI/zlength; // wave number is 1/[rad]
+  BOUT_FOR(i, zShift.getRegion("RGN_ALL")) {
+    BoutReal yupShift = zShift(i.x(),i.y()) - zShift(i.x(),i.y()+1);
```
d7919: Index with `i` and `i.yp()` possible here I think.

src/mesh/parallel/shiftedmetric.cxx
```diff
@@ -109,50 +105,44 @@ void ShiftedMetric::calcYUpDown(Field3D &f) {
   Field3D& yup = f.yup();
   yup.allocate();
 
-  for(int jx=0;jx<mesh.LocalNx;jx++) {
-    for(int jy=mesh.ystart;jy<=mesh.yend;jy++) {
-      shiftZ(&(f(jx,jy+1,0)), yupPhs[jx][jy], &(yup(jx,jy+1,0)));
-    }
+  BOUT_FOR(i, f.getRegion("RGN_NOX")) {
+    shiftZ(&(f(i.x(),i.y()+1,0)), yupPhs[i.x()][i.y()], &(yup(i.x(),i.y()+1,0)));
```
d7919: Again I think you can use `i.yp()`.

src/mesh/parallel/shiftedmetric.cxx
```diff
@@ -109,50 +105,44 @@ void ShiftedMetric::calcYUpDown(Field3D &f) {
...
 
   Field3D& ydown = f.ydown();
   ydown.allocate();
 
-  for(int jx=0;jx<mesh.LocalNx;jx++) {
-    for(int jy=mesh.ystart;jy<=mesh.yend;jy++) {
-      shiftZ(&(f(jx,jy-1,0)), ydownPhs[jx][jy], &(ydown(jx,jy-1,0)));
-    }
+  BOUT_FOR(i, f.getRegion("RGN_NOX")) {
+    shiftZ(&(f(i.x(),i.y()-1,0)), ydownPhs[i.x()][i.y()], &(ydown(i.x(),i.y()-1,0)));
```
d7919: `i.ym()`.

src/mesh/parallel/shiftedmetric.cxx
```diff
@@ -109,50 +105,44 @@ void ShiftedMetric::calcYUpDown(Field3D &f) {
...
   Field3D result(&mesh);
   result.allocate();
-  
-  for(int jx=0;jx<mesh.LocalNx;jx++) {
-    for(int jy=0;jy<mesh.LocalNy;jy++) {
-      shiftZ(f(jx,jy), phs[jx][jy], result(jx,jy));
-    }
+
+  BOUT_FOR(i, f.getRegion(region)) {
+    shiftZ(f(i.x(),i.y()), phs[i.x()][i.y()], result(i.x(),i.y()));
```
d7919: Could you replace `f(i.x(), i.y())` with `&f(i.x(), i.y(), 0)` and the same for `result`. Note I think you want `f.getRegion2D(region)` (or `f.getMesh()->getRegion2D(region)`) as currently you're looping over `z` as well.

src/mesh/parallel/shiftedmetric.cxx
```diff
@@ -177,18 +167,23 @@ void ShiftedMetric::shiftZ(const BoutReal *in, const std::vector<dcomplex> &phs,
...
-      shiftZ(f(jx,jy), mesh.LocalNz, zangle(jx,jy), result(jx,jy));
-    }
+  // We only use methods in ShiftedMetric to get fields for parallel operations
+  // like interp_to or DDY.
+  // Therefore we don't need x-guard cells, so do not set them.
+  // (Note valgrind complains about corner guard cells if we try to loop over
+  // the whole grid, because zShift is not initialized in the corner guard
+  // cells.)
+  BOUT_FOR(i, f.getRegion(region)) {
+    shiftZ(f(i.x(), i.y()), mesh.LocalNz, zangle(i.x(),i.y()), result(i.x(), i.y()));
```
d7919: See previous comments.

src/mesh/parallel/shiftedmetric.cxx
johnomotani: @d7919: This loop should be using `getRegion2D`, so can't use `i.yp` for `f`. I've updated to follow the rest of your comments.

johnomotani:
`test-yupdown` is failing with OpenMP. The test passes if I change the loop in this `shiftZ` method from `BOUT_FOR` to `BOUT_FOR_SERIAL`
https://github.com/boutproject/BOUT-dev/blob/baa630ce774f167b1593e3db8f3672e0c8194826/src/mesh/parallel/shiftedmetric.cxx#L138-L152

I can't see why the loop should not parallelize though. The loop body for each `i` should be unaffected by the result for any other `i`... Anyone got any hints for debugging?

d7919:
I don't have the code in front of me but done of these methods use a class member working space (vector) which would lead to race competition etc. I think this could probably be removed from class and made a routine local Array without performance penalty. Can look later if any help

d7919:
I've pushed a branch at https://github.com/boutproject/BOUT-dev/tree/region-arguments-tofromFieldAligned-shiftZ-openmp which might help fix this.

src/mesh/parallel/shiftedmetric.cxx
d7919: @johnomotani  it's possible to index a `Field3D` with an `Ind2D`, you just need to specify the `z` index as well. This uses a multiply and add to get the corresponding `Ind3D` as opposed to three divides and a modulus to call `i.x()` and `i.y()`, as such I'd expect the `Ind2D` approach to be more efficient. I'll add these changes to the branch that fixes the OpenMP stuff

ZedThree:
This needs updating now #1345 is in

johnomotani:
I've merged `next` now.

This PR does two things:
1. adds and makes use of `REGION` arguments for the `toFieldAligned` `fromFieldAligned` methods.
2. Uses `Tensor` for the cached phases in `ShiftedMetric` instead of `arr3Dvec`

I think 1 is good and should be merged. @ZedThree do you want 2, and if so should I split it into a separate PR?

johnomotani:
This is failing with OpenMP enabled again. If I switch the `BOUT_FOR` to a `BOUT_FOR_SERIAL` in this loop, it passes
https://github.com/boutproject/BOUT-dev/blob/e1857ec05183461f9a819645011c9fb845058819/src/mesh/parallel/shiftedmetric.cxx#L193-L213
Can anyone see the race condition or other error?

d7919:
Probably unrelated but

https://github.com/boutproject/BOUT-dev/blob/e1857ec05183461f9a819645011c9fb845058819/src/mesh/parallel/shiftedmetric.cxx#L205

doesn't set the `jz=0` point which I think we probably should as it's not initialised anywhere (this could just be `shifted_temp[0] = 0` before the loop)

d7919:
The only place inside the loop that I could see an issue arising is with

https://github.com/boutproject/BOUT-dev/blob/e1857ec05183461f9a819645011c9fb845058819/src/mesh/parallel/shiftedmetric.cxx#L211

As we're writing to `current_result` and this is a shared object. However, it looks like it should be fine as `i.yp(phase.y_offset)` should be unique for each `i` which is the OpenMP loop variable.

ZedThree:
> Probably unrelated but
> 
> [BOUT-dev/src/mesh/parallel/shiftedmetric.cxx](https://github.com/boutproject/BOUT-dev/blob/e1857ec05183461f9a819645011c9fb845058819/src/mesh/parallel/shiftedmetric.cxx#L205)
> 
> Line 205 in [e1857ec](/boutproject/BOUT-dev/commit/e1857ec05183461f9a819645011c9fb845058819)
> 
>  for (int jz = 1; jz < nmodes; ++jz) { 
> doesn't set the `jz=0` point which I think we probably should as it's not initialised anywhere (this could just be `shifted_temp[0] = 0` before the loop)

L232 has been clobbered, it should look like:

https://github.com/boutproject/BOUT-dev/blob/ace3f77c9b3d153b3c27330dc0c651564f0ae488/src/mesh/parallel/shiftedmetric.cxx#L232

johnomotani:
@ZedThree I'd changed `f_fft` to a `Tensor` so couldn't get an `Array` from it to initialize `shifted_temp` with. I thought it would be OK to move getting values out of `f_fft` down into the loop instead, using `shifted_temp[jz] = f_fft(i.x(), i.y() + phase.y_offset, jz) * phase.phase_shift(i.x(), i.y(), jz);` instead of `shifted_temp[jz] *= phase.phase_shift(i.x(), i.y(), jz);`.

I've just pushed a commit replacing `Tensor<dcomplex> f_fft` with `Matrix<Array<dcomplex>> f_fft` so the structure looks a lot more like it did previously. The unit tests now pass with OpenMP. I still don't understand what the problem was though... I tried adding `shifted_temp.ensureUnique()` and `shifted_temp[0] = 0.`, giving the method below, but still got the same errors
```
std::vector<Field3D>
ShiftedMetric::shiftZ(const Field3D& f,
                      const std::vector<ParallelSlicePhase>& phases) const {

  const int nmodes = mesh.LocalNz / 2 + 1;

  // FFT in Z of input field at each (x, y) point
  Tensor<dcomplex> f_fft(mesh.LocalNx, mesh.LocalNy, nmodes);

  BOUT_FOR(i, mesh.getRegion2D("RGN_ALL")) {
    rfft(&f(i, 0), mesh.LocalNz, &f_fft(i.x(), i.y(), 0));
  }

  std::vector<Field3D> results{};

  for (auto& phase : phases) {
    // In C++17 std::vector::emplace_back returns a reference, which
    // would be very useful here!
    results.emplace_back(&mesh);
    auto& current_result = results.back();
    current_result.allocate();
    current_result.setLocation(f.getLocation());
  
    BOUT_FOR(i, mesh.getRegion2D("RGN_NOY")) {
      // Deep copy the FFT'd field
      Array<dcomplex> shifted_temp(nmodes);
      shifted_temp.ensureUnique();
    
      shifted_temp[0] = 0.;
      for (int jz = 1; jz < nmodes; ++jz) {
        shifted_temp[jz] = f_fft(i.x(), i.y() + phase.y_offset, jz)
                           * phase.phase_shift(i.x(), i.y(), jz);
      }
  
      irfft(shifted_temp.begin(), mesh.LocalNz,
            &current_result(i.yp(phase.y_offset), 0));
    }
  } 
     
  return results;
} 
```

johnomotani:
The previous test failure was because I'd messed up an index in `shiftZ`; I fixed it in c2f9eff, but was concerned that it hadn't been caught by the unit tests. It was missed because the input for the unit tests was constant in x and y, so I swapped some pairs of entries in the input field, and corresponding ones in the expected results. The updated unit test would catch the error.

bendudson:
Is this PR good to go in now, if the conflict is fixed?

include/bout/mesh.hxx
```diff
@@ -702,18 +702,18 @@ class Mesh {
   ///////////////////////////////////////////////////////////
 
   /// Transform a field into field-aligned coordinates
-  const Field3D toFieldAligned(const Field3D &f) {
-    return getParallelTransform().toFieldAligned(f);
+  const Field3D toFieldAligned(const Field3D &f, const REGION region = RGN_NOX) {
```
bendudson: Should the default just be `RGN_ALL` since (I think) that's what it was before?

tests/unit/mesh/parallel/test_shiftedmetric.cxx
```diff
@@ -38,28 +43,28 @@ class ShiftedMetricTest : public ::testing::Test {
     Field3D input_temp{mesh};
 
     fillField(input_temp, {{{1., 2., 3., 4., 5.},
+                            {2., 1., 3., 4., 5.},
+                            {1., 3., 2., 4., 5.},
+                            {1., 2., 4., 3., 5.},
+                            {1., 2., 3., 5., 4.},
```
ZedThree: Why have these changed? Was there a (potential) bug the previous pattern was hiding? This might need a comment

src/mesh/parallel/shiftedmetric.cxx
```diff
@@ -66,35 +66,21 @@ void ShiftedMetric::cachePhases() {
...
-      }
+  // Allocate storage for our 3d phase information.
+  fromAlignedPhs = Tensor<dcomplex>(mesh.LocalNx, mesh.LocalNy, nmodes);
+  toAlignedPhs = Tensor<dcomplex>(mesh.LocalNx, mesh.LocalNy, nmodes);
+
+  //To/From field aligned phases
+  BOUT_FOR(i, mesh.getRegion2D("RGN_ALL")) {
+    for(int jz=0;jz<nmodes;jz++) {
+      BoutReal kwave=jz*2.0*PI/zlength; // wave number is 1/[rad]
+      fromAlignedPhs(i.x(), i.y(), jz) =
```
ZedThree: Is it worth caching the values of `i.x()`, `i.y()` outside the z-loop? Probably won't make much difference in the constructor, but could be more important in the other methods

(also clang-format please :slightly_smiling_face:)

tests/unit/mesh/parallel/test_shiftedmetric.cxx
johnomotani: There was a bug, fixed in c2f9eff, that the previous pattern missed because the input was constant in y. Will add a comment.

PR #1433, opened 2018-12-10
Location checking in ShiftedMetric::shiftZ()

johnomotani:
Fixing the low-level `shiftZ` method means that `toFieldAligned`/`fromFieldAligned` also do the right thing.

Check that input to shiftZ is at same location as the shift angle `zShift` (i.e. `CELL_CENTRE`). Set location of result returned from `zShift` to be the same as the location of the input.

Extracted from #1176, which will probably not now be merged.


d7919:
Should we do the same thing in `calcYupYdown`?

ZedThree:
If we now move the parallel transform into `Coordinates`, will this be outdated, or just need updating?

bendudson:
Codecov seems to comment on the most random things. It seems very excited about an extra 6 lines of coverage! Is there method to this madness? (actually maybe I don't want to know; it's quite charming really).

d7919:
> Codecov seems to comment on the most random things. It seems very excited about an extra 6 lines of coverage! Is there method to this madness? (actually maybe I don't want to know; it's quite charming really).

We explicitly try to disable these comments in the codecov yaml so it is a bit odd! The only thing I can think of is it might be triggered by the presence of certain phrases in the commit message (like we can use to ask travis to skip tests for a particular commit etc.) but I don't think this is likely.

ZedThree:
It seems to be failing to find the yaml file, as it also doesn't ignore the unit tests: https://codecov.io/gh/boutproject/BOUT-dev/tree/dfb5215ebdc5695a757ee473edb6a72f9724f40c

I raised an issue with their support the last time this happened, but I'm yet to hear a response

ZedThree:
Deleted codecov comment as it's not helpful

PR #1434, opened 2018-12-10
Fix FV::Div_par_K_Grad_par

### LINKED ISSUE #1429: FV::Div_par_K_Grad_par region bug?
Do we know if `FV::Div_par_K_Grad_par` is working? I happened to look at it, and the loop is over `"RGN_ALL"` where I would expect `"RGN_NOBNDRY"`
https://github.com/boutproject/BOUT-dev/blob/ea2986bdbb250ce257290888150da53a8b11299b/src/mesh/fv_ops.cxx#L138

Last time the loop was written out as a nested loop (at 72f1ce2), the bounds were `xstart`/`xend` and `ystart`/`yend`
https://github.com/boutproject/BOUT-dev/blob/72f1ce2f88f55c19d0dbac9c274c6a9173b137b6/src/mesh/fv_ops.cxx#L118-L120

johnomotani:
Should loop over RGN_NOBNDRY, previously loop was over RGN_ALL, which causes some indexes to go out of bounds.

Fixes #1429.

### DIFF
--- a/src/mesh/fv_ops.cxx
+++ b/src/mesh/fv_ops.cxx
@@ -135,7 +135,7 @@ namespace FV {
     
     Coordinates *coord = fin.getCoordinates();
 
-    BOUT_FOR(i, result.getRegion("RGN_ALL")) {
+    BOUT_FOR(i, result.getRegion("RGN_NOBNDRY")) {
       // Calculate flux at upper surface
       
       const auto iyp = i.yp();


d7919:
Are there any example cases that we could turn into a small test to help catch issues with the FV operators?

bendudson:
There is an MMS test for this diffusion operator here: https://github.com/boutproject/BOUT-dev/tree/master/examples/finite-volume/diffusion
Apparently that was not sufficient to spot this out of bounds error; I guess the values which were calculated using out of bounds data are never used.

For the `FV::Div_par` operator there is an MMS test here: https://github.com/boutproject/BOUT-dev/tree/master/examples/finite-volume/fluid

d7919:
With high enough checking the diffusion case does actually throw (but I think the required check level has to be set manually).

Is it worth moving these into the automatically run tests?

d7919:
> There is an MMS test for this diffusion operator here: https://github.com/boutproject/BOUT-dev/tree/master/examples/finite-volume/diffusion
> Apparently that was not sufficient to spot this out of bounds error; I guess the values which were calculated using out of bounds data are never used.
> 
> For the `FV::Div_par` operator there is an MMS test here: https://github.com/boutproject/BOUT-dev/tree/master/examples/finite-volume/fluid

Also whilst I can run either the test case or the python file this doesn't actually seem to be an mms test as it doesn't seem to explore either convergence or difference between expected results -- am I missing something with these?

bendudson:
I think I remember why this was not RGN_ALL originally. Inside the loop the flux through the upper edge of each cell is calculated, then removed from one cell and added to the other. This calculation needs to be done in the first point in the guard cell at the bottom (ystart-1), or the flux from the guard cell into the domain will be missed. 

This might be a good place to demo custom regions: RGN_ALL + 1 layer of points in y.

johnomotani:
@bendudson the behaviour you describe doesn't seem to be what's implemented any more. I think the fluxes through upper and lower edges are calculated for each point independently. I guess this means the calculation of fluxes is duplicated, which could be avoided with the custom region you suggest.
https://github.com/boutproject/BOUT-dev/blob/ea2986bdbb250ce257290888150da53a8b11299b/src/mesh/fv_ops.cxx#L138

bendudson:
@johnomotani yes you are right. The number of calculations is doubled, which could be avoided with a custom region.

PR #1440, opened 2018-12-11
Removes explicit request for timezone

### LINKED ISSUE #1428: Duplicate time zone
The output of `std::string toString<>(const time_t& time)` seems to duplicate the time zone -- for instance lines like `Run started at  : Mon 10 Dec 2018 12:23:38 GMT GMT` appear in the output of jobs (note the duplicate `GMT`). 

I suspect the format used in the `strftime(buffer, 80, "%Ec %Z", tm);` call probably just needs to drop the `%Z`, however this may be LOCALE dependent?

d7919:
The "Ec" requests time in the current locale's alternate format -- this may or may not contain a timezone, however I guess now that we're using locales we should probably try to respect the locale's preferred format for things where possible.

Fixes #1428

### DIFF
--- a/src/sys/utils.cxx
+++ b/src/sys/utils.cxx
@@ -150,6 +150,6 @@ const std::string toString<>(const time_t& time) {
 
   // Older compilers
   char buffer[80];
-  strftime(buffer, 80, "%Ec %Z", tm);
+  strftime(buffer, 80, "%Ec", tm);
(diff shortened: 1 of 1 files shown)


d7919:
Testing this locally I now don't get the time zone at all and the original case that had two GMT doesn't seem repeatable so I'm closing this.

Issue #829, opened 2018-02-02
All Fields should have a location


dschwoerer:
Currently only Field3D supports it.

See #64 

See https://github.com/boutproject/BOUT-dev/pull/818#discussion_r165302485

d7919:
Field2D has location now, is this enough to close the issue or should FieldPerp have a location as well?

johnomotani:
I'd vote for FieldPerp having a location too. I think it's a 'nice to have' rather than important though. For example, if someone calls `Laplacian::solve(FieldPerp rhs)`, at the moment we can't check the location. Not the best example because `solve(Field3D)` 'should' be called, but still...

ZedThree:
Fixed in #1198 and #1441

PR #1441, opened 2018-12-12
Add FieldPerp location

d7919:
Aimed at addressing #829 

It might also be worth exploring if this might allow us to squash `FieldPerp` into the generated fieldops code now.

Initially marked as a work in progress, as whilst I've (hopefully) added appropriate location code to all the routines in FieldPerp there may be additional places we need to do checks and/or remember to set location on the result.

### DIFF
--- a/include/field.hxx
+++ b/include/field.hxx
@@ -32,9 +32,10 @@ class Field;
 #include <cstdio>
 
 #include "bout_types.hxx"
+#include "boutexception.hxx"
+#include "msg_stack.hxx"
 #include "stencils.hxx"
 #include <bout/rvec.hxx>
-#include "boutexception.hxx"
 
 #include "unused.hxx"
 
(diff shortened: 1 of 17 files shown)


d7919:
@johnomotani I think the only places that are left to change/update is in the Laplacian code -- I'm not quite sure how the Laplacian location systems works/what checks we should make. Do you think you'd get chance to have a look at this?

johnomotani:
It should be simple, I can take a look at it :+1: 

Quick summary is: each Laplacian object has a `CELL_LOC location` member that's set in the constructor (it's not a const member at the moment but maybe should be...). All the inputs and outputs should be at `location`. [Having said that, I'm not sure whether `CELL_XLOW` is actually supported properly in the Laplacian solvers...]

d7919:
Thanks @johnomotani

johnomotani:
Just to note: I don't think we have tests for any `Laplacian` with staggered grids. We should make some, and I don't think there's any particular difficulty to convert the existing tests to add a staggered-grid case, but it's a little more work than just adding the location check/set calls that I did in the last commit.

PR #1446, opened 2018-12-12
Pass f_aligned to standardDerivative in DDY/D2DY2/D4DY4

johnomotani:
Previously the branch of y-derivatives that transforms to field-aligned coordinates was passing the input field (in orthogonal coordinates) to the derivative instead of the transformed `f_aligned`.

Also adds a check of the field-aligned branch of derivatives to `test-yupdown` and makes `test-yupdown` part of the standard test-suite.


d7919:
Good spot, thanks. I thought this was covered by the test suite but it seems not!

PR #1447, opened 2018-12-12
Verbose output

ZedThree:
Adds `output_verbose` for when you sometimes want more information, but don't want to recompile the library. Comes on with `./program -v -v`, but could come on lower?

I've had to modify `DummyOutput` as that tried to throw if it was enabled which seems to defeat the purpose a little bit.

Only currently used in registering regions and derivatives as they are a little noisy at the minute

### DIFF
--- a/src/bout++.cxx
+++ b/src/bout++.cxx
@@ -338,13 +338,14 @@ int BoutInitialise(int &argc, char **&argv) {
       return 1;
     }
   }
-  
-  output_error.enable(verbosity>0);
-  output_warn.enable(verbosity>1);
-  output_progress.enable(verbosity>2);
-  output_info.enable(verbosity>3);
-  output_debug.enable(verbosity>4); //Only actually enabled if also compiled with DEBUG
-  
+
+  output_error.enable(verbosity > 0);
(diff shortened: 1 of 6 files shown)


src/bout++.cxx
```diff
@@ -338,13 +338,14 @@ int BoutInitialise(int &argc, char **&argv) {
...
-  output_info.enable(verbosity>3);
-  output_debug.enable(verbosity>4); //Only actually enabled if also compiled with DEBUG
-  
+
+  output_error.enable(verbosity > 0);
+  output_warn.enable(verbosity > 1);
+  output_progress.enable(verbosity > 2);
+  output_info.enable(verbosity > 3);
+  output_debug.enable(verbosity > 4); // Only actually enabled if also compiled with DEBUG
+  output_verbose.enable(verbosity > 5);
```
d7919: Should this be the same level as debug (or even debug > verbose)? The choice of debug or not is configure time whilst verbose is just the verbosity?

ZedThree: Yes, probably makes sense for debug to be at 5 and verbose at 4

Issue #1452, opened 2018-12-14
Deprecate invert_laplace free functions?


johnomotani:
There's a comment above the `invert_laplace()` functions saying:
https://github.com/boutproject/BOUT-dev/blob/b30d255341fc7544ae56ba91e4c792c2e6df68b8/include/invert_laplace.hxx#L237-L238
`laplace_tridag_coefs` is still used in `Delp2()` and some boundary conditions, so we probably need to keep that, but can we remove/deprecate the `invert_laplace()` free functions?

bendudson:
I'm in favour of deprecating them for now, then removing them in the next release

ZedThree:
Deprecated in #1453

PR #1453, opened 2018-12-14
Deprecate invert_laplace() free functions

johnomotani:
Addresses #1452.


ZedThree:
Ah, definitely! I assumed we had already dropped all uses of them. It looks like the last one is used all over the place. We should replace all uses -- make sure there are no deprecated warnings from the library either

johnomotani:
`invert_laplace` has also been removed from the tests. The only exception `test-laplace` which tests both a `Laplacian` object and `invert_laplace` functions, so makes sense to keep there.

Also updated the boundary conditions in `tests/MMS/elm-pb` to get the test to run: replaced the `shifted` modifier which no longer exists with `toFieldAligned` which I think does the same thing. The test is still running, so don't know if it passes yet...

johnomotani:
The last place `invert_laplace` is used in `BOUT-dev` is in the `gyroPade0` and `gyroPade1` functions. Those functions only seem to be used in the the `gyro-gem` example and `test-gyro` test, so may not be worth updating. I guess the 'fix' would be to have a `gyroAverage` class with the `gyroPade*` functions as methods, something like
```
class gyroAverage {
public:
  gyroAverage(std::string opts_section = "gyroaverage",
              CELL_LOC location = CELL_CENTRE, Mesh* mesh_in = mesh)
              : laplace_solver(nullptr) {
    laplace_solver = Laplacian::create(Options::getRoot()->getSection(opts_section),
                                       location, mesh_in);
  }

  void setGlobalFlags(int flags) {
    laplace_solver->setGlobalFlags(flags);
  }

  void setInnerBoundaryFlags(int flags) {
    laplace_solver->setInnerBoundaryFlags(flags);
  }

  void setOuterBoundaryFlags(int flags) {
    laplace_solver->setOuterBoundaryFlags(flags);
  }

  const Field3D gyroPade0(const Field3D &f, BoutReal rho) {
    laplace_solver.setCoefA(1.);
    laplace_solver.setCoefD(-rho*rho);

    // Invert, leaving boundaries unchanged
    return laplace_solver->solve(f);
  }
  ...
private:
  Laplacian* laplace_solver;
}
```

bendudson:
I think there's enough here to be worth merging now. We can revisit the gyroPade operators in another PR.

PR #1459, opened 2018-12-16
Coordinate system labelling

johnomotani:
As a preliminary to the `ShiftToFieldAligned` implementation of `ParallelTransform`, this PR introduces labelling of the coordinate system used by `Field3D`s. This is now stored as an `enum class COORDINATE_SYSTEM` member of `Field3D`. The main aim is for `toFieldAligned()` to be able to skip transforming if a field is already field-aligned. I think it's also helpful for record-keeping, `Field3D`s as it allows the coordinate system of a `Field3D` to be kept with the field, both in the C++ code and in output files (where it is saved as an attribute).

~~In order to be able to get the `COORDINATE_SYSTEM` for new fields from the `ParallelTransform` while initializing the `ParallelTransform`, I've split the initilization of `ParallelTransform`s into a minimal constructor and an `initialize` method: in `initialize` the `Mesh` has a valid `ParallelTransform` so `ParallelTransform::getCoordinateSystem()` can be called.~~

Uses the changes in #1432, so using `region-arguments-tofromFieldAligned` as the base (until that PR is merged).

### DIFF
--- a/include/bout/index_derivs_interface.hxx
+++ b/include/bout/index_derivs_interface.hxx
@@ -96,6 +96,9 @@ T flowDerivative(const T& vel, const T& f, CELL_LOC outloc, const std::string& m
   T result(localmesh);
   result.allocate(); // Make sure data allocated
   result.setLocation(outloc);
+  if (std::is_base_of<Field3D, T>::value) {
+    result.setCoordinateSystem(f.getCoordinateSystem());
+  }
 
   // Apply method
   derivativeMethod(vel, f, result, region);
@@ -162,6 +165,9 @@ T standardDerivative(const T& f, CELL_LOC outloc, const std::string& method,
   T result(localmesh);
   result.allocate(); // Make sure data allocated
   result.setLocation(outloc);
+  if (std::is_base_of<Field3D, T>::value) {
+    result.setCoordinateSystem(f.getCoordinateSystem());
+  }
 
   // Apply method
   derivativeMethod(f, result, region);
--- a/include/bout/paralleltransform.hxx
+++ b/include/bout/paralleltransform.hxx
@@ -48,7 +48,7 @@ public:
   /*!
    * Return the default coordinate system for Field3Ds when using this ParallelTransform
    */
-  virtual COORDINATE_SYSTEM getCoordinateSystem() const = 0;
+  virtual CoordinateSystem getCoordinateSystem() const = 0;
 
   virtual bool canToFromFieldAligned() = 0;
 
@@ -89,8 +89,8 @@ public:
     return f;
   }
 
-  COORDINATE_SYSTEM getCoordinateSystem() const override {
-    return COORDINATE_SYSTEM::FieldAligned;
+  CoordinateSystem getCoordinateSystem() const override {
+    return CoordinateSystem::FieldAligned;
   }
 
   bool canToFromFieldAligned() override{
@@ -134,8 +134,8 @@ public:
    */
   const Field3D fromFieldAligned(const Field3D &f, const REGION region=RGN_NOX) override;
 
-  COORDINATE_SYSTEM getCoordinateSystem() const override {
-    return COORDINATE_SYSTEM::Orthogonal;
+  CoordinateSystem getCoordinateSystem() const override {
+    return CoordinateSystem::Orthogonal;
   }
 
   bool canToFromFieldAligned() override{
--- a/include/bout/paralleltransform.hxx
+++ b/include/bout/paralleltransform.hxx
@@ -24,6 +24,8 @@ class Mesh;
  */
 class ParallelTransform {
 public:
+  /// Constructor
+  ParallelTransform(Mesh& mesh_in = *mesh) : thismesh(mesh_in) {}
   virtual ~ParallelTransform() {}
 
   /// Given a 3D field, calculate and set the Y up down fields
@@ -49,6 +51,9 @@ public:
   virtual COORDINATE_SYSTEM getCoordinateSystem() const = 0;
 
   virtual bool canToFromFieldAligned() = 0;
+
+protected:
+  Mesh& thismesh; ///< Mesh that this ParallelTransform belongs to
 };
 
 
@@ -59,6 +64,9 @@ public:
  */
 class ParallelTransformIdentity : public ParallelTransform {
 public:
+  /// Constructor
+  ParallelTransformIdentity(Mesh& mesh_in = *mesh)
+    : ParallelTransform(mesh_in) {}
   /*!
    * Merges the yup and ydown() fields of f, so that
    * f.yup() = f.ydown() = f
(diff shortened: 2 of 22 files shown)


include/bout/index_derivs_interface.hxx
```diff
@@ -96,6 +96,9 @@ T flowDerivative(const T& vel, const T& f, CELL_LOC outloc, const std::string& m
   T result(localmesh);
   result.allocate(); // Make sure data allocated
   result.setLocation(outloc);
+  if (std::is_base_of<Field3D, T>::value) {
```
d7919: It would be good to avoid special handling for particular field types so would it be possible to ensure all field types can provide this information? (also in the long run we might need to be able to identify if a Field2D is compatible with a Field3D etc.)

include/bout/paralleltransform.hxx
```diff
@@ -93,7 +102,13 @@ public:
...
 public:
   ShiftedMetric() = delete;
-  ShiftedMetric(Mesh &mesh);
+  ShiftedMetric(Mesh &m) : mesh(m) {}
+
+  /// initialize after constructor is finished, so that
+  /// localmesh->getCoordinateSystem() can call
+  /// ParallelTransform::getCoordinateSystem() for fields belonging to the
+  /// ParallelTransform
+  void initialize();
```
d7919: This sort of two stage setup has advantages for a number of things, but it is an important design decision so I think there would need to be quite a bit of input on this from @bendudson, @ZedThree  and others.

include/field2d.hxx
```diff
@@ -111,6 +111,24 @@ class Field2D : public Field, public FieldData {
    */
   int getNz() const override {return 1;};
 
+  /*!
+   * Return the coordinate system of this field
+   */
+  COORDINATE_SYSTEM getCoordinateSystem() const {
+#if CHECK > 0
```
d7919: I think this should always throw as it's a sign of an invalid executable and there's no performance penalty (i.e. it's not a check that typically passes in a valid program).

include/bout/paralleltransform.hxx
johnomotani: @d7919 agreed. I did consider calling `setCoordinateSystem` explicitly for all the fields in each `ParallelTransform`, but it got quite messy. It could make more sense to put off this kind of redesign, e.g. if we will make `ParallelTransform`s belong to `Coordinates`, it would make sense to design the initialization of both together.

For what I need at the moment, I think could get away with reducing the amount of checking: I only really want the checks in `to/fromFieldAligned`, and it could be OK in the short term to allow some fields to float around with `coordinate_system = COORDINATE_SYSTEM::None`, especially since they would only be in the `ParallelTransforms` and are probably only accessed by index rather than with `Field3D` operations. I'll have a think if there's a sensible, simpler initial version to start with, so we can give longer consideration to something like the implementation here, which is a significant re-design.

include/bout/index_derivs_interface.hxx
johnomotani: I'm not sure what the right design would be. Things to consider:
* `Field2D` and `Field3D` probably need different labelling - `Field2D` is the same for `Orthogonal` and `FieldAligned` coordinate systems
  * could have different `COORDINATE_SYSTEM` enum classes for `Field3D` and `Field2D` plus some methods/functions to compare them
  * or extend `COORDINATE_SYSTEM` to something like `enum class COORDINATE_SYSTEM = {None, FieldAligned, Orthogonal, FieldAligned2D, FCI}` where `FieldAligned2D` would be used for `Field2D` and compatible with both `FieldAligned` and `Orthogonal` (suggestions for better naming welcome if we want to go with a scheme like this)
* In the final implementation, `Field2D` members of `Coordinates` need `coordinate_system` setting. Best thing to do might be split initialization with 'uninitialized' `ParallelTransform` objects constructed early on (so we can `getCoordinateSystem()` from them), before most of the initialization of `Coordinates`.

include/bout/paralleltransform.hxx
johnomotani: I've moved the split initialization into #1463. This PR still seems to work (at least passes the tests) without it. The split initialization is, I think, necessary for *all* `Field3D`s to have their `coordinate_system` set correctly, but I think the only ones that don't get one set are in `ShiftedMetric` or `FCI` and only get accessed via indices, where no check is (or can really be) performed.

I think something like it was going to be more necessary when I was trying to get `Field2D` to also have a `coordinate_system`: then all the `Field2D`s in `Coordinates` need to get set properly (which #1463 actually doesn't help with in the end...) because they multiply `Field3D`s in the differential operators, etc., so if they had `coordinate_system = COORDINATE_SYSTEM::None` then checks would fail, or some horrible hacks would be necessary, like preventing copying the `coordinate_system` from a `Field*D` which has `coordinate_system = COORDINATE_SYSTEM::None`.

ZedThree: I would much prefer to avoid two-phase initialisation if possible, as it increases the surface area for potential bugs.

Would moving the parallel transform from `Mesh` into `Coordinates` help? I think that's probably where it really belongs anyway. 
Another thing that might help would be to use another type instead of `Field*`s in `Coordinates`/`ParallelTransform` that aren't "on" coordinates (though they'd still need to be "on" meshes).

include/bout/index_derivs_interface.hxx
ZedThree: `Field2D` could have `COORDINATE_SYSTEM::AxisSymmetric`?

Implementing `operator==(COORDINATE_SYSTEM, COORDINATE_SYSTEM)` would give us control over comparisons.

include/field.hxx
```diff
@@ -46,6 +46,28 @@ extern Mesh * mesh; ///< Global mesh
...
+/*!
+ * Coordinate systems that Field3Ds can be stored in
+ *
+ * None - special value to be used for Field3Ds initialized before the Mesh and
+ * ParallelTransform have finished being initialized, because then we cannot
+ * get the COORDINATE_SYSTEM to use from the ParallelTransform, but Field3Ds
+ * created then must not need to know their COORDINATE_SYSTEM, or must set it
+ * explicitly later.
+ */
+enum class COORDINATE_SYSTEM { None, FieldAligned, Orthogonal, FCI };
```
ZedThree: http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#enum5-dont-use-all_caps-for-enumerators

We're a bit inconsistent here, but I think this should be `CoordinateSystem`

include/bout/paralleltransform.hxx
johnomotani: I can't actually remember now exactly what made me think two-phase initialization was necessary. I've added a few commits trying to sort this a different way, adding constructors which take an explicit CoordinateSystem for the ParallelTransform constructors to use. Also attempting to add coordinate_system to Field2D. Still needs testing, probably tweaking a bit too, but I've got to go now. I'll look over it again later, hopefully this week.

johnomotani:
Closing this proposal: we're working on a different approach to tagging fields with coordinate types.

PR #1460, opened 2018-12-16
New shifted metric 2

johnomotani:
Implementation of `ParallelTransform` for shifted-metric method, transforming to globally field-aligned coordinates to take y-derivatives.

Includes a function to correct twist-shift boundary guard cells of `zShift` by adding/subtracting `ShiftAngle`, so twist-shift does not have to be applied during communications. This could be copied to `ShiftedMetric` too. Implementing this function required adding the `hasBranchCutUp/Down` methods to Mesh.

Currently allows parallel derivatives to be taken by going through the branch that transforms to/from field-aligned coordinates. Next step is to merge #1459 so the user can save a field-aligned field and pass that into parallel derivatives. Finally (possibly) would be to auto-magically cache the field-aligned version of the field somewhere so the user does not have to add extra `Field3D`s to their code.

Replaces #1177.

### DIFF
--- a/include/bout/mesh.hxx
+++ b/include/bout/mesh.hxx
@@ -322,7 +322,15 @@ class Mesh {
   /// \param[in] jx   The local (on this processor) index in X
   /// \param[out] ts  The Twist-Shift angle if periodic
   virtual bool periodicY(int jx, BoutReal &ts) const = 0;
+
+  /// Is there a branch cut at the lower y edge of this processor's grid points
+  /// at x-index jx?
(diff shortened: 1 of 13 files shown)


d7919:
Should we close #1177 if this replaces it?

johnomotani:
This new `ParallelTransform` will not be needed. Functionality for shifting between orthogonal and field-aligned coordinates with staggered coordinates will be provided by `ShiftedMetric` when `ParallelTransform` objects are moved to be members of `Coordinates` and have a location.
See https://github.com/boutproject/BOUT-dev/projects/5

PR #1462, opened 2018-12-17
Use function local storage in shiftZ rather than member storage

d7919:
Those should help improve OpenMP safety.


d7919:
There's also a commit that replaces the use of nested vectors with `Tensor`

d7919:
If `Tensor` accepted `Ind3D` and `Ind2D` indexing we could remove all calls to `i.x()` and `i.y()` (which can be expensive).

PR #1463, opened 2018-12-17
Separate initialization of ParallelTransform objects from constructors

johnomotani:
Allows `localmesh->getCoordinateSystem()` to call `ParallelTransform::getCoordinateSystem()` during initialization of the `ParallelTransform`.

Something like this would be nice to have to ensure that all `Field3D`s get the right `coordinate_system` set, instead of ever having `COORDINATE_SYSTEM::None`, but probably makes more sense to include something like it as part of refactoring `ParallelTransform` (e.g. making `ParallelTransform` a member of `Coordinates` rather than `Mesh`).

See discussion in #1459.

### DIFF
--- a/include/bout/paralleltransform.hxx
+++ b/include/bout/paralleltransform.hxx
@@ -101,7 +101,13 @@ public:
 class ShiftedMetric : public ParallelTransform {
 public:
   ShiftedMetric() = delete;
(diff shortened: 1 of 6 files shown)


johnomotani:
Updates to #1459 removed the need for this change.

Issue #1134, opened 2018-06-22
Use more forward declarations


dschwoerer:
Currently many includes could be replaced by forward declarations, e.g. in mesh.hxx:
```
#include "field2d.hxx"
#include "field3d.hxx"
#include "datafile.hxx"
```
could be replaced by
```
class Field2D;
class Field3D;
class Datafile;
```

That should speed up the compilation of bout, as only the things that are needed get included.

At the same time, we should provide a file that includes ether bout_all - so that projects that use bout can simply include that and get everything they need, rather than including a few files, which then includes everything that gets needed ...

This is going to be a breaking change - so I am not sure this is something we want do ...

Another reason, besides compile performance, is that the current way can be quite horrible to write templates in BOUT++.
I need to forward declare interp_to in newfield, to have it available in the template.
This breaks for the aiolos mesh - as interp_to becomes an inline function - which I cannot copy paste, as I cannot include bout/mesh.hxx - as that results in circular dependencies.
I gave up resolving that, as it seems everything is including everything ... and as mentioned changing that would break a lot of code ...

ZedThree:
I rather suspect 5.0 is going have some radical changes to a lot of things, and I'm very happy to break things in major version changes.

I've played with [include-what-you-use](https://include-what-you-use.org/) a few times, which can work out when we only need to forward declare things. Unfortunately, we actually don't include as many headers as we "should" in most places, so accepting everything it says might be a net negative.

A single header to include in physics models would be great, and I think we could do that at the same time as moving all the headers under a proper namespace (as well as maybe using C++ namespace as well).

After 4.2, which is coming I promise, I think we can start breaking things again.

PR #1464, opened 2018-12-17
Start to use more forward declarations

d7919:
Work contributes towards resolving #1134 

Alongside potential compilation speed improvements (that I've not observed yet), these changes (particularly once fully rolled out) should hopefully help make things a bit more flexible/easy to extend.

### DIFF
--- a/include/bout/mesh.hxx
+++ b/include/bout/mesh.hxx
@@ -45,36 +45,27 @@ class Mesh;
 
 #include "mpi.h"
 
-#include <bout/deprecated.hxx>
-#include <bout/deriv_store.hxx>
-#include <bout/index_derivs_interface.hxx>
-
-#include "field_data.hxx"
 #include "bout_types.hxx"
-#include "field2d.hxx"
-#include "field3d.hxx"
-#include "datafile.hxx"
-#include "options.hxx"
-
 #include "fieldgroup.hxx"
-
-#include "boundary_region.hxx"
-#include "parallel_boundary_region.hxx"
-
-#include "sys/range.hxx" // RangeIterator
-
-#include <bout/griddata.hxx>
-
-#include "coordinates.hxx"    // Coordinates class
-
-#include "paralleltransform.hxx" // ParallelTransform class
-
 #include "unused.hxx"
+#include <bout/deprecated.hxx>
+#include <bout/index_derivs_interface.hxx>
+#include <bout/paralleltransform.hxx>
 
-#include <bout/region.hxx>
+class Field3D;
+class Field2D;
+class RangeIterator;
+class Coordinates;
+class GridDataSource;
+class FieldData;
+class Datafile;
+class Options;
+class BoundaryRegion;
+class BoundaryRegionPar;
+template <typename T>
+class Region;
 
-#include <list>
-#include <memory>
+//#include <memory>
 #include <map>
 
 /// Type used to return pointers to handles
--- a/include/boundary_op.hxx
+++ b/include/boundary_op.hxx
@@ -5,12 +5,14 @@ class BoundaryModifier;
 #ifndef __BNDRY_OP__
 #define __BNDRY_OP__
 
-#include "boundary_region.hxx"
-#include "field2d.hxx"
-#include "field3d.hxx"
+#include "bout_types.hxx"
+#include "unused.hxx"
 #include "vector2d.hxx"
 #include "vector3d.hxx"
-#include "unused.hxx"
+
+class BoundaryRegion;
+class Field2D;
+class Field3D;
 
 #include <cmath>
 #include <string>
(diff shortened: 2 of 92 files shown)


d7919:
In terms of optimising the compilation I've done a bit of timing which shows that including `bout_types.hxx` has a cost of about 0.8s, which compares to a time to compile a single source file of somewhere between 2 and 6 seconds (for the majority of source files). As bout_types is included (directly or indirectly) in a large fraction of our source files we might be able to get a decent saving by splitting this header into separate parts. About 3/4 of the time comes from defining the maps used to convert enums to strings and the associated utility routines (with the other quarter coming from the includes of language standard libraries).

Just declaring these maps without initialising them in the header cuts the time from 0.8s to about 0.45s. It doesn't sound much but I think it could be a reasonable fraction saving across a full build (note configure+compile on travis takes about 10 minutes, with the tests then taking 2-7 minutes).

include/bout/coordinates.hxx
```diff
@@ -35,10 +35,13 @@ class Coordinates;
 #ifndef __COORDINATES_H__
 #define __COORDINATES_H__
 
-#include "mesh.hxx"
-#include "datafile.hxx"
-#include "utils.hxx"
 #include <bout_types.hxx>
+#include <field2d.hxx>
+#include <field3d.hxx>
+#include <fieldperp.hxx>
```
bendudson: I'm confused about this... should we use `<>` for system headers and `""` for BOUT++ headers?

d7919: I tend to view `<>` as being something provided by a library (including `bout`) and `""` for local files. Or in other words `<>` for things in an include path and `""` for things that aren't. This is just my interpretation though. I think `""` _can_ search more places, but if this is an advantage or a disadvantage may be a matter of opinion.

d7919:
@johnomotani thanks, you're right -- I've now added the new files, thanks!

ZedThree:
I think it's now just missing `#include "dcomplex.hxx"` in `slepc.cxx`

include/bout/mesh.hxx
```diff
@@ -451,57 +442,15 @@ class Mesh {
...
-    case (DIRECTION::YOrthogonal):
-    case (DIRECTION::YAligned):
-      return CELL_YLOW;
-    case (DIRECTION::Z):
-      return CELL_ZLOW;
-    default:
-      throw BoutException("Unhandled direction encountered in getAllowedStaggerLoc");
-    }
-  };
+  CELL_LOC getAllowedStaggerLoc(DIRECTION direction) const;
```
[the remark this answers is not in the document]
d7919: Yes, this is one of the reasons I think this PR might be more for discussion/experimentation to decide on a policy. I'd prefer runtime performance over compilation performance but there may be places we can get both.

ZedThree:
While we're tidying up, we should probably rename all the include guards as macros beginning with `__` or `_X` are technically reserved by the preprocessor.

include/bout/solver.hxx
```diff
@@ -35,17 +35,29 @@
 
 class Solver;
```
ZedThree: We should be able to delete this

include/bout/coordinates.hxx
ZedThree: One reason to prefer `""` for all BOUT++ headers: clang-format and include-what-you-use can group headers together, and it might be nice to separate the system headers from ours.

ZedThree:
I used `include-what-you-use` plus some manual fiddling to make sure everything in headers is inside include guards: https://travis-ci.org/boutproject/BOUT-dev/builds/475492401

Also made all BOUT++ headers included with `""` and used clang-format to group headers together (main header (i.e. `field3d.hxx` for `field3d.cxx`-> system headers -> `bout/` headers -> everything else)

Any gain over this PR is in the noise, but it would at least be a consistent treatment of headers. At the cost of 2k+ lines changed!

I'm in favour of merging this PR at least, it's about a 20% speedup of compile times on Travis

d7919:
> I used `include-what-you-use` plus some manual fiddling to make sure everything in headers is inside include guards: https://travis-ci.org/boutproject/BOUT-dev/builds/475492401
> 
> Also made all BOUT++ headers included with `""` and used clang-format to group headers together (main header (i.e. `field3d.hxx` for `field3d.cxx`-> system headers -> `bout/` headers -> everything else)
> 
> Any gain over this PR is in the noise, but it would at least be a consistent treatment of headers. At the cost of 2k+ lines changed!
> 
> I'm in favour of merging this PR at least, it's about a 20% speedup of compile times on Travis

For reference I've made a branch that just includes the changes to `bout_types` and the travis build is at https://travis-ci.org/boutproject/BOUT-dev/builds/475809020?utm_source=github_status&utm_medium=notification -- this hasn't finished yet but I suspect this should get much of the compilation gain from this PR.

bendudson:
Is this one for tidying and merging, or closing?

ZedThree:
The tidying is likely going to be a nightmare so probably closing and rebasing. I did have a go last week, I'll check it still goes in cleanish.

bendudson:
Closing this because it's been so long and has too many conflicts. Sorry @d7919 !

PR #1466, opened 2018-12-18
Ensure the location is initialised

d7919:
The location passed into the LaplaceXZ constructor was ignored leading to unitialised locations being used.

### DIFF
--- a/include/bout/invert/laplacexz.hxx
+++ b/include/bout/invert/laplacexz.hxx
@@ -38,7 +38,8 @@
 
 class LaplaceXZ {
 public:
-  LaplaceXZ(Mesh *UNUSED(m), Options *UNUSED(options), const CELL_LOC UNUSED(loc)) {}
+  LaplaceXZ(Mesh* UNUSED(m), Options* UNUSED(options), const CELL_LOC loc = CELL_CENTRE)
+      : location(loc) {}
   virtual ~LaplaceXZ() {}
 
   virtual void setCoefs(const Field2D &A, const Field2D &B) = 0;


include/bout/invert/laplacexz.hxx
```diff
@@ -38,7 +38,8 @@
 
 class LaplaceXZ {
 public:
-  LaplaceXZ(Mesh *UNUSED(m), Options *UNUSED(options), const CELL_LOC UNUSED(loc)) {}
+  LaplaceXZ(Mesh* UNUSED(m), Options* UNUSED(options), const CELL_LOC loc = CELL_CENTRE)
```
johnomotani: Don't think we need the default argument here, because the constructor(s) should only ever be called through `LaplaceXZ::create`, which already has the default argument set.

d7919: We'd like users to use `create` I guess, but there's nothing currently stopping a user from using the exposed constructor.

d7919: I suspect this code may need a refactor in the long run to look a bit more like our usual factory methods.

d7919:
This was noticed by @bshanahan when trying to run the `test-laplacexz` case -- the call to get `Coordinates` passes the location and this was then triggering an exception as we ended up trying to interpolate as the passed location was not equal to CELL_CENTRE (indeed it was full of junk!).

include/bout/invert/laplacexz.hxx
johnomotani: OK, that's a good point, and having default arguments in the constructors is also what we do for Laplacian and it's implementations. We need default arguments for the implementations' constructors then, I'll add those while I'm at it.

Agree that a refactor to be like the usual factory methods would be good :+1:

include/bout/invert/laplacexy.hxx
```diff
@@ -49,7 +49,7 @@
  */
 class LaplaceXY {
  public:
-  LaplaceXY(Mesh *m, Options *opt = nullptr, const CELL_LOC = CELL_CENTRE) {
+  LaplaceXY(Mesh *m = mesh, Options *opt = nullptr, const CELL_LOC = CELL_CENTRE) {
```
d7919: I think this should default to `nullptr` and then the initialisation statement can use the usual pattern of checking if it's nullptr and if so setting to the global mesh. Two motivations being 1) I think this is closer to what we do in most of the rest of the code (probably not consistently though), 2) we can in the future change that logic to do something else without changes to the header (i.e. without changing the interface).

src/invert/laplacexy/laplacexy.cxx
```diff
@@ -514,10 +521,13 @@ LaplaceXY::~LaplaceXY() {
 const Field2D LaplaceXY::solve(const Field2D &rhs, const Field2D &x0) {
   Timer timer("invert");
   
+  ASSERT1(rhs.getLocation() == location);
+  ASSERT1(x0.getLocation() == location);
+
```
d7919: Meshes are compatible.

src/invert/laplacexz/impls/cyclic/laplacexz-cyclic.cxx
```diff
@@ -59,20 +63,23 @@ LaplaceXZcyclic::~LaplaceXZcyclic() {
 void LaplaceXZcyclic::setCoefs(const Field2D &A2D, const Field2D &B2D) {
   TRACE("LaplaceXZcyclic::setCoefs");
   Timer timer("invert");
+
```
d7919: Mesh compatible?

src/invert/laplacexz/impls/cyclic/laplacexz-cyclic.cxx
```diff
@@ -157,18 +164,21 @@ void LaplaceXZcyclic::setCoefs(const Field2D &A2D, const Field2D &B2D) {
 Field3D LaplaceXZcyclic::solve(const Field3D &rhs, const Field3D &x0) {
   Timer timer("invert");
 
-  Mesh *mesh = rhs.getMesh();
+  ASSERT1(rhs.getLocation() == location);
+  ASSERT1(x0.getLocation() == location);
+
+  Mesh *localmesh = rhs.getMesh();
```
d7919: Shouldn't this be using the class member localmesh rather than shadowing it?

src/invert/laplacexz/impls/cyclic/laplacexz-cyclic.hxx
```diff
@@ -15,7 +17,7 @@ public:
   using LaplaceXZ::solve;
   Field3D solve(const Field3D &b, const Field3D &x0) override;
 private:
-  Mesh *mesh;   ///< The mesh this operates on, provides metrics and communication
+  Mesh *localmesh;   ///< The mesh this operates on, provides metrics and communication
```
d7919: Is it better to put this into the base class?

include/bout/invert/laplacexz.hxx
```diff
@@ -38,15 +38,16 @@
...
+      : location(loc) {}
   virtual ~LaplaceXZ() {}
 
   virtual void setCoefs(const Field2D &A, const Field2D &B) = 0;
   virtual void setCoefs(const Field3D &A, const Field3D &B) { setCoefs(DC(A), DC(B)); }
 
   virtual Field3D solve(const Field3D &b, const Field3D &x0) = 0;
 
-  static LaplaceXZ *create(Mesh *m, Options *opt = nullptr, const CELL_LOC loc = CELL_CENTRE);
+  static LaplaceXZ *create(Mesh *m = mesh, Options *opt = nullptr, const CELL_LOC loc = CELL_CENTRE);
```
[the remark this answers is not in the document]
johnomotani: The mesh pointer was stored in both of the implementations, but seeing as it's passed to the base class constructor, it probably makes more sense to store it in the base class...

src/invert/laplacexz/impls/petsc/laplacexz-petsc.cxx
```diff
@@ -276,6 +276,9 @@ void LaplaceXZpetsc::setCoefs(const Field3D &Ain, const Field3D &Bin) {
 
   TRACE("LaplaceXZpetsc::setCoefs");
 
+  ASSERT1(Ain.getLocation() == location);
+  ASSERT1(Bin.getLocation() == location);
+
```
d7919: Check meshes are compatible?

src/invert/laplacexz/impls/petsc/laplacexz-petsc.cxx
```diff
@@ -578,6 +581,9 @@ Field3D LaplaceXZpetsc::solve(const Field3D &bin, const Field3D &x0in) {
 
   TRACE("LaplaceXZpetsc::solve");
 
+  ASSERT1(bin.getLocation() == location);
+  ASSERT1(x0in.getLocation() == location);
+
```
d7919: Mesh.

include/bout/invert/laplacexz.hxx
```diff
@@ -38,21 +38,29 @@
 
 class LaplaceXZ {
 public:
-  LaplaceXZ(Mesh *UNUSED(m), Options *UNUSED(options), const CELL_LOC UNUSED(loc)) {}
+  LaplaceXZ(Mesh* m = nullptr, Options* UNUSED(options) = nullptr,
+      const CELL_LOC loc = CELL_CENTRE)
+  : localmesh(m), location(loc) {
```
d7919: This could use `localmesh(m==nullptr ? mesh : m)` in the initialiser list to leave the function body as empty.

src/invert/laplacexz/impls/cyclic/laplacexz-cyclic.cxx
```diff
@@ -41,14 +41,18 @@ LaplaceXZcyclic::LaplaceXZcyclic(Mesh *m, Options *options, const CELL_LOC loc)
   k1d_2 = Array<dcomplex>((m->LocalNz) / 2 + 1);
 
   // Create a cyclic reduction object, operating on dcomplex values
-  cr = new CyclicReduce<dcomplex>(mesh->getXcomm(), nloc);
+  cr = new CyclicReduce<dcomplex>(localmesh->getXcomm(), nloc);
```
d7919: Should this use `make_unique` as is done in `LaplaceXY`?

include/bout/invert/laplacexz.hxx
johnomotani: Yes, definitely. Sorry, I was being dumb!

src/invert/laplacexz/impls/cyclic/laplacexz-cyclic.cxx
johnomotani: I think so. At the moment `cr` causes a memory leak, since it is not deleted in the destructor...

Issue #1468, opened 2018-12-18
Namespace for globals?


johnomotani:
Would it be a good idea to put the global variables defined in `globals.hxx` (`mesh` and `dump`) into a namespace like `bout::globals`?

I've come across bugs a few times caused by parts of the library using the global `mesh` when they should have been using a local one.

If we put `using namespace bout::globals` into `physicsmodel.hxx` then user code could use the global namespaces transparently and (once #1464 is merged) the `using` statement would not propagate into any library code except `physicsmodel.hxx/cxx`. This would help root out any remaining unintentional uses of the global variables in the library code, and places that we do want them (like default arguments) can use e.g. `bout::globals::mesh` to avoid any possibility of name conflicts.

d7919:
We've been starting to use namespaces a bit more in recent PRs (see #1415 and the recent derivatives changes, for example) and I think it would be great to expand this effort.

As you've noted, by itself this could break code but there are ways to avoid this. I can think of (at least) two ways we could manage this in the interim:

1. We namespace things in headers, try to compile the code (all tests and examples as well as library), find where it breaks, update to use the namespace, commit  and then finally add a `using .....` to the header. This would ensure user code would continue to work whilst allowing us to update the codebase to only use the namespaced routines. The downside of this is that it only catches non-namespaced use at one point in time -- we'd have to remain vigilant in PRs to ensure they don't sneak back in (harder than it sounds!) and there's also no obvious way to automatically alert users that something should be changed in their code to ensure it will continue to work in the future (when we remove the using statements).
2. As with one except we add the using statements to a new header like `bout_all.hxx` (probably should just be `bout.hxx`, but this already exists). This has a possible advantage that user code wishing to use the library just has to include a single header (see #1134 for some comments on this). The downside is that this single header might be quite slow to include (as it would have to include all the headers that define the public api*). It might also be difficult to inform users of the change, however this change would break users models with the simple fix being just replace your includes with this magic include line.

johnomotani:
Implemented by #1470.

Issue #897, opened 2018-02-22
Make `Field` more powerful


dschwoerer:
Currently it is quite hard to use Field for anything.
It cannot be used to checkdata, copy a field, or iterate over it.
Therefore the solver e.g. needs to maintain a separate list for field3d and field2d - fieldperp or other fields are not supported.
This breaks the usefulness of the fields being derived from field.
Therefore fixing #208 is rather useless - as it would require much work or be of limited use.

`checkData(Field)` could be implemented to use RTTI to do the job - but that is not a nice solution - and would be trivially avoided if checkData() would be a member function of `Field` - then the vtable would resolve to the appropriate implementation of checkData at runtime.
Field3D shouldn't be (much) slower, if it is marked as final ...

To summarize:
 * Do we want other Fields?
 * Should we use RTTI or make fiend functions member functions?
 * Is there a better way to make use use Field more often, rather then being limited to Field3D?

bendudson:
Field and FieldData should be merged, in my opinion. It's hard to make much use of Field in any performance-critical parts of the code, due to the vtable indirection needed. Generally I think we can use polymorphism at compile time: checkData overloaded for different field types; polymorphism less commonly needed at runtime. There are some places which are not performance critical where it would be nice though.

ZedThree:
I think this mostly done in #2313. as well as more use of templates.

There's still places where this could be improved (e.g. `Solver`'s use of separate vectors for each field type), but we're getting there

PR #1470, opened 2018-12-19
Use namespace for global variables 'mesh' and 'dump'

johnomotani:
* Create the namespace `bout::globals` containing `Mesh* mesh` and `Datafile dump`.
* In `bout.hxx` include statement `using namespace bout::globals` so no changes are needed to user code.
* Remove references to global `mesh` pointer in several places in the library code.
* The global pointer (renamed to `bout::globals::mesh`) is still used in many places as a default argument or if `nullptr` is passed as a `Mesh*`.
* The last places depending on `bout::globals::mesh` are: the solver code (see #1471); `FieldData` (which should be refactored, possibly merging `FieldData` and `Field`, see also #897).
* SlepcSolver uses `bout::globals::dump` directly because it needs to write output without calling BoutMonitor. Seems acceptable to leave this as a special case because SlepcSolver needs a different output scheme as an eigenvalue rather than initial-value solver.
* The only changes to the tests are that the unit tests need to include `using namespace bout::globals` because they use the global mesh.
* There is an unrelated change to `test-petsc_laplace`, just getting it to work and removing the binary grid file.

This implements the suggestion in #1468.


include/field.hxx
```diff
@@ -37,7 +37,11 @@ class Field;
 
 class Mesh;
 class Coordinates;
-extern Mesh * mesh; ///< Global mesh
+namespace bout {
+namespace globals {
+  extern Mesh* mesh; ///< Global mesh
```
d7919: Is there a reason not to include `globals.hxx` here?

include/field.hxx
```diff
@@ -83,10 +87,10 @@ class Field {
 #endif
 
   virtual Mesh * getMesh() const{
-    if (fieldmesh){
-      return fieldmesh;
+    if (fieldmesh == nullptr) {
```
d7919: I think we generally prefer the true branch to be the most common case, which I'd guess is that `fieldmesh` is not `nullptr`. Rewriting as `return fieldmesh != nullptr? fieldmesh : bout::globals::mesh` would be one way to go.

include/field_data.hxx
```diff
@@ -81,6 +82,7 @@ public:
   FieldGeneratorPtr getBndryGenerator(BndryLoc location);
 
 protected:
+  Mesh* fielddatamesh;
```
d7919: This seems to be a separate addition than the main purpose of this PR (related to #1319 which I think was being delayed).

src/invert/laplace/impls/cyclic/cyclic_laplace.hxx
```diff
@@ -44,7 +44,7 @@ class LaplaceCyclic;
  */
 class LaplaceCyclic : public Laplacian {
 public:
-  LaplaceCyclic(Options *opt = nullptr, const CELL_LOC loc = CELL_CENTRE, Mesh *mesh_in = mesh);
+  LaplaceCyclic(Options *opt = nullptr, const CELL_LOC loc = CELL_CENTRE, Mesh *mesh_in = bout::globals::mesh);
```
d7919: Should this default to `nullptr` for mesh and then get handled in the usual initialiser list way?

src/mesh/boundary_standard.cxx
```diff
@@ -546,6 +551,8 @@ void BoundaryDirichlet::apply_ddt(Field2D &f) {
 }
 
 void BoundaryDirichlet::apply_ddt(Field3D &f) {
+  Mesh* mesh = bndry->localmesh;
+  ASSERT1(mesh = f.getMesh());
```
d7919: If we require `bndry->localmesh == f.getMesh()` can we remove the `localmesh` member of `bndry` or do we need it for something else?

src/mesh/difops.cxx
```diff
@@ -452,6 +452,7 @@ const Field3D Div_par_LtoC(const Field3D &var) {
   Field3D result;
   result.allocate();
 
+  Mesh* mesh = var.getMesh();
```
d7919: Should we get this `mesh` out before we create `result` so that we can do `Field3D result(mesh);`?

src/physics/gyro_average.cxx
```diff
@@ -89,15 +89,15 @@ const Field2D gyroPade1(const Field2D &f, const Field2D &rho, int flags) {
...
+  result.getMesh()->communicate(result);
   result = 0.5*rho*rho*Delp2( result );
   result.applyBoundary("dirichlet");
   return result;
 }
 
 const Field3D gyroPade2(const Field3D &f, const Field2D &rho, int flags) {
   Field3D result = gyroPade1(gyroPade1(f, rho, flags), rho, flags);
-  mesh->communicate(result);
+  result.getMesh()->communicate(result);
```
d7919: Just a thought looking at this syntax I wonder if it would be worth introducing a free function `communicate(Field3D &var) { var.getMesh()->communicate(var); return;}` -- probably not in this PR though.

src/physics/sourcex.cxx
```diff
@@ -16,57 +16,63 @@ BoutReal TanH(BoutReal a) {
...
 
   return result;
 }
 
 // create radial buffer zones to set jpar zero near radial boundaries
-const Field3D sink_tanhx(const Field2D &UNUSED(f0), const Field3D &f, BoutReal swidth,
+const Field3D sink_tanhx(const Field2D &f0, const Field3D &f, BoutReal swidth,
                          BoutReal slength, bool UNUSED(BoutRealspace)) {
-  Field3D result;
+  Mesh* localmesh = f0.getMesh();
```
d7919: Could this be `f.getMesh()` to leave `f0` unused?

src/solver/impls/arkode/arkode.cxx
```diff
@@ -135,15 +135,25 @@ int ArkodeSolver::init(int nout, BoutReal tstep) {
...
-  int MXSUB = mesh->xend - mesh->xstart + 1;
+
+  // Compute band_width_default from actually added fields, to allow for multiple Mesh objects
+  //
+  // Previous implementation was equivalent to:
+  //   int MXSUB = mesh->xend - mesh->xstart + 1;
+  //   int band_width_default = n3Dvars()*(MXSUB+2);
+  int band_width_default = 0;
+  for (auto fvar : f3d) {
+    band_width_default += fvar.var->getNx();
```
d7919: Isn't `MXSUB+2 /= Field3D{}.getNx()` and if so is this change intentional (or rather why is it intended)?

include/field.hxx
johnomotani: I don't think so... At the moment including `globals.hxx` includes `bout/mesh.hxx`, `datafile.hxx` and `bout/macro_for_each.hxx`, but I guess we can just forward-declare `Mesh` and `Datafile`.

johnomotani: ah, don't think we can forward-declare `Datafile` because `dump` is a variable, not a pointer. So including `globals.hxx` will have to include `datafile.hxx` too, which might not be necessary...

include/field_data.hxx
johnomotani: That's right, but is related to this PR in the sense that it's necessary in order to get rid of uses of the global mesh pointer. #1319 has been closed; @ZedThree, @bendudson discussed at the workshop a better re-design of the boundary conditions following the design of the new derivatives. This change is tangential to the redesign of the boundary conditions too; actually I think it may belong more naturally here. The 'correct' final solution may be something like merging `FieldData` into `Field` so we don't have the multiple-inheritance, but as far as I know nobody has immediate plans for doing that...

include/field.hxx
d7919: Ah ok -- https://github.com/boutproject/BOUT-dev/pull/1464/commits/cd6929ce5f4e402bdfb90ce723bae48f28def63c provides another way to avoid this (as mentioned earlier). 

Similar(ish) changes to `bout_types.hxx`  in https://github.com/boutproject/BOUT-dev/pull/1464/commits/c5d37d4b6cc6c9f57fa9e2673f59fcb9a4946f48 (and bug fix in https://github.com/boutproject/BOUT-dev/pull/1464/commits/52e72e2405cf1163bb1a2542ffa650daaeab6e59) seem to have given a decent speedup in travis builds (maybe 1-2 minutes) which is motivating me to look into shifting more stuff to source rather than header where possible.

I guess this might be something for another time/PR

src/invert/laplace/impls/cyclic/cyclic_laplace.hxx
johnomotani: :+1: I'd left them 'as-is' on the first pass... will update in a new commit.

include/field_data.hxx
d7919: Yes it's coming back to me now -- although the primary goal of this PR isn't to remove uses of the global mesh, just to move the global mesh/dump into a namespace?

johnomotani: In my mind the aim was to do both, because in making the namespace all the places where the global mesh was used get flagged up, so I thought it made sense to 'fix' them as I was going through, at least for 'small' fixes.

Adding a `Mesh*` member and a new method to `FieldData` isn't really a 'small' change. Would be happy to back out of it, leave the global mesh pointer for now and fix another time (possibly by merging `Field`/`FieldData`). That would make sense, we'd have two 'design' issues left after the PR, this and the `Solver`, to finally get rid of global mesh pointer from the library. What would people prefer: 1 keep this change as a stop-gap; 2 revert the changes to `FieldData` and fix properly in a redesign (addressing #897); 3 something else?

d7919: My preference would be to back out on this for now (we don't support no global mesh until everywhere supports it anyway, so I guess this isn't really blocking that) and address it in a redesign alongside the updates to boundary (as this may impact on how we want to do this anyway).

src/mesh/boundary_standard.cxx
johnomotani: `BoundaryRegion::localmesh` is used in several places in the initialization of `BoundaryRegion`s. At that point they do not have a `Field`, so I think do need their own `localmesh`. It may well be that we can get rid of the requirement as part of the boundary re-design, e.g. by using `Region`s.

src/physics/gyro_average.cxx
johnomotani: I like the idea. Could tidy up user code, reduce uses of global mesh pointer, and allow handling of fields with different meshes in the same call (i.e. if `f` and `g` have different `fieldmesh`es, `communicate(f, g)` could handle calling `communicate` from each mesh).

src/solver/impls/arkode/arkode.cxx
johnomotani: I'm a bit confused about what this setting is doing, but you're right, I should leave the current behaviour.

include/field_data.hxx
johnomotani: I've taken out the changes to FieldData. Now it just uses `bout::globals::mesh` to use the global mesh pointer.

bendudson:
Sorry missed that this is going into `forward_declares` rather than `next`. Can it go into `next` then merge `next` into `forward_declares`?

d7919:
#1464 was kind of an experiment to see if there were any immediate gains -- I think there are some (both in terms of compilation time as well as flexibility) but the commit history might be a bit noisier than required as I approached this in a roundabout way. @ZedThree has looked into tools that might be able to automate some of these include changes (although the compilation time gains would not have been achieved with this -- it required some insight to see how to make a gain).

johnomotani:
Righto, in that case it does make sense to rebase this on next. Will do now...

johnomotani:
I've rebased this PR on `next` instead of `forward_declares` now, and tidied up the commit history a little bit. I think the PR should be ready to merge now, unless anyone has more comments?

include/bout/solver.hxx
```diff
@@ -71,7 +71,9 @@ typedef int (*TimestepMonitorFunc)(Solver *solver, BoutReal simtime, BoutReal la
 #include "vector2d.hxx"
 #include "vector3d.hxx"
 
+#define NO_GLOBAL_NAMESPACE
```
ZedThree: #1464 will forward declare `PhysicsModel` and remove the include, so presumably this won't be needed after that goes in?

include/bout.hxx
```diff
@@ -65,6 +66,13 @@
 
 const BoutReal BOUT_VERSION = BOUT_VERSION_DOUBLE;  ///< Version number
 
+#ifndef NO_GLOBAL_NAMESPACE
```
ZedThree: `NO_GLOBAL_NAMESPACE` -> `BOUT_NO_GLOBAL_NAMESPACE` :)

include/bout/solver.hxx
```diff
@@ -290,7 +292,7 @@ class Solver {
   ///
   /// @param[inout] outputfile   The file to add variable to
   /// @param[in] save_repeat    If true, add variables with time dimension
-  void outputVars(Datafile &outputfile, bool save_repeat=true);
+  virtual void outputVars(Datafile &outputfile, bool save_repeat=true);
```
ZedThree: Is this a separate bugfix?

src/physics/physicsmodel.cxx
```diff
@@ -28,7 +28,11 @@
  *
  **************************************************************************/
 
+#define NO_GLOBAL_NAMESPACE
```
ZedThree: Is this needed here? Later we do `bout::globals::mesh` anyway

include/invert_parderiv.hxx
```diff
@@ -63,16 +63,16 @@ public:
    * with pure virtual members, so can't be created directly.
    * To create an InvertPar object call the create() static function.
    */ 
-  InvertPar(Options *UNUSED(opt), Mesh *mesh_in = mesh)
-    : localmesh(mesh_in) {}
+  InvertPar(Options *UNUSED(opt), Mesh *mesh_in = nullptr)
+    : localmesh(mesh_in==nullptr ? bout::globals::mesh : localmesh) {}
```
ZedThree: `localmesh` -> `mesh_in`

src/bout++.cxx
```diff
@@ -83,6 +83,9 @@ void bout_signal_handler(int sig);  // Handles signals
 
 #include <output.hxx>
 
+// We initialize the global variables here, so allow their use
+using namespace bout::globals;
```
ZedThree: I'm not sure I understand the comment? We're using the variables a lot here, so the `using namespace` is for convenience?

src/solver/impls/petsc/petsc.cxx
```diff
@@ -376,14 +385,15 @@ int PetscSolver::init(int NOUT, BoutReal TIMESTEP) {
     PetscInt dof = n3Dvars();
 
     // Maximum allowable size of stencil in x is the number of guard cells
-    PetscInt stencil_width = mesh->xstart;
+    PetscInt stencil_width_estimate;
+    OPTION(options, stencil_width_estimate, bout::globals::mesh->xstart);
```
ZedThree: I know the rest of the file uses the old style, but could this be

```
PetscInt stencil_width_estimate = options["stencil_width_estimate"].withDefault(bout::globals::mesh->xstart);
```

It's nice to immediately initialise variables wherever possible

src/solver/impls/power/power.hxx
```diff
@@ -48,6 +48,14 @@ class PowerSolver : public Solver {
...
   
   int run() override;
+
+  void outputVars(Datafile &outputfile, bool save_repeat=true) override {
+    // Include base class functionality
+    this->Solver::outputVars(outputfile, save_repeat);
+
+    // Save the eigenvalue to the output
+    outputfile.add(eigenvalue, "eigenvalue", true);
+  }
```
ZedThree: Is this orthogonal to this PR?

include/bout/solver.hxx
johnomotani: With the forward declared `PhysicsModel` the include will move to `solver.cxx` and the `#define`/`#undef` wrapper should go with it.

My intention is to (transparently) let user code keep using `mesh` and `dump` as global names. So:
* put a `using namespace bout::globals` statement in `bout.hxx`
    * `bout.hxx` is included in `physicsmodel.hxx` and `boutmain.hxx` and now nowhere else (`bout.hxx` was also in `difops.cxx` before this PR)
    * User code must include at least one of `bout.hxx`, `physicsmodel.hxx` or `boutmain.hxx`, so get the `using` statement included.
* don't want the library code to have the `using namespace bout::globals` statement so that we ensure that the global mesh pointer is only ever `bout::globals::mesh`
    * this is what the `NO_GLOBAL_NAMESPACE` macro is for - every include of `physicsmodel.hxx` or `bout.hxx` in library code should be wrapped in `#define NO_GLOBAL_NAMESPACE`/`#undef NO_GLOBAL_NAMESPACE` to cut out the `using` statement

It's not great to need the macro, but there are only a couple of places where `physicsmodel.hxx` is included. Still maybe there is a neater way to handle this?

ZedThree:
Thanks John, this is great and definitely the direction we should be going in! Some minor nit-picking above :)

src/physics/physicsmodel.cxx
johnomotani: Sorry, badly named macro. Defining the macro means we have to write out `bout::globals::mesh` in full, see comment above about `using namespace bout::globals`. I'll rename it, thinking of changing to `BOUT_NO_USING_NAMESPACE_BOUTGLOBALS`.

johnomotani:
> Is it worth picking one and being consistent (answer might be no!)? I only mention this as we're touching all these places

:+1: for being consistent. I favour using `Mesh* localmesh` for the local pointers to be double clear they aren't the global mesh. I used `Mesh* mesh` sometimes when it would reduce the number of lines I had to touch here.

I'm happy to go through and tidy all the `Mesh*`s up. I propose the naming scheme:
* `localmesh` when local to the body of a function/method
* `thismesh` when a member of a class (but would something class-specific be better, e.g. `fieldmesh` in `Field`, `coordinatesmesh` in `Coordinates`?)
* `mesh_in` as a constructor argument

What does everyone think? Should we use underscores `local_mesh` or camel-case `localMesh`? I think I'll wait for this to merge and do the tidy-up in a new PR though...

include/bout/solver.hxx
johnomotani: This change and the changes in `src/solver/impls/power/power.hxx,power.cxx` are introduced in 
6c70779. The purpose is to remove a use of the global `dump` in `PowerSolver` by overriding `Solver::outputVars` instead of using `dump.add(eigenvalue, "eigenvalue", true)`.

src/bout++.cxx
johnomotani: It was for convenience, but on second thoughts I don't like the `using namespace` here. Will get rid of it.

johnomotani:
@ZedThree I think I've addressed all your comments except the one about `Mesh*` naming consistency.

BTW I made the changes as a rebase because there are already rather a lot of commits in this PR. Do we have a style preference for (a) 'fixing' existing commits or (b) adding new commits with small changes/fixes to a PR?

ZedThree:
Thanks! I'll look through them now.

> BTW I made the changes as a rebase because there are already rather a lot of commits in this PR. Do we have a style preference for (a) 'fixing' existing commits or (b) adding new commits with small changes/fixes to a PR?

I prefer (b) as it's then possible to review just those commits, although if it's something like a typo in the last commit and I catch it early, I sometimes amend and force-push.

src/solver/impls/slepc/slepc.cxx
```diff
@@ -606,10 +606,10 @@ void SlepcSolver::monitor(PetscInt its, PetscInt nconv, PetscScalar eigr[],
       output<<formatEig(reEigBout,imEigBout)<<endl;
       if(eigenValOnly){
         simtime=reEigBout;
-        dump.write();
+        bout::globals::dump.write();
```
ZedThree: Not for this PR, but it might be nice for `SlepcSolver` to have it's own output file (which is `bout::globals::dump` by default)?

ZedThree:
Looks good to me, but @d7919 had a bunch of comments too -- are you happy those are resolved?

include/invert_laplace.hxx
```diff
@@ -192,7 +192,7 @@ public:
    * 
    * @param[in] opt  The options section to use. By default "laplace" will be used
    */
-  static Laplacian *create(Options *opt = nullptr, const CELL_LOC loc = CELL_CENTRE, Mesh *mesh_in = mesh);
+  static Laplacian *create(Options *opt = nullptr, const CELL_LOC loc = CELL_CENTRE, Mesh *mesh_in = bout::globals::mesh);
```
d7919: Should `mesh_in` really be `nullptr` here? Presumably this is passed to the constructor noted above (or equivalent) which can then choose what to do if it receives `nullptr`?

src/mesh/parallel/fci.cxx
```diff
@@ -53,8 +53,9 @@ inline BoutReal sgn(BoutReal val) { return (BoutReal(0) < val) - (val < BoutReal
 
 // Calculate all the coefficients needed for the spline interpolation
 // dir MUST be either +1 or -1
-FCIMap::FCIMap(Mesh &mesh, int dir, bool zperiodic)
-  : dir(dir), boundary_mask(mesh), corner_boundary_mask(mesh), y_prime(&mesh) {
+FCIMap::FCIMap(Mesh &mesh_in, int dir, bool zperiodic)
+  : mesh(mesh_in), dir(dir), boundary_mask(mesh), corner_boundary_mask(mesh),
```
d7919: We use `mesh_in` to set `mesh` and then use `mesh` to set things like the `boundary_mask`. Whilst this should work ok it's sensitive to order -- would it be safer to use `mesh_in` everywhere here? @ZedThree may have some comments one way or the other.

d7919:
Have you checked that we can compile and run an appropriately updated example without using the `using` statements?

src/mesh/parallel/fci.cxx
ZedThree: The variables are always initialised in the order they are declared in the class, so there is already an order dependency there. There should be a warning generated if this is wrong -- if there's no warning, it should be safe!

Note that this will conflict with #1345, so might be best to wait for that to go in then revisit this.

d7919: In that case I think I'd prefer the use of `mesh_in` everywhere here -- it's easy for someone to change the order of variable declarations in the separate header without realising they're being used like this.

include/invert_laplace.hxx
ZedThree: Agreed -- `nullptr` is a better default, it's what we do in other bits of the code, as well as giving us some space to potential remove the global mesh entirely.

ZedThree:
Might be wise to merge in `next` now 4.2.1 is in and verify everything still works

include/invert_laplace.hxx
johnomotani: We can't just pass `nullptr` here because `LaplaceFactory` uses `mesh_in` to check whether we can use a serial solver
https://github.com/boutproject/BOUT-dev/blob/08cc9f225a207c96bd54aacaa6355ac7f7b4d98e/src/invert/laplace/laplacefactory.cxx#L46-L47
I think this was originally to allow the default Laplace solver to switch between `serial_tri` and `spt` depending on whether a serial or parallel solver was needed, but now the default is `cyclic` in either case. I think it would be better now to allow any Laplace solver to be created, and move the check into the constructor of the serial solvers. Does that sound reasonable, and should I do it here or create a separate PR?

johnomotani:
> Have you checked that we can compile and run an appropriately updated example without using the `using` statements?

@d7919 The integrated tests run without needing any update. Examples and user code should not need updating either. I've just checked `examples/blob2d` to be sure.

include/invert_laplace.hxx
ZedThree: If `LaplaceFactory` gets a `nullptr`, it could just fetch the global mesh though?

`LaplaceFactory` should get rewritten anyway (e.g. using `GenericFactory`), so I would leave that to another PR

PR #1479, opened 2018-12-26
Fix bugs in boututils.calculus.deriv2D

bendudson:
It may be best to merge this into `master`, then `master` into `next`, rather than merging #1478 and then getting conflicts when merging `next` into `master` later.

Using index without casting to int after divide, and calling len() on a scalar value.

Thanks to @rheinonen for finding


ZedThree:
Agreed this -> master, then master -> next. I think we're pretty much ready for a bugfix release.

Is it worth investing some time in writing unit tests for `boututils` now? Or should we just do it as we move functionality into the new xarray project?

bendudson:
I vote for writing tests as functionality is moved to xarray. Better to write tests first based on what we want the functionality to be, rather than try to write tests for the functionality which the code currently has, which may or may not be what we want it to be.

PR #1482, opened 2019-01-03
Documentation for making releases

ZedThree:
Adds a [checklist](https://github.com/boutproject/BOUT-dev/blob/release-docs/manual/RELEASE_HOWTO.md) to help the release process.

Also includes some helpful utilities with `make` targets:

- Update the changelog, requires [github_changelog_generator][gcg] and access token to be setup:

      make changelog TOKEN=<token> LAST_VERSION=v4.2.0 RELEASE_BRANCH=master

- Make a tarball, including submodules, for the current version (can be changed by setting `BOUT_VERSION` on the command line)

      make dist

This is straight into master to help with the release of v4.2.1

[gcg]: https://github.com/github-changelog-generator/github-changelog-generator

### DIFF
--- a/externalpackages/bin/git-archive-all.sh
+++ b/externalpackages/bin/git-archive-all.sh
@@ -1,0 +1,315 @@
+#!/bin/bash -
+#
+# File:        git-archive-all.sh
+#
+# Description: A utility script that builds an archive file(s) of all
+#              git repositories and submodules in the current path.
+#              Useful for creating a single tarfile of a git super-
+#              project that contains other submodules.
+#
+# Examples:    Use git-archive-all.sh to create archive distributions
+#              from git repositories. To use, simply do:
+#
+#                  cd $GIT_DIR; git-archive-all.sh
+#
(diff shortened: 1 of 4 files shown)


externalpackages/bin/git-archive-all.sh
```diff
@@ -0,0 +1,315 @@
+#!/bin/bash -
```
bendudson: One thing: This file doesn't say who it's by. Probably should credit https://github.com/meitar (interesting character btw https://maymay.net)

ZedThree: Ah maybe I should just add it as a submodule?

bendudson: Is it a hard dependency? If not then perhaps a submodule might be better (?) No strong feelings about it. Otherwise maybe just adding a line to the file with a link to the original repository.

ZedThree:
The PR test is pending but Travis says it's completed

PR #1484, opened 2019-01-07
Include what we use (header tidy)

ZedThree:
Includes #1464. Done automatically using [include-what-you-use](https://github.com/include-what-you-use/) plus some manual tinkering. I can add the mapping file I used, but it has hardcoded absolute paths in it.

- Include all needed headers for each file
- Use forward declarations wherever possible
- Consistent use of `""` when including our headers
- Consistent position of include guards on outside of files
- Use `clang-format` to sort headers

This is an enormous change, but it should be entirely backwards compatible -- `bout/physicsmodel.hxx` includes more headers than it actually needs in order to maintain this.
The compilation speedup is about 20%, but it does also make all our headers consistent (albeit using a sledgehammer!)

I think we should do this at some point, along with moving all the headers inside `include/bout/`, but we could put both off till 5.0. They can (mostly) be done automatically now too, so it's not a great effort to re-do at a later point.


ZedThree:
As expected, this has a million conflicts. Given my changes were either from a tool or basically just mechanical, I'm closing this and will revisit later.

PR #1485, opened 2019-01-09
Silence remaining warnings on master

ZedThree:
This should silence (almost) all the remaining warnings at all CHECK
levels, including from the tests (but not all the examples). There are
some deprecation warnings from a couple of integrated tests, but these
are fixed in next.

This should at least silence all the warnings for the Travis builds.

About half of this PR already made it into next -- I'd like it into
master though to make that as quiet as possible.

### DIFF
--- a/src/solver/impls/petsc/petsc.cxx
+++ b/src/solver/impls/petsc/petsc.cxx
@@ -669,15 +669,22 @@ PetscErrorCode solver_rhsjacobian(TS UNUSED(ts), BoutReal UNUSED(t), Vec UNUSED(
   PetscFunctionReturn(0);
 }
 #else
-PetscErrorCode solver_rhsjacobian(TS ts,BoutReal t,Vec globalin,Mat *J,Mat *Jpre,MatStructure *str,void *f_data) {
+PetscErrorCode solver_rhsjacobian(MAYBE_UNUSED(TS ts), MAYBE_UNUSED(BoutReal t),
+                                  MAYBE_UNUSED(Vec globalin), Mat *J, Mat *Jpre,
+                                  MAYBE_UNUSED(MatStructure *str),
+                                  MAYBE_UNUSED(void *f_data)) {
   PetscErrorCode ierr;
 
   PetscFunctionBegin;
-  ierr = MatAssemblyBegin(*Jpre, MAT_FINAL_ASSEMBLY);CHKERRQ(ierr);
-  ierr = MatAssemblyEnd(*Jpre, MAT_FINAL_ASSEMBLY);CHKERRQ(ierr);
-  if (*J != *Jpre){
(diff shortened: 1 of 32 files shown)


src/solver/impls/petsc/petsc.cxx
```diff
@@ -669,15 +669,22 @@ PetscErrorCode solver_rhsjacobian(TS UNUSED(ts), BoutReal UNUSED(t), Vec UNUSED(
...
   PetscErrorCode ierr;
 
   PetscFunctionBegin;
-  ierr = MatAssemblyBegin(*Jpre, MAT_FINAL_ASSEMBLY);CHKERRQ(ierr);
-  ierr = MatAssemblyEnd(*Jpre, MAT_FINAL_ASSEMBLY);CHKERRQ(ierr);
-  if (*J != *Jpre){
-    ierr = MatAssemblyBegin(*J, MAT_FINAL_ASSEMBLY);CHKERRQ(ierr);
-    ierr = MatAssemblyEnd(*J, MAT_FINAL_ASSEMBLY);CHKERRQ(ierr);
+  ierr = MatAssemblyBegin(*Jpre, MAT_FINAL_ASSEMBLY);
+  CHKERRQ(ierr);
```
bendudson: This is probably automatic formatting, but having `CHKERRQ` on the same line is the pattern usually used in PETSc code.

src/sys/derivs.cxx
```diff
@@ -327,7 +328,8 @@ const Field2D D2DYDZ(const Field2D &f, CELL_LOC UNUSED(outloc),
   return tmp;
 }
 
-const Field3D D2DYDZ(const Field3D &f, CELL_LOC outloc, DIFF_METHOD method, REGION UNUSED(region)) {
+const Field3D D2DYDZ(const Field3D &f, CELL_LOC outloc, MAYBE_UNUSED(DIFF_METHOD method),
+                     REGION UNUSED(region)) {
```
bendudson: Is the fact that `region` is not used a bug? Seems like this might be surprising.

ZedThree: There is a commented out section that begins `// TODO: use region aware implementation`. I'm not sure why it's not used -- @d7919 any ideas?

src/solver/impls/petsc/petsc.cxx
ZedThree: Ah, good point. Might need to find a way to tell clang-format to leave PETSc stuff alone

src/sys/derivs.cxx
d7919: I think this may have just been waiting for the `SpecificInd` offsetting.

ZedThree: Ok, we can fix that in a separate PR

PR #1487, opened 2019-01-10
Hypnotoad fixes

bendudson:
Some important fixes for Hypnotoad, mainly found by @HarukiST .  Particularly important for people using proportional density and temperature profiles, since the old code gave the wrong pressure profile.

Emphasises the need for:
1. More (some) testing of input grid generation code (definitely)
2. Splitting hypnotoad out into a separate repository (probably)


bendudson:
After `primary_opt = ind` is set on line 269 (https://github.com/boutproject/BOUT-dev/blob/master/tools/tokamak_grids/gridgen/analyse_equil.pro#L269), `ind` could be modified by line 324 (https://github.com/boutproject/BOUT-dev/blob/master/tools/tokamak_grids/gridgen/analyse_equil.pro#L324). It's not obvious because `MIN` modifies its second argument, but I'm surprised it didn't cause more problems.

PR #1491, opened 2019-01-11
Add Matrix/Tensor::resize; check Arrays are of positive size

ZedThree:
- Add `Matrix`/`Tensor::resize()`
- Use `Array`/`Matrix`/`Tensor::resize()` instead of the copy
  assignment operator everywhere
- Add `Array::size_type = int`
- Add check that `Array` is of strictly positive size

I've put the positivity check for `Array` in `Array::get`: this so
everything above it gets the check without repeating the check in lots
of places (either code duplication or literally checking twice). This
won't catch e.g. `Matrix(-1, -1)`, but I don't believe this is a big
source of bugs -- willing to be convinced the check should be moved up
into the call stack though.

### DIFF
--- a/include/bout/array.hxx
+++ b/include/bout/array.hxx
@@ -69,6 +69,7 @@ template<typename T>
 class Array {
 public:
   typedef T data_type;
+  using size_type = int;
     
   /*!
    * Create an empty array
@@ -82,7 +83,7 @@ public:
   /*!
    * Create an array of given length
    */
-  Array(int len) {
+  Array(size_type len) {
     ptr = get(len);
   }
   
@@ -121,7 +122,7 @@ public:
   /*!
    * Resize the array to \p new_size
    */
-  void resize(int new_size) {
+  void resize(size_type new_size) {
     release(ptr);
     ptr = get(new_size);
   }
@@ -176,7 +177,7 @@ public:
   /*!
    * Return size of the array. Zero if the array is empty.
    */
-  int size() const noexcept {
+  size_type size() const noexcept {
     if(!ptr)
       return 0;
 #ifdef BOUT_ARRAY_WITH_VALARRAY
@@ -269,7 +270,7 @@ public:
    * or if ind is out of bounds. For efficiency no checking is performed,
    * so the user should perform checks.
    */
-  T& operator[](int ind) {
+  T& operator[](size_type ind) {
     ASSERT3(0 <= ind && ind < size());
 #ifdef BOUT_ARRAY_WITH_VALARRAY
     return ptr->operator[](ind);
@@ -277,7 +278,7 @@ public:
     return ptr->data[ind];
 #endif    
   }
-  const T& operator[](int ind) const {
+  const T& operator[](size_type ind) const {
     ASSERT3(0 <= ind && ind < size());
 #ifdef BOUT_ARRAY_WITH_VALARRAY
     return ptr->operator[](ind);
@@ -303,10 +304,10 @@ private:
    * Handles the allocation and deletion of data
    */
   struct ArrayData {
-    int len;    ///< Size of the array
+    size_type len;    ///< Size of the array
     T *data;    ///< Array of data
     
-    ArrayData(int size) : len(size) {
+    ArrayData(size_type size) : len(size) {
       data = new T[len];
     }
     ~ArrayData() {
@@ -336,11 +337,11 @@ private:
    */
   dataPtrType ptr;
 
-  typedef std::map< int, std::vector<dataPtrType> > storeType;
-  typedef std::vector< storeType > arenaType;
+  using storeType = std::map<size_type, std::vector<dataPtrType>>;
+  using arenaType = std::vector<storeType>;
 
   /*!
-   * This maps from array size (int) to vectors of pointers to ArrayData objects
+   * This maps from array size (size_type) to vectors of pointers to ArrayData objects
    *
    * By putting the static store inside a function it is initialised on first use,
    * and doesn't need to be separately declared for each type T
@@ -393,7 +394,7 @@ private:
    * Returns a pointer to an ArrayData object with no
    * references. This is either from the store, or newly allocated
    */
-  dataPtrType get(int len) {
+  dataPtrType get(size_type len) {
     dataPtrType p;
 
     auto& st = store()[len];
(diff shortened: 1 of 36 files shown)


d7919:
I think it's right to have the check in `Array::get` but I think we should really put checks into `Matrix` and `Tensor` as well for safety (e.g. in the case that you mentioned).

include/bout/array.hxx
```diff
@@ -390,10 +391,14 @@ private:
   }
   
   /*!
-   * Returns a pointer to an ArrayData object with no
+   * Returns a pointer to an ArrayData object of size \p len with no
    * references. This is either from the store, or newly allocated
+   *
+   * Expects \p len >= 0
```
d7919: Is it worth adding a comment about what `len=0` means?

d7919:
This is probably a comment better suited to the earlier PR but I guess `resize` here (on `Array` really) acts a bit different from `std::vector` resize in that it invalidates all the data so you can't use it to naturally grow storage without doing some extra work copying data around. I guess that's fine for our typical use case but probably worth documenting this clearly in manual/comments and through tests?

d7919:
> I think it's right to have the check in `Array::get` but I think we should really put checks into `Matrix` and `Tensor` as well for safety (e.g. in the case that you mentioned).

This could be avoided if we actually made `Matrix` etc. store things using a nested structure. For example I wonder if it's worth introducing `Vector` (possibly not because of confusion with with `Vector?D`, maybe `Array1D` instead) as a 1D version of `Matrix` so `Matrix`  storage is `Array<Vector<T>>` and `Tensor` is `Array<Matrix<T>>`. `Tensor.resize(a,b,c)` would then consume `a` and call `Matrix.resize(b,c)` etc. This makes it easier to extract a subset of the data (but only a particular subset). A common use case is to want something like `Tensor[a,b,:]` e.g. for the FFT routines in shiftz. I'm not sure if this design is performant though.

ZedThree:
Good point, that would be a nasty gotcha. Best to document all of them

include/utils.hxx
```diff
@@ -166,13 +181,30 @@ public:
...
   Tensor(const Tensor &other) : n1(other.n1), n2(other.n2), n3(other.n3), data(other.data) {
     // Prevent copy on write for Tensor
     data.ensureUnique();
   }
 
+  /// Resize the Tensor to \p new_size_1 by \p new_size_2 by \p new_size_3
+  ///
+  /// Note that this invalidates the existing data!
+  void resize(size_type new_size_1, size_type new_size_2, size_type new_size_3) {
+    ASSERT2(new_size_1 >= 0);
```
d7919: I'm not sure what it means for one of `n1`, `n2` and `n3` to be zero here -- clearly the effect is the same as for `Array` and we do explicitly initialise `n*` to 0 in the empty constructor so I guess it just means empty. Should we add a comment/note to explain this.

include/bout/array.hxx
```diff
@@ -120,8 +121,10 @@ public:
 
   /*!
    * Resize the array to \p new_size
+   *
+   * Note that this invalidates the existing data!
    */
-  void resize(int new_size) {
+  void resize(size_type new_size) {
     release(ptr);
     ptr = get(new_size);
```
d7919: Note if `size()==new_size` then I guess there's a good chance (but no promise - or maybe we can promise this due to only one thread having access to this store at one time) that we actually end up getting the same block back off the store that we just released back to the store. Might be worth making a note of this somewhere -- I could see this leading to confusion/hiding problems potentially. 

Is it worth having some sort of `resizeInPlace` routine that promises to keep the first `min(size(), new_size)` elements of data that would behave a bit more like `resize` on `std::vector`? If we kept track of how big we thought the size of the array was we could make shrinking resizes not potentially do an allocation (we use our idea of size for bounds checking and responding to external size queries -- a shrinking resize just changes this without releasing/getting, a growing resize checks if the new size is bigger than the actual size, if it isn't we can just change our tracked upper bound).

ZedThree: For the first point, I'm not sure I see the problem. Do we just need to be more explicit in the docstring that you shouldn't use the values the `Array` immediately after a call to `resize`?

For the second point, I can see the appeal but:
1. this would require a fair amount of changes to the `Array` internals.
2. the major use-case of `Array::resize` is to set the size of a default-initialised member variable. `Array` is not quite a general-purpose container like `std::vector` and I'm not sure we want it to be (similarly for `Matrix/Tensor`).

include/utils.hxx
ZedThree: I'll make a note of this in the `Matrix/Tensor` docstring

include/bout/array.hxx
ZedThree: Maybe `resize` is the wrong name, and this should really be `reallocate`?

bendudson: Yes, this makes sense: `resize` does imply that the original contents will be kept.

ZedThree: Renamed `resize` -> `reallocate`

src/invert/laplace/impls/mumps/mumps_laplace.cxx
```diff
@@ -211,21 +211,21 @@ LaplaceMumps::LaplaceMumps(Options *opt, const CELL_LOC loc, Mesh *mesh_in = mes
     localrhssize += localmesh->xstart*(localmesh->LocalNz);
     
     int nxpe = localmesh->NXPE;
-    localrhs_size_array = Array<int>(nxpe);
+    localrhs_size_array.resize(nxpe);
```
bendudson: reallocate?

ZedThree: Oops, thanks! Due to me relying on the compiler and not actually compiling everything. Luckily we have code review! :smile:

bendudson:
Also lucky we don't (as far as I know) use mumps laplace solver... Thanks @ZedThree !

ZedThree:
One of the CyclicReduce unit tests failed as the error was 1.1e-15 when the tolerance is 1.0e-15. Might need to loosen the tolerance a little bit (although worrying if this test is non-deterministic when serial?)

ZedThree:
Test is still failing after rerunning. I'll loosen the tolerance in a separate PR then merge this after that goes in

PR #1493, opened 2019-01-15
specify python package versions for Travis

d7919:
Avoids surprises -- numpy 1.16 doesn't provide a whl for python 3.4 (not supported) so we were spending >2 minutes for each build compiling numpy.


bendudson:
Agree this is currently wasting time on Travis. The only concern I have is now remembering to update this occasionally. It's also nice (I think) if we try to support the latest version available. It would be good to support multiple versions of course.

PR #1494, opened 2019-01-15
FieldFactory: Separate parsing, looping stages in create{2,3}D

bendudson:
To create a field in the RHS function, for example sources as a function of time, currently we have to do something like:

```
FieldFactory::get()->create3D( Options::root()["vorticity"]["source"],
                               &Options::root()["vorticity"],
                               mesh,
                               CELL_CENTRE,
                               time );
```

which involves looking up an options string, parsing it and then evaluating the resulting generator over the mesh. This isn't as bad as it looks, because the string is hashed in the parser so is in a lookup table.

To clean this up a little, this splits `create3D` into the parse step, and the part which calls the generator. This adds additional functions to the API which take a `FieldGeneratorPtr` rather than a `std::string`.

The above example becomes:

```
FieldGeneratorPtr generator =
       FieldFactory::get()->parse(Options::root()["vorticity"]["source"],
                                  &Options::root()["vorticity"]);

...

FieldFactory::get()->create3D( generator,
                               mesh,
                               CELL_CENTRE,
                               time );
```

Adds some documentation but no tests yet.


d7919:
Is this going to be a breaking change or should this all be internal only?

src/field/field_factory.cxx
```diff
@@ -99,30 +99,31 @@ FieldFactory::~FieldFactory() {
...
+  
+  Field2D result(0., localmesh);
 
-  FieldGeneratorPtr gen = parse(value, opt);
-  if(!gen) {
-    output << "FieldFactory error: Couldn't create 2D field from '"
-           << value
-           << "'" << endl;
-    return result;
+  if (localmesh->StaggerGrids == false) {
```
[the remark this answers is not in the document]
ZedThree: I've removed this, and also replaced the initialisation of `result` with just `allocate`

ZedThree:
I've added unit tests that cover the majority of the `FieldFactory` and `FieldGenerators`, also the following changes:

- Pass `location` straight through to `setLocation` in `create2D`
- Allocate result instead of initialising it in `create2D`
- Changed the type of the exception thrown in `FieldRound::clone` to match the other `FieldGenerators`

include/field_factory.hxx
```diff
@@ -57,30 +57,49 @@ public:
...
-  // These functions called by the parser
+  /// These functions called by the parser to resolve unknown symbols.
+  /// This is used to enable options to be referred to in expressions.
   FieldGeneratorPtr resolve(std::string &name) override;
 
 private:
-  Mesh *fieldmesh;  
-  const Options *options;
+  Mesh *fieldmesh;        ///< The default mesh for create functions.
+  const Options *options; ///< Set in parse() and used in resolve()
```
d7919: Is this fully accurate (i.e. does `create2D`/`3D` also set this due to the call to `parse` internal to those routines)?

ZedThree: Yes, if you call the `create2D/3D` overload that takes an `Options`, that instance is used to parse the input string. It's then restored to the previous instance. Any expressions that are used get cached, so it doesn't need to hang on it afterwards.

Maybe "Can be overridden by argument to `parse`"?

I think ideally, this would be something like `vector<shared_ptr<Options>>`, but we'd have to be passing about `shared_ptr<Options>` in more places to do this correctly.

src/field/field_factory.cxx
```diff
@@ -99,30 +99,29 @@ FieldFactory::~FieldFactory() {
...
 
-  if(localmesh == nullptr)
+const Field2D FieldFactory::create2D(FieldGeneratorPtr gen, Mesh* localmesh, CELL_LOC loc,
+                                     BoutReal t) {
+  AUTO_TRACE();
+  
+  if (localmesh == nullptr)
     localmesh = fieldmesh;
-  if(localmesh == nullptr)
+  if (localmesh == nullptr)
```
d7919: This double `localmesh` check looks a bit odd -- I can see what it is doing (second check checks `fieldmesh/=nullptr` essentially) but this is a bit unusual in terms of what we do elsewhere so it might be good to add a comment. Also the exception message could perhaps be a little more descriptive (to note what we've tried, say). Should we finally try to fall back to global mesh before deciding to throw?

tests/unit/test_extras.hxx
```diff
@@ -225,10 +225,10 @@ public:
   void addBoundary(BoundaryRegion* region) {boundaries.push_back(region);}
   std::vector<BoundaryRegion *> getBoundaries() { return boundaries; }
   std::vector<BoundaryRegionPar *> getBoundariesPar() { return std::vector<BoundaryRegionPar *>(); }
-  BoutReal GlobalX(int UNUSED(jx)) const { return 0; }
-  BoutReal GlobalY(int UNUSED(jy)) const { return 0; }
-  BoutReal GlobalX(BoutReal UNUSED(jx)) const { return 0; }
-  BoutReal GlobalY(BoutReal UNUSED(jy)) const { return 0; }
+  BoutReal GlobalX(int jx) const { return jx; }
```
d7919: Are these sensible values or could it cause problems if they are used for anything?

ZedThree: Any value should be sensible. This is just how does index-x on this processor map onto some global value of bout-space-x.

src/field/field_factory.cxx
ZedThree: Yes, it's a bit of an odd pattern, but I think we do occasionally use this elsewhere (not that that's an argument in favour of it!).

We normally fall back to the global mesh in the ctor, which is probably the best place to do so.

Does anyone have an opinion on which of these is more readable:

```cpp
if ((localmesh == nullptr) and (fieldmesh == nullptr)) {
  throw BoutException("No valid mesh found: FieldFactory not created with mesh and no mesh passed in");
}

if (localmesh == nullptr) {
  localmesh = fieldmesh;
}
```
or:
```cpp
if (localmesh == nullptr) {
  if (fieldmesh == nullptr) {
    throw BoutException("No valid mesh found: FieldFactory not created with mesh and no mesh passed in");
  } else {
    localmesh = fieldmesh;
  }
}
```

The first makes the precondition more explicit, while the second avoids the repeated check.

johnomotani: I find the second one easier to read. I think that's because when I look at the first one it makes me think about the values of `localmesh` and `fieldmesh` at the same time, whereas in the second I think about the state of `localmesh` first, and then about `fieldmesh`.

ZedThree: I'll make these changes in separate PR, as they're orthogonal to this one

PR #1495, opened 2019-01-15
Bug fix: communications in Coordinates ctor may cause hang

### LINKED ISSUE #1369: getCoordinates() can hang

ZedThree:
Fixes #1369

`geometry()` is no longer called from the `Coordinates`
constructor. This means that the Christoffel symbols and non-uniform
corrections will not be available until either `geometry()` or the new
`calcChristoffelSymbols()`/`calcNonUniformCorrections()` are called.

These functions are called on demand in the library functions where
the relevant variables are needed.

Note that this is technically a breaking change, but user code is
unlikely to be affected. If it is, the fix is a one liner:

```cpp
mesh->getCoordinates()->geometry();
```

### DIFF
--- a/include/bout/coordinates.hxx
+++ b/include/bout/coordinates.hxx
@@ -68,7 +68,7 @@ public:
   BoutReal zlength() const { return dz * nz; } ///< Length of the Z domain. Used for FFTs
 
   /// True if corrections for non-uniform mesh spacing should be included in operators
-  bool non_uniform;
+  bool non_uniform{true};
   Field2D d1_dx, d1_dy;  ///< 2nd-order correction for non-uniform meshes d/di(1/dx) and d/di(1/dy)
   
   Field2D J; ///< Coordinate system Jacobian, so volume of cell is J*dx*dy*dz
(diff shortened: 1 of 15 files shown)


johnomotani:
There is still communication in the `Coordinates` constructors, so #1369 is not totally fixed yet:
* `Mesh::get(Field2D,std::string,BoutReal)` contains a `communicate()` call
* `dx` is communicated if `periodicX` (note that communication in `Mesh::get()` means that this is redundant) https://github.com/boutproject/BOUT-dev/blob/9a242eb3b3a6ab27ce5b55604d34cdc31b2c5f39/src/mesh/coordinates.cxx#L36-L39
* `interpolateAndNeumann(Field2D,CELL_LOC)` contains a `communicate()` call

ZedThree:
Ah, I thought the issue was just due to needing the `Coordinates` at staggered locations so the non-staggered constructor was fine to communicate. I also hadn't appreciated that `interpolateAndNeumann` communicates.

Do you have a test that demonstrates the problem? It would be good to add it to the test suite in any case.

ZedThree:
So this is not a useful fix for the original bug, but could still be useful (in `next`) as it allows us to get away with not calculating the Christoffel symbols until needed. There are a lot of them (21) and they need derivatives and communication, and once we move to 3D metrics will have a significant memory footprint.

PR #1497, opened 2019-01-15
Show available options in Factory

dschwoerer:


include/bout/generic_factory.hxx
```diff
@@ -59,7 +59,14 @@ public:
     if (index != std::end(type_map)) {
       return index->second(std::forward<Args>(args) ...);
     }
-    throw BoutException("Could not find '%s'", name.c_str());
+    // List available options in error
+    const char* const delim = "\n";
```
ZedThree: Unused variable

dschwoerer: Thanks, removed that

include/bout/generic_factory.hxx
```diff
@@ -59,7 +59,13 @@ public:
     if (index != std::end(type_map)) {
       return index->second(std::forward<Args>(args) ...);
     }
-    throw BoutException("Could not find '%s'", name.c_str());
+    // List available options in error
+    std::string available;
```
d7919: Do we need to / should we initialise this to an empty string for the unusual case where the factory has no available methods? This would avoid us passing an uninitialised string to the `BoutException` later although there might still be a bit of confusion from the user in parsing the resulting message. I guess we could check in the available methods list is empty and throw with a specific message in that case -- it's possible we already check this elsewhere.

d7919: One option would be to initialise it to "\n" and remove it from the fixed part of the exception message, i.e. `"Available:\n%s"` --> `""Available:%s"`

ZedThree: Just as a note, `std::string` has a default constructor which creates the empty string, so this is ok. Although an explicit initialisation might be clearer

d7919: ok, if all is fine I'll mark as resolved.

PR #1506, opened 2019-01-16
Two stage add coordinates

d7919:
Proposed/potential fix for #1369. Tested on cases provided by @johnomotani but wider testing would be helpful.


d7919:
I realised I branched off next for this -- should I form another PR for master?

d7919:
This fails the unit tests on travis but passes them locally, which is a bit confusing.

src/field/field2d.cxx
```diff
@@ -130,9 +130,8 @@ void Field2D::setLocation(CELL_LOC new_location) {
       new_location = CELL_CENTRE;
     }
 
-    fieldCoordinates = getCoordinates();
-
     location = new_location;
+    fieldCoordinates = getCoordinates();
```
johnomotani: These calls could be just `getCoordinates()`. `fieldCoordinates` is already set inside `getCoordinates()`.

d7919: Good point!

tests/unit/test_extras.hxx
```diff
@@ -195,6 +204,18 @@ public:
...
     mesh->createDefaultRegions();
     output_info.enable();
+
+    // Delete any existing mesh_staggered
+    if (mesh_staggered != nullptr) {
+      delete mesh_staggered;
+      mesh_staggered = nullptr;
+    }
+    mesh_staggered = new FakeMesh(nx, ny, nz);
+    mesh_staggered->StaggerGrids = true;
```
d7919: Is this line the only difference between `mesh` and `mesh_staggered`?

d7919:
If we need four procs for the hang tests then I think we might need to force one openmp thread for the test on travis as we'd be massively oversubscribing which causes things to crawl to a halt and is likely why https://travis-ci.org/boutproject/BOUT-dev/jobs/480756720#L1735 is happening (although it's hard to tell if this is just slow and has timed out or if it is hanging).

tests/integrated/test-coordinates-initialization/runtest
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+set -e
+
+make
+# Kill the test after 3 seconds, if it hasn't already exited
+mpirun -n 4 timeout 3s ./test-coordinates-initialization
```
d7919: Should we `mpirun` the `timeout` or `timeout` the `mpirun` (is there any difference, e.g. in the error message)?

As a sidenote openmpi provides `--timeout` option to mpirun and `mpich` provides `MPIEXEC_TIMEOUT` environment variable, unfortunately the standard doesn't mandate anything similar.

tests/integrated/test-coordinates-initialization2/test-coordinates-initialization.cxx
```diff
@@ -0,0 +1,31 @@
...
+ * Test of initialization of Coordinates objects in Mesh::coords_map
+ */
+
+#include "bout.hxx"
+#include "optionsreader.hxx"
+
+int main(int argc, char** argv) {
+
+  // Initialize BOUT++
+  BoutInitialise(argc,argv);
```
d7919: It's not clear what extra this test brings over the other -- presumably if the other doesn't fail this one shouldn't (beyond being slower)?

tests/unit/field/test_vector3d.cxx
```diff
@@ -163,8 +166,12 @@ TEST_F(Vector3DTest, SetLocationZLOW) {
...
-  Vector3D vector;
-  vector.x.getMesh()->StaggerGrids = true;
+  FakeMesh local_mesh{Vector3DTest::nx,Vector3DTest::ny,Vector3DTest::nz};
+  local_mesh.StaggerGrids = true;
+  local_mesh.setCoordinates(nullptr, CELL_XLOW);
+  local_mesh.setCoordinates(nullptr, CELL_YLOW);
+  local_mesh.setCoordinates(nullptr, CELL_ZLOW);
+  Vector3D vector(&local_mesh);
   EXPECT_EQ(vector.getLocation(), CELL_CENTRE);
   EXPECT_NO_THROW(vector.setLocation(CELL_VSHIFT));
```
d7919: Note `FakeMesh` has `xstart=1` so if we tried to let the actual coordinates be created here we'd find an assert fails. So whilst this (setting nullptr coordinates) does allow us to proceed with the test in some sense it's hiding a potential issue -- for the unit tests here it's probably OK but is something we might need to address for other interpolation tests.

tests/unit/test_extras.hxx
johnomotani: The other (more important?) difference is the `setCoordinates` call just below.

tests/integrated/test-coordinates-initialization/runtest
ZedThree: I would suggest `timeout 3s mpirun ...`

tests/integrated/test-coordinates-initialization2/test-coordinates-initialization.cxx
johnomotani: > if the other doesn't fail this one shouldn't

That's true. Sorry, I probably shouldn't have put both tests in.
The reason for having this test is to show that if we require `BoutInitialise()` to be called, then 
the `getCoordinates()` call in `BoutMesh::load()` (added in ed81860) is not 'necessary' (see discussion on #1369). If we're happy with ed81860, then `test-coordinates-initialization2` is indeed redundant.

tests/unit/test_extras.hxx
```diff
@@ -14,6 +14,12 @@
 
 const BoutReal BoutRealTolerance = 1e-15;
 
+#ifndef GLOBALORIGIN
+extern Mesh* mesh_staggered;
+#else
+Mesh* mesh_staggered = nullptr;
+#endif
```
ZedThree: Can't `mesh_staggered` just be a member of `FakeMeshFixture`?

tests/integrated/test-coordinates-initialization/runtest
johnomotani: I found that `timeout 3s mpirun...` left processes running when timeout tried to stop the run, whereas `mpirun timeout...` finished cleanly but with non-zero exit code.

ZedThree:
> I realised I branched off next for this -- should I form another PR for master?

I would just put this straight into `master`. After this bugfix, v4.2.1 should be good to go, so will get merged into `next` pretty soon

> Ah it seems like there might be some relation to the test order, which suggests it's possibly something to do with preserved state between tests -- any ideas @ZedThree ?

The tests should be completely independent. If you run `./serial_tests --gtest_shuffle --gtest_repeat=1000` you should be able to see if the order matters

tests/unit/mesh/test_interpolation.cxx
```diff
@@ -54,6 +54,9 @@ class Field3DInterpToTest : public ::testing::Test {
     mesh->ystart = 2;
     mesh->xend = nx - 3;
     mesh->yend = ny - 3;
+    dynamic_cast<FakeMesh*>(mesh)->setCoordinates(nullptr, CELL_XLOW);
+    dynamic_cast<FakeMesh*>(mesh)->setCoordinates(nullptr, CELL_YLOW);
+    dynamic_cast<FakeMesh*>(mesh)->setCoordinates(nullptr, CELL_ZLOW);
```
ZedThree: This is a bit surprising! We should really have `setCoordinates` in the base `Mesh` class

johnomotani:
> we might need to force one openmp thread for the test on travis

Is the best way to do this just to put
```
OMP_NUM_THREADS=1 mpirun -n 4 timeout 60s ./test-coordinates-initialization
```
in `runtest`?

d7919:
> > we might need to force one openmp thread for the test on travis
> 
> Is the best way to do this just to put
> 
> ```
> OMP_NUM_THREADS=1 mpirun -n 4 timeout 60s ./test-coordinates-initialization
> ```
> in `runtest`?
> 
> On my laptop, the test does eventually pass with OpenMP enabled and 2 or 4 threads, but it's very slow (20+ seconds).

The tests now pass in about 1s on Travis which is a good sign. I think 10-15s per test (for optimised build) is about where we've set the limit for on by default or not.

johnomotani:
> The tests now pass in about 1s on Travis which is a good sign.
(note that I did push an update setting `OMP_NUM_THREADS=1`)

d7919:
Looking at the changes to setLocation, I think we should move the getCoordinates call to the last line of the function such that it does the same thing for staggered and non-staggered cases.

johnomotani:
> I would just put this straight into master.

I had a go at putting the changes onto `master`. It turned out to be a pain to get the unit tests to work, because `master` doesn't have `FakeMesh::setCoordinates()` or the shared `FakeMeshFixture`. Also the unit tests of swap methods need updating, but aren't present in `master`. `FieldPerp::setLocation()` is also not present in `master`.

I think it would be easier, and maybe better (?), to put a dumber fix into `master`, like adding (temporarily) to the end of `BoutMesh::load()`
```
getCoordinates(CELL_CENTRE);
getCoordinates(CELL_XLOW);
getCoordinates(CELL_YLOW);
getCoordinates(CELL_ZLOW);
```

d7919:
> > I would just put this straight into master.
> 
> I had a go at putting the changes onto `master`. It turned out to be a pain to get the unit tests to work, because `master` doesn't have `FakeMesh::setCoordinates()` or the shared `FakeMeshFixture`. Also the unit tests of swap methods need updating, but aren't present in `master`. `FieldPerp::setLocation()` is also not present in `master`.
> 
> I think it would be easier, and maybe better (?), to put a dumber fix into `master`, like adding (temporarily) to the end of `BoutMesh::load()`
> 
> ```
> getCoordinates(CELL_CENTRE);
> getCoordinates(CELL_XLOW);
> getCoordinates(CELL_YLOW);
> getCoordinates(CELL_ZLOW);
> ```

Good point, I think that should be fine -- I guess the process will be something like

1. Get the above fix into master.
2. Make release 4.2.1 from master.
3. Merge master into next to ensure it is based off the newest release.
4. Update this PR to also remove the fix added in step 1.
5. Review and merge this PR

but @ZedThree will know the details.

tests/unit/mesh/test_interpolation.cxx
johnomotani: Do we need `setCoordinates` in `Mesh` though? Using the pointer from `getCoordinates()` the user can make any modifications they need to the staggered `Coordinates` objects. Is there a reason someone would want to construct a new `Coordinates` explicitly to pass to `setCoordinates`?

johnomotani:
I realized that the 'simple' fix for `master` that I proposed above is not quite enough. We also need to update the staggered `Coordinates` if the `CELL_CENTRE` ones are changed; I think updating when `Coordinates::geometry()` is called is good enough. To make that happen in #1509, I added a new method `Mesh::recalculateStaggeredCoordinates()` that re-interpolates any staggered `Coordinates` that have been added to `Mesh::coords_map`.

Unless we want to restrict `Coordinates::geometry()` to be called by the user only before `setLocation()` is first called, then we may want to keep this change (or a similar one) in `next`.

I think we can simplify the merging process slightly, I had pushed an attempt before, but rebased it away when I realized #1509 needed fixing. I think it works though:
1. #1509 is based on the common ancestor of `master` and `next`
2. Merge #1509 when it's finished into `master`
3. Merge #1509 in here
4. Here revert the commit changing `BoutMesh::load()` from #1509, plus anything else we want to remove.
5. Merge `master` and this PR to next in either order. Since the commits from #1509 are identical in both, there shouldn't be a conflict.

include/bout/mesh.hxx
```diff
@@ -444,11 +444,26 @@ class Mesh {
...
       // (circular dependency between Mesh and Coordinates)
-      coords_map.emplace(location, createDefaultCoordinates(location));
+      coords_map[location] = std::shared_ptr<Coordinates>(nullptr);
+      coords_map[location] = createDefaultCoordinates(location);
       return coords_map[location].get();
     }
   }
 
+  /// Re-calculate staggered Coordinates, useful if CELL_CENTRE Coordinates are changed
+  void recalculateStaggeredCoordinates() {
```
johnomotani: Another issue about checking that we don't do something unexpected: I think we expect `geometry()` to be called for staggered `Coordinates` only when they are being interpolated from the `CELL_CENTRE` `Coordinates`. I think we should do like 3. above and throw an exception if `geometry()` is called for `location != CELL_CENTRE`, unless an argument (different from the one for the exception above?) is passed to suppress the exception. We can pass that argument in the `Coordinates(Mesh*,CELL_LOC,Coordinates*)` constructor to allow it to call `geometry()`, and a user can if they *really* want to call `geometry()` on a staggered `Coordinates`.

d7919: This seems to be separate to the original goal of this PR (to prevent hangs) -- it's obviously important that we think about how to ensure our coordinates are consistent but this could involve a number of changes and requires more design decisions which will likely hold up the original bug fix. This is probably fine as we have an alternative fix for master, but just thought I'd mention it.

tests/unit/test_extras.hxx
```diff
@@ -195,16 +198,32 @@ public:
...
+
+    // Delete any existing mesh_staggered
+    if (mesh_staggered != nullptr) {
+      delete mesh_staggered;
+      mesh_staggered = nullptr;
+    }
+    mesh_staggered = new FakeMesh(nx, ny, nz);
+    mesh_staggered->StaggerGrids = true;
+    output_info.disable();
+    dynamic_cast<FakeMesh*>(mesh_staggered)->setCoordinates(nullptr, CELL_XLOW);
```
d7919: I think this can be a `static_cast` as we know all the types are fixed in advance?

include/bout/mesh.hxx
johnomotani: :+1: I'd agree it's better to move this discussion to another PR/Issue, to avoid delaying and complicating this one.

tests/integrated/.gitignore
```diff
@@ -35,3 +35,5 @@ BOUT.settings
 /test-petsc_laplace_MAST-grid/test_petsc_laplace_MAST_grid
 /test-stopCheck/test_stopCheck
 /test-yupdown/test_yupdown
+/test-coordinates-initialization/test-coordinates-initialization
+/test-coordinates-initialization2/test-coordinates-initialization
```
ZedThree: Does this second one still exist?

tests/unit/test_extras.hxx
```diff
@@ -255,16 +258,32 @@ public:
...
     bout::globals::mesh = nullptr;
+    delete mesh_staggered;
+    mesh_staggered = nullptr;
   }
 
   static constexpr int nx = 3;
   static constexpr int ny = 5;
   static constexpr int nz = 7;
+
+  Mesh* mesh_staggered = nullptr;
```
ZedThree: I think I must be being a bit dim; why do we need to make a separate mesh for the tests involving staggering?

johnomotani: It seemed neater to me. The tests with staggering were doing things like
```
  auto backup = mesh->StaggerGrids;
  mesh->StaggerGrids = true;
  ...
  mesh->StaggerGrids = backup;
```
and they also needed `mesh->setCoordinates(nullptr CELL_XLOW);` calls adding. By having `mesh_staggered` we only need to do the setup in one place.

If you prefer, I'd be happy to have staggering enabled in `mesh` for all the tests, but since `mesh->StaggerGrids=false` is the default, I assumed it would be better to have the tests using that unless they needed staggering.

johnomotani: @ZedThree do you want any changes to the unit tests with staggering? And are there any other issues left? Would be nice to get this PR merged :-)

d7919:
Is this ready for another review?

johnomotani:
> Is this ready for another review?

I think so :+1:

PR #1508, opened 2019-01-18
Text backup

### LINKED ISSUE #1163: Do not delete files?
In case reading of the input file fails (#1161), it would probably best to not write a `BOUT.settings` - as in this case the command line arguments are not read. I got lucky as `BOUT.settings` contains that the settings file is `BOUT.settings`, but the parser changes that to `bout.settings`, so `BOUT.settings` survived, but this is less then ideal that my settings file nearly got overwritten ...

From this experience, I have been wondering whether it would be worth to not overwrite text files in general?
I guess they should be sufficiently small that in most cases we don't care about the size. In that case the log and setting files etc should be safe from accidental deletion. Rather then overwriting move the old file to a numbered backup ...

dschwoerer:
Resolves #1163 

I prefer the integrated test, to make sure that that it works for a normal physics model.

Should I change / add documentation? Or rather, what should be updated?


src/sys/output.cxx
```diff
@@ -153,4 +155,21 @@ ConditionalOutput output_error(Output::getInstance());
 ConditionalOutput output_verbose(Output::getInstance(), false);
 ConditionalOutput output(Output::getInstance());
 
+void backupfile(const char* filename) {
```
ZedThree: It would be nicer to use `const std::string&` rather than `const char*`

src/sys/output.cxx
```diff
@@ -153,4 +155,21 @@ ConditionalOutput output_error(Output::getInstance());
...
+  struct stat buffer;
+  if (stat(filename, &buffer) == 0){
```
ZedThree: I know `stat` is POSIX, but it's not standard C++ and we should prefer to keep to the standard. An alternative would be `std::ifstream`, which is convertable to `bool`:

```cpp
bool fileExists(const std::string& filename) {
  return static_cast<bool>(std::ifstream(filename));
}
```

If we eventually get to C++17, this could just be `std::filesystem::exists`

src/sys/output.cxx
```diff
@@ -153,4 +155,21 @@ ConditionalOutput output_error(Output::getInstance());
...
+    char filename_[strlen(filename)+10];
```
ZedThree: `char[]` -> `std::string`

And instead of `sprintf`, we can do `filename_ = filename + "~" + std::to_string(j);`

src/sys/output.cxx
```diff
@@ -153,4 +155,21 @@ ConditionalOutput output_error(Output::getInstance());
...
+    int j=0;
+    do {
+      if (j > 1000000) {
```
ZedThree: Magic number should be `constexpr int` and named -- `max_backup_files`?

Also this is a pretty big number, and quite possibly more than can be supported in a single directory, depending on the filesystem. I would be surprised if a program silently made a million backups for me! Maybe this should just be one or two?

tests/integrated/test-backup/test.cxx
```diff
@@ -0,0 +1,24 @@
+#include <bout/physicsmodel.hxx>
+#include "unused.hxx"
+
+class Test : public PhysicsModel {
```
ZedThree: `Test` -> `TestFileBackup`

Similarly for the filename, please :)

src/sys/output.cxx
dschwoerer: std::ifstream checks whether the file is readable - while stat actually checks the existence.
Thus stat returns true for folders/symlinks/... - while std::ifstream would not 'detect' them.
Undetected symlinks could even lead to privilege escalation, while I don't think it is a issue here, I still think stat is the right tool for this, until we get C++17.

Are there hones attempts to port BOUT++ to non-posix systems?

src/sys/output.cxx
dschwoerer: Changed to a warning with much lower number :+1:

dschwoerer:
@ZedThree  thanks for the review, I should have addressed everything now.

src/sys/output.cxx
```diff
@@ -153,4 +155,24 @@ ConditionalOutput output_error(Output::getInstance());
...
+    std::string filename_;
+    int j=0;
+    do {
+      filename_=filename + "~" + std::to_string(j);
+      j+=1;
+    }
+    while (stat(filename_.c_str(), &buffer) == 0);
+    constexpr int backup_files_warning = 10;
+    if (j > backup_files_warning) {
+      printf("There are currently %d backup files for %s - please delete them if you don't need them anymore.",j,filename.c_str());
```
ZedThree: `printf` -> `output_warn`

ZedThree:
I'm still uneasy about essentially making unlimited number of backup files. It doesn't look like any of the tests clean up the log files, so we're quickly going to generate a *lot* of useless files -- I just ran the test suite once, and this PR generated 32 log files for test-delp2. Ok, it's not the end of the world, but it is surprising that running the tests a few times leaves hundreds of extra files about.

Why not just limit it to a single backup file? Do we need to backup the log files as well?

ZedThree:
Another thing might be to make it a runtime option: `backup:always_clobber`/`backup:never_clobber` or something. Of course, what the default should be is a matter of opinion.

What I'm trying to get at it is that sometimes backups are useful/wanted and sometimes not! A one-size fits all approach is probably not going to happen, so lets think about exactly what want and when we want it

dschwoerer:
> I think you might have accidentally removed the test file instead of renaming it. `git mv` will do the "right thing" and immediately stage the rename, btw.

Nope. The makefile does not compile this, but rather makes sure that some other file is compiled. That way we can get a relative huge speedup. I did an `rm`, and then removed the file also from git.

dschwoerer:
Anyway, I will close this.

I had #1163 open for half a year, tagged as discussion. I would have loved if we could have the discussion before the implementation.

PR #1510, opened 2019-01-19
Move getLocation() and setLocation() to Field

johnomotani:
Now that location is implemented for `Field3D`, `Field2D` and `FieldPerp` we can move the implementation of `getLocation()` and `setLocation()` into `Field`. This removes code duplication.

PR based on `two_stage_add_coordinates` to avoid merge conflict with the change to `setLocation()` in #1506.

### DIFF
--- a/include/fieldperp.hxx
+++ b/include/fieldperp.hxx
@@ -59,8 +59,9 @@ class FieldPerp : public Field {
    * will be shared (non unique)
    */
   FieldPerp(const FieldPerp& f)
-      : Field(f.fieldmesh), yindex(f.yindex), nx(f.nx), nz(f.nz), data(f.data),
-        location(f.location) {}
+      : Field(f.fieldmesh), yindex(f.yindex), nx(f.nx), nz(f.nz), data(f.data) {
+    location = f.location;
+  }
 
   /*!
    * Move constructor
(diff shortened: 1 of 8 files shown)


src/field/field.cxx
```diff
@@ -44,6 +44,38 @@ Field::Field(Mesh *localmesh) : fieldmesh(localmesh) {
 // constructors.
 }
 
+void Field::setLocation(CELL_LOC new_location) {
+  AUTO_TRACE();
+  if (getMesh()->StaggerGrids) {
+    if (new_location == CELL_VSHIFT) {
+      throw BoutException(
+          "Field3D: CELL_VSHIFT cell location only makes sense for vectors");
```
dschwoerer: typo: should now be `Field`

johnomotani:
The tests are failing for the OpenMP case with segfaults in `test-naulin-laplace` and `test-gyro`. I can't reproduce the failures. Any ideas?

johnomotani:
Correction: I can reproduce the segfault. It seems to be happening because `AUTO_TRACE()` is not thread-safe (and I guess `TRACE` and `MsgStack` in general are not either). `Field::getLocation()` was called inside some OMP parallelized loops (in particular from `tridagMatrix`), and I think that was causing the segfaults.

d7919:
Yes if we could merge these two things I think it would make life easier for some other parts of the code.

d7919:
As noted in PR message this effectively depends on #1506 (i.e. can't go straight into next).

johnomotani:
Note: tests fail now, need to merge `next` after bugfixes from `master` are merged.

dschwoerer:
As there are loads of merge conflicts - should this be rebased onto next?
I think conflicts arising from that shouldn't be worse, than what is here?

johnomotani:
This got clobbered when I force-pushed #1506. Re-basing now to fix...

include/fieldperp.hxx
```diff
@@ -59,8 +59,9 @@ class FieldPerp : public Field {
    * will be shared (non unique)
    */
   FieldPerp(const FieldPerp& f)
-      : Field(f.fieldmesh), yindex(f.yindex), nx(f.nx), nz(f.nz), data(f.data),
-        location(f.location) {}
+      : Field(f.fieldmesh), yindex(f.yindex), nx(f.nx), nz(f.nz), data(f.data) {
+    location = f.location;
```
ZedThree: Might be worth adding a constructor for `Field` that takes the location and then passing `f.location` to that

johnomotani: That's taken care of in #1624. In fact this update is implemented independently in #1624. Might be better just to close this one?

ZedThree:
Changes are now in #1624

PR #1513, opened 2019-01-21
Delp2 test requires fftw

dschwoerer:
This is a bug fix, but I think it is mostly triggered by not having rerun configure (at least that how I triggered it) - but this also requires to rerun configure to get the new bout-config, thus I think next is fine ...


ZedThree:
Autoconf hasn't had a release in many, many years, and some distros have taken to packaging `HEAD`. So autoconf 2.69 might be any number of things

d7919:
Just to note there is a non-fft based `Delp2` available if we wanted to make sure we run this test all the time.

ZedThree:
Oh, we should probably make sure we test both!

PR #1514, opened 2019-01-21
Try to provide some thread safety to msg_stack

### LINKED ISSUE #1511: TRACE is not thread-safe?
`TRACE` seems not to be thread-safe, and can cause segfaults if called from inside an OpenMP parallelized loop, see #1510. This is unfortunate because it means any function/method that might ever be called inside an OpenMP loop cannot use `TRACE`/`AUTO_TRACE`. I think this behaviour at least needs documenting. I don't know if there is a simple fix; could we perhaps make a separate `msg_stack` for each thread (maybe as a `std::vector` whose size is the maximum number of threads)?

d7919:
A better longer term fix is probably worth it but hopefully this will
help in the interim.

Attempts to fix #1511

### DIFF
--- a/src/sys/msg_stack.cxx
+++ b/src/sys/msg_stack.cxx
@@ -24,30 +24,33 @@
  *
  **************************************************************************/
 
+#include "bout/openmpwrap.hxx"
+#include <cstdarg>
 #include <msg_stack.hxx>
 #include <output.hxx>
-#include <cstdarg>
 #include <string>
 
 #if CHECK > 1
 int MsgStack::push(const char *s, ...) {
   va_list ap; // List of arguments
+  BOUT_OMP(critical(MsgStack::push)) {
+    if (s != nullptr) {
+      va_start(ap, s);
+      vsnprintf(buffer, MSG_MAX_SIZE, s, ap);
+      va_end(ap);
+    } else {
+      buffer[0] = '\0';
+    }
 
-  if (s != nullptr) {
-    va_start(ap, s);
-    vsnprintf(buffer, MSG_MAX_SIZE, s, ap);
-    va_end(ap);
-  } else {
-    buffer[0] = '\0';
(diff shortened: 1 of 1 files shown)


src/sys/msg_stack.cxx
```diff
@@ -24,30 +24,33 @@
...
+    if (position >= stack.size()) {
+      stack.emplace_back(buffer);
+    } else {
+      stack[position] = buffer;
+    }
 
-  return position++;
+    position++;
+  };
+  return position;
```
d7919: position-1

src/sys/msg_stack.cxx
```diff
@@ -58,25 +61,32 @@ int MsgStack::setPoint() {
...
 
-  if (id > static_cast<int>(position))
-    return;
+  BOUT_OMP(critical(MsgStack::pop)) {
+    if (id <= static_cast<int>(position))
+      position = id;
+  };
 
-  position = id;
+  return;
```
ZedThree: `return` is superfluous here :)

d7919:
The section name isn't strictly required but is strongly recommended to prevent locks.

d7919:
@ZedThree should this be for 4.2.1 or into next?

d7919:
There's perhaps a concern about the performance impact of these critical sections if `TRACE` is used in a tight loop -- the CI times don't seem to indicate a big change but these are not the most reliable performance indicators. It might be worth checking the performance with and without this PR for some realistic cases using significant numbers of threads?

ZedThree:
This is an existing problem in `master`, so yes, should be in to `v4.2.1-rc` please.

It would be nice to get a feel for the performance impact, but `TRACE` is really a debugging tool and can always be turned off, so I'm not overly concerned.

d7919:
Replaced by #1517

PR #1515, opened 2019-01-21
Fix listing available derivative methods

ZedThree:
Currently, it's possible to get error messages like:

```
Unknown option c4.
Available options are:
U1
U2
C2
W2
W3
C4
U3
U4
S2
FFT
SPLIT
```

Instead, just list the options that are implemented for the particular
derivative, along with the name e.g.:

```
Unknown option w3.
Available options are:
U1: First order upwinding
U2: Second order upwinding
C2: Second order central
C4: Fourth order central
```

### DIFF
--- a/src/mesh/index_derivs.cxx
+++ b/src/mesh/index_derivs.cxx
@@ -521,7 +521,10 @@ DiffLookup lookupFunc(DiffLookup *table, const std::string & label){
   // No exact match, so throw
   std::string avail{};
   for (int i = 0; DiffNameTable[i].method != DIFF_DEFAULT; ++i) {
-    avail += DiffNameTable[i].label;
+    if (!isImplemented(table, DiffNameTable[i].method)) {
+      continue;
+    }
+    avail += DiffNameTable[i].label + std::string(": ") + DiffNameTable[i].name;
     avail += "\n";
   }
   throw BoutException("Unknown option %s.\nAvailable options are:\n%s", label.c_str(),


d7919:
I think the changes to derivatives in `next` should already address this (not super relevant comment but just to flag that this _should_ have been sorted in `next`).

ZedThree:
Yes, good point! This should only be relevant for `master`

d7919:
Is the message able to say for what derivative type/direction the method is unknown (i.e. which case failed)?

ZedThree:
I did briefly investigate this, and I think it would involve adding more metadata to some of the structures, which seems a bit pointless in light of the changes in 4.3.

On the other hand, the last normal output is exactly what is trying to be set:

```
 Third order WENO (W3)
        Stag. Upwind:   Option mesh:ddy:upwind = w3 (grid8/BOUT.inp)
Error encountered during initialisation: ====== Exception thrown ======
Unknown option w3.
Available options are:
U1: First order upwinding
U2: Second order upwinding
C2: Second order central
C4: Fourth order central
====== Back trace ======
 -> derivs_set() on line 562 of 'index_derivs.cxx'
 -> Initialising derivatives on line 612 of 'index_derivs.cxx'
Stacktrace not enabled.
```

Less readable than it could be, partly due to the "Exception thrown" header, but the information is there.

Although I get output from the exception twice as we explicitly catch exceptions during initialisation, print the exception message, and then rethrow. This then aborts the run, which results in the exception message being printed again.

https://github.com/boutproject/BOUT-dev/blob/450f4d1b1148b4f94af8dcba4253361714022b3e/src/bout%2B%2B.cxx#L453-L456

I'm not sure what's best here:

1) Just print "Error during initialisation" and rethrow
2) Don't catch anything, we'll get the message regardless
3) Immediately call `std::terminate` instead of rethrowing

dschwoerer:
I am against 3) -The only reason (at least that I am aware of) a library should terminate, is if it is in a security critical state, and tries to prevent very nasty things, e.g. remote execution. Besides that, a library shouldn't terminate, throw and let the calling code handle the exception.
Currently I am only aware of the python library that uses the bout library not as a stand-alone code, but I think we should stick to good practice.

dschwoerer:
> Yes, that's true, although I don't tend to think of `BoutInitialise` and `BoutFinalise` as "true" library code as they also do lots of stuff that's really designed for executable physics models and is less useful for e.g. the python API -- anyway, that's not important for this PR!

Let us discuss in #1521

PR #1522, opened 2019-01-22
Move googletest into externalpackages

d7919:
As we may expand the number of external packages we are using it makes sense to try to keep these tidied into a specific location.

### DIFF
--- a/tests/unit/makefile
+++ b/tests/unit/makefile
@@ -6,8 +6,8 @@ BOUT_TOP = ../..
 
 BOUT_TEST_DIR = .
 
-GTEST_DIR = $(BOUT_TOP)/googletest/googletest
-GMOCK_DIR = $(BOUT_TOP)/googletest/googlemock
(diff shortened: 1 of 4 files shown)


tests/unit/makefile
```diff
@@ -31,7 +31,7 @@ include $(BOUT_TOP)/make.config
 LDFLAGS += -pthread
 
 # Existence of this file indicates that GTEST has been downloaded
-GTEST_SENTINEL = $(BOUT_TOP)/googletest/README.md
+GTEST_SENTINEL = $(BOUT_TOP)/externalpackages/googletest/README.md
```
ZedThree: We should probably just change this to use `GTEST_DIR` while we're at it

d7919: This wouldn't work as `GTEST_DIR = $(BOUT_TOP)/externalpackages/googletest/googletest` -- note the extra `googletest`

d7919: We could introduce a GTEST_BASE which everything is defined from though.

PR #1523, opened 2019-01-23
Fix typo in petsc installation instructions

bshanahan:
Fixes a small typo in the installation instructions for PESTc
* --download-f-blas-lapack=1 should be --download-fblas-lapack=1

### DIFF
--- a/manual/sphinx/user_docs/advanced_install.rst
+++ b/manual/sphinx/user_docs/advanced_install.rst
@@ -417,7 +417,7 @@ debugging.
               --download-mumps \
               --download-scalapack \
               --download-blacs \
-              --download-f-blas-lapack=1 \
+              --download-fblas-lapack=1 \
               --download-parmetis \
               --download-ptscotch \
               --download-metis


bendudson:
It looks like PETSc changed this at some point. The docs for PETSc 3.1 include `--download-f-blas-lapack`
https://www.mcs.anl.gov/petsc/petsc-3.1/docs/installation.html

bshanahan:
Ah, then perhaps we should check where it changed?  I only know that this
is the case for 3.9.

bendudson:
I had a look, but my google-fu failed to find any useful changelog for this. It seems that this is what is used for recent versions of PETSc, which is probably all we can realistically support anyway.

d7919:
I think it was 3.5 https://www.mcs.anl.gov/petsc/documentation/changes/35.html

bshanahan:
In that case, since the manual specifically shows v 3.4, we could:
1.) leave it as-is, with a note of the change for v3.5+
or
2.) have the example in the manual use a newer version.

ZedThree:
The example in the manual should really use a newer version, preferably the latest (assuming 3.10 works for us)

bendudson:
Thanks @d7919. Unfortunately that seems to add another option, `--download-fblaslapack` to go with `--download-fblas-lapack` and `--download-f-blas-lapack`. Unless it's a typo of course... 

Agree that latest version we support should be there. Perhaps with a note saying that PETSc options occasionally change.

dschwoerer:
For BEST I compile 3.4 .. 3.9 - and only 3.4 uses the old syntax

Issue #1499, opened 2019-01-15
 test MMS/advection/weno3 is broken


dschwoerer:
Should be fixed or disabled/removed

https://omadesala.alfa1024.info/BEST/2019-01-15_20:01/

Related: #1049

dschwoerer:
```
advection.cxx: In member function ‘virtual int AdvectMMS::init(bool)’:
advection.cxx:12:17: warning: unused parameter ‘restarting’ [-Wunused-parameter]
   int init(bool restarting) {
                 ^~~~~~~~~~
At global scope:
cc1plus: warning: unrecognized command line option ‘-Wno-cast-function-type’
rm: cannot remove 'data/BOUT.dmp.*.nc': No such file or directory
getmpirun: using the default mpirun -np
Making MMS advection test
  Running with method=0 mesh:ddx:upwind=w3 mesh:ddz:upwind=w3 mesh:nx=68 mesh:dx=0.09817477042468103 MZ=64
 64 |  12.657 s | ?
  -> Error norm: l-2 0.000457 l-inf 0.002288
  Running with method=0 mesh:ddx:upwind=w3 mesh:ddz:upwind=w3 mesh:nx=132 mesh:dx=0.04908738521234052 MZ=128
 128 |  19.825 s | 2.400913
  -> Error norm: l-2 0.000086 l-inf 0.000590
Convergence order = 2.400913
............ FAIL
 => Some failed tests
```

ZedThree:
There's at least two different things going on here. One is the boundary conditions: they are currently `dirichlet_o2`. Changing this to `dirichlet_o4` surprisingly makes things worse, dropping the `l_2` order from 2.4 to 2.2 for the test. Maybe `dirichlet_o4` is not actually 4th order?

The second thing is that WENO3 is 3rd order at best and drops to lower order in regions of steep gradients. I'm not sure what counts as "steep", so I looked at the scaling of the error in the regions where the bracket term was less/more than either 1.0 or 0.5 (where the max value for this test is 3.0), and the order is mostly > 3 for the small gradient regions, except when nx > 128.

I'm not convinced about the latter though, as the function is pretty smooth. I've not found a good reference for a WENO scheme that's been MMS'd for an example function. The papers I have found have been a bit hit and miss about whether they actually get the expected order or not too.

dschwoerer:
I tired dirichlet_nu_o4 from #1179 - and got convergence of 1.3
[...]

PR #1528, opened 2019-01-24
Use third order B.C. in MMS/advection/weno3 test case.

d7919:
With this change recover third order convergence

### DIFF
--- a/tests/MMS/advection/runtest
+++ b/tests/MMS/advection/runtest
@@ -99,7 +99,7 @@ def run_mms(options,exit=True):
         order = log(error_2[-1] / error_2[-2]) / log(dx[-1] / dx[-2])
         print("Convergence order = %f" % (order))
 
-        if exp_order-.25 < order < exp_order+.25:
+        if exp_order-.25 < order < exp_order+0.75:
             print("............ PASS")
         else:
             success = False


ZedThree:
Fixes #1499 

I think that just switching to `dirichlet_o3` implies that `dirichlet_o4` is probably broken

johnomotani:
I think I tested `dirichlet_o4` successfully in the `test-twistshift-with-BoundaryOp-refactor` branch, using the test in https://github.com/boutproject/BOUT-dev/blob/test-twistshift-with-BoundaryOp-refactor/tests/MMS/difops/runtest.py
That test used the `BoundaryOp-refactor` branch, but the stencils for `dirichlet_o4` were the same (I just checked the `CELL_CENTRE` versions anyway). #1315 is not very nice, but @ZedThree's comment does suggest that a nicer version that can MMS both difops and boundary conditions would be nice to have...

dschwoerer:
I also tested `dirichlet_nu_o4`, and that gave the same result as `dirichlet_o4` - thus I am also sceptical that dirichlet_o4 is actually broken ...

Could it be that dirichlet_o3 is broken, and thus gives 3rd order convergence? Assuming it introduces an sufficiently large error, that decreases with 3rd order, that would falsely give 3rd order convergence?

d7919:
> I also tested `dirichlet_nu_o4`, and that gave the same result as `dirichlet_o4` - thus I am also sceptical that dirichlet_o4 is actually broken ...
> 
> Could it be that dirichlet_o3 is broken, and thus gives 3rd order convergence? Assuming it introduces an sufficiently large error, that decreases with 3rd order, that would falsely give 3rd order convergence?

I think that's certainly possible.

dschwoerer:
Running it over a wider range, shows we are transitioning from 3.3 to ???, rather then having 3rd order convergence:
```
  16 |   2.074 s | ?
  32 |   3.835 s | 3.339857
  64 |  13.989 s | 3.359458
 128 |  24.719 s | 2.943565
 256 |  21.445 s | 2.194482
 512 |  95.321 s | 1.484496
```

d7919:
> Running it over a wider range, shows we are transitioning from 3.3 to ???, rather then having 3rd order convergence:
> 
> ```
>   16 |   2.074 s | ?
>   32 |   3.835 s | 3.339857
>   64 |  13.989 s | 3.359458
>  128 |  24.719 s | 2.943565
>  256 |  21.445 s | 2.194482
>  512 |  95.321 s | 1.484496
> ```

Ah, yes that looks bad. Have you tried this including the fix in #1529 ? I found this didn't make much difference at the 128 scale but it may help as we go beyond that.

dschwoerer:
Yes, I thought it wasn't but as that the derivatives are templated, switching to this branch didn't trigger a rebuild, so the patch was still included.

d7919:
It seems that if we use fourth order methods for the central methods in conjunction with this change to the boundaries we get at least third order convergence over an extended range.

d7919:
Switching to the fourth order boundaries the convergence order is reported as worse but the l2 error seems to be getting relatively small (~1.0e-5) and "bottoming out". If we reduce the size of the hyper-diffusion this bottoming out happens at a smaller scale (and we can recover third order if we're lucky).

d7919:
With third order BC the l2 error is smaller than with fourth order. The convergence order from 32 to 1024 points is >3 for every step (in powers of two), which I think is promising. This is true with the hyper-diffusion completely disabled and enabled, but the magnitude of the error is somewhat smaller with the hyper-diffusion disabled.

d7919:
For clarity this is without the change in #1529

PR #1529, opened 2019-01-24
Flip the sign in the rightwards velocity branch

d7919:
This should then match the equations given in Jiang and Peng (1999).

### DIFF
--- a/include/bout/index_derivs.hxx
+++ b/include/bout/index_derivs.hxx
@@ -308,7 +308,7 @@ DEFINE_UPWIND_DERIV(VDDX_WENO3, "W3", 2, DERIV::Upwind) { // No vec
     r = (WENO_SMALL + SQ(f.pp - 2.0 * f.p + f.c))
         / (WENO_SMALL + SQ(f.p - 2.0 * f.c + f.m));
 
-    deriv = (-f.m + 3. * f.c - 3. * f.p + f.pp);
+    deriv = -(-f.m + 3. * f.c - 3. * f.p + f.pp);
   }
 
   w = 1.0 / (1.0 + 2.0 * r * r);


include/bout/index_derivs.hxx
```diff
@@ -308,7 +308,7 @@ DEFINE_UPWIND_DERIV(VDDX_WENO3, "W3", 2, DERIV::Upwind) { // No vec
     r = (WENO_SMALL + SQ(f.pp - 2.0 * f.p + f.c))
         / (WENO_SMALL + SQ(f.p - 2.0 * f.c + f.m));
 
-    deriv = (-f.m + 3. * f.c - 3. * f.p + f.pp);
+    deriv = -(-f.m + 3. * f.c - 3. * f.p + f.pp);
```
johnomotani: This doesn't look right to me. In the limit of smooth f, r->1. Then the left stencil gives `1./6.*f.mm - f.m + 1./2.*f.c + 1./3.*f.p`, which is a 3rd order first derivative (see http://web.media.mit.edu/~crtaylor/calculator.html with points at -2,-1,0,1). Before this change the right stencil is similar, but after it the right stencil would give `-2./3.*f.m + 1./2.*f.c + 1./6.*f.pp`, which isn't a 3rd order first derivative. Maybe there was a typo in the paper you cited @d7919?

d7919: Yes that's definitely possible.

d7919: (in the other limit both branches reduce to a second order central difference, I'm a bit surprised it's second order rather than first order and one sided but that's without thinking about it too hard)

d7919:
Closing following input from @johnomotani combined with fix to MMS test in #1528 which suggests we do get 3rd order without the change proposed here.

PR #1531, opened 2019-01-25
Write job execution information to dump file

JosephThomasParker:
Write job execution data (like wall clock times and number of RHS evaluations) to the dump file.  When looking at code performance, it's easier to have this data available in the dump file than to have to extract it from the log files!

This PR makes a `run_metrics` class to contain timing and iteration counts.  These were previously just local variables in `bout++.cxx`.


include/bout/run_metrics.hxx
```diff
@@ -0,0 +1,65 @@
...
+///
+/// The `Run Metrics` class holds non-physics information about the job execution,
+/// such as wall clock times and number of iterations.
+
+class RunMetrics;
+
+#ifndef __RUNMETRICS_H__
+#define __RUNMETRICS_H__
+
+#include "datafile.hxx"
```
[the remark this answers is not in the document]
ZedThree: The forward declaration of `RunMetrics` can go

include/bout/run_metrics.hxx
```diff
@@ -0,0 +1,65 @@
...
+
+
+class RunMetrics {
+  public:
+  BoutReal simtime;      // cumulative simulation time
```
ZedThree: Please could you move the comments to the line before the variable and use triple-slashes for doxygen: `///`

JosephThomasParker: Thanks - I've Doxyfied them.

include/bout/run_metrics.hxx
```diff
@@ -0,0 +1,65 @@
...
+  BoutReal t_elapsed;    // cumulative wall clock time in seconds
+  BoutReal wtime;        // time step's wall clock time in seconds
+
+  int ncalls;            // number of RHS calls
+  int ncalls_e;          // number of RHS calls for fast timescale
+  int ncalls_i;          // number of RHS calls for slow timescale
+
```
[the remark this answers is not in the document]
JosephThomasParker: Thanks - added.

src/sys/run_metrics.cxx
```diff
@@ -0,0 +1,73 @@
...
+  file.add(wtime_io, "wtime_io", true);
+  file.add(wtime_per_rhs, "wtime_per_rhs", true);
+  file.add(wtime_per_rhs_e, "wtime_per_rhs_e", true);
+  file.add(wtime_per_rhs_i, "wtime_per_rhs_i", true);
+}
+
+void RunMetrics::calculateDerivedMetrics() {
+
+  wtime_per_rhs = wtime / ncalls;
+  wtime_per_rhs_e = wtime / ncalls_e;
```
d7919: I'm not sure this and `wtime_per_rhs_i` will be useful as `wtime` is the total time for the output step, which will have some non-trivial split between the explicit and implicit parts. If `ncalls==ncalls_e+ncalls_i` in this case then the first metric is still partially useful. The second two might be more useful as `ncalls_e/ncalls` and `ncalls_i/ncalls` here so that they're fraction of calls that are explicit/implicit. @bendudson may be better positioned to comment here though as I've not worked with the split schemes for a while.

include/bout/run_metrics.hxx
```diff
@@ -0,0 +1,87 @@
...
+
+class Datafile;
+#include "bout_types.hxx"
+
+
+class RunMetrics {
+  public:
+  /// cumulative wall clock time in seconds
+  BoutReal t_elapsed;
```
ZedThree: I would initialise all these to 0

ZedThree:
Do we need the `RunMetrics` class outside of `BoutMonitor`? If not, I might suggest doing the following:

- Move the definition of `RunMetrics` into `BoutMonitor` and make it a `struct`
- Make the methods free functions in `bout++.cxx`

This would leave the same functionality, without exposing the details of how we're doing it

include/bout/run_metrics.hxx
JosephThomasParker: Thanks - done.

JosephThomasParker:
I think `RunMetrics` has to be outside `Monitor`, as members of `RunMetrics` get `add`ed to the dump file during initialization, while the `Monitor` itself doesn't exist until `bout_run` gets called.

d7919:
It seems strange that the tests fail with these changes.

src/bout++.cxx
```diff
@@ -87,6 +87,7 @@ void bout_signal_handler(int sig);  // Handles signals
 #include <output.hxx>
 
 BoutReal simtime;
+RunMetrics run_data;
```
ZedThree: Can this be a member of `BoutMonitor` instead of being a global? You'd need to also add a constructor to `BoutMonitor` and move the `run_data.outputVars` call to there.

PR #1533, opened 2019-01-31
Remove checkData from copy constructor and assignment operator in Fields

d7919:
This allows us to copy and assign from empty fields.

### DIFF
--- a/src/field/field3d.cxx
+++ b/src/field/field3d.cxx
@@ -62,10 +62,6 @@ Field3D::Field3D(const Field3D& f) : Field(f.fieldmesh), data(f.data) {
 
   TRACE("Field3D(Field3D&)");
 
-#if CHECK > 2
-  checkData(f);
-#endif
-
   if (fieldmesh) {
     nx = fieldmesh->LocalNx;
     ny = fieldmesh->LocalNy;
@@ -269,9 +265,6 @@ Field3D & Field3D::operator=(const Field3D &rhs) {
     return(*this); // skip this assignment
 
   TRACE("Field3D: Assignment from Field3D");
-  
-  /// Check that the data is valid
-  checkData(rhs);
   
   // Copy the data and data sizes
   fieldmesh = rhs.fieldmesh;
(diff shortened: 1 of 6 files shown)


ZedThree:
I think we test for this behaviour, so you'll likely need to remove those tests as well

d7919:
> I think we test for this behaviour, so you'll likely need to remove those tests as well

It passed all the unit tests locally!

ZedThree:
Not sure why CodeCov has chimed in!

Any major objections to this? I think this is the right thing to do (see #1385)

d7919:
One possible objection is operator= with mixed types (e.g 3d = 2d) have to do checkdata but same types (3d=3d) don't.

johnomotani:
I agree this is the right thing to do :+1:

ZedThree:
For the mixed types assignment, I think we need *at least* the check the rhs is allocated (else the copy won't work), but we could change what we do with that information, e.g. if the rhs is not allocated, deallocate the lhs and return. Otherwise, do the copy without checking if the rhs is finite.

d7919:
> For the mixed types assignment, I think we need _at least_ the check the rhs is allocated (else the copy won't work), but we could change what we do with that information, e.g. if the rhs is not allocated, deallocate the lhs and return. Otherwise, do the copy without checking if the rhs is finite.

Yes I thought about that a bit but decided for now that I'd just do the simplest thing (i.e. leave the checks in place). Happy to replace those with an `isAllocated` type check though.

d7919:
> I think having an `isAllocated` check, and invalidating the LHS if the RHS is invalid is probably the most consistent approach. I have some worry that not checking for invalid data will make tracking down bugs harder, but a user can always add calls to `checkData` for debugging.

I hope this shouldn't impact debugging too much as we still rigorously `checkData` on the input and output of most operations so at the point of doing any calculation with the fields we should catch this.

src/field/field3d.cxx
```diff
@@ -279,10 +279,10 @@ Field3D & Field3D::operator=(const Field3D &rhs) {
 
 Field3D & Field3D::operator=(const Field2D &rhs) {
   TRACE("Field3D = Field2D");
-  
-  /// Check that the data is valid
-  checkData(rhs);
- 
+
+  /// Check that the data is allocated
+  ASSERT1(rhs.isAllocated());
```
d7919: Instead of `ASSERT` here we could just choose to return the unallocated `Field3D` so that we allow assignment from unallocated fields -- I'm not sure that would be good but is an option.

PR #1534, opened 2019-01-31
Derivative unit tests

ZedThree:
Basic sanity checks for all the first and second derivative kernels

Also a couple of useful helper utilities:

- `EnableIfField<T>` for enabling functions only if `T` is derived from `Field`
- `makeField<T>` returns a field with the result of calling a function at each point in index space

These are just in the unit tests at the mo, but could be moved into the main library if they're more generally useful

### DIFF
--- a/tests/unit/include/test_derivs.cxx
+++ b/tests/unit/include/test_derivs.cxx
@@ -21,23 +21,26 @@
 // variation in the other directions.
 //
 // This is one of the more complicated uses of googletest! We need to
-// test all combinations of methods and directions. Unfortunately, Z
-// has one more method than X and Y -- FFT. This means we can't just
-// use the provided `Combine` to produce the Cartesian product of
-// directions and methods as the latter depends on the
-// former. Instead, we instantiate the tests separately for each
-// direction and test all the methods for that direction in that
-// instantiation.
+// test all combinations of methods, directions and (standard)
+// derivative types. Unfortunately, Z has one more method than X and Y
+// (FFT), and the different orders have different sets of
+// methods. This means we can't just use the provided `Combine` to
+// produce the Cartesian product of methods, directions, and types as
+// the first depends on the second two. Instead, we instantiate the
+// tests separately for each direction and order and test all the
+// methods for that direction in that instantiation.
 
 namespace {
-// These are merely sanity checks, so don't expect them to agree very well!
-constexpr BoutReal derivatives_tolerance{1.e-2};
+// These are merely sanity checks, so don't expect them to agree very
+// well! This has to be sufficiently loose for the least accurate
+// method to pass, or we need to also match up tolerances to methods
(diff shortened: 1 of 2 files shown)


d7919:
Is it worth adding the fourth order derivatives as well as these have the same signature as first/second?

tests/unit/include/test_derivs.cxx
```diff
@@ -0,0 +1,222 @@
...
+// one more method than X and Y -- FFT. This means we can't just use
+// the provided `Combine` to produce the Cartesian product of
+// directions and methods as the latter depends on the
+// former. Instead, we instantiate the tests separately for each
+// direction and test all the methods for that direction in that
+// instantiation.
+
+namespace {
+// These are merely sanity checks, so don't expect them to agree very well!
+constexpr BoutReal derivatives_tolerance{1.e-2};
```
d7919: How small can we go currently?

tests/unit/include/test_derivs.cxx
```diff
@@ -0,0 +1,222 @@
...
+#include <vector>
+
+// Some basic sanity checks for the derivative kernels. Checks the
+// derivatives of sin(R) where R = {X, Y, Z} for each R
+// individually. To make this as fast as possible, we use only a
+// couple of points in the non-tested directions -- not just one
+// though, as this allows us to check we're not introducing spurious
+// variation in the other directions.
+//
+// This is one of the more complicated uses of googletest! We need to
```
d7919: Missing word "need to..."

tests/unit/include/test_derivs.cxx
```diff
@@ -0,0 +1,222 @@
...
+// These are merely sanity checks, so don't expect them to agree very well!
+constexpr BoutReal derivatives_tolerance{1.e-2};
+}
+
+class DerivativesTest
+    : public ::testing::TestWithParam<std::pair<DIRECTION, std::string>> {
+public:
+  DerivativesTest() : input{mesh}, first_order_expected{mesh} {
+
+    // Make sure fft functions are quiet by setting fft_measure to false
```
d7919: Has the side effect of making them deterministic.

tests/unit/include/test_derivs.cxx
ZedThree: Smallest we can do for all the first derivatives is half this.

- `W2` fails below 5.e-3
- `C2` and `W3` fail below 2.e-4
- `S2` fails below 8.e-4
- `C4` fails below 5.e-7
- `FFT` fails below 1.e-14

We could pair the tolerances with the methods, but that would add additional complexity for what's supposed to be a basic test and I think would be more fragile.

I wouldn't be opposed to dropping the tolerance to 5.e-3, but I would be wary of doing so.

d7919: This is used as an absolute tolerance I think -- do you know what a relative tolerance limit would look like?

PR #1538, opened 2019-02-01
Enable FPE checking by default

dschwoerer:
As some checks get disabled in #1533 - it may be beneficially to catch floating point exceptions in a different way.

This enables by default SIG_FPE - if the required fenv header is available.

### DIFF
--- a/configure
+++ b/configure
@@ -6027,16 +6027,36 @@ $as_echo "$as_me: Field name tracking disabled" >&6;}
 
 fi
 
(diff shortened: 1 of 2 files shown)


dschwoerer:
I thought that there should be no impact, but I failed to find a source.

I wrote a small test, and I could find no significant bias, whether disabling or enabling is faster.

This still needs work, as parts of the field may not be set, (due to the region flag) and that may cause issues ...

I will close this, until I have a patch ...

Issue #1503, opened 2019-01-15
test MMS/spatial/advection is broken


dschwoerer:
Should be fixed or disabled/removed

https://omadesala.alfa1024.info/BEST/2019-01-15_20:01/

dschwoerer:
Cvode doesn't converge:
```
advection.cxx: In member function ‘virtual int AdvectMMS::init(bool)’:
advection.cxx:12:17: warning: unused parameter ‘restarting’ [-Wunused-parameter]
   int init(bool restarting) {
            ~~~~~^~~~~~~~~~
rm: cannot remove 'data/BOUT.dmp.*.nc': No such file or directory
/usr/lib64/python3.6/site-packages/h5py/__init__.py:36: FutureWarning: Conversion of the second argument of issubdtype from `float` to `np.floating` is deprecated. In future, it will be treated as `np.float64 == np.dtype(float).type`.
  from ._conv import register_converters as _register_converters
getmpirun: using the default mpirun -np
Making MMS steady-state advection test
  Running with method=2 mesh:nx=20 mesh:dx=0.39269908169872414 MZ=16
  -> Error norm: l-2 0.383543 l-inf 0.983098
  Running with method=2 mesh:nx=36 mesh:dx=0.19634954084936207 MZ=32
  -> Error norm: l-2 0.046825 l-inf 0.133331
Convergence order = 3.034030
... Failure
  Running with method=0 mesh:ddx:upwind=u1 mesh:ddz:upwind=u1 mesh:nx=20 mesh:dx=0.39269908169872414 MZ=16
  -> Error norm: l-2 0.408177 l-inf 1.012092
  Running with method=0 mesh:ddx:upwind=u1 mesh:ddz:upwind=u1 mesh:nx=36 mesh:dx=0.19634954084936207 MZ=32
  -> Error norm: l-2 0.096668 l-inf 0.253686
Convergence order = 2.078086
... Failure
  Running with method=0 mesh:ddx:upwind=c2 mesh:ddz:upwind=c2 mesh:nx=20 mesh:dx=0.39269908169872414 MZ=16
  -> Error norm: l-2 0.388361 l-inf 0.998959
  Running with method=0 mesh:ddx:upwind=c2 mesh:ddz:upwind=c2 mesh:nx=36 mesh:dx=0.19634954084936207 MZ=32
  -> Error norm: l-2 0.047559 l-inf 0.131730
Convergence order = 3.029612
... Failure
[...]

PR #1540, opened 2019-02-04
Replace MMS/spatial/advection with simpler/more complete test

ZedThree:
Instead of time-evolving a steady-state solution, just calculate the bracket term directly and compare that to the analytic answer.

Also checks the bracket operator with all the upwind derivatives (except WENO3 which has problematic behaviour, see https://github.com/boutproject/BOUT-dev/issues/1503#issuecomment-459805592)


tests/MMS/spatial/advection/runtest
```diff
@@ -30,99 +26,102 @@ shell_safe("make > make.log")
...
+    # sizes, which is suspicious. It does seem to work correctly in
+    # other contexts, so not overly worrying that it doesn't work
+    # exactly as expected here
+    # ("method=1 mesh:ddx:upwind=w3 mesh:ddz:upwind=w3", "3rd order WENO", "-s", 3),
+]
 
 # List of NX values to use
-nxlist = [16, 32]
-#nxlist = [16, 32, 64, 128, 256, 512, 1024]
+nxlist = [16, 32, 64, 128, 256]
```
dschwoerer: what is the point of running them all, if we only check the last 2 points?

tests/MMS/spatial/advection/runtest
```diff
@@ -30,99 +26,102 @@ shell_safe("make > make.log")
...
-    if not ( exp_ord + .2 > order > exp_ord - .2):
-        success=False
+    if not (exp_ord + .2 > order > exp_ord - .2):
+        success = False
         print("... Failure")
     else:
         print("... Success")
 
     # plot errors
     if False:
```
dschwoerer: replace with `if '-p' in sys.argv:`?

tests/MMS/spatial/advection/runtest
```diff
@@ -132,61 +131,62 @@ with open("advection.pkl", "wb") as output:
...
-plt.xscale('log')
-    
-plt.xlabel(r'Mesh spacing $\delta x$')
-plt.ylabel(r'$l^\infty$ error norm')
-plt.savefig("advection_norm_linf.pdf")
-plt.close()
-
-######
+# Plot all errors
+if False:
```
dschwoerer: replace with `if '-p' in sys.argv:`?

tests/MMS/spatial/advection/runtest
ZedThree: It's useful to see the error over those points, especially if you're plotting the error.

Also this test is only run on `all_tests`, and not doing the first three points only saves us a few seconds anyway, as the last run takes the majority of the time.

dschwoerer: Why don't you check these points as well?

I am not object to run them, I am just objecting to run them, and ignore them. That is all.

ZedThree: We're pretty inconsistent about whether we check the full range or the last two points. I've added a module `boututils.check_scaling` that provides some functions so we can be more consistent. They're only used in this test so far

tools/pylib/boututils/check_scaling.py
```diff
@@ -0,0 +1,70 @@
...
+
+    return (full_range[0], small_spacing)
+
+
+def check_order(actual_order, expected_order, tolerance=2.e-1):
+    """Check if the actual_order is sufficiently close to the
+    expected_order within a given tolerance
+
+    """
+    return isclose(actual_order, expected_order, rtol=tolerance)
```
dschwoerer: I would not have expected that the tolerance increases with the order.
Do we intend it to be this way?
`rtol=0, atol=tolerance` is what I would have expected.
[From the documentation](https://docs.scipy.org/doc/numpy/reference/generated/numpy.isclose.html)
```
    absolute(a - b) <= (atol + rtol * absolute(b))
```

tools/pylib/boututils/check_scaling.py
```diff
@@ -0,0 +1,70 @@
...
+
+    """
+    full_range = polyfit(log(grid_spacing), log(errors), 1)
+
+    small_spacing = log(errors[-2] / errors[-1]) / log(grid_spacing[-2] / grid_spacing[-1])
+
+    return (full_range[0], small_spacing)
+
+
+def check_order(actual_order, expected_order, tolerance=2.e-1):
```
dschwoerer: I like the idea of a unified api, but I would prefer if the API would take the L2/Linf errors, the grid spacing (optional) and the (absolute) tolerance (also optional) and return success.

NB: The grid spacing is not really needed in (basically) all cases I have seen, as we are going from coarse to fine grid, doubling the number of grid points, thus this is repetetive and can be skipped, as the absolute values don't matter.

tools/pylib/boututils/check_scaling.py
```diff
@@ -0,0 +1,70 @@
...
+
+    Returns
+    -------
+    tuple of float
+        The first value is the error scaling over the full range of
+        grid spacings; the second value is the scaling over the last
+        two points
+
+    """
+    full_range = polyfit(log(grid_spacing), log(errors), 1)
```
dschwoerer: Wouldn't a point-by-point comparison be the strictest?
I would prefer if that would be the default way to check the error.

tests/MMS/spatial/advection/runtest
```diff
@@ -13,180 +13,172 @@ from builtins import str
...
-    order = log(error_2[-1] / error_2[-2]) / log(dx[-1] / dx[-2])
-    print("Convergence order = %f" % (order))
+    order = check_scaling.get_order(dx, error_2)
+    print("Convergence order = {:f} ({:f} at small spacing)".format(*order))
 
-    if not ( exp_ord + .2 > order > exp_ord - .2):
-        success=False
-        print("... Failure")
-    else:
+    success = check_scaling.check_order(order[0], exp_ord)
```
dschwoerer: note that this will overwrite the previous success state, thus potentially leading to a clean exit, when it shouldn't.
```
if check_scaling...
    print("... Success")
else:
   success = False
   print("... Failure")
```
should fix it

tools/pylib/boututils/check_scaling.py
ZedThree: Again, we're not very consistent with this. In some places we use an absolute tolerance and in others a relative tolerance. I don't think it makes a great deal of difference either way, to be honest.

tools/pylib/boututils/check_scaling.py
ZedThree: Splitting it into calculating the error and checking it allows us to do things in-between, such as print the order. Combining the two functions into a one-liner is also trivial, so I much prefer this split

tools/pylib/boututils/check_scaling.py
ZedThree: That might be strictest way, but that doesn't give you a single value for the order. We could take the average, but that would be prone to the same errors as fitting.

tools/pylib/boututils/check_scaling.py
ZedThree: Sorry, I meant practically it doesn't seem to make any difference. I'm not sure why you would expect it to behave one way or the other anyway?

PR #1542, opened 2019-02-04
Improve make-script example

dschwoerer:
* Cache bout-config results
* Bail-out early if bout-config is not in path

### DIFF
--- a/examples/make-script/makefile
+++ b/examples/make-script/makefile
@@ -1,5 +1,6 @@
 # Example makefile. Sources are listed in SOURCEC
 # and optionally a TARGET executable name can be given
+# It requires that bout-config is in $PATH
 
 SOURCEC = test.cxx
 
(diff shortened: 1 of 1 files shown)


examples/make-script/makefile
```diff
@@ -9,10 +10,17 @@ OBJ = $(SOURCEC:%.cxx=%.o)
...
-CXX=`bout-config --cxx`
-CFLAGS=`bout-config --cflags`
-LD=`bout-config --ld`
-LDFLAGS=`bout-config --libs`
+CXX:=$(shell bout-config --cxx)
+
+ifneq "$(.SHELLSTATUS)" "0"
+which=$(shell which bout-config)
+$(info Output of which bout-config: $(which))
+$(error Failed to run bout-config.)
```
ZedThree: Maybe combine these two lines and put something like "is bout-config in $PATH"? Just to make the error a bit nicer

PR #1546, opened 2019-02-05
Fix bug in DST option for cyclic solve

bendudson:
Was putting data into `bcmplx` rather than `bcmplx3D`, so the cyclic solver with `dst=true` always returned zero.


ZedThree:
Should this be into `master`?

PR #1547, opened 2019-02-05
Replace IsField\*Equal\* predicates with templated versions

ZedThree:
Doesn't save many lines but does make usage a bit more readable


tests/unit/test_extras.hxx
```diff
@@ -66,6 +46,66 @@ inline std::ostream& operator<< (std::ostream &out, const SpecificInd<N> &index)
   return out << index.ind;
 }
 
+/// Helpers to get the type of a Field as a string
+auto inline getFieldType(MAYBE_UNUSED(const Field2D& field)) -> std::string {
```
d7919: Can we be stricter and switch `MAYBE_UNUSED` --> `UNUSED`?

ZedThree: I think we should prefer to use `MAYBE_UNUSED` as it tries to use machinery that's in the standard. `UNUSED` tries to use a GCC extension

d7919: But don't they do different things -- one is essentially a convenience for saying just don't raise a warning if we don't use it whilst the other says it's an error to try to use it. It probably doesn't matter here but I think we might need to be careful with MAYBE_UNUSED -- it hides the warning, but the warning might actually indicate a real problem whilst saying UNUSED is a clear declaration that it's expected that this thing isn't used.

PR #1548, opened 2019-02-06
Swap coordinates when adding

### LINKED ISSUE #1544: fieldCoordinates pointers not updated if mesh->coords_map is changed
`coords_map` in a `Mesh` may be changed, in particular if `Coordinates::geometry()` is called to update the `Coordinates` objects, which calls `Mesh::recalculateStaggeredCoordinates()`. This changes the pointers in `coords_map`, so any `Coordinates*` pointers that have previously been stored as `fieldCoordinates` in `Field`s become out of date. I think there are two possible solutions
1. in `recalculateStaggeredCoordinates()` create new `Coordinates` objects and then copy them into the existing objects in `coords_map`. Requires adding a `Coordinates::operator=(Coordinates)` method. This way the pointers in `coords_map` don't change, so existing `fieldCoordinates` variables are OK. For a test implementation see https://github.com/boutproject/BOUT-dev/tree/fix_coordinates_recalculation
2. don't store `fieldCoordinates` pointers in `Field`s. Instead look up using `fieldmesh->getCoordinates(getLocation())` each time `Field::getCoordinates()` is called. For test implementation see https://github.com/boutproject/BOUT-dev/tree/fix_coordinates_recalculation2

In both of the branches linked above, I've added checks in field operations, e.g. `ASSERT1(lhs.getCoordinates() == rhs.getCoordinates());` which make the bug throw an exception, for example in `tests/integrated/test-drift-instability-staggered` until the final, bugfix commit is applied.

I favour option 2 as a fix, because it also means we don't need to worry about updating `fieldCoordinates`, for example if the location of a `Field` is changed. There is a performance penalty from having the lookup from a `std::map` instead of storing the pointer, but I guess that would be small?

Whichever option we choose as the fix, I guess it should go into `master`. I'll make a PR with just the fix and not the extra checks when we decide.

d7919:
Fixes #1544 

`Field` now stores a `std::shared_ptr<Coordinates>` rather than a raw `Coordinates*`. In the long run this may be what we want to return from `Field::getCoordinates` but I've not changed this here in order to maintain backwards compatibility. Similarly I had to add an extra method to `Mesh` to be able to request the smart pointer rather than the raw pointer.

As it's a bug fix we may also want this to go into master (I figure we'd like it in next as well as we've just had a bug fix release so unlikely to get fixes in master into next for a while).

### DIFF
--- a/include/bout/mesh.hxx
+++ b/include/bout/mesh.hxx
@@ -435,17 +435,22 @@ class Mesh {
 
   /// Coordinate system
   Coordinates *getCoordinates(const CELL_LOC location = CELL_CENTRE) {
+    return getCoordinatesSmart(location).get();
+  };
+
+  std::shared_ptr<Coordinates>
+  getCoordinatesSmart(const CELL_LOC location = CELL_CENTRE) {
     ASSERT1(location != CELL_DEFAULT);
     ASSERT1(location != CELL_VSHIFT);
 
     if (coords_map.count(location)) { // True branch most common, returns immediately
-      return coords_map[location].get();
+      return coords_map[location];
     } else {
       // No coordinate system set. Create default
       // Note that this can't be allocated here due to incomplete type
       // (circular dependency between Mesh and Coordinates)
       coords_map.emplace(location, createDefaultCoordinates(location));
-      return coords_map[location].get();
+      return coords_map[location];
     }
   }
 
@@ -513,7 +518,7 @@ class Mesh {
         continue;
       }
 
-      i.second = createDefaultCoordinates(location);
+      std::swap(*coords_map[location], *createDefaultCoordinates(location));
     }
   }
 
--- a/include/field.hxx
+++ b/include/field.hxx
@@ -30,6 +30,7 @@ class Field;
 #define __FIELD_H__
 
 #include <cstdio>
+#include <memory>
 
 #include "bout_types.hxx"
 #include "boutexception.hxx"
@@ -122,7 +123,7 @@ class Field {
 
 protected:
   Mesh* fieldmesh{nullptr};
-  mutable Coordinates* fieldCoordinates{nullptr};
+  mutable std::shared_ptr<Coordinates> fieldCoordinates{nullptr};
 };
 
 /// Unary + operator. This doesn't do anything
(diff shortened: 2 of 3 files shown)


johnomotani:
I've made another PR to add the bugfix to master. Should I force-push this branch to `next` with `swap_coordinates_when_adding_master` merged? I think that would result in a cleaner history when we eventually merge `master` into `next`.

d7919:
> I've made another PR to add the bugfix to master. Should I force-push this branch to `next` with `swap_coordinates_when_adding_master` merged? I think that would result in a cleaner history when we eventually merge `master` into `next`.

I think we should generally try to avoid force-push wherever we can -- if the other PR is just a cherry-pick of this commit then I think the history should be fine but @ZedThree can probably comment.

ZedThree:
I strongly recommend not force-pushing, especially to branches with open PRs.

I wouldn't worry about clean history, the merge commits already see to that!

johnomotani:
This PR does not compile on my laptop, I get an error
```
  Compiling  coordinates.cxx
In file included from ../../include/bout/coordinates.hxx:38:0,
                 from coordinates.cxx:9:
../../include/bout/mesh.hxx: In member function ‘void Mesh::recalculateStaggeredCoordinates()’:
../../include/bout/mesh.hxx:521:75: error: no matching function for call to ‘swap(std::__shared_ptr_access<Coordinates, (__gnu_cxx::_Lock_policy)2, false, false>::element_type&, std::__shared_ptr_access<Coordinates, (__gnu_cxx::_Lock_policy)2, false, false>::element_type&)’
       std::swap(*coords_map[location], *createDefaultCoordinates(location));
```
I think it might be because `Coordinates` is an incomplete type when the compiler tries to compile `recalculateStaggeredCoordinates()`, because if I move the implementation into `mesh.cxx` then it compiles ok. We do include `coordinates.hxx` in `mesh.hxx` though, so I'm not sure why it fails, especially as it compiles on Travis. My gcc is 7.3.0, OpenMPI is 2.1.1.

Maybe it would be safer to move the implementation to `mesh.cxx`: @d7919 @ZedThree I can push a commit if you agree?

johnomotani:
Closing, this fix was added to `master` in #1549 and has now been merged to `next` with v4.2.2.

PR #1549, opened 2019-02-06
Swap coordinates when adding - to master

johnomotani:
Add the bugfix in #1548 to `master`.

### DIFF
--- a/include/bout/mesh.hxx
+++ b/include/bout/mesh.hxx
@@ -431,17 +431,22 @@ class Mesh {
 
   /// Coordinate system
   Coordinates *getCoordinates(const CELL_LOC location = CELL_CENTRE) {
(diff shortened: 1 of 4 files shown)


johnomotani:
Tests fail with debugging switched on because of the `checkData` in `Field2D`'s copy constructor (which has been removed in `next`. `swap(Coordinates,Coordinates)` calls the copy constructor for every member, but `IntShiftTorsion` is not set to anything unless `mesh->IncIntShear` is true. Not sure what the best thing to do is. One workaround would be to set `IntShiftTorsion = 0.` in `master` when creating staggered `Coordinates`, and then revert that change in `next`.

d7919:
Yes I think that probably makes sense, it will increase memory usage slightly but I think it's better that the Field is initialised.

PR #1550, opened 2019-02-06
Remove old examples and change to use new options style

### LINKED ISSUE #1457: Examples that don't compile
It seems that both `non-local_1d` and `test-difops` rely on aspects of BOUT++ that no longer exist (and were possibly experimental). Should we retire/remove these examples or seek to update to be compatible with recent BOUT++ versions (if this is even possible)?

d7919:
Removes two examples demonstrating approaches not currently supported (fixes #1457) and removes use of the `OPTIONS` macro in favour of the newer approach.


examples/2Dturbulence_multigrid/esel.cxx
```diff
@@ -20,17 +20,17 @@ class ESEL : public PhysicsModel {
...
-    int bracket; 
-    OPTION(options, bracket, 2);
-    OPTION(options, test_laplacian, false);
-    
+
+    auto options = Options::root()["esel"];
+
+    zeta = options["zeta"].withDefault(2.15e-3);
+    D = options["D"].withDefault(1.97e-3);
+    //   chi = options["chi"].withDefault(4.61e-3) ;
```
ZedThree: If we're going to touch all this, may as well remove commented out code too!

examples/2Dturbulence_multigrid/esel.cxx
```diff
@@ -20,17 +20,17 @@ class ESEL : public PhysicsModel {
...
+    mu = options["mu"].withDefault(3.88e-2);
+    int bracket;
+    bracket = options["bracket"].withDefault(2);
```
ZedThree: This could be merged with the previous line. Not sure how to do it automatically, although there don't appear to be many instances like this

examples/6field-simple/elm_6f.cxx
```diff
@@ -383,45 +383,62 @@ int physics_init(bool restarting) {
...
-  OPTION(options, Sheath_width, 1);  // Sheath boundary width in grid number
-
-  OPTION(options, density, 1.0e19);   // Number density [m^-3]
-  OPTION(options, Zi, 1);             // ion charge number
-  OPTION(options, continuity, false); // use continuity equation
-
-  OPTION(options, evolve_jpar, false);    // If true, evolve J raher than Psi
-  OPTION(options, phi_constraint, false); // Use solver constraint for phi
+  n0_fake_prof = options["n0_fake_prof"].withDefault(
+      false); // use the hyperbolic profile of n0. If both  n0_fake_prof and
```
ZedThree: Clang-format usually handles these a bit better if you move the comment to the previous line

Issue #1557, opened 2019-02-07
wall_limit not working as expected


dschwoerer:
I have noticed in one run that wall_limit did not trigger on all threads.
Some files have:
```
6.000e+01      25860       5.72e+03    56.9    2.0    1.7    0.0   39.4
Only 1.022924e+04 seconds (1.79 steps) left. Quitting
Monitor signalled to quit
Error encountered in solver run
```
But some have just
```
6.000e+01      25860       5.72e+03    57.1    2.0    1.6    0.0   39.3
```
I didn't notice at first, because everything seemed fine - but things got strange when I restarted the simulation:
```
Run started at  : Wed Feb  6 10:07:20 2019

        Option :restart = true (Command line)
        Option :append = true (Command line)
        Option :dump_on_restart = false  (default)
slurmstepd: error: *** JOB 3575425 ON r145c05s02 CANCELLED AT 2019-02-07T10:07:32 DUE TO TIME LIMIT ***
slurmstepd: error: *** STEP 3575425.0 ON r145c05s02 CANCELLED AT 2019-02-07T10:07:32 DUE TO TIME LIMIT ***
```
i.e. the simulation was doing nothing for 24 hours.

I am wondering whether the monitor triggered an hard-quit, and thus some processes got killed before they finished writing - and thus the dump files are in an inconsistent state? Preventing the simulation from running?

This is an older version of next, but I am not sure how to reproduce ...

dschwoerer:
Fixed by #1558 in next.
Should it also be fixed in master?

dschwoerer:
Fixed by #1610 in master

PR #1558, opened 2019-02-07
Prevent data corruption

dschwoerer:
Returning 1 throws an exception, which can cause to corrupted data.
This makes wall_time more reliable.
I just had a simulation that restarted from a simulation stopped by wall_limit that didn't do anything for 24 hours. Collect also fails, as the number of time slices is different for different dump files.

The factor of 2 is to incorporate varying duration of time-steps.

### DIFF
--- a/src/bout++.cxx
+++ b/src/bout++.cxx
@@ -702,11 +702,10 @@ int BoutMonitor::call(Solver *solver, BoutReal t, int iter, int NOUT) {
     // Check if enough time left
 
     BoutReal t_remain = mpi_start_time + wall_limit - MPI_Wtime();
-    if (t_remain < wtime) {
+    if (t_remain < wtime*2) {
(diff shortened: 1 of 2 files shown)


dschwoerer:
While it does prevent data-corruption - it will only terminate one step later.
I will change it such that a non-zero return will not throw, but cause a clean shutdown.

dschwoerer:
I tried simply passing on the return 1 to the solver, but that fails if not all processes return 1;
A monitor can now set `user_requested_exit=true` to get a clean exit.

I guess we could remove the first `MPI_Allreduce` - because we anyway cannot guarantee that the monitors get called with `iter == NOUT - 1` ...

dschwoerer:
Can we merge this?

This fixes #1557 
Or should it go directly to master?

PR #1559, opened 2019-02-07
Inform about gettext's license

dschwoerer:
By default BOUT++ compiles with gettext, thus to my understanding BOUT++ is most often GPL, rather then LGPL. If that understanding is correct, should we inform in the README?

The LGPL is GPL compatible, so re-licensing as GPL should be no issue ...

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,8 +182,8 @@ used in the creation of [configure](configure) from
 developers. You are free to substitute them with other autoconf macros
 that provide equivalent functionality.
 
-The natural language support (NLS) extension uses gettext, which is
-licensed as GPL. Thus if you compile BOUT++ with NLS, BOUT++ is
-automatically licensed as GPL. Thus if you want to use BOUT++ with
-non-GPL compatible code, make sure to compile without NLS by
-configuring as: `./configure --disable-nls`
+BOUT++ links by default with gettext and fftw, both are licensed as
+GPL. Thus if you compile BOUT++ with ether of them, or any other GPLed
+code, BOUT++ will automatically be licensed as GPL. Thus if you want
+to use BOUT++ with non-GPL compatible code, make sure to compile
+without GPLed code.
--- a/README.md
+++ b/README.md
@@ -181,3 +181,9 @@ used in the creation of [configure](configure) from
 [configure.ac](configure.ac), and are provided as a courtesy to
 developers. You are free to substitute them with other autoconf macros
 that provide equivalent functionality.
+
+The natural language support (NLS) extension uses gettext, which is
+licensed as GPL. Thus if you compile BOUT++ with NLS, BOUT++ is
+automatically licensed as GPL. Thus if you want to use BOUT++ with
+non-GPL compatible code, make sure to compile without NLS by
+configuring as: `./configure --disable-nls`
--- a/README.md
+++ b/README.md
@@ -182,8 +182,7 @@ used in the creation of [configure](configure) from
 developers. You are free to substitute them with other autoconf macros
 that provide equivalent functionality.
 
-BOUT++ links by default with gettext and fftw, both are licensed as
-GPL. Thus if you compile BOUT++ with ether of them, or any other GPLed
-code, BOUT++ will automatically be licensed as GPL. Thus if you want
-to use BOUT++ with GPL non-compatible code, make sure to compile
-without GPLed code.
+BOUT++ links by default with some GPL licensed libraries. Thus if you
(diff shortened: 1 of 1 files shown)


ZedThree:
Note that this only applies to distributing compiled BOUT++ binaries. This isn't a problem if you only distribute source files or don't share your compiled library or physics models with anyone.

Also this applies to FFTW as well, which is GPL.

dschwoerer:
https://www.gnu.org/licenses/gpl-faq.html#MereAggregation

> If the modules are included in the same executable file, they are definitely combined in one program. If modules are designed to run linked together in a shared address space, that almost surely means combining them into one program.

If BOUT++ does not run without fftw (before the recent change) I think it was effectively GPL.

I couldn't find anywhere that the GPL only convers compiled code.
To my understanding that would defeat the purpose of the GPL, as you could ship your non-free code, and link on the users computer with GPL code.
Where did you find that the GPL only covers compiled code?

ZedThree:
I'm not a lawyer, but my understanding is all of this is only important to things that get distributed. Distributing the source code of BOUT++ does not include anything from FFTW or gettext, therefore we're fine licensing BOUT++ as LGPL.

Once you compile libbout++, that may or may not be linking against GPL libraries, and if it is, the compiled code is now GPL. But this is only important if you send the compiled code to someone else.

The licences are purposefully vague about defining things like "program", but given the distinction that they make between "program" and "source code", I take that to mean "derivative work" (when talking about linking) only applies to the compiled program/library.

> To my understanding that would defeat the purpose of the GPL, as you could ship your non-free code, and link on the users computer with GPL code.

Well in that particular case, the user must have a copy of the source code in order to compile it, so I think Stallman would be semi-happy! :smile: I believe in order to actually compile that code though, it would have to be GPL-compatible.

---

I'm not worried about any of this though, because I don't think anyone actually distributes compiled forms of BOUT++!

dschwoerer:
https://www.gnu.org/licenses/gpl-faq.html#IfLibraryIsGPL

Well, after long search, I found the above FAQ:
> If a library is released under the GPL (not the LGPL), does that mean that any software which uses it has to be under the GPL or a GPL-compatible license? (#IfLibraryIsGPL)
>
> Yes, because the program actually links to the library. As such, the terms of the GPL apply to the entire combination. The software modules that link with the library may be under various GPL compatible licenses, but the work as a whole must be licensed under the GPL. See also: What does it mean to say a license is “compatible with the GPL”? 

Thus if you use BOUT++ chances are, it is GPL, thus if you publish a physics model, it needs to be GPL compatible.

dschwoerer:
I forgot to add:
There was even a talk at the talk at the workshop, about binary distribution: docker images.

Besides that packaging BOUT++ and providing binaries seems useful to me, thus I think it is beneficial to make clear that BOUT++ is GPL in that case.

bendudson:
Thanks @dschwoerer It seems that it's ok to license the BOUT++ code itself as LGPL, but we should advise users that if they use FFTW or gettext then the result will be GPL. In principle BOUT++ can be configured without either of these libraries, or a company could use BOUT++ while paying for FFTW with a different license, in which case it would be LGPL.

dschwoerer:
I wrote it more general, as e.g. `bin/bout-pylib-cmd-to-bin` uses readline, which again is GPL.

I was rather thinking about EIRENE - not sure what license that uses, but it seems to be non-free.

bendudson:
Yes, best not to ask what license EIRENE is under. Definitely non-free of some kind, but this is normal for older scientific code: no money is exchanged for it, so most of the time it doesn't matter.

README.md
```diff
@@ -181,3 +181,9 @@ used in the creation of [configure](configure) from
 [configure.ac](configure.ac), and are provided as a courtesy to
 developers. You are free to substitute them with other autoconf macros
 that provide equivalent functionality.
+
+BOUT++ links by default with gettext and fftw, both are licensed as
```
d7919: Could we make this a bit more generic rather than specifying some of the GPL licensed things we can use so that we don't have to remember to update this as and when we add more links to external GPL licensed things? This would help us avoid accidentally misleading by omission.

README.md
```diff
@@ -181,3 +181,9 @@ used in the creation of [configure](configure) from
...
+GPL. Thus if you compile BOUT++ with ether of them, or any other GPLed
```
d7919: `ether` --> `either`

README.md
dschwoerer: I now removed the references to the libraries :+1:

PR #1561, opened 2019-02-08
Hypnotoad: handle case when break of contour is very close to x-point

johnomotani:
This PR addresses a couple of minor bugs with Hypnotoad:
* If the beginning and end of the separatrix contour happened to be right at the primary x-point, Hypnotoad couldn't find two intersections with a loop drawn around the x-point because the separatrix contour was missing a segment that should have been crossing the loop.
  * fixed by passing `info[i].type` (which says whether the contour is open or closed) to `line_crossings`
* Also when the separatrix contour ended at the X-point, Hypnotoad couldn't follow the contour in both directions away from the x-point
  * fixed by wrapping around to the end when the beginning of the contour is reached when constructing `si`
* in `leg_separatrix2` there was an error if no extremum was found on the contour
  * fixed by skipping truncation of `sepri` and `sepzi` in this case

### DIFF
--- a/tools/tokamak_grids/gridgen/leg_separatrix2.pro
+++ b/tools/tokamak_grids/gridgen/leg_separatrix2.pro
@@ -75,9 +75,10 @@ FUNCTION leg_separatrix2, interp_data, R, Z, xpt_ri, xpt_zi, $
       oplot, bndry[0,*], bndry[1,*], color=2, thick=2
     ENDIF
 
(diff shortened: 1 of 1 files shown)


johnomotani:
I couldn't figure out how to get this case to work with the non-orthogonal grid generation. Is anyone interested in helping @bshanahan @JarrodLeddy @nick-walkden?

PR #1562, opened 2019-02-08
Hypnotoad: add checkbox to output metrics for orthogonal coordinates

johnomotani:
Add a checkbox to the 'Output' tab, which if selected outputs metrics for orthogonal coordinates (i.e. using ShiftedMetric), by setting the integrated shear to zero, I=0, when calculating the metric components.

### DIFF
--- a/tools/tokamak_grids/gridgen/hypnotoad.pro
+++ b/tools/tokamak_grids/gridgen/hypnotoad.pro
@@ -445,7 +445,8 @@ PRO event_handler, event
       IF info.rz_grid_valid AND info.flux_mesh_valid THEN BEGIN
         ; Get settings
         settings = {calcp:info.calcp, calcbt:info.calcbt, $
-                    calchthe:info.calchthe, calcjpar:info.calcjpar}
+                    calchthe:info.calchthe, calcjpar:info.calcjpar, $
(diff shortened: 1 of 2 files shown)


johnomotani:
I think grid files should say what type of coordinates they use for g**. I've had a stab at adding a flag to the grid file to specify the type of coordinates: https://github.com/boutproject/BOUT-dev/compare/next...hypnotoad-shiftedmetric-output-coord-system-labelling.

That branch uses 'generic' labels. It adds a `coordinates_type` to the gridfile (as an attribute, because that's how netCDF prefers to deal with strings), which is allowed to take values of `"field_aligned"` (for `ParallelTransformIdentity`, `"orthogonal"` (for `ShiftedMetric`) or `"fci"` (for `FCITransform`). Would it be better to be BOUT-specific and use instead something like `paralleltransform`={`"identity"`,`"shiftedmetric"`,`"fci"`}?

Agree some kind of version numbering would be good. We could:
1. Write the commit hash to the grid file
2. Write a version number for Hypnotoad, like '1.0.1' as an array of ints, and update the version number manually when we update Hypnotoad.

2's easier to check if the version is newer/older than something, but I don't see a way of making it automatic. Any preferences or other options?

PR #1564, opened 2019-02-08
Allow Coordinates on staggered grids to be read from grid files

johnomotani:
If grid files contain fields with names like "`g11_xlow`" for `CELL_XLOW` or "`g11_ylow`" for `CELL_YLOW` then the staggered fields can be loaded from the grid file instead of being interpolated. Adds an extra argument to the staggered `Coordinates` constructor to force them to be calculated by interpolation rather than loaded.

Also adds a `bool` argument to `Coordinates::geometry()` to turn off recalculation of staggered `Coordinates` by interpolation, in case a user needs to, e.g., renormalize the `Coordinates` and will handle all the locations explicitly.

This PR is fairly simple, but should probably wait for #1506 to go in first in case of merge conflicts.


src/mesh/impls/bout/boutmesh.cxx
```diff
@@ -844,13 +844,13 @@ int BoutMesh::load() {
...
     if (xstart >= 2) {
-      getCoordinates(CELL_XLOW);
+      //getCoordinates(CELL_XLOW);
     }
     if (ystart >= 2) {
       getCoordinates(CELL_YLOW);
     }
     if (LocalNz > 3) {
-      getCoordinates(CELL_ZLOW);
+      //getCoordinates(CELL_ZLOW);
```
ZedThree: Why is this commented out?

johnomotani: Sorry I shouldn't have left that in. Was a hack to get around an exception being thrown when interpolating to get the xlow coordinates, which I didn't actually want. Will fix...

src/mesh/coordinates.cxx
```diff
@@ -249,57 +268,165 @@ Coordinates::Coordinates(Mesh *mesh, const CELL_LOC loc, const Coordinates* coor
...
+      break;
+    }
+  case CELL_YLOW: {
+      suffix = "_ylow";
+      break;
+    }
+  case CELL_ZLOW: {
+      // geometrical quantities are Field2D, so CELL_ZLOW version is the same
+      // as CELL_CENTRE
+      suffix = "";
```
ZedThree: Is it worth implementing this now? @d7919 @bshanahan opinions?

src/mesh/coordinates.cxx
```diff
@@ -249,57 +268,165 @@ Coordinates::Coordinates(Mesh *mesh, const CELL_LOC loc, const Coordinates* coor
...
+
+    /// Find covariant metric components
+    // Check if any of the components are present
+    if (mesh->sourceHasVar("g_11"+suffix) or mesh->sourceHasVar("g_22"+suffix) or
+        mesh->sourceHasVar("g_33"+suffix) or mesh->sourceHasVar("g_12"+suffix) or
+        mesh->sourceHasVar("g_13"+suffix) or mesh->sourceHasVar("g_23"+suffix)) {
+      // Check that all components are present
+      if (mesh->sourceHasVar("g_11"+suffix) and mesh->sourceHasVar("g_22"+suffix) and
+          mesh->sourceHasVar("g_33"+suffix) and mesh->sourceHasVar("g_12"+suffix) and
+          mesh->sourceHasVar("g_13"+suffix) and mesh->sourceHasVar("g_23"+suffix)) {
```
ZedThree: I think these two `if` statements could be simplified with `std::any_of` and `std::all_of`:

```cpp
auto metric_components = {"g_11", "g_22", ...};
auto source_has_component = [&suffix] (const std::string& name) {
  return mesh->sourceHasVar(name + suffix);
};
if (std::any_of(begin(metric_components), end(metric_components),
                      source_has_component)) {
  if (std::all_of(begin(metric_components), end(metric_components),
                        source_has_component)) {
```

Ok, still a tiny bit verbose, but the logic is maybe more explicit/readable

src/mesh/coordinates.cxx
```diff
@@ -249,57 +268,165 @@ Coordinates::Coordinates(Mesh *mesh, const CELL_LOC loc, const Coordinates* coor
...
+
+        getAtLoc(mesh, g_11, "g_11", suffix, location);
+        getAtLoc(mesh, g_22, "g_22", suffix, location);
+        getAtLoc(mesh, g_33, "g_33", suffix, location);
+        getAtLoc(mesh, g_12, "g_12", suffix, location);
+        getAtLoc(mesh, g_13, "g_13", suffix, location);
+        getAtLoc(mesh, g_23, "g_23", suffix, location);
```
ZedThree: Plausible `std::for_each` or `std::transform` can simplify this as well.

johnomotani: This is doable with `std::transform`, but it doesn't save any lines of code, and I find the `std::transform` version harder to read (might be just not enough familiarity with the std library though).
Original version:
```
        getAtLoc(mesh, g_11, "g_11", suffix, location);
        getAtLoc(mesh, g_22, "g_22", suffix, location);
        getAtLoc(mesh, g_33, "g_33", suffix, location);
        getAtLoc(mesh, g_12, "g_12", suffix, location);
        getAtLoc(mesh, g_13, "g_13", suffix, location);
        getAtLoc(mesh, g_23, "g_23", suffix, location);
```
New version (with `getAtLoc` modified to return the `Field2D` rather than taking by reference as an argument):
```
        auto covariant_components = {g_11, g_22, g_33, g_12, g_13, g_23}; 
        auto getatloc_component = [&mesh, &suffix, &location] (const std::string name) {
          return getAtLoc(mesh, name, suffix, location);
        }
        std::transform(begin(covariant_component_names),
                       end(covariant_component_names),
                       begin(covariant_components), getatloc_component);
```
It would make more sense to use `std::transform` if we could also use `getatloc_component` for the contravariant components above, but the contravariant components need default values (1.0 for diagonal and 0.0 for off-diagonal components), which would add another layer of complication.

src/mesh/coordinates.cxx
johnomotani: One reason not to implement now is that ideally we would treat 2d and 3d gridfiles differently here (and we don't currently have a flag to switch between the two). We want to read the metrics from the gridfile whenever possible (see #1607), and if we make `suffix` `_zlow` here, then a 2d grid file would have to store an extra set of fields (that are just duplicates of the `CELL_CENTRE` fields) to avoid calling `interpolateAndNeumann` (or `interpolateAndExtrapolate` if #1607 goes in) which would overwrite the boundary cells. The point being that we want `suffix = ""` for 2d grid files (so we're not just wasting disk space) but `suffix = "_zlow"` for 3d grid files. [although this is all academic as I think nobody's using CELL_ZLOW at the moment]

We could add a method to `GridDataSource` like `GridDataSource::isAxisymmetric` in anticipation of 3d metrics. For now it would just `return true`?

johnomotani: Could look something like this: https://github.com/boutproject/BOUT-dev/compare/read-staggered-grids...read-staggered-grids-isaxisymmetric-switch

ZedThree: Ok, I'm happy to leave it as-is for now, and we'll revisit when the 3D metrics stuff goes in.

src/mesh/coordinates.cxx
ZedThree: Fair enough, if it was cleaner with the contravariant components as well, that it would make sense to use it. (There's also the complication of getting the right container to store the metric fields. Most of the standard containers don't store references, so they'd have to be pointers.)

It's always worth checking if there's a [standard algorithm](https://en.cppreference.com/w/cpp/algorithm) that can do the job!

ZedThree:
@johnomotani Please could you investigate getting some tests on this? I think it might now be possible to  do so with unit tests, but an integrated test would be ok if that's not possible.

tests/unit/mesh/data/test_gridfromoptions.cxx
```diff
@@ -11,240 +11,278 @@
...
+        },
+        &mesh_from_options)};
   }
 
   ~GridFromOptionsTest() {
     Options::cleanup();
     output_info.enable();
+    output_progress.enable();
     output_warn.enable();
+    // note GridFromOptions* griddata will be deleted by the ~Mesh() destructor
```
ZedThree: Ah, I was wondering why you made `griddata` a pointer -- I bet that was a fun discovery!

One day, we'll make `Mesh` store a `unique_ptr` rather than a raw pointer

tests/unit/mesh/data/test_gridfromoptions.cxx
```diff
@@ -11,240 +11,278 @@
...
-} // namespace globals
-} // namespace bout
-
 // The unit tests use the global mesh
 using namespace bout::globals;
 
 class GridFromOptionsTest : public FakeMeshFixture {
 public:
-  GridFromOptionsTest() : FakeMeshFixture(), options(), griddata(&options) {
+  GridFromOptionsTest() : FakeMeshFixture(), options(), griddata(nullptr) {
```
ZedThree: This now uses its own `mesh_from_options` -- could remove the inheritance from `FakeMeshFixture`? (would need to be replaced with inheritance from `::testing::Test`)

tests/unit/mesh/data/test_gridfromoptions.cxx
```diff
@@ -11,240 +11,278 @@
...
+    options["g11"] = expected_string + " + 5.";
+    options["g22"] = expected_string + " + 4.";
+    options["g33"] = expected_string + " + 3.";
+    options["g12"] = expected_string + " + 2.";
+    options["g13"] = expected_string + " + 1.";
+    options["g23"] = expected_string;
+
+    mesh_from_options.getCoordinates();
+
+    expected_2d = {makeField<Field2D>(
```
ZedThree: Not a problem, just a comment -- the `{}` around the call to `makeField` aren't needed here, but they don't do anything bad

tests/unit/mesh/data/test_gridfromoptions.cxx
```diff
@@ -254,43 +292,202 @@ TEST_F(GridFromOptionsTest, GetVectorBoutRealZ) {
...
+  EXPECT_TRUE(IsFieldEqual(coords->g13, expected_xlow + 1., "RGN_NOBNDRY", this_tolerance));
+  EXPECT_TRUE(IsFieldEqual(coords->g23, expected_xlow, "RGN_NOBNDRY", this_tolerance));
+}
+
+TEST_F(GridFromOptionsTest, CoordinatesXlowRead) {
+  // *_xlow fields added to options, will be read to initialise Coordinates
+
+  // Note '(9 - x)' here because FakeMesh::GlobalX(int jx) returns jx, not a
+  // global position between 0 and 1 (in grid cells, <0 or >1 in boundaries),
+  // like a Mesh is supposed to.
```
ZedThree: Ah sorry, I just did the simplest thing for `FakeMesh::GlobalX` that allowed me to write tests!

tests/unit/test_extras.hxx
```diff
@@ -169,14 +169,24 @@ public:
...
+  void setGridDataSource(GridDataSource* source_in) {
+    source = source_in;
+  }
+
+  // Use this if the FakeMesh needs x- and y-boundaries
+  void createBoundaries() {
+    addBoundary(new BoundaryRegionXIn("core", ystart, yend, this));
+    addBoundary(new BoundaryRegionXOut("sol", ystart, yend, this));
+    addBoundary(new BoundaryRegionYUp("upper_target", xstart, xend, this));
+    addBoundary(new BoundaryRegionYDown("lower_target", xstart, xend, this));
```
ZedThree: The names might be a misnomer, but I guess that's not really a problem

tests/unit/mesh/data/test_gridfromoptions.cxx
johnomotani: I think it's sensible behaviour for the tests :+1: Just took me a while to figure out what was going on!

PR #1566, opened 2019-02-11
Spanish translation

estaremp:
Translated what was already in the .po file into Spanish. 

Will add more strings, _( ), into the file as I get an idea on how the code is structured ;)

### DIFF
--- a/locale/spa/libbout.po
+++ b/locale/spa/libbout.po
@@ -77,7 +77,7 @@ msgid ""
 "\t -> leg region ny_inner-jyseps2_1-1 (%d-%d-1 = %d) must be a multiple of "
 "MYSUB (%d)\n"
 msgstr ""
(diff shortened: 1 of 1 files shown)


bendudson:
Cool! Thank you very much! The test failures seem to be something to do with Codecov, not the changes made here.

ZedThree:
Coverage tests are failing because the PR is coming from a fork and codecov doesn't know it's associated with the project.

We might need to mark those tests as "allowed to fail" as everything important actually passed.

estaremp:
Mmm...I see, I was wondering if that was the reason. 
I can just do a PR from outside my fork, that's probably the easiest.

ZedThree:
It's not a problem, I can see that all the tests are passing! It's just I need to click one link in order to do so.

bendudson:
Replaced by #1579

PR #1569, opened 2019-02-12
Fix Field3D::swap

dschwoerer:
y*_field == this has a special meaning, that is broken by swap.
Thus we need to manually check for this condition, and changing it to
the new this. Otherwise free is not happy.

Merge until #1345 is in?


johnomotani:
@ZedThree Possibly this patch should go into `master`, and we can leave it to #1345 to fix `next`?

dschwoerer:
there is no swap in master, so no bug ether ...

johnomotani:
Ah, my mistake. Thanks @dschwoerer :+1:

ZedThree:
Fixed in #1345

PR #1570, opened 2019-02-12
Avoid out-of-range access for invalid regions

dschwoerer:


src/sys/bout_types.cxx
```diff
@@ -20,6 +20,7 @@ const std::string& DIFF_METHOD_STRING(DIFF_METHOD location) {
 }
 
 const std::string& REGION_STRING(REGION region) {
+  ASSERT2(region >= 0 && region <= 4);
```
d7919: I think this could go below the `REGIONtoString` declaration below and could then read `ASSERT2(region >=0 && region <= REGIONtoString.size())`. Alternatively we could do `ASSERT2(REGIONtoString.count(region) == 1)` to ask the map if it can find the requested region or not -- this then doesn't assume anything about the `enum` numbering etc.

ZedThree: Cleanest is probably to change the `at` to `find`, and throw `BoutException` if not found

dschwoerer: Is REGION meant to stay? Or will it be deprecated soon?

ZedThree: We're planning on deprecating `REGION` in the next release.

dschwoerer: I have changed to at - in all cases, to be consistent. To avoid code duplication I wrote a template, that has the small disadvantage of not saying what enum we are trying to convert, but I think that should be fine, as the backtrace should make that clear ...

src/sys/bout_types.cxx
```diff
@@ -1,13 +1,22 @@
+#include <bout/assert.hxx>
 #include <bout_types.hxx>
-
 #include <map>
 
+template <typename T>
+const std::string& saveat(const std::map<T, std::string>& mymap, T t) {
+  auto found = mymap.find(t);
```
d7919: I'm not sure I understand the origin of the name `saveat` -- should this be `safeAt` or something else?

It might help to add an `AUTO_TRACE` to the start of this routine (also to the `*_STRING` methods I guess) so it's clearer which enum type we're looking at.

src/sys/bout_types.cxx
```diff
@@ -1,13 +1,22 @@
+#include <bout/assert.hxx>
```
d7919: Should this now be `#include <boutexception.hxx>`?

src/sys/bout_types.cxx
dschwoerer: Yes, thanks, I have fixed the name, and added AUTO_TRACE

src/sys/bout_types.cxx
dschwoerer: Yes, updated, thanks

src/sys/bout_types.cxx
bendudson: Sorry, perhaps `safeAt` rather than `saveAt` since exception safety seems to be the reason for doing this?

dschwoerer: Of course :sweat_smile: - thanks :+1:

PR #1571, opened 2019-02-12
Codecov tweaks

ZedThree:
Use `lcov` to generate a report file and upload that to codecov.io. This makes the reporting a bit more accurate, especially for templates and headers. Compare:

Before: https://codecov.io/gh/boutproject/BOUT-dev/src/0dd8975afd8950ec0e7997102b1a2c5e6471f6b4/include/utils.hxx

After: https://codecov.io/gh/boutproject/BOUT-dev/src/c85466c87b83f20ac25106f629a7c0ec5c4acf64/include/utils.hxx

This is running exactly the same tests, but it actually reports most of `Matrix` getting hit.

### DIFF
--- a/.codacy_coverage.sh
+++ b/.codacy_coverage.sh
@@ -1,6 +1,13 @@
 #First install the reporter tool
 wget -q -O ~/codacy-coverage-reporter-assembly-latest.jar https://oss.sonatype.org/service/local/repositories/releases/content/com/codacy/codacy-coverage-reporter/4.0.1/codacy-coverage-reporter-4.0.1-assembly.jar > /dev/null
 
+# Create all the .gcov files
+# Run gcov in the same directory as the source file for the main
+# library (because we used recursive make)
+find . -type f -name *.gcno -execdir gcov -pb -r {} +
+# but just in the unit tests top-level dir as we ran make from there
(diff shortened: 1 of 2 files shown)


.codacy_coverage.sh
```diff
@@ -1,6 +1,13 @@
 #First install the reporter tool
 wget -q -O ~/codacy-coverage-reporter-assembly-latest.jar https://oss.sonatype.org/service/local/repositories/releases/content/com/codacy/codacy-coverage-reporter/4.0.1/codacy-coverage-reporter-4.0.1-assembly.jar > /dev/null
 
+# Create all the .gcov files
+# Run gcov in the same directory as the source file for the main
+# library (because we used recursive make)
+find . -type f -name *.gcno -execdir gcov -pb -r {} +
```
d7919: Should this be `find ./src` (or some list explicitly excluding the unit test directory)?

ZedThree: It doesn't seem to make a difference in terms of what `.gcov` files get generated. I'll have to wait till codacy actually generates something for this PR to see if it affects that

ZedThree: Ah, I'm not overly concerned about that, more worried about under-counting!

PR #1575, opened 2019-02-13
Hypnotoad: output coordinate system labelling

johnomotani:
Following discussion on #1562 this PR is a proposal for recording the coordinate system used to generate metric components in grid files: the `ParallelTransform` used for a simulation must be compatible with the coordinate system (field-aligned for `ParallelTransformIdentity`, orthogonal for `ShiftedMetric` or FCI-type for `FCITransform`) so it's good to be able to require this.


tools/idllib/file_write_string.pro
```diff
@@ -0,0 +1,36 @@
...
+  CATCH, errcode
+  
+  IF errcode NE 0 THEN BEGIN
+    ; error occurred
+    PRINT, "Error occurred in file_write: "+!ERROR_STATE.MSG
+    PRINT, "=> Couldn't write variable '"+varname+"'"
+    RETURN, 1
+  ENDIF
+
+  IF handle.type EQ 1 THEN BEGIN
```
bendudson: I don't think we support PDB anymore, though this might be useful for older versions of BOUT(++)

src/mesh/parallel/fci.cxx
```diff
@@ -294,6 +294,26 @@ const Field3D FCIMap::integrate(Field3D &f) const {
...
+FCITransform::FCITransform(Mesh &mesh, bool zperiodic)
+    : ParallelTransform(mesh), forward_map(mesh, +1, zperiodic), backward_map(mesh, -1, zperiodic),
+      zperiodic(zperiodic) {
+  // check the coordinate system used for the grid data source
+  checkInputGrid();
+}
+
+void FCITransform::checkInputGrid() {
+  std::string coordinates_type = "";
+  if (mesh.get(coordinates_type, "coordinates_type")) {
```
ZedThree: Can you just reverse this conditional to avoid the empty branch?

johnomotani:
Do we want to do something about storing a version number or git hash while I'm making these changes?

> Agree some kind of version numbering would be good. We could:
>
> 1. Write the commit hash to the grid file
> 2. Write a version number for Hypnotoad, like '1.0.1' as an array of ints, and update the version number manually when we update Hypnotoad.
>
> 2's easier to check if the version is newer/older than something, but I don't see a way of making it automatic. Any preferences or other options?

tools/idllib/file_write_string.pro
johnomotani: I haven't checked if the PDB branch works. The function was copied from `file_write.pro`, which had PDB support, so I just left it in.

ZedThree:
A version number is definitely nicer and easier to compare. I'm not sure what you mean by automatic though?

I'd make that change in another PR at any rate.

PR #1576, opened 2019-02-13
Tidy derivatives

d7919:
* Move registration/declaration of methods into source file -- this is required to prevent problems from including `include/bout/index_derivs.hxx` more than once in the project.
* Provide a routine to setup the default default derivative methods and call from `DerivStore` constructor.


ZedThree:
We could maybe remove `initialDefaultMethods` from `DerivStore::initialise` so we don't need to worry about keeping them in sync. It would change the logic in `initialise` a little.

d7919:
> We could maybe remove `initialDefaultMethods` from `DerivStore::initialise` so we don't need to worry about keeping them in sync. It would change the logic in `initialise` a little.

Yes I thought about that -- the main reason why I didn't is that it makes the output of the default type here slightly messier as we're not certain to have it but I think it probably is the better option. Will push shortly

PR #1577, opened 2019-02-13
Use Xenial on Travis

ZedThree:
- Update Travis distro and various packages
- Build with PETSc and SLEPc in all the jobs -- we now actually test PETSc! :tada: 
- Tweak the test-solver settings for PETSc < 3.7 -- this could go into master too

Still no SUNDIALS on Travis as Xenial only has the serial version :(


d7919:
This seems to be quite a bit slower (~30%) -- do you know what the source of this is?

ZedThree:
This is fixed now and I don't know why. By the time I managed to get enough debugging output printed on Travis, it had mysteriously resolved itself, and now I can't even reproduce it on Travis!

I've also not managed to establish if it is actually slower, or if it's just a contention thing. Certainly the gcc-7 job is sometimes *very* slow, which appears to be because the PPA server is *very* slow.

I might try to push some things to branches off this to see if the errors come back.

ZedThree:
Needs #1587 -- waiting for 4.2.2 release, which will likely be soon given the number of bug fixes that have just gone into master

ZedThree:
I think these jobs are sometimes slower mostly due to apt. We might be able to cache the files though: https://stackoverflow.com/a/47722940/2043465

Will investigate separately

Issue #1572, opened 2019-02-12
master is broken with gcc 8


ZedThree:
I think we just need to provide `swap(Coordinates&, Coordinates&)`. I'm not really sure how this compiled to begin with.

```make
In file included from ../../include/bout/coordinates.hxx:38,
                 from coordinates.cxx:9:
../../include/bout/mesh.hxx: In member function ‘void Mesh::recalculateStaggeredCoordinates()’:
../../include/bout/mesh.hxx:463:77: error: no matching function for call to ‘swap(std::__shared_ptr_access<Coordinates, (__gnu_cxx::_Lock_policy)2, false, false>::element_type&, std::__shared_ptr_access<Coordinates, (__gnu_cxx::_Lock_policy)2, false, false>::element_type&)’
       std::swap(*coords_map[location], *createDefaultCoordinates(location));
                                                                           ^
```

EDIT: Actually we don't even need `swap`, just a copy ctor or `noexcept` move ctor will do

d7919:
I thought as all members have a defined swap this should hiding swap each individual member (like certain default methods), but I may well have misunderstood something

d7919:
Any idea why it works with 7 and older (and clang)?

johnomotani:
This looks like the same issue I had on #1548? That problem seemed to be fixed by moving the implementation of `recalculateStaggeredCoordinates` to the .cxx file. My guess is that then `Coordinates` is a complete type, so the compiler knows it has a copy constructor (the default one) and `std::swap` can work. Don't know why the original version works on older compilers though!

ZedThree:
Yes, I think you're right. This comes about due to the inclusion of `mesh.hxx` in `coordinates.hxx`. A forward declaration of `class Mesh` in `coordinates.hxx` also fixes things.

PR #1581, opened 2019-02-15
More printf

dschwoerer:
We can tell gcc to find #1580 for us :+1: 

It did not compile - but I think that is #1572 

This probably should be made with a nice macro, and go into next?

I already spotted another bug, where the argument was ignored ... so probably should incorporate this one way or the other ...

### DIFF
--- a/include/bout/sys/expressionparser.hxx
+++ b/include/bout/sys/expressionparser.hxx
@@ -194,7 +194,8 @@ private:
 
 class ParseException : public std::exception {
 public:
-  ParseException(const char *, ...);
+  ParseException(const char *, ...)
+    __attribute__ ((format (printf, 2, 3)));
   ~ParseException() override {}
 
   const char *what() const noexcept override;
--- a/include/bout/sys/expressionparser.hxx
+++ b/include/bout/sys/expressionparser.hxx
@@ -41,6 +41,8 @@ class ParseException;
 #include <memory>
 #include <exception>
 
+#include "bout/format.hxx"
+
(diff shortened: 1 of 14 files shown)


include/bout/sys/expressionparser.hxx
```diff
@@ -194,7 +194,8 @@ private:
 
 class ParseException : public std::exception {
 public:
-  ParseException(const char *, ...);
+  ParseException(const char *, ...)
+    __attribute__ ((format (printf, 2, 3)));
```
d7919: Isn't this `GCC` specific so may not work with any other compiler family?

ZedThree: Yes, this is non-portable, although it seems clang and Intel both accept it. It should be wrapped up in a macro.

One day we'll actually move to `{fmt}` and we'll get proper type-checking

dschwoerer: Probably, but if we can fix them at least with one compiler, than this is fine.

Also, gcc is available on most systems.

I tagged it as WIP, so we can guard it/put it in a macro if it causes issues for other compilers.

dschwoerer: thoughts on the macro name?
`_BOUT_FORMAT(i,j)?

ZedThree: We should avoid names with leading underscores followed by capitals. Maybe just `BOUT_FORMAT`, or `BOUT_FORMAT_ARGS`

dschwoerer: is master ok, or should it go into next only?

PR #1586, opened 2019-02-17
Check buffer length

dschwoerer:
Not doing so can trigger assert errors in the stl library.

I am not sure how we should treat a single `-` - so I decided to throw ...

The bug can be observed by compiling with `-D_GLIBCXX_ASSERTIONS`

### DIFF
--- a/src/sys/optionsreader.cxx
+++ b/src/sys/optionsreader.cxx
@@ -94,7 +94,7 @@ void OptionsReader::parseCommandLine(Options* options, int argc, char** argv) {
     if (buffer[0] == '-') {
       buffer = buffer.substr(1); // Remove the first character (-)
       if (buffer.length() == 0) {
-        throw BoutException(_("single `-` found - do not know what to do"));
+        throw BoutException(_("Invalid command line option '-' found - maybe check whitespace?"));
       }
     }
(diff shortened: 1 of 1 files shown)


src/sys/optionsreader.cxx
```diff
@@ -74,46 +74,50 @@ void OptionsReader::write(Options *options, const char *file, ...) {
...
+      continue;
+    }
     // Test if name starts with a '-', and remove if found
-    if (buffer[0] == '-')
-      buffer = buffer.substr(1);  // Remove the first character (-)
-    
+    if (buffer[0] == '-') {
+      buffer = buffer.substr(1); // Remove the first character (-)
+      if (buffer.length() == 0) {
+        throw BoutException(_("single `-` found - do not know what to do"));
```
ZedThree: Maybe "Invalid command line option '-'"? Possibly a reminder to check whitespace too? I guess this is most likely to happen with a typo like `- d`

ZedThree:
Thanks, good spot! I think throwing is the best thing to do here.

dschwoerer:
Thanks, I found it while packaging for fedora, where various hardening flags are set.

I have updated the error message, and it should be good to get merged.

PR #1588, opened 2019-02-19
Provide formatting string

dschwoerer:
This guards against potential issues, if strings contain %...

I think there are still even more issues - I am thinking of enabling `-Werror=format` in travis, to catch these issues. Any objections to enabling that?

### DIFF
--- a/src/fileio/datafile.cxx
+++ b/src/fileio/datafile.cxx
@@ -997,7 +997,7 @@ bool Datafile::write(const char *format, ...) const {
   // Create a new datafile
   Datafile tmp(*this);
 
-  tmp.openw(filename);
+  tmp.openw("%s",filename);
   bool ret = tmp.write();
   tmp.close();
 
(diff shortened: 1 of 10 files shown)


src/fileio/datafile.cxx
```diff
@@ -997,7 +997,7 @@ bool Datafile::write(const char *format, ...) const {
   // Create a new datafile
   Datafile tmp(*this);
 
-  tmp.openw(filename);
+  tmp.openw("%s",filename);
```
bendudson: Is this needed? It seems very odd to have to pass a format string here.

bendudson:
Ah yes, I see that if the string contains % then bad things could happen. On the other hand the syntax here is just a bit awful. Is there a way that we can detect if there are no varargs, and in that case avoid doing any string formatting?

When can we start using {fmt}?

dschwoerer:
Yes, I think it is not nice ether. Automatically detecting isn't trivial, we cannot provide 2 function interfaces, one with varargs, and one without - at least not that I am aware of.

We could provide the function without varargs with a different name, but I am not sure that is a nice solution ether, as we need to do change the interface in several places.

fmt sounds like a better solution :+1:

dschwoerer:
Is anybody working on porting to fmt?

Otherwise it would be nice if this could go in.
Some of the changes are required to build for fedora, as `-Werror=format-security` is part of the standard flags ...

PR #1589, opened 2019-02-19
Timer tests and refactor

ZedThree:
- Add tests for `Timer`
- Refactor `Timer` to store values in map rather than pointers. This removes the need to `delete` the map contents

The tests have failed on Travis before, by a factor of more than 3. We could switch to using one of the `std::chrono` clocks instead of `MPI_Wtime`, but I think all our options are basically implementation defined, and liable to be pretty crap in virtualised environments.

Maybe the tests should just check the result is at least the expected size, with no upper bound?

### DIFF
--- a/src/sys/timer.cxx
+++ b/src/sys/timer.cxx
@@ -3,12 +3,12 @@
 
 using namespace std;
 
-Timer::Timer() : timing{getInfo("")} {
+Timer::Timer() : timing(getInfo("")) {
   timing.started = MPI_Wtime();
   timing.running = true;
 }
 
-Timer::Timer(const std::string& label) : timing{getInfo(label)} {
+Timer::Timer(const std::string& label) : timing(getInfo(label)) {
   timing.started = MPI_Wtime();
   timing.running = true;
 }
--- a/src/sys/timer.cxx
+++ b/src/sys/timer.cxx
@@ -1,18 +1,17 @@
-#include <bout/sys/timer.hxx>
-#include <mpi.h>
+#include "bout/sys/timer.hxx"
 
 Timer::Timer() : timing(getInfo("")) {
-  timing.started = MPI_Wtime();
+  timing.started = clock_type::now();
   timing.running = true;
 }
 
 Timer::Timer(const std::string& label) : timing(getInfo(label)) {
-  timing.started = MPI_Wtime();
+  timing.started = clock_type::now();
   timing.running = true;
 }
 
 Timer::~Timer() {
-  double finished = MPI_Wtime();
+  auto finished = clock_type::now();
   timing.running = false;
   timing.time += finished - timing.started;
 }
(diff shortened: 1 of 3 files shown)


src/sys/timer.cxx
```diff
@@ -1,82 +1,72 @@
-#include <mpi.h>
 #include <bout/sys/timer.hxx>
+#include <mpi.h>
```
d7919: Can we use `boutcomm` here instead (promises a header guard)? It might be nice for `boutcomm` to provide a `time` method rather than calling `MPI_Wtime` directly here as this would allow us to easily switch to a non-mpi based timer should we need to (e.g. if we wanted to support a serial build).

src/sys/timer.cxx
```diff
@@ -1,82 +1,72 @@
...
-  timing->time += finished - timing->started;
+  timing.running = false;
+  timing.time += finished - timing.started;
 }
 
 double Timer::getTime() {
-  if(timing->running)
-    return timing->time + (MPI_Wtime() - timing->started);
-  return timing->time;
+  // If we've got a Timer by value, it must be running
```
d7919: I'm not sure why you've got rid of the running check here but not in the next routine (`resetTime`). If `running = true` is the only possibility then why do we need it? I guess it's not the only possibility but in that case it's not obvious why we can avoid the check here (even with the comment).

d7919: (also could/should this use `return getTime("")`?)

src/sys/timer.cxx
ZedThree: I would not be opposed to just replacing `MPI_Wtime` with either `std::chrono::high_resolution` or `std::chrono::steady_clock`. I don't think there's any advantage to the MPI function over the C++ ones

d7919: I seem to recall previously seeing guidance that `MPI_Wtime` was generally the best choice timer where available but I can't remember why so this may have changed.

d7919: (also http://www.cplusplus.com/forum/general/187899/#msg913531 might be handy -- we're measuring intervals so probably want a steady timer)

src/sys/timer.cxx
ZedThree: Good point. I just spotted I could remove this one due to the coverage not hitting the other branch.

> (also could/should this use `return getTime("")` ?)

Not exactly, but a common function can be pulled out, yes

src/sys/timer.cxx
ZedThree: Thanks! I wasn't sure which to prefer, but that recipe is very nice.

Any objections to replacing `MPI_Wtime` with a C++ steady timer?

bendudson: No not really; it's not critical to the code working correctly, just a useful performance hint.

ZedThree:
I've replaced `MPI_Wtime` with a standard C++ `chrono` clock.

As far as I can tell, `MPI_Wtime` is necessary in C, as the standard clocks are not great, but C++ is likely able to access the same hardware timers as MPI.

PR #1592, opened 2019-02-19
More unit tests for derivatives

ZedThree:
- Add tests for `bout::derivatives::index::DDX` etc.
- Add more tests for `DerivativeStore`
- Fix a small bug when registering methods -- if registration failed,
  the name would still be registered (though the method wouldn't be)
- Rename `DerivativeStore::reset` -> `clear` and add `reset` that
  restores the instance to its initial state (i.e. original defaults)
  
Apologies for the barrage of PRs -- I've split up what was one large
branch into more manageable chunks!


d7919:
Looks good, looks like just a small error in the test case thrown up by travis (but ok in clang)

PR #1595, opened 2019-02-20
Hypnotoad version numbering

johnomotani:
This PR makes Hypnotoad save the BOUT++ version, git hash of BOUT-dev and a new Hypnotoad version number to grid files.

My proposal for Hypnotoad numbering is that this version will be 1.1.1: The major version should increase for substantial changes that change the format of the output grid files; The minor version should increase when new features are added; The patch number will now increase when bugs are fixed or minor tweaks are made.

Version history:
* 1.0.0 - original version of hypnotoad
* 1.1.0 - non-orthogonal grid generation added
* 1.1.1 - Hypnotoad version number added here, and now saved to grid files

The version number is defined/saved in `hypnotoad_version.pro`, which (if this PR is accepted/merged) should be updated in every PR that changes Hypnotoad.

Is this OK, or are there other options we should consider?

### DIFF
--- a/tools/tokamak_grids/gridgen/process_grid.pro
+++ b/tools/tokamak_grids/gridgen/process_grid.pro
@@ -1627,9 +1627,16 @@ retrybetacalc:
   ; still worth saving as a sanity check
   hypnotoad_info = ROUTINE_INFO('hypnotoad', /SOURCE)
   hypnotoad_path = FILE_DIRNAME(hypnotoad_info.path)
-  SPAWN, STRJOIN([hypnotoad_path, PATH_SEP(), '..', PATH_SEP(), '..', PATH_SEP(), '..', PATH_SEP(), 'bin/bout-config --git']), bout_git_hash, EXIT_STATUS=status
+
+  ; BOUT++ git hash
+  ;; old version got git hash from bout-config, but this may be out of date
+  ;SPAWN, STRJOIN([hypnotoad_path, PATH_SEP(), '..', PATH_SEP(), '..', PATH_SEP(), '..', PATH_SEP(), 'bin/bout-config --git']), bout_git_hash, EXIT_STATUS=status
+  ;IF status THEN BEGIN
+  ;  PRINT, "WARNING: Failed to get Git hash for BOUT++, could not run '../../../bin/bout-config --git'. Have you configured BOUT++ successfully?"
+  ; new version gets current git hash directly from git-describe
+  SPAWN, STRJOIN(['cd ',hypnotoad_path, '&& git describe --always --abbrev=0 --dirty --match "NOT A TAG"']), bout_git_hash, EXIT_STATUS=status
   IF status THEN BEGIN
-    PRINT, "WARNING: Failed to get Git hash for BOUT++, could not run '../../../bin/bout-config --git'. Have you configured BOUT++ successfully?"
+    PRINT, 'WARNING: Failed to get Git hash for BOUT++, could not run < git describe --always --abbrev=0 --dirty --match "NOT A TAG" > in Hypnotoads directory.'
   ENDIF ELSE BEGIN
     ; bout_git_hash as returned from SPAWN seems to have some funny character
     ; in, maybe a trailing newline. This character causes an error when trying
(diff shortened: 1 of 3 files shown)


johnomotani:
Following @ZedThree's suggestion, changing the base from `next` to `master` so this can go into 4.2.2.

tools/tokamak_grids/gridgen/process_grid.pro
```diff
@@ -1619,6 +1619,45 @@ retrybetacalc:
...
+    ; to write as a NetCDF attribute. STRJOIN seems to fix this.
+    bout_git_hash = STRJOIN(bout_git_hash, '')
+
+    s = file_write_attribute(handle, "git_hash", bout_git_hash)
+  ENDELSE
+
+  ; BOUT++ version number
+  SPAWN, STRJOIN([hypnotoad_path, PATH_SEP(), '..', PATH_SEP(), '..', PATH_SEP(), '..', PATH_SEP(), 'bin/bout-config --version']), bout_version, EXIT_STATUS=status
+  IF status THEN BEGIN
+    PRINT, "WARNING: Failed to get version number for BOUT++, could not run '../../../bin/bout-config --version'. Have you configured BOUT++ successfully?"
```
ZedThree: Would it be better to try and call `bout-config` without a relative path, and ask the user to check it's in `$PATH`?

tools/tokamak_grids/gridgen/process_grid.pro
```diff
@@ -1619,6 +1619,45 @@ retrybetacalc:
...
+  hypnotoad_info = ROUTINE_INFO('hypnotoad', /SOURCE)
+  hypnotoad_path = FILE_DIRNAME(hypnotoad_info.path)
+
+  ; BOUT++ git hash
+  ;; old version got git hash from bout-config, but this may be out of date
+  ;SPAWN, STRJOIN([hypnotoad_path, PATH_SEP(), '..', PATH_SEP(), '..', PATH_SEP(), '..', PATH_SEP(), 'bin/bout-config --git']), bout_git_hash, EXIT_STATUS=status
+  ;IF status THEN BEGIN
+  ;  PRINT, "WARNING: Failed to get Git hash for BOUT++, could not run '../../../bin/bout-config --git'. Have you configured BOUT++ successfully?"
+  ; new version gets current git hash directly from git-describe
+  SPAWN, STRJOIN(['cd ',hypnotoad_path, '&& git describe --always --abbrev=0 --dirty --match "NOT A TAG"']), bout_git_hash, EXIT_STATUS=status
```
ZedThree: We might need to check it's a git repo first -- BOUT++ might have been installed

tools/tokamak_grids/gridgen/process_grid.pro
```diff
@@ -1619,6 +1619,45 @@ retrybetacalc:
...
+  ; BOUT++ version information: this is set when BOUT++ is configured.
+  ; Hypnotoad doesn't require BOUT++ to have been configured, and IDL code may
+  ; have changed since 'configure' was run, so this is not 100% reliable, but
+  ; still worth saving as a sanity check
+  hypnotoad_info = ROUTINE_INFO('hypnotoad', /SOURCE)
+  hypnotoad_path = FILE_DIRNAME(hypnotoad_info.path)
+
+  ; BOUT++ git hash
+  ;; old version got git hash from bout-config, but this may be out of date
+  ;SPAWN, STRJOIN([hypnotoad_path, PATH_SEP(), '..', PATH_SEP(), '..', PATH_SEP(), '..', PATH_SEP(), 'bin/bout-config --git']), bout_git_hash, EXIT_STATUS=status
```
ZedThree: Commented out code can go :)

tools/tokamak_grids/gridgen/process_grid.pro
johnomotani: The warning needs improving... The path it actually uses is relative to the location of the Hypnotoad that is actually running, which seems sensible to me, unless BOUT++ is installed. If its installed I guess we can assume that bout-config should be in the `$PATH`. I'll try to update to handle:
* normal git repo: get git hash from `git describe`, version from `bout-config` using relative path
* tarball download: get git hash and version from `bout-config` using relative path
* installed as library: get both from `bout-config` in `$PATH`, giving an informative warning if they can't be found.

ZedThree:
LGTM. This is going into master with version number 1.1.1 -- is that correct?

johnomotani:
1.1.1 into `master` is right. I'll bump the version number in `next` at some point after `master` gets merged in.

PR #1597, opened 2019-02-21
FieldFactory and ExpressionParser tidy

ZedThree:
Involves clang-format, so probably best viewed with "ignore whitespace" on.

- Don't return `const` values
- `const`-correctness in `FieldFactory` and `ExpressionParser`
- Use `override`
- Use `= default` instead of empty constructors/destructors
- Use global mesh/options if `nullptr` passed to `FieldFactory` constructor
- Remove some extraneous comments -- the code was clear enough already

### DIFF
--- a/include/bout/sys/expressionparser.hxx
+++ b/include/bout/sys/expressionparser.hxx
@@ -67,7 +67,7 @@ public:
   virtual double generate(double x, double y, double z, double t) = 0;
 
   /// Create a string representation of the generator, for debugging output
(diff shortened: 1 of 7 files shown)


ZedThree:
Removing the `const` from `FieldFactory::create2D` meant it ended up calling the `default`ed `Field2D` move assignment operator, which clobbered `Field2D::deriv`, which in turn broke the solver!

I've removed the `Field2D::operator(Field2D&&)`. I don't think this is a bug so much as something that can be unsafe to use.

PR #1600, opened 2019-02-25
Fix Coordinates initialization at boundaries

johnomotani:
At the moment when metric components are loaded in `Coordinates` boundary cells are set equal to the last grid cell (zero gradient boundary conditions) and they are communicated. This causes two problems:

1. the boundary conditions mean that the gradients of the metric components may be discontinuous at the boundaries. Consequences:
    * `G1`, `G2` and `G3` get large, incorrect values at the last grid cells.
    * interpolated metric components on staggered grids are badly calculated in the last two grid cells. This can be enough to make the determinant of the metric negative, throwing an exception.
2. the communication means that when using `ParallelTransformIdentity` so that integrated shear `I` is non-zero, the guard cells at the poloidal 'boundary' of the core region are wrong. The integrated shear is not continuous across the poloidal boundary, therefore neither are the metric components that depend on it. With these guard cells set by communication, parallel gradients see the discontinuity, so `G1`, `G2`, `G3` as well as the Christoffel terms have large errors.

This PR fixes these problems by extrapolating the metric components into the boundary cells, ensuring their gradients are continuous and smooth. If using `ParallelTransformIdentity` we also extrapolate at the core poloidal boundary.

### DIFF
--- a/src/mesh/coordinates.cxx
+++ b/src/mesh/coordinates.cxx
@@ -228,12 +228,32 @@ namespace {
             )/16.;
         }
 
-        // extrapolate into boundary guard cells
-        for(int i=extrap_start;i<bndry->width;i++) {
-          int xi = bndry->x + i*bndry->bx;
-          int yi = bndry->y + i*bndry->by;
-          result(xi, yi) = 3.0*result(xi - bndry->bx, yi - bndry->by) - 3.0*result(xi - 2*bndry->bx, yi - 2*bndry->by) + result(xi - 3*bndry->bx, yi - 3*bndry->by);
-        }	
+        // set boundary guard cells
+        if ((bndry->bx != 0 && localmesh->GlobalNx-2*bndry->width >= 3) || (bndry->by != 0 && localmesh->GlobalNy-2*bndry->width >= 3)) {
+          if (bndry->bx != 0 && localmesh->LocalNx == 1 && bndry->width == 1) {
+            throw BoutException("Not enough points in the x-direction on this "
+                "processor for extrapolation needed to use staggered grids. "
+                "Increase number of x-guard cells MXG or decrease number of "
+                "processors in the x-direction NXPE.");
+          }
+          if (bndry->by != 0 && localmesh->LocalNy == 1 && bndry->width == 1) {
+            throw BoutException("Not enough points in the y-direction on this "
+                "processor for extrapolation needed to use staggered grids. "
+                "Increase number of y-guard cells MYG or decrease number of "
+                "processors in the y-direction NYPE.");
+          }
+          // extrapolate into boundary guard cells if there are enough grid points
+          for(int i=extrap_start;i<bndry->width;i++) {
+            int xi = bndry->x + i*bndry->bx;
+            int yi = bndry->y + i*bndry->by;
+            result(xi, yi) = 3.0*result(xi - bndry->bx, yi - bndry->by) - 3.0*result(xi - 2*bndry->bx, yi - 2*bndry->by) + result(xi - 3*bndry->bx, yi - 3*bndry->by);
+          }
+        } else {
+          // not enough grid points to extrapolate, set equal to last grid point
+          for(int i=extrap_start;i<bndry->width;i++) {
+            result(bndry->x + i*bndry->bx, bndry->y + i*bndry->by) = result(bndry->x - bndry->bx, bndry->y - bndry->by);
+          }
+        }
       }
     }
 
--- a/src/mesh/coordinates.cxx
+++ b/src/mesh/coordinates.cxx
@@ -194,33 +194,46 @@ Coordinates::Coordinates(Mesh *mesh)
 namespace {
   /// Interpolate a Field2D to a new CELL_LOC with interp_to.
   /// Communicates to set internal guard cells.
-  /// Boundary guard cells are set equal to the nearest grid point (equivalent to
-  /// 2nd order accurate Neumann boundary condition).
+  /// Boundary guard cells are set by extrapolating from the grid, like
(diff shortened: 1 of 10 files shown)


src/mesh/coordinates.cxx
```diff
@@ -46,7 +172,10 @@ Coordinates::Coordinates(Mesh *mesh)
       G1_23(mesh), G2_11(mesh), G2_22(mesh), G2_33(mesh), G2_12(mesh), G2_13(mesh),
       G2_23(mesh), G3_11(mesh), G3_22(mesh), G3_33(mesh), G3_12(mesh), G3_13(mesh),
       G3_23(mesh), G1(mesh), G2(mesh), G3(mesh), ShiftTorsion(mesh),
-      IntShiftTorsion(mesh), localmesh(mesh), location(CELL_CENTRE) {
+      IntShiftTorsion(mesh), localmesh(mesh), options(options_in), location(CELL_CENTRE) {
+
+  std::string paralleltransform = options["paralleltransform"].withDefault(std::string("identity"));
+  bool extrapolate_periodic_boundary = (paralleltransform == "identity");
```
bendudson: Is this extrapolation to avoid communication? If the domain is periodic, surely we don't need to extrapolate?

d7919: I'm a bit nervous of code that has to know what the specific type of an object is -- ideally we'd do the same thing for all types or at least ask the instance if it can do an operation if it only makes sense to do it for certain cases (rather than it's possible for one case but not for another due to a lack of information, but would make sense for both if we had all the information available).

johnomotani: My choice of variable names was bad. I should have called this something like `extrapolate_branch_cut`. The point is that at the poloidal branch cut on closed field lines, with a field-aligned grid the integrated shear `I` isn't continuous, so some of the metric components are also not continuous. But the discontinuity is a branch cut like \theta=2\pi->0 so gradients near the branch cut should not be affected.

If we communicate in the field-aligned case, the guard cells at the branch cut boundary come from the wrong side of the branch cut. I think the most straightforward way to get around this is to extrapolate into the guard cells at the branch cut. It would be more accurate not to extrapolate, but we'd need to save some extra information in the grid file and that doesn't seem worthwhile as if you need accuracy near the branch cut you shouldn't be using the identity transform anyway.

johnomotani: @d7919 I agree this test should be a method of `ParallelTransform` rather than reading the option here. I was trying to avoid the need to add to the interface, but that was the wrong choice here.

src/mesh/parallel/shiftedmetric.cxx
```diff
@@ -34,8 +34,6 @@ ShiftedMetric::ShiftedMetric(Mesh& m) : ParallelTransform(m), zShift(&m) {
         "ShiftedMetric usually requires the option TwistShift=true\n"
         "    Set ShiftWithoutTwist=true to use ShiftedMetric without TwistShift");
   }
-
-  cachePhases();
```
ZedThree: I'm not overly keen on this, as it means calling this constructor results in a partly initialised object, and we should strive to avoid that as much as possible

src/mesh/coordinates.cxx
ZedThree: I'm still a bit confused -- which branch cut are you talking about?

Presumably this only applies in the BOUT++ tokamak coordinates, and not simple geometries like slab?

Would a simpler solution be to just generate quantities like `I` in the guard cells?

johnomotani: It's the branch cut where twist-shift is applied.

Only really applies in tokamak coordinates, although a sheared slab would have a similar issue.

Generating `I` in the guard cells would be a solution, but to do it generally we'd have to change the structure of the grid files, because if there's a private flux region, then in the global, cartesian grid the PFRs are adjacent to the y-boundaries of the core region
(see https://bout-dev.readthedocs.io/en/latest/user_docs/input_grids.html#advanced). So it's not straightforward to add guard cells to the grid input that could store the extra pieces of `I`. [We'd also have to add back some functionality to generate metric components from `I`, `Bxy`, etc., but that shouldn't be a problem.]

I got hit by this because I'm trying to update a test of the differential operators that includes non-trivial metrics and varying grid spacings. I'm using a large-aspect-ratio tokamak-like geometry (from analytic expressions). Possibly I should give up on that and just generate some analytical metrics that are non-constant but periodic to test the field-aligned case, and only test tokamak geometry with ShiftedMetric. In that case I think we should add a warning somewhere that field-aligned simulations in tokamak geometry may have issues, especially as `paralleltransform = identity` is the default.

src/mesh/parallel/shiftedmetric.cxx
johnomotani: yes, good point. I think it's only `zlength` that `ShiftedMetric` needs from `Coordinates`. Would it be better to just pass `zlength` to the constructor?

ZedThree: If that's all that's needed, I think that would be cleaner

src/mesh/coordinates.cxx
ZedThree: I guess the identity transform also has problems with x-derivatives for tokamak geometry as well, so you're likely to run into multiple problems. I'm not sure where the best place for such a warning is -- the manual somewhere?

Providing some functionality to automatically set-up tokamak geometry that defaults the transform to shifted would be nice too.

johnomotani:
I think this PR was trying to do too much. I'll make a new PR for using extrapolation rather than Neumann boundary conditions. The thing about metrics at the branch cut is an issue for another time...

PR #1602, opened 2019-02-26
Initial profile tests and tidy

ZedThree:
I'm not sure why I didn't just do these in the `FieldFactory` tests PR! As `FieldFactory` is now well tested, we just need to test that `options["field"]["function"]` is read correctly, and we can scale the result.

This is a legacy interface and could possibly be deprecated in favour of just using the `FieldFactory`?


bendudson:
Yes, this is a legacy interface which should probably be removed at some point. The solver currently calls these routines to initialise variables, but should just use `FieldFactory` directly.

PR #1607, opened 2019-02-27
Fix Coordinates initialization at boundaries

johnomotani:
At the moment when metric components are loaded in Coordinates boundary cells are set equal to the last grid cell (zero gradient boundary conditions). The gradients of the metric components may be discontinuous at the boundaries. Consequences:
* G1, G2 and G3 get large, incorrect values at the last grid cells.
* interpolated metric components on staggered grids are badly calculated in the last two grid cells. This can be enough to make the determinant of the metric negative, throwing an exception.

This PR extrapolates the metric components into the boundary cells, ensuring their gradients are continuous and smooth.


src/mesh/coordinates.cxx
```diff
@@ -17,6 +17,95 @@
...
+    // would try to create a new Coordinates object since we have not finished
+    // initializing yet, leading to an infinite recursion.
+    // Also, here we interpolate for the boundary points at xstart/ystart and
+    // (xend+1)/(yend+1) instead of extrapolating.
+    for (auto bndry : localmesh->getBoundaries()) {
+      int extrap_start = 0;
+      if ( (location == CELL_XLOW) && (bndry->bx>0) )
+        extrap_start = 1;
+      else if ( (location == CELL_YLOW) && (bndry->by>0) )
+        extrap_start = 1;
```
ZedThree: Please could you use braces on these `if` statements, and then clang-format this function before this goes in?

johnomotani:
Work in progress while I check that metric quantities are converging correctly...

johnomotani:
I found a bug while working on an MMS test for differential operators (PR for that coming soon hopefully). Bug is fixed by edc3ef4.

Point for discussion: this PR makes `Coordinates` extrapolate metric components, etc., into the boundary cells at every boundary, even for the `CELL_CENTRE` coordinates that are loaded directly rather than interpolated. It's like this at the moment because the y-boundaries are not normally generated, and for a double-null grid the y-boundaries can't be loaded from a grid file. So there is a need to extrapolate the y-boundary cells and I've extrapolated the x-boundary ones too to be consistent, although at the moment in `master`/`next` the x-boundary cells are loaded from the grid file*.

It would be better to load all the boundary cells from grid files for the `CELL_CENTRE` `Coordinates` (and maybe for staggered ones where possible, see #1564). This wouldn't be so hard to do, because the global array can be split at the upper target boundary to insert some extra boundary cells.
1. The simplest way to implement this would be to allow `GridFile` to depend on `BoutMesh` so it can get `jyseps2_1`, `jyseps1_2` and `ny_inner` to put the boundary points in the right place.
2. An alternative, which seems nicer from the point of view of the interface structure, would be to add some extra arrays to the grid file, containing the boundary cell values. We'd need to add some new `BoundaryRegion`s to loop over each target individually as the current `BoundaryRegionYUp` and `BoundaryRegionYDown` each include two targets in a double-null case.

I'll leave this as `WIP` while we decide whether to do 1, 2, neither or some other alternative.

PS In case we decide not to do anything fancier, I think it would be better to update so we don't overwrite the x-boundary values when they were loaded from a grid file. It's silly really to add extrapolation where it's not absolutely necessary.

--
\* Actually there is a small potential bug here: `GridFile::get` allows reading from grid files that don't have x-boundary cells as well as ones that do. However, it doesn't initialize the x-boundary cells if they were not loaded, which is done for the y-boundary cells.

johnomotani:
I've implemented a fix for the issue in the previous comment:
* update Hypnotoad to (optionally) save y-boundary cells. New attribute in grid files `y_boundary_guards` records how many have been saved.
  * note if `y_boundary_guards > 0` the y-dimension of the arrays saved in the grid file is `>ny` - it is `ny+2*y_boundary_guards` for a single-null and `ny+4*y_boundary_guards` for a double-null.
* Update `GridFile` to read the y-boundary cells if present
  * Gridfile needs to know where the upper target is for double-null equilibria. I added a member variable `int numberOfXPoints` to `Mesh` so `Gridfile` can check whether the equilibrium is simple, single-null or double-null; could have read `jyseps*` in the `Gridfile` constructor instead (and duplicated some logic from `BoutMesh`), but this seemed simpler. `Gridfile` does read `ny_inner`, and uses it to know where the upper targets should be if the equilibrium is double-null.
  * Also update `Gridfile` to handle grid files that don't have x-boundary cells - previously this was partially implemented, completing it made the code more symmetric between x- and y-directions, which seemed like a good idea.
* Add some methods to `GridDataSource` and `Mesh` to check if x- and y-boundary cells have been read.
* In the `Coordinates` constructor, if the x- and y-boundary cells have been read, then keep them. Only if they were not read (in which case Neumann boundary conditions were used instead), then extrapolate into the boundary cells to ensure the geometry fields are smooth.

Still `WIP` because I need to try generating grids with and without y-boundary cells, loading them into BOUT++ and checking that only the y-boundary cells change.

johnomotani:
The final version of the changes in this PR got a bit long. I've split it into #1646 for the changes to BOUT++ and #1647 for the changes to Hypnotoad.

PR #1614, opened 2019-02-28
Unit tests for GridFromOptions

ZedThree:
Plus some small improvements:

- Consistent warnings when variable isn't in options
- The `GridFromOptions::get` overloads that get vectors now return
  empty vectors, rather than leaving the out parameter untouched,
  in-line with the other overloads


src/mesh/data/gridfromoptions.cxx
```diff
@@ -11,50 +11,54 @@
...
     return false;
   }
-  
-  // Fetch expression as a string 
-  std::string expr;
-  options->get(name, expr, "0");
 
-  // Parse, and evaluate with x,y,z,t = 0
-  std::shared_ptr<FieldGenerator> gen = FieldFactory::get()->parse(expr, options);
+  auto expr = (*options)[name].withDefault(std::string{"0"});
```
bendudson: I think this can now be replaced with `rval = (*options)[name].withDefault(0.0);`
since `Options` now parses strings and generates values with `x=y=z=t=0`.

ZedThree: Ah thanks, I thought there might be a nicer way! I'll change it shortly

PR #1615, opened 2019-03-01
Add FDDX_U2 implementation, plus minor bugfixes

johnomotani:
Second order upwind option for flux derivatives. This makes the available options more consistent with the upwind options.

Also a few very minor bugfixes:
1. Setting the type of flux derivatives was not quite right because of a hard-coded `DERIV::Upwind`
2. Upwind and flux derivatives require the input field `f` to be at the same location as the output, only the velocity can be staggered. Here add an ASSERT1 to ensure this.
3. `Field2D` versions of `VDDZ` and `FDDZ` had a question `// Should we take location from v or f?`. Taking 2 into account, makes most sense to use `outloc` if it's set, and default to `f`'s location otherwise.

### DIFF
--- a/src/mesh/index_derivs.cxx
+++ b/src/mesh/index_derivs.cxx
@@ -274,6 +274,18 @@ REGISTER_FLUX_DERIVATIVE(FDDX_U1, "U1", 1, DERIV::Flux) { // No vec
   return result - std::get<0>(vSplit) * f.c + std::get<1>(vSplit) * f.p;
 }
 
+REGISTER_FLUX_DERIVATIVE(FDDX_U2, "U2", 2, DERIV::Flux) { // No vec
+
+  // Velocity at lower end
+  BoutReal vs = 0.5 * (v.m + v.c);
+  BoutReal result = (vs >= 0.0) ? vs * (1.5*f.m - 0.5*f.mm) : vs * (1.5*f.c - 0.5*f.p);
+  // and at upper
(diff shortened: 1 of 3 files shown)


src/mesh/index_derivs.cxx
```diff
@@ -58,6 +58,7 @@ STAGGER Mesh::getStagger(const CELL_LOC inloc, const CELL_LOC outloc,
 STAGGER Mesh::getStagger(const CELL_LOC vloc, MAYBE_UNUSED(const CELL_LOC inloc),
                          const CELL_LOC outloc, const CELL_LOC allowedStaggerLoc) const {
   TRACE("Mesh::getStagger -- four arguments");
+  ASSERT1(inloc == outloc);
```
d7919: I'm not sure this is needed/right -- `vloc==inloc` in the line below so if `outloc==inloc` this says everything has to be at exactly the same location but I thought we were slightly more flexible than this.

johnomotani: We are more flexible: `vloc==inloc` is only one possibility in the line below. `vloc` is also allowed to be staggered relative to `inloc` and `outloc`, as long as one of `vloc` or `inloc` is CELL_CENTRE. The `vloc==inloc` case lets us also do unstaggered but not-`CELL_CENTRE` cases, e.g. `vloc=inloc=outloc=CELL_XLOW` for a y-derivative.

PR #1618, opened 2019-03-05
Use std::map::find to avoid double lookup in getCoordinates

johnomotani:
Following pattern used elsewhere in BOUT++, in getCoordinates use `std::map::find` instead of `std::map::count`. This allows us to avoid searching coords_map twice.


ZedThree:
Travis failed to pip install netCDF4 but carried on anyway -- we should fail if we can't install any of the python scripts.

ZedThree:
There was a new release of netCDF4 18 hours ago -- I suspect it's that. I'll make a PR pinning the version on Travis

PR #1620, opened 2019-03-05
Fix printf formats in Ncxx4/H5Format::setAttribute(..., BoutReal)

ZedThree:


ZedThree:
Tests failing due to this python netCDF problem -- all the tests that don't need it pass though, so I'm merging now and will check `next` isn't broken afterwards

PR #1621, opened 2019-03-05
Pin pip packages on Travis

ZedThree:
Use "known good" versions. We should periodically review the package versions


bendudson:
Is this something for the release to-do list?

ZedThree:
That is wise, thanks!

PR #1622, opened 2019-03-05
set -e

dschwoerer:
Most of the time we don't want to ignore errors.
This sets the default to not ignore errors - they can be ignored by appending `|| :`

### DIFF
--- a/.pip_install_for_travis.sh
+++ b/.pip_install_for_travis.sh
@@ -1,4 +1,7 @@
 #!/bin/bash
+
+set -e
+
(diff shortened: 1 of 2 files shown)


ZedThree:
Push tests failing due to change in netCDF4 version -- the PR tests should now pass; I'll just wait for them to rerun before merging

ZedThree:
Ah, it appears Travis hasn't picked up the new commit in `next`. It should work though, so I'll merge anyway and fix it if not

PR #1624, opened 2019-03-05
Direction tagging and emptyFrom

johnomotani:
Adds members to `Field` to tag the type of grid directions. In particular, this is useful to tag fields as either `YOrthogonal` (i.e. defined on a grid where y and z directions are orthogonal, as for `FCITransform` and the default case for `ShiftedMetric`) or `YAligned` (i.e. defined on a field-aligned grid where y and z directions are not orthogonal, as for `ParallelTransformIdentity` or fields which have been transformed with `toFieldAligned` by `ShiftedMetric`).

Expands constructors for `Field`, `Field3D`, `Field2D` and `FieldPerp` with extra arguments to set the new members, and also `location`. Also adds a new `T emptyFrom(const T&f)` function to create a new field with meta-data copied from an existing one, and data allocated but not initialized. Using `emptyFrom` means the new long constructors are hardly ever needed explicity. Modifies the 'setter' methods to return a reference to the field, allowing method chaining if something needs to be changed, for example `Field3D result{emptyFrom(f).setLocation(outloc)};`.

There are a fair few files that are only changed by using `emptyFrom` for new fields and `fieldsCompatible` for assert checks. Here's a link to compare up to e195604, before I added most of those: https://github.com/boutproject/BOUT-dev/compare/next...e195604babcf0fac9897ff413e0f611f82e3d34b

Todo list (feel free to add to this):
- [ ] add some explanation about `YOrthogonal`/`YAligned`, `emptyFrom` and `fieldsCompatible` to the manual


include/fieldperp.hxx
```diff
@@ -254,16 +257,13 @@ class FieldPerp : public Field {
   
 private:
   /// The Y index at which this FieldPerp is defined
-  int yindex{-1};
+  int yindex;
```
ZedThree: Why the removal of the default value?

johnomotani: That was a mistake. I'll fix it.

src/mesh/interpolation/bilinear.cxx
```diff
@@ -28,7 +28,11 @@
 #include <vector>
 
 Bilinear::Bilinear(int y_offset, Mesh *mesh)
-  : Interpolation(y_offset, mesh), w0(localmesh), w1(localmesh), w2(localmesh), w3(localmesh) {
+  : Interpolation(y_offset, mesh),
+  w0(localmesh, CELL_CENTRE, DIRECTION::X, DIRECTION::YOrthogonal, DIRECTION::Z),
+  w1(localmesh, CELL_CENTRE, DIRECTION::X, DIRECTION::YOrthogonal, DIRECTION::Z),
+  w2(localmesh, CELL_CENTRE, DIRECTION::X, DIRECTION::YOrthogonal, DIRECTION::Z),
+  w3(localmesh, CELL_CENTRE, DIRECTION::X, DIRECTION::YOrthogonal, DIRECTION::Z) {
```
ZedThree: Do these need explicit values?

src/mesh/interpolation/hermite_spline.cxx
```diff
@@ -28,9 +28,15 @@
...
-      h11_z(localmesh) {
+    : Interpolation(y_offset, mesh),
+      h00_x(localmesh, CELL_CENTRE, DIRECTION::X, DIRECTION::YOrthogonal, DIRECTION::Z),
+      h01_x(localmesh, CELL_CENTRE, DIRECTION::X, DIRECTION::YOrthogonal, DIRECTION::Z),
+      h10_x(localmesh, CELL_CENTRE, DIRECTION::X, DIRECTION::YOrthogonal, DIRECTION::Z),
+      h11_x(localmesh, CELL_CENTRE, DIRECTION::X, DIRECTION::YOrthogonal, DIRECTION::Z),
+      h00_z(localmesh, CELL_CENTRE, DIRECTION::X, DIRECTION::YOrthogonal, DIRECTION::Z),
+      h01_z(localmesh, CELL_CENTRE, DIRECTION::X, DIRECTION::YOrthogonal, DIRECTION::Z),
+      h10_z(localmesh, CELL_CENTRE, DIRECTION::X, DIRECTION::YOrthogonal, DIRECTION::Z),
+      h11_z(localmesh, CELL_CENTRE, DIRECTION::X, DIRECTION::YOrthogonal, DIRECTION::Z) {
```
ZedThree: Similarly, can these just use the defaults? I'm sure there's a reason you've done this, but it's not obvious to me, sorry!

src/mesh/parallel/shiftedmetric.cxx
```diff
@@ -136,18 +136,47 @@ void ShiftedMetric::cachePhases() {
...
-  return shiftZ(f, toAlignedPhs, region);
+  switch (f.getDirectionY()) {
+  case (DIRECTION::YOrthogonal):
+    return shiftZ(f, toAlignedPhs, DIRECTION::YOrthogonal, region);
+  case (DIRECTION::YAligned):
+    // f is already in field-aligned coordinates
+    return f;
+  default:
+#if CHECK > 0
+    throw BoutException("Unrecognized y-direction type for Field3D passed to "
```
ZedThree: Can we just always do this? This is essentially free, so I don't see a downside to removing the `#if CHECK`

src/mesh/parallel/shiftedmetric.cxx
```diff
@@ -136,18 +136,47 @@ void ShiftedMetric::cachePhases() {
...
 const Field3D ShiftedMetric::fromFieldAligned(const Field3D& f, const REGION region) {
-  return shiftZ(f, fromAlignedPhs, region);
+  switch (f.getDirectionY()) {
+  case (DIRECTION::YAligned):
+    return shiftZ(f, fromAlignedPhs, DIRECTION::YOrthogonal, region);
+  case (DIRECTION::YOrthogonal):
+    // f is already in orthogonal coordinates
+    return f;
+  default:
+#if CHECK > 0
```
ZedThree: See above -- this check is free, let's always do it

src/sys/bout_types.cxx
```diff
@@ -73,3 +81,88 @@ const std::string& DERIV_STRING(DERIV deriv) {
...
+bool isZDirectionType(DIRECTION z) {
+  switch (z) {
+  case (DIRECTION::Z):
+    return true;
+  default:
+    return false;
+  }
+}
+
+bool compatibleDirections(DIRECTION d1, DIRECTION d2) {
```
ZedThree: How about the following?

```cpp
return (isXDirectionType(d1) && isXDirectionType(d2))
    || (isYDirectionType(d1) && isYDirectionType(d2))
    || (isZDirectionType(d1) && isZDirectionType(d2))
    || (isSpecialDirectionType(d1) || isSpecialDirectionType(d1));
```

(needs `isSpecialDirectionType` to be implemented)

src/sys/derivs.cxx
```diff
@@ -229,8 +229,8 @@ const Field3D D2DZ2(const Field3D &f, CELL_LOC outloc, const std::string &method
 
 const Field2D D2DZ2(const Field2D &f, CELL_LOC UNUSED(outloc), const std::string &UNUSED(method),
                     REGION UNUSED(region)) {
-  auto tmp = Field2D(0., f.getMesh());
-  tmp.setLocation(f.getLocation());
+  Field2D tmp{emptyFrom(f)};
+  tmp = 0.;
```
ZedThree: Could just do `return (emptyFrom(f) = 0.)`, but it does look a bit weird

tests/unit/mesh/parallel/test_shiftedmetric.cxx
```diff
@@ -37,6 +37,14 @@ class ShiftedMetricTest : public ::testing::Test {
                        {2., 4., 6., 8., 10., 12., 14.},
                        {3., 6., 9., 12., 15., 18., 21.}});
 
+    static_cast<FakeMesh*>(mesh)->setCoordinates(std::make_shared<Coordinates>(
+        mesh, Field2D{1.0}, Field2D{1.0}, BoutReal{1.0}, Field2D{1.0}, Field2D{0.0},
+        Field2D{1.0}, Field2D{1.0}, Field2D{1.0}, Field2D{0.0}, Field2D{0.0},
+        Field2D{0.0}, Field2D{1.0}, Field2D{1.0}, Field2D{1.0}, Field2D{0.0},
+        Field2D{0.0}, Field2D{0.0}, Field2D{0.0}, Field2D{0.0}, false));
+
+    mesh->setParallelTransform(Mesh::PTptr(new ShiftedMetric(*mesh, zShift)));
```
ZedThree: Use `bout::utils::make_unique<ShiftedMetric>(*mesh, zShift)` here

tests/unit/test_extras.hxx
```diff
@@ -149,6 +149,7 @@ public:
     maxregionblocksize = MAXREGIONBLOCKSIZE;
 
     setCoordinates(nullptr);
+    setParallelTransform(PTptr(new ParallelTransformIdentity(*this)));
```
ZedThree: `make_unique`

include/bout/paralleltransform.hxx
```diff
@@ -92,6 +94,11 @@ public:
   bool canToFromFieldAligned() override{
     return true;
   }
+
+  DIRECTION getDefaultYDirectionType() const override {
+    return DIRECTION::YAligned;
```
ZedThree: I'm not sure about this at all. This says that the identity transform is synonymous with a field-aligned grid, and I'm not convinced that's true.

What does saying the identity transform has field-aligned fields by default get us?

ZedThree:
Lots of comments, sorry! This is definitely the direction we want to be going in. There's a few things we should iron out first though.

I'm quite wary of `ParallelTransform::getDefaultYDirectionType`. It adds some complications in some places, and I can't see the benefit. What does it get us, and is that worth the complexity it introduces?

include/bout/empty_from.hxx
```diff
@@ -0,0 +1,21 @@
+#ifndef __NEW_EMPTY_FIELD_H__
+#define __NEW_EMPTY_FIELD_H__
+
+#include "field.hxx"
+#include "fieldperp.hxx"
+
+/// Return an empty shell field of some type derived from Field, with metadata
+/// copied but empty data array
+template<typename T>
+inline T emptyFrom(const T& f) {
```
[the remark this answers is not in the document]
johnomotani: That's probably better. The reason for having a separate file was that I wanted to put the general version and the `FieldPerp` specialization in the same place; I couldn't include `fieldperp.hxx` in `field.hxx` because of the circular dependency, so put them both in a new file. In hindsight it would be better to have `emptyFrom` in `field.hxx` and the `FieldPerp` specialization in `fieldperp.hxx`.

include/field.hxx
```diff
@@ -95,6 +114,11 @@ class Field {
     if (fieldmesh){
       return fieldmesh;
     } else {
+      // Don't set fieldmesh=mesh here, so that fieldmesh==nullptr until
+      // allocate() is called in one of the derived classes. fieldmesh==nullptr
+      // indicates that some initialization that would be done in the
+      // constructor if fieldmesh was a valid Mesh object still needs to be
+      // done.
```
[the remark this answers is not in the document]
johnomotani: I agree this hasn't ended up being a good implementation, but I don't think we can just return `nullptr` here:
* If `fieldmesh==nullptr` then when the field is allocated, we will set `fieldmesh=bout::globals::mesh`, and since `getMesh()` might be called before the field is allocated, we want to return `bout::globals::mesh`.
* The reason we can't set it here is that we use `fieldmesh==nullptr` to indicate that `allocate()` needs to do some additional set-up that couldn't be done in the constructor because we didn't have a valid mesh then: namely setting, if needed for the child class, `nx`/`ny`/`nz` and default values of `xDirectionType`/`yDirectionType`/`zDirectionType`.

Would be nice for this to be able to require fields to have a valid mesh in the constructor, but that would break the interface too much.

Would it be better to add something like `bool fullyInitialised{false}` to `Field` and check/update that in `allocate()`? Then the intention would be clearer from the variable name and we could set `fieldmesh = bout::globals::mesh` in `getMesh()` too.

include/interpolation.hxx
```diff
@@ -62,138 +62,140 @@ const T interp_to(const T& var, CELL_LOC loc, REGION region = RGN_ALL) {
...
-  if ((loc != CELL_CENTRE && loc != CELL_DEFAULT) && (fieldmesh->StaggerGrids == false)) {
+  if ((loc != CELL_CENTRE) && (fieldmesh->StaggerGrids == false)) {
     throw BoutException("Asked to interpolate, but StaggerGrids is disabled!");
   }
 
-  if (fieldmesh->StaggerGrids && (var.getLocation() != loc)) {
+  if (var.getLocation() == loc) {
+    // Nothing to do - just return unchanged
+    // Copying into result to return as returning var may increase the number of
+    // references to the var data whilst returning result doesn't
```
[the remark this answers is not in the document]
johnomotani: This was a suggestion from @d7919 in https://github.com/boutproject/BOUT-dev/pull/921#discussion_r172671112:
> I think we may still want to copy var into result. As it stands returning var may increase the number of references to the var data whilst returning result doesn't. This may or may not be an important difference.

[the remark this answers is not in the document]
d7919: (I may have been very wrong)

src/field/field.cxx
```diff
@@ -32,14 +32,58 @@
...
+             DIRECTION xDirectionType_in, DIRECTION yDirectionType_in,
+             DIRECTION zDirectionType_in)
+    : fieldmesh(localmesh==nullptr ? bout::globals::mesh : localmesh),
+      location(location_in), xDirectionType(xDirectionType_in),
+      yDirectionType(yDirectionType_in), zDirectionType(zDirectionType_in) {
+
+  // Need to check for nullptr again, because the fieldmesh might still be
+  // nullptr if the global mesh hasn't been initialized yet
+  if (fieldmesh != nullptr) {
+    // get Coordinates for our location from fieldmesh
```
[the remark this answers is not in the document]
johnomotani: I think it needed a better comment, like `// sets fieldCoordinates by getting Coordinates for our location from fieldmesh`. Since `getCoordinates()` is for getting `fieldCoordinates` *from* the field, this is (to me) a non-obvious use, in that the fact that `getCoordinates()` *sets* `fieldCoordinates` is a sort of side-effect. I'll update the comment.

johnomotani:
There's a few points that I think tie together:
* I think `ParallelTransformIdentity` means a field-aligned grid/coordinate system, because it says we can do `Grad_par` as a y-derivative without any transformation, and we don't necessarily have to have orthogonal x-z coordinates in a simulation using `ParallelTransformIdentity` (e.g. sheared slab). That's what motivated me to make `YAligned` the default for `ParallelTransformIdentity`.
    * practically, it would be just as good at the moment to have `YOrthogonal` the default for `ParallelTransformIdentity` because it doesn't change the y-derivatives for `ParallelTransformIdentity`. As @ZedThree says, having `YOrthogonal` the default for everything would simplify things a fair bit.
* The reason for having `DIRECTION::Null` and using it for defaults is to support having default direction types that could depend on the type of `ParallelTransform`, `Mesh`, `Coordinates`, etc. I still think this is a potentially useful feature, but it's true we could do without it for the moment.

So at the moment we don't need the feature to set the default `yDirectionType`, we could just use `YOrthogonal`. I guess we're not likely to add a new `ParallelTransform` any time soon, that might require a different default. I can imagine it would be more likely that we could add some functionality that required x-z to be orthogonal (`g13==0`), in which case we might want to set `YAligned` as the default for `ParallelTransformIdentity` so we could add checks assuming that `YOrthogonal` really means x-z are orthogonal.

What do people prefer?

include/field3d.hxx
```diff
@@ -167,7 +168,10 @@ class Field3D : public Field, public FieldData {
    * Note: the global "mesh" can't be passed here because
    * fields may be created before the mesh is.
    */
-  Field3D(Mesh *localmesh = nullptr);
+  Field3D(Mesh *localmesh = nullptr, CELL_LOC location_in=CELL_CENTRE,
+          DIRECTION xDirectionType_in=DIRECTION::Null,
+          DIRECTION yDirectionType_in=DIRECTION::Null,
+          DIRECTION zDirectionType_in=DIRECTION::Null);
```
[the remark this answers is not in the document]
johnomotani: The reason is to ensure that the `fieldsCompatible` check would fail if setNullDirectionTypesToDefault() hasn't been called, which it should have been if the field has been allocated. Using `X`, `Y`, `Z` for defaults this wouldn't be true because `Y` is compatible with `Y`, `YOrthogonal` or `YAligned.

src/field/fieldperp.cxx
```diff
@@ -378,12 +338,11 @@ const FieldPerp sliceXZ(const Field3D& f, int y) {
   // Source field should be valid
   checkData(f);
 
-  FieldPerp result(f.getMesh());
+  FieldPerp result(f.getMesh(), f.getLocation(), y, f.getDirectionX(),
+                   f.getDirectionY(), f.getDirectionZ());
```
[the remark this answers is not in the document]
johnomotani: Here we can't use `emptyFrom` because `f` is a `Field3D` while `result` is a `FieldPerp`. This only happens in a couple of places, so it didn't seem worth adding a template or set of functions to handle it.

src/mesh/interpolation/bilinear.cxx
johnomotani: In the present implementation, we have to set explicit values here, because these fields need to be allocated before the `ParallelTransform` has finished being constructed, so we can't get the default `yDirectionType` from the `ParallelTransform` without creating a circular call. Because this class is only used by `FCITransform` we know what the default should be, so we can set it explicitly.

src/sys/bout_types.cxx
johnomotani: Almost, need to add another condition so that we return `false` if `d1==YAligned` and `d2==YOrthogonal` or vice versa. Still think this looks nicer than the way I have it at the moment.

johnomotani: actually, I'm not so sure about that. I want `YOrthogonal` to be compatible with `YOrthogonal` or `Y` but not `YAligned`, and `YAligned` to be compatible with `YAligned` or `Y` but not `YOrthogonal`. The single conditional gets a bit convoluted. Do you think it's still nicer than the nested switch-case statements?

include/interpolation.hxx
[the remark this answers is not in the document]
ZedThree: I think if we did `result.ensureUnique` before returning it, that would be the case, but also a pessimisation as we would have to copy the actual data. We should be able to safely `return var`.

src/mesh/parallel/shiftedmetric.cxx
```diff
@@ -136,18 +136,41 @@ void ShiftedMetric::cachePhases() {
  * and Y is then field aligned.
  */
 const Field3D ShiftedMetric::toFieldAligned(const Field3D& f, const REGION region) {
-  return shiftZ(f, toAlignedPhs, region);
+  switch (f.getDirectionY()) {
+  case (DIRECTION::YOrthogonal):
+    return shiftZ(f, toAlignedPhs, DIRECTION::YOrthogonal, region);
```
ZedThree: This should use `DIRECTION::YAligned`

ZedThree:
> I think `ParallelTransformIdentity` means a field-aligned grid/coordinate system, because it says we can do Grad_par as a y-derivative without any transformation

But simulations without magnetic fields aren't field-aligned, so I don't think that's quite right. The identity transform just means "don't do a parallel transform for parallel derivatives", which isn't quite the same as saying that the grid is field-aligned.
I'd really rather as much of the code as possible doesn't care about what transform is being used.

A few more points I've been thinking about:
- Is it possible to drop `DIRECTION::YOrthogonal` completely? We put it in originally in order to handle derivatives using the parallel slices slightly differently. As far as I can tell, the only place it's actually _used_ is in `populateStencil`, where it could be replaced with `f.hasYUpDown()`?
- Possibly separately from the above, does it make more sense to move `Special` and `YAligned` out of `DIRECTION` into separate enums? We might then have something like `enum class YDirectionType {Full, Slice, Aligned}` and `enum class ZDirectionType {Full, Average}` (or perhaps combine them into a single enum). It currently feels a bit redundant to say that the x-direction-type is x.

include/bout_types.hxx
```diff
@@ -49,10 +49,12 @@ enum REGION {RGN_ALL, RGN_NOBNDRY, RGN_NOX, RGN_NOY, RGN_NOZ};
 const std::string& REGION_STRING(REGION region);
 
 /// To identify particular directions (in index space)
-enum class DIRECTION { X = 0, Y = 1, Z = 3, YAligned = 4, YOrthogonal = 5 };
+enum class DIRECTION { X = 0, Y = 1, Z = 3, YAligned = 4, YOrthogonal = 5, Special = 6, Null = 7 };
```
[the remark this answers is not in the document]
ZedThree: Also I realise you've only updated this bit, but note that we don't need the explicit initialisers here, we could just have `enum class DIRECTION {X, Y, Z, ...}`

src/mesh/parallel/shiftedmetric.cxx
johnomotani: Good spot! Guess this shows I should add more unit tests...

include/bout_types.hxx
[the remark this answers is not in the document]
johnomotani: Sounds right to me. @d7919 was there a reason for skipping `2` when initializing the values?

d7919:
> > I think `ParallelTransformIdentity` means a field-aligned grid/coordinate system, because it says we can do Grad_par as a y-derivative without any transformation
> 
> But simulations without magnetic fields aren't field-aligned, so I don't think that's quite right. The identity transform just means "don't do a parallel transform for parallel derivatives", which isn't quite the same as saying that the grid is field-aligned.
> I'd really rather as much of the code as possible doesn't care about what transform is being used.
> 
> A few more points I've been thinking about:
> 
> * Is it possible to drop `DIRECTION::YOrthogonal` completely? We put it in originally in order to handle derivatives using the parallel slices slightly differently. As far as I can tell, the only place it's actually _used_ is in `populateStencil`, where it could be replaced with `f.hasYUpDown()`?
> * Possibly separately from the above, does it make more sense to move `Special` and `YAligned` out of `DIRECTION` into separate enums? We might then have something like `enum class YDirectionType {Full, Slice, Aligned}` and `enum class ZDirectionType {Full, Average}` (or perhaps combine them into a single enum). It currently feels a bit redundant to say that the x-direction-type is x.

I think the original idea of not using `hasYupYdown` in populate stencil was to avoid introducing a runtime branch into this code as it's right at the bottom of the nested loops -- using the enum class as a template parameter/known at compile time for each call should allow the branching to be determined at compile time (I think).

include/bout_types.hxx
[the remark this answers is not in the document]
d7919: I don't remember I'm afraid -- I suspect not. Probably I had something like `{X, Y, YOther, Z}` and then wanted to expand to more `Y` types and was easiest to just stick both "special" cases at the end rather than leave `YOther` where it was and add another at the end.

ZedThree:
> I think the original idea of not using `hasYupYdown` in populate stencil was to avoid introducing a runtime branch into this code as it's right at the bottom of the nested loops -- using the enum class as a template parameter/known at compile time for each call should allow the branching to be determined at compile time (I think).

Ah yes, good point! This is why we still need to create separate derivatives for `Y` and `YOrthogonal`. Ok, we probably do need to keep it then.

johnomotani:
> But simulations without magnetic fields aren't field-aligned, so I don't think that's quite right. The identity transform just means "don't do a parallel transform for parallel derivatives", which isn't quite the same as saying that the grid is field-aligned.

I'd not considered simulations with no magnetic field. Even then though, isn't the grid's y-direction in some sense aligned to whatever `Grad_par` is parallel to (if using the identity transform)?

> I'd really rather as much of the code as possible doesn't care about what transform is being used.

Agree with this. Would it be better to say that regardless of which transform we're using fields all start out just in default state (i.e. `yDirectionType = Y`), and then change the `yDirectionType` if we do something special like transforming to field-aligned?

I agree it would make sense to make separate `enum class YDirectionType` and `ZDirectionType`. It would simplify things, and since it makes no sense to have `yDirectionType = DIRECTION::X` the general `DIRECTION` doesn't really fit for what we've used it for here.

johnomotani:
@ZedThree @d7919 do you think it's useful to have a `YDirectionType::Slice`? I guess it would indicate that a `Field3D` has parallel slices, so we could use it to replace `hasYUpDown()`. I don't see an obvious benefit of doing that over leaving `yDirectionType=YDirectionType::Y` and using `hasYUpDown()` as we do at the moment.

johnomotani:
I've had a go at implementing some of what we were just discussing. There are now `YDirectionType` and `ZDirectionType` enum classes, and they get sensible defaults `*DirectionType::Standard`*, independent of the `ParallelTransform`. I also decided to store the direction types in `DirectionTypes` struct, because checking compatibility of directions now needs both direction types: `YDirectionType::Aligned` is compatible with `YDirectionType::Standard` if the `Standard` field has `ZDirectionType::Average`.

@ZedThree I think this addresses most of your remaining comments?

--
\* except `Field2D` which uses `ZDirectionType::Average`

include/bout_types.hxx
```diff
@@ -48,21 +48,49 @@ enum REGION {RGN_ALL, RGN_NOBNDRY, RGN_NOX, RGN_NOY, RGN_NOZ};
...
+enum class DIRECTION { X, Y, Z, YAligned, YOrthogonal };
 
 const std::string& DIRECTION_STRING(DIRECTION direction);
 
-void swap(DIRECTION& first, DIRECTION& second);
+/// Identify kind of a field's y-direction
+/// - Standard is the default for the Mesh/Coordinates/ParallelTransform
+/// - Aligned indicates that the field has been transformed to field-aligned
+///   coordinates
+enum class YDirectionType { Standard, Aligned };
```
ZedThree: Should this also include something like `Slice` for `FieldPerp`?

include/bout_types.hxx
```diff
@@ -48,19 +48,49 @@ enum REGION {RGN_ALL, RGN_NOBNDRY, RGN_NOX, RGN_NOY, RGN_NOZ};
...
+
+/// Container for direction types
+struct DirectionTypes {
+  YDirectionType y;
+  ZDirectionType z;
+};
+
+/// Check whether direction types are compatible, so two fields with attributes
+/// d1 and d2 respectively can be added, subtracted, etc.
+bool compatibleDirections(const DirectionTypes& d1, const DirectionTypes& d2);
```
ZedThree: If I could bikeshed this a little, function names should be verbs where practical. So maybe `areDirectionsCompatible`?

include/field.hxx
```diff
@@ -121,11 +140,65 @@ class Field {
...
+  friend void swap(Field& first, Field& second) noexcept {
+    using std::swap;
+    swap(first.name, second.name);
+    swap(first.fieldmesh, second.fieldmesh);
+    swap(first.fieldCoordinates, second.fieldCoordinates);
+    swap(first.location, second.location);
+    swap(first.directions, second.directions);
+  }
+
+  friend bool fieldsCompatible(const Field& field1, const Field& field2) {
```
ZedThree: Does this need to be a friend? There are getters for the directions, so it would be cleaner to use those

include/field.hxx
```diff
@@ -121,11 +140,65 @@ class Field {
...
+        // The following is a possible alternative to
+        // checking the two meshes and location are the same
+        // It is slightly stricter if we decide that coordinates
+        // could differ in more than just location.
+        field1.getCoordinates() == field2.getCoordinates() &&
+        // In the unit tests fieldCoordinates get set to nullptr, so we still
+        // need to check fieldmesh and location, at least for now
+        field1.getMesh() == field2.getMesh() &&
+        field1.getLocation() == field2.getLocation() &&
+        // Compatible directions
```
ZedThree: I think the code is clear enough you can drop all these comments :)

include/field.hxx
```diff
@@ -121,11 +140,65 @@ class Field {
...
+  void setDirectionZ(ZDirectionType z_type) {
+    directions.z = z_type;
+  }
+
+private:
+  DirectionTypes directions{YDirectionType::Standard, ZDirectionType::Standard};
 };
 
+/// Return an empty shell field of some type derived from Field, with metadata
+/// copied but empty data array
```
ZedThree: Please could you add something explicit here about the field being allocated but uninitialised?

src/field/field3d.cxx
```diff
@@ -72,21 +74,18 @@ Field3D::Field3D(const Field3D& f) : Field(f.fieldmesh), data(f.data) {
...
   nz = fieldmesh->LocalNz;
 
-  location = f.getLocation();
-  fieldCoordinates = nullptr;
-
   *this = f;
 }
 
-Field3D::Field3D(const BoutReal val, Mesh* localmesh) : Field(localmesh) {
+Field3D::Field3D(const BoutReal val, Mesh* localmesh) : Field3D(localmesh) {
```
ZedThree: This now delegates to the `Field3D` ctor that sets `nx` etc. so we can probably remove L92-94 below that also set them

src/field/field3d.cxx
```diff
@@ -143,8 +145,8 @@ void Field3D::splitYupYdown() {
   }
 
   for (int i = 0; i < fieldmesh->ystart; ++i) {
-    yup_fields.emplace_back(fieldmesh);
-    ydown_fields.emplace_back(fieldmesh);
+    yup_fields.emplace_back(emptyFrom(*this));
+    ydown_fields.emplace_back(emptyFrom(*this));
```
ZedThree: Note that this now immediately allocates. I'm not sure we want to do this here as it should be done in the parallel transforms.

Although maybe it doesn't hurt as it shouldn't actually allocate memory in most cases.

include/bout_types.hxx
johnomotani: `FieldPerp` is different depending on whether it's a slice of a standard `Field3D` or a field-aligned one. At the moment it just uses the same `YDirectionTypes` as the `Field3D`. I think that's the simplest way: we could have `Slice` and `SliceAligned` values too and add code to translate/compare those, but I can't think of anywhere we'd need to use or be able to check the difference between `Slice`/`Standard` or `SliceAligned`/`Aligned`. Do you have something in mind?

src/field/field3d.cxx
```diff
@@ -263,13 +233,16 @@ Field3D & Field3D::operator=(const Field3D &rhs) {
...
-  fieldmesh = rhs.fieldmesh;
-  nx = rhs.nx; ny = rhs.ny; nz = rhs.nz; 
-  
-  data = rhs.data;
+  nx = rhs.nx;
+  ny = rhs.ny;
+  nz = rhs.nz;
 
-  setLocation(rhs.location);
+  ASSERT1(fieldsCompatible(*this, rhs));
```
ZedThree: I think this be done immediately after `TRACE` and before `copyFieldMembers`. It seems likely this test will always pass in this location

src/mesh/parallel/fci.cxx
```diff
@@ -73,11 +73,16 @@ FCIMap::FCIMap(Mesh& mesh, int offset_, BoundaryRegionPar* boundary, bool zperio
   auto k_corner = Tensor<int>(map_mesh.LocalNx, map_mesh.LocalNy, map_mesh.LocalNz);
 
   // Index-space coordinates of forward/backward points
-  Field3D xt_prime(&map_mesh), zt_prime(&map_mesh);
+  Field3D xt_prime(&map_mesh);
+  Field3D zt_prime{emptyFrom(xt_prime)};
```
ZedThree: Do we need `emptyFrom` here if we're just going to use `Mesh::get`? Unless that checks that the fields are compatible?

src/sys/bout_types.cxx
```diff
@@ -52,6 +53,36 @@ const std::string& DIRECTION_STRING(DIRECTION direction) {
...
+}
+
+bool compatibleDirections(const DirectionTypes& d1, const DirectionTypes& d2) {
+  if (d1.y == d2.y && d1.z == d2.z) {
+    // direction types are the same, most common case, return immediately
+    return true;
+  }
+
+  if (d2.z == ZDirectionType::Average && d2.y == YDirectionType::Standard
+      && (d1.y == YDirectionType::Standard || d1.y == YDirectionType::Aligned)) {
```
ZedThree: This second part will always be true. Can we just combine these two `if` statements?

tests/unit/test_extras.hxx
```diff
@@ -153,6 +153,7 @@ public:
     maxregionblocksize = MAXREGIONBLOCKSIZE;
 
     setCoordinates(nullptr);
+    setParallelTransform(bout::utils::make_unique<ParallelTransformIdentity>(*this));
```
ZedThree: This is undefined behaviour, btw, as `this` is not a complete `Mesh` object until _after_ the constructor has finished. If we do need to do this, it should be done in `FakeMeshFixture`

ZedThree:
Sorry @johnomotani, more comments!

src/field/field3d.cxx
johnomotani: Still, it wouldn't hurt to just use the full constructor here to avoid the `allocate()`. It might save a bit of memory usage in some cases. I'll update.

ZedThree: I'm not even sure we need the full constructor -- just some kind of object in the vector that we can assign to. The transforms don't (or shouldn't?) update the field, but clobber it completely.

src/field/field3d.cxx
johnomotani: If `copyFieldMembers` is working correctly then this test should indeed always pass. So it's only a sanity check that `copyFieldMembers` and `areFieldsCompatible` haven't somehow become inconsistent. It's probably better to check that in a unit test than here...

src/mesh/parallel/fci.cxx
johnomotani: Ah, this use of `emptyFrom` was only useful when I was calling the constructor for `xt_prime` with more arguments, that then wanted to be copied. It doesn't add anything any more, so will revert.

src/sys/bout_types.cxx
johnomotani: We could at the moment, but I think this way is stricter in case we ever add more `YDirectionType`s. I think actually I should check `d1.z == ZDirectionType::Standard` here as well, to be as strict as possible.

include/bout/fv_ops.hxx
```diff
@@ -194,7 +193,8 @@ namespace FV {
 
     Coordinates *coord = f_in.getCoordinates();
 
-    Field3D result = 0.0;
+    Field3D result{emptyFrom(f_in)};
+    result = 0.0;
```
ZedThree: We actually do this a fair bit, use `emptyFrom` and then set to 0.0. Is it worth having another helper function like `zeroFrom(f)`?

johnomotani: How about another modifier method? So we could write something like `Field3D result{emptyFrom(f_in).fillZeros();}?

johnomotani: ...on the other hand, I think `zeroFrom` would take less code to implement, and require less typing to use. Seems like a good reason to add a new function instead of more methods :+1:

ZedThree:
@johnomotani Please could you also add unit tests for `emptyFrom` and any other new functions/methods? I think you have covered most of them already, but I didn't spot tests for that

include/bout/fv_ops.hxx
ZedThree: Yeah, I think another free function is better than a method. It looks like we currently only need a `zeroFrom(Field)` rather than a more generic `fillFrom(BoutReal, Field)`?

johnomotani: Yep, all the places we use `emptyFrom` we only ever set to zero afterwards, not to any non-zero value.

include/bout/index_derivs_interface.hxx
```diff
@@ -83,8 +83,7 @@ T flowDerivative(const T& vel, const T& f, CELL_LOC outloc, const std::string& m
   const int nPoint = localmesh->getNpoints(direction);
 
   if (nPoint == 1) {
-    T tmp{emptyFrom(f).setLocation(outloc)};
-    tmp = 0.;
+    T tmp{zeroFrom(f).setLocation(outloc)};
```
ZedThree: You could go even further and just `return zeroFrom(f).setLocation(outloc)`

src/sys/derivs.cxx
```diff
@@ -229,8 +229,7 @@ const Field3D D2DZ2(const Field3D &f, CELL_LOC outloc, const std::string &method
 
 const Field2D D2DZ2(const Field2D &f, CELL_LOC UNUSED(outloc), const std::string &UNUSED(method),
                     REGION UNUSED(region)) {
-  Field2D tmp{emptyFrom(f)};
-  tmp = 0.;
+  Field2D tmp{zeroFrom(f)};
```
ZedThree: `return zeroFrom(f);` :)

include/bout_types.hxx
ZedThree: Nothing in particular, just trying to think through everything

include/field.hxx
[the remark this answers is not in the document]
ZedThree: Ok, looks like we need to do some more thinking about what we should/want to be doing here. Definitely future work though

include/bout/fv_ops.hxx
```diff
@@ -344,10 +343,10 @@ namespace FV {
    */
   template<typename CellEdges = MC>
   const Field3D Div_f_v(const Field3D &n_in, const Vector3D &v, bool bndry_flux) {
-    ASSERT2(n_in.getLocation() == v.getLocation());
+    ASSERT1(n_in.getLocation() == v.getLocation());
+    ASSERT1(areFieldsCompatible(n_in, v.x));
```
ZedThree: Do we also need the location check on `v` if we're checking `v.x`?

johnomotani: It does: the location check on `v` makes sure its location isn't `CELL_VSHIFT`, which isn't supported by this function.

ZedThree:
:tada: Thanks @johnomotani ! A really great PR, I think `emptyFrom` especially leads to a lot cleaner code. Sorry for my million comments!

I suspect we need some experience actually using the direction tagging before being completely confident we've got exactly the right design, but I'm sure we're most of the way there already.

johnomotani:
Thanks @ZedThree. All the comments made the final code a lot simpler, more logical and cleaner!

src/physics/smoothing.cxx
```diff
@@ -45,7 +45,7 @@
 const Field3D smooth_x(const Field3D &f) {
   TRACE("smooth_x");
   Mesh *mesh = f.getMesh();
-  Field3D result(mesh);
+  Field3D result{emptyFrom(f)};
   result.allocate();
```
d7919: This allocate should be removed

src/sys/derivs.cxx
```diff
@@ -329,18 +331,19 @@ const Field3D D2DXDZ(const Field3D &f, CELL_LOC outloc, const std::string &metho
...
 }
 
 const Field3D D2DYDZ(const Field3D& f, CELL_LOC outloc,
                      MAYBE_UNUSED(const std::string& method), REGION UNUSED(region)) {
   Coordinates *coords = f.getCoordinates(outloc);
 
-  Field3D result(f.getMesh());
+  Field3D result{emptyFrom(f)};
   ASSERT1(outloc == CELL_DEFAULT || outloc == f.getLocation());
   result.allocate();
```
d7919: Remove please.

src/mesh/difops.cxx
```diff
@@ -482,8 +457,7 @@ const Field2D Div_par_CtoL(const Field2D &var) {
 const Field3D Div_par_CtoL(const Field3D &var) {
   Mesh* mesh = var.getMesh();
 
-  Field3D result(mesh);
-  result.allocate();
+  Field3D result{emptyFrom(var).setLocation(CELL_CENTRE)};
```
d7919: This looks like the wrong location

PR #1626, opened 2019-03-06
Bug in calculation of ShiftAngle (qloop)

bendudson:
When the integral of the field line pitch (qinty, qloop) is calculated
in `my_int_y`, the values of the loop integral were set in all
regions. There are multiple surfaces for each xi index (PF regions,
and core) so this variable would be overwritten multiple times.

In practice it seems that the value it ended up with was the value in
the core, at least for the MAST double null g014220.00200 case I
tried. Hopefully this means that the bug didn't actually have any
effect on the resulting grids.

To try to catch bugs where qloop (->ShiftAngle) is used, the values
are set to NaN if they do not correspond to closed field-line regions.


johnomotani:
I think we should also bump the `patch_number` in `hypnotoad_version.pro` and add a comment what the bug fixed was (or maybe just note the PR #).

ZedThree:
Tests are failing due to the netcdf pip thing. Seeing as this only touches hypnotoad, it should be fine.

I'll get the netcdf pip fix into master now

PR #1630, opened 2019-03-07
Some small improvements to unit tests

ZedThree:
- Add `WithQuietOutput` to disable an output during a scope (e.g. just during a single test)
- Make sure all test fixtures have virtual destructors
- Fix some printf format warnings in unit tests
- Update googletest to latest version and update deprecated macros

### DIFF
--- a/tests/unit/mesh/test_boundary_factory.cxx
+++ b/tests/unit/mesh/test_boundary_factory.cxx
@@ -45,7 +45,7 @@ public:
     region = new BoundaryRegionXIn{"test_region", 0, 1, mesh};
   }
 
-  ~BoundaryFactoryTest() {
+  virtual ~BoundaryFactoryTest() {
     delete mesh;
     mesh = nullptr;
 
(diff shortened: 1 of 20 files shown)


tests/unit/mesh/test_boundary_factory.cxx
```diff
@@ -34,29 +34,22 @@ class TestBoundary : public BoundaryOp {
 class BoundaryFactoryTest : public ::testing::Test {
 public:
   BoundaryFactoryTest() {
-    if (mesh != nullptr) {
-      delete mesh;
-      mesh = nullptr;
-    }
+    delete mesh;
```
bendudson: Do we know it's always safe to delete the mesh here? Probably worth documenting why

ZedThree: Yep -- if it's `nullptr`, it's always safe to delete. Otherwise, it's an actual value and if it's ours, we're good. If it's something we shouldn't delete, then we're in trouble as there's nothing we can do to check that.

bendudson: Ah I hadn't realised `nullptr` was safe to delete. Thanks!

PR #1631, opened 2019-03-07
Write Field directions as attributes in output files

johnomotani:
A few follow-ups to #1624:
* Write out `Field::directions.y` and `Field::directions.z` as attributes of the variable in output files.
* Adds a paragraph to the manual about direction types, `emptyFrom` and `zeroFrom`.
* Update `FieldPerp::setIndex` to return `FieldPerp&`, allowing method chaining.

### DIFF
--- a/include/bout_types.hxx
+++ b/include/bout_types.hxx
@@ -65,12 +65,16 @@ const std::string& DIRECTION_STRING(DIRECTION direction);
 ///   coordinates
 enum class YDirectionType { Standard, Aligned };
 
+const std::string& convertYDirectionTypeToString(YDirectionType d);
+
 /// Identify kind of a field's z-direction
 /// - Standard is the default
 /// - Average indicates that the field represents an average over the
 ///   z-direction, rather than having a particular z-position (i.e. is a
 ///   Field2D)
 enum class ZDirectionType { Standard, Average };
+
+const std::string& convertZDirectionTypeToString(ZDirectionType d);
 
 /// Container for direction types
 struct DirectionTypes {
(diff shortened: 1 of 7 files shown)


include/bout_types.hxx
```diff
@@ -65,13 +65,17 @@ const std::string& DIRECTION_STRING(DIRECTION direction);
 ///   coordinates
 enum class YDirectionType { Standard, Aligned };
 
+const std::string& convertYDirectionTypeToString(YDirectionType d);
```
ZedThree: I know this just following the existing naming convention for enum -> string conversion functions in this header, but we do have `toString(T)` in [utils.hxx](https://github.com/boutproject/BOUT-dev/blob/master/include/utils.hxx#L442). I think it would be nice if we just overloaded `toString` for all the enums.

I'll make a PR now for the others

johnomotani: I'll update this PR to follow the style of your PR when you're done. Trying to dumbly replace `convertYDirectionTypeToString` with `toString` failed because we can't include `utils.hxx` in `bout_types.hxx`...

include/datafile.hxx
```diff
@@ -135,6 +135,11 @@ class Datafile {
   bool write_f2d(const std::string &name, Field2D *f, bool save_repeat);
   bool write_f3d(const std::string &name, Field3D *f, bool save_repeat);
 
+  /// Write out the meta-data of a field as attributes of the variable in
+  /// 'file'.
+  template <typename T>
+  void writeFieldAttributes(const std::string name, T* f);
```
ZedThree: I think this could drop the template and just take `const Field&` (note the call sites would need updating, and you'd need to forward-declare `class Field` at the top of this file).

Also you should take `std::string` by `const&` too.

PR #1635, opened 2019-03-08
Make ParallelTransform location-aware

johnomotani:
Main change:
* Shift ownership of `ParallelTransform` objects from `Mesh` to `Coordinates`
   * means `ParallelTransform`s have an associated `CELL_LOC location`. This is currently only used by `ShiftedMetric`, which is passed a `zShift` interpolated to the correct `location`.

Other `ParallelTransform` updates:
* Move headers of `ParallelTransformIdentity` and `ShiftedMetric` into separate files under `src/mesh/parallel` instead of having them in `include/bout/paralleltransform.hxx`.
* Rename `*YUpDown` methods to `*ParallelSlices` - more natural naming since there are now multiple parallel slices in each direction.
* Make `calcParallelSlices()` a method of `Field3D` instead of calling through `Mesh` or `Coordinates`.
* Make `toFieldAligned` and `fromFieldAligned` free functions instead of calling through `Mesh` or `Coordinates`.
* Correct `zShift` in the guard cells at branch cuts - `ShiftedMetric` works with `TwistShift==false`, avoiding extra FFT interpolations at the branch cut location (cf discussion following https://github.com/boutproject/BOUT-dev/issues/1294#issuecomment-427793452)
    * add methods to `Mesh`/`BoutMesh` to test where branch cut locations are
    * (attempt to) load `ShiftAngle` even when `TwistShift==false`


ZedThree:
Rather than moving the parallel transform stuff out of `Mesh`, please could you just deprecate it and have them call the appropriate functions instead?

tests/unit/field/test_field_factory.cxx
```diff
@@ -24,12 +24,28 @@ template <typename T>
 class FieldFactoryCreationTest : public FakeMeshFixture {
 public:
   FieldFactoryCreationTest() : FakeMeshFixture{}, factory{mesh} {
-    // We need a parallel transform as FieldFactory::create3D wants to
-    // un-field-align the result
-    mesh->setParallelTransform(
+    // We need Coordinates so a parallel transform is available as
+    // FieldFactory::create3D wants to un-field-align the result
+    static_cast<FakeMesh*>(mesh)->setCoordinates(std::make_shared<Coordinates>(
```
ZedThree: It looks like we're remaking the exact same `shared_ptr<Coordinates>` object several times here. I'd probably just make it once and save it as a member of the test fixture, and then pass that to `setCoordinates`

include/bout/mesh.hxx
```diff
@@ -332,6 +330,20 @@ class Mesh {
   /// \param[in] jx   The local (on this processor) index in X
   /// \param[out] ts  The Twist-Shift angle if periodic
   virtual bool periodicY(int jx, BoutReal &ts) const = 0;
+
+  /// Is there a branch cut at this processor's lower boundary?
```
ZedThree: I guess this only applies to the Y-boundaries, right?

include/bout/mesh.hxx
```diff
@@ -332,6 +330,20 @@ class Mesh {
...
+  ///
+  /// @param[in] jx             The local (on this processor) index in X
+  /// @param[out] global_shift  The total zShift for a 2pi poloidal circuit,
+  ///                           non-zero if there is a branch-cut
+  virtual bool hasBranchCutLower(int jx, BoutReal& global_shift) const = 0;
```
ZedThree: I'd prefer if this returned `pair<bool, BoutReal>` rather than have a `BoutReal&` as an out parameter. Probably `optional<BoutReal>` is really the correct type, but that's C++17.

include/bout/paralleltransform.hxx
```diff
@@ -54,185 +51,4 @@ protected:
   Mesh &mesh; ///< The mesh this paralleltransform is part of
 };
 
-
-/*!
- * This class implements the simplest form of ParallelTransform
- * where the domain is a logically rectangular domain, and 
- * yup() and ydown() refer to the same field.
- */
-class ParallelTransformIdentity : public ParallelTransform {
```
ZedThree: I know @d7919 has feelings on this, but I'm not overly keen on moving these types out of the public API and into the private API. From the point of view of the library, it doesn't matter, and we should treat all `ParallelTransform`s the same. However, from the point of view of the user, they are different beasts that may need to be treated differently (e.g. we need zShift to create a `ShiftedMetric`).

bendudson:
Strange... sorry, my first review wasn't showing up so I assumed it was lost. I doubly approve this PR!

include/bout/paralleltransform.hxx
johnomotani: OK. No strong feelings from me about it. I'll put them back in `paralleltransform.hxx`.

include/bout/coordinates.hxx
```diff
@@ -210,6 +231,13 @@ private:
   int nz; // Size of mesh in Z. This is mesh->ngz-1
   Mesh * localmesh;
   CELL_LOC location;
+
+  /// Handles calculation of yup and ydown
+  std::unique_ptr<ParallelTransform> transform{nullptr};
+
+  /// Set the parallel (y) transform from the options file.
+  /// Usual way to create the transform object.
+  void setParallelTransform(Options* options);
```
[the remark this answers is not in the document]
johnomotani: I was thinking it would be the usual way to create a `ParallelTransform`, but only ever be called in the `Coordinates` constructor. I guess I don't expect users to have a need to change the `ParallelTransform` once `Mesh` and `Coordinates` are first created; isn't it better to minimise the public interface? If someone really wants to do this they can use the `setParallelTransform(std::unique_ptr<ParallelTransform> pt)` method. Happy to move to public if you think it's better though, I've not got a strong opinion on it. Could also make it `protected` in case we might at some point have derived classes of `Coordinates` which need access, e.g. https://github.com/boutproject/BOUT-dev/blob/f0d55cd749587143008a770e3db6495416483a38/include/bout/coordinates.hxx#L244-L245

Should I move the method to `public`, move to `protected` or just change the comment?

tests/integrated/test-smooth/test_smooth.cxx
```diff
@@ -17,7 +17,7 @@ int main(int argc, char **argv) {
   Field2D input2d = f.create2D("1 + sin(2*y)");
   Field3D input3d = f.create3D("gauss(x-0.5,0.2)*gauss(y-pi)*sin(3*y - z)");
   
-  mesh->getParallelTransform().calcYUpDown(input3d);
+  input3d.getCoordinates()->getParallelTransform().calcParallelSlices(input3d);
```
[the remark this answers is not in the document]
johnomotani: Oops, we do have that method already, I must have missed updating this call!

src/field/field_factory.cxx
```diff
@@ -222,10 +222,11 @@ Field3D FieldFactory::create3D(FieldGeneratorPtr gen, Mesh* localmesh, CELL_LOC
   }
   };
 
-  if (transform_from_field_aligned and localmesh->canToFromFieldAligned()) {
+  if (transform_from_field_aligned
+        and result.getCoordinates()->getParallelTransform().canToFromFieldAligned()) {
```
ZedThree: Looks like this could do with being a free function as well

ZedThree:
Thanks @johnomotani ! Looks like we end up with a lot cleaner code after this

src/field/field_factory.cxx
johnomotani: Free function would be nicer, but this is the only place it's called in the whole library - is it worth having? In practice `canToFromFieldAligned` is checking whether we're using FCI or not. I guess a user code would either use FCI or not, so wouldn't be likely to need a check like this?

PR #1640, opened 2019-03-14
Options variant update

ZedThree:
The huge number of commits are due to merging in `next` to fix conflicts. Interesting commits start at 1b351f973.

This is going into #1425 as there's a fair few changes. Summary of changes on top of that PR:

- Fix conflicts:
  - Resolve circular dependency by forward-declaring `Options` in datafile.hxx
  - Add new constructor arguments for `Field2D/Field3D` (`DirectionType`)
  - Use `bout::globals::mesh`
  - Use `toString` instead of `CELL_LOC_STRING`
- Avoid narrowing conversion in `Array/Matrix/Tensor` constructors (I used `()` instead of `{}`, but could've also used explicit `static_cast<Array::size_type>`)
- Fix printf format warning
- Use `mesh_staggered` in unit tests
- Replace `_StaticCastOrThrow` with `StaticCastOrThrow::operator()` overloads on `true/false_type` (this was originally to avoid the `_Captial` naming which is reserved, but also happens leads to more concise code)
- Move argument to move constructor
- Use `= default` on empty constructors/destructors
- Clean up some unit tests by silencing uninteresting output and always removing temporary files
- Add mpark.variant as a submodule; install it; remove unneeded files
- Add a getter for `Matrix/Tensor::data` and make the member private again

Tests are currently failing, possibly due to a bug in the version of NetCDF on Travis. I think #1425 wasn't running those unit tests as it was on Trusty which had the legacy C++ interface only, and we're now using Xenial which has the CXX4 API.
It might be possible to use a workaround, will investigate


ZedThree:
Test failure is due to a bug present in the combination of NetCDF C and C++ libraries (4.4.0 and 4.2.0 respectively, fixed in 4.4.1 and/or 4.3.0). I'll try to find a PPA or something to install a newer version on Xenial

bendudson:
Thank you very much @ZedThree  for dealing with this. Sometimes NetCDF does seem more buggy and trouble than it's worth.

ZedThree:
Yes, that was a bit painful to debug and fix!

Issue #1638, opened 2019-03-12
googletest update


johnomotani:
After #1630 went in, the unit tests failed to compile for me because googletest did not automatically update itself to the new version. Doing:
```
git submodule update --init --recursive
```
fixed the problem for me. Is there a nicer solution?

johnomotani:
Checking further, the `rm` and `git checkout` that I had in my original post were not necessary, just `git submodule update --init --recursive` is enough to update googletest. Would be nice if this could be automatic or we could print some kind of helpful warning? Otherwise, this issue gives the workaround for anyone who has this problem.

PS the error messages when trying to compile tests with the old version of google test started like this:
```
field/test_field_factory.cxx:70:17: error: expected constructor, destructor, or type conversion before ‘(’ token
 TYPED_TEST_SUITE(FieldFactoryCreationTest, Fields);
                 ^
In file included from ../../externalpackages/googletest/googletest/include/gtest/gtest.h:66:0,
                 from field/test_field_factory.cxx:1:
field/test_field_factory.cxx:72:1: error: ‘gtest_type_params_FieldFactoryCreationTest_’ was not declared in this scope
 TYPED_TEST(FieldFactoryCreationTest, CreateFromValueGenerator) {
 ^
field/test_field_factory.cxx:72:1: error: template argument 3 is invalid
 TYPED_TEST(FieldFactoryCreationTest, CreateFromValueGenerator) {
```

ZedThree:
I'm not sure how automatic we can make this as it mostly depends on what version of git you have, how you cloned the repo, and your personal git settings.

I think most submodule problems can be solved by running:

    git config --global submodule.recurse true

which then makes the update automatic, but this requires git 2.14.

ZedThree:
This is very likely "can't/won't fix", see: https://stackoverflow.com/a/1899826/2043465
Another answer to the same question: https://stackoverflow.com/a/43854593/2043465

Someone else on a different question suggested making an alias for `git checkout && git submodule update --init` for earlier versions of git.

In order to make this automatic for us, we'd likely need to keep track of the submodule commit id separately from git, and I don't fancy that. We might like to add something to the developer's documentation though.

johnomotani:
:+1: the `git config` option fixes this issue for me. Would be good to add it to the developer documentation.

PR #1643, opened 2019-03-18
Add order-only dependency on gtest sentinel to unit test objects

### LINKED ISSUE #1642: Build failure: missing gtest/gtest.h
A clean build can fail with missing gtest.

I tried to add it as dependency for compilation, but at least the naive approach fails (https://github.com/boutproject/BOUT-dev/pull/1629)

Log:
https://omadesala.physics.dcu.ie/BEST/2019-03-15_02:00/build.log

Relevant part:
```
  Compiling  invert/test_fft.cxx
invert/test_fft.cxx:1:10: fatal error: gtest/gtest.h: No such file or directory
 #include "gtest/gtest.h"
          ^~~~~~~~~~~~~~~
compilation terminated.
make[1]: *** [makefile:102: invert/test_fft.o] Error 1
make: *** [makefile:60: build-check-unit-tests] Error 2the last 5 days.
```

I could reproduce this using docker:
```
 podman run  -it --rm fedora:29 bash
dnf install -y git m4 zlib-devel autoconf autoconf-archive  gettext-devel automake environment-modules netcdf-devel netcdf-cxx4-devel hdf5-devel fftw-devel make python3 python3-numpy python3-Cython python3-netcdf4 python3-scipy blas-devel lapack-devel gcc-c++ python3-jinja2 doxygen python3-sphinx mpich-devel openmpi-devel
git clone https://github.com/boutproject/BOUT-dev.git
cd BOUT-dev/ && source /usr/share/Modules/init/bash && module load mpi/mpich-x86_64
git checkout origin/next && ./configure --enable-shared && make -j 2 build-check
```

I haven't tested further, but it seems to depend on the number of threads used for building, on whether it succeeds or not ...

ZedThree:
Fixes #1642 ~~#1638~~

I've checked this works in various scenarios, and it seems to always do the right thing. The sentinel gets checked before any compilation is done, so this works regardless of the number of make jobs.
Order-only dependency as we don't care about the timestamp, so this will hopefully not cause any weirdness with fresh clones.

### DIFF
--- a/tests/unit/makefile
+++ b/tests/unit/makefile
@@ -43,7 +43,7 @@ clean ::
 
 # Download Google Test
 $(GTEST_SENTINEL) :
-	echo "Downloading Google Test"
+	@echo "Downloading Google Test"
 	git submodule update --init --recursive
 
 # For simplicity and to avoid depending on Google Test's
@@ -78,6 +78,8 @@ $(TARGET): makefile $(BOUT_TOP)/make.config $(OBJ) $(SUB_LIBS) bout_test_main.a
 
 TEST_SOURCES = $(shell find $(BOUT_TEST_DIR) -type f -name "test_*.cxx" 2> /dev/null)
 TEST_OBJECTS = $(TEST_SOURCES:%.cxx=%.o)
+
+$(TEST_OBJECTS): | $(GTEST_SENTINEL)
 
 serial_tests: makefile $(BOUT_TOP)/make.config $(OBJ) $(LIB) checklib \
               $(SUB_LIBS) $(TEST_OBJECTS) bout_test_main.a


johnomotani:
I don't think this fixes #1638. If an old version of googletest exists, then `externalpackages/googletest/README.md` will exist already, so `GTEST_SENTINEL` will not trigger a download of the new version.

To reproduce (using an old enough commit to have a different version of googletest, but still in `externalpackages/googletest`):
```
git checkout 771815cbcae1a06250e5ca48159c5468b02c498f
git submodule update --init --recursive
git checkout gtest-sentinel-on-objects
make check-unit-tests
```
At this point for me the unit tests fail to link, and updating googletest I get
```
$ git submodule update --init --recursive
Submodule path 'externalpackages/googletest': checked out '8b6d3f9c4a774bef3081195d422993323b6bb2e0'
```

PS I have git version 2.7.4

ZedThree:
Right! I'm not sure why I thought that it did, apologies!

PR #1645, opened 2019-03-20
Update examples/staggered_grid for v4.2

johnomotani:
Stricter checking on staggered grid functionality had broken this example.

Thanks to @jmbols for finding this.

Should this go into `master` instead? Not sure if errors in examples count as 'bugs'. Should be trivial to re-base.


ZedThree:
Yeah, probably makes sense to go into master as well.

I wonder if it would be possible to "deprecate" options set in the wrong section? Something like `Options::deprecatedGet` that printed a warning if the value was found in that section?

PR #1647, opened 2019-03-20
Hypnotoad: double precision and  y-boundary guard cells

johnomotani:
Adds the option to write y-boundary guard cells from Hypnotoad. This includes the 'upper' target in double null configurations; the number of y-points in the gridfile may therefore be `ny+2*n_yguards` in single-null or `ny+4*n_yguards` in double null configurations. Requires the updates to `GridFile` in #1646 to read these grid files correctly if `n_yguards != 0`, so the number of y-guards defaults to zero.

Upgrades Hypnotoad to use double-precision arithmetic everywhere. The motivation for this is that the change to number of y-boundary guard cells changes some of the indexing in Hypnotoad, which effectively chages some of the rounding errors. I generated one grid file with a version of Hypnotoad with double-precision but without the y-boundary changes here and another with this version: the relative errors between the two were mostly below `1.e-6`, but `bxcvx` and `ShiftTorsion` were `~3e-6`. That was for an 'orthogonal' grid - field-aligned grids were significantly worse because of the integrated shear appearing in metric components. Comparing the single-precision version in `next` to the double precision (with or without the y-boundary changes) the relative errors are up to `1.8` in the metric components!

Also a few other fixes:
* turn on refinement of the starting contour for non-orthogonal grids in all cases
* pass through the `simple` setting if `create_grid` is re-run
* rename `gen_surface` to `gen_surface_hypnotoad` so it doesn't get masked by the version from `idllib` (only matters because `gen_surface_hypnotoad` needed changing in this PR)
* `H` was calculated as the y-derivative of `qinty`, but `qinty` is a y-integral, so can just set `H` from the integrand
* calculate curvature with `I` instead of `sinty` so that when writing out x-z orthogonal (non-field-aligned) grids the curvature does not include the integrated shear (the option to write x-z orthogonal output sets `I=0`)
* fix a place where `nnpol` was re-calculated wrongly. This did not cause a problem because the value would always be too big, and only had the effect of making a few output arrays too long (but they were just filled with harmless zeros).
* don't smooth `beta` - not sure this is a fix as such, but smoothing propagates differences from the boundary points further into the grid and made it harder to compare grids with/without y-boundary guard cells - hopefully with double precision `beta` won't be too noisy.

### DIFF
--- a/tools/tokamak_grids/gridgen/adjust_jpar.pro
+++ b/tools/tokamak_grids/gridgen/adjust_jpar.pro
@@ -45,9 +45,9 @@ PRO adjust_jpar, grid, smoothp=smoothp, jpar=jpar, noplot=noplot
   ; Matching here rather than outboard produces more realistic results
   ; (current doesn't reverse direction at edge)
   mid_ind = -1
-  status = gen_surface(mesh=data) ; Start generator
+  status = gen_surface_hypnotoad(mesh=data) ; Start generator
   REPEAT BEGIN
-    yi = gen_surface(last=last, xi=xi, period=period)
+    yi = gen_surface_hypnotoad(last=last, xi=xi, period=period)
   
     IF period THEN BEGIN
       mr = MIN(data.rxy[xi, yi], mid_ind)
@@ -89,9 +89,9 @@ PRO adjust_jpar, grid, smoothp=smoothp, jpar=jpar, noplot=noplot
   IF count GT 0 THEN dj[w] = 0.0 ; just zero in this region
 
   jpar = ps
-  status = gen_surface(mesh=data) ; Start generator
+  status = gen_surface_hypnotoad(mesh=data) ; Start generator
   REPEAT BEGIN
-    yi = gen_surface(last=last, xi=xi, period=period)
+    yi = gen_surface_hypnotoad(last=last, xi=xi, period=period)
     
     IF NOT period THEN BEGIN
       ; Due to multi-point differencing, dp/dx can be non-zero outside separatrix
--- a/tools/tokamak_grids/gridgen/adjust_jpar.pro
+++ b/tools/tokamak_grids/gridgen/adjust_jpar.pro
@@ -23,7 +23,7 @@
 
 
 FUNCTION grad_par, var, mesh
-  dtheta = 2.*!PI / FLOAT(TOTAL(mesh.npol))
+  dtheta = 2.D*!DPI / DOUBLE(TOTAL(mesh.npol))
   RETURN, (mesh.Bpxy / (mesh.Bxy * mesh.hthe)) * ddy(var, mesh)*dtheta / mesh.dy
 END
 
@@ -62,7 +62,7 @@ PRO adjust_jpar, grid, smoothp=smoothp, jpar=jpar, noplot=noplot
   
   ; Calculate 2*b0xk dot Grad P
   
-  kp = 2.*data.bxcvx*DDX(data.psixy, data.pressure)
+  kp = 2.D*data.bxcvx*DDX(data.psixy, data.pressure)
   
   ; Calculate B^2 Grad_par(Jpar0)
   
@@ -85,8 +85,8 @@ PRO adjust_jpar, grid, smoothp=smoothp, jpar=jpar, noplot=noplot
   m = MAX(ABS(dj), ind)
   s = SIGN(dj[ind])
 
-  w = WHERE(dj * s LT 0.0, count) ; find where contribution reverses
-  IF count GT 0 THEN dj[w] = 0.0 ; just zero in this region
+  w = WHERE(dj * s LT 0.0D, count) ; find where contribution reverses
+  IF count GT 0 THEN dj[w] = 0.0D ; just zero in this region
 
   jpar = ps
   status = gen_surface_hypnotoad(mesh=data) ; Start generator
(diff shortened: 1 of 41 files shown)


bendudson:
Thanks @johnomotani  this looks good. My one worry is what happens if a new grid with `n_yguards` != 0 is read into an older version of the code. Is there a way to make it fail rather than reading incorrectly? Is there a way to catch this with version number checking? (I guess not for older versions of BOUT++).

johnomotani:
@bendudson I think the `n_yguards != 0` case should be mostly caught even by old versions of BOUT++ (hooray for input checking!):

https://github.com/boutproject/BOUT-dev/blob/1c8d7699c096a98dd28af393bd1ac32e4ffec6af/src/mesh/data/gridfromfile.cxx#L200-L210

Single null cases with 0 or 2 y-boundary guards will read correctly, double null cases with 2 y-boundary guards will throw an exception (because `field_dimensions[1] > GlobalNy`). The only case that would be read incorrectly is a double-null with 1 y-boundary guard cell, which would slip into the first case.

I can't think of a way to check more strictly? I meant to make people think before using grids with `n_yguards > 0` with the comment about backward compatibility here
https://github.com/boutproject/BOUT-dev/blob/3049f4a8e3f3cdfb994ea6a35354728ccfac2b79/tools/tokamak_grids/gridgen/hypnotoad.pro#L1072-L1075

tools/tokamak_grids/gridgen/adjust_jpar.pro
```diff
@@ -113,7 +113,7 @@ PRO adjust_jpar, grid, smoothp=smoothp, jpar=jpar, noplot=noplot
     !P.multi=[0,2,2,0,0]
     SURFACE, data.jpar0, tit="Input Jpar0", chars=2
     SURFACE, jpar, tit="New Jpar0", chars=2
-    PLOT, data.jpar0[0,*], tit="jpar at x=0. Solid=input", yr=[MIN([data.jpar0[0,*],jpar[0,*]]), $
+    PLOT, data.jpar0[0,*], tit="jpar at x=0.D Solid=input", yr=[MIN([data.jpar0[0,*],jpar[0,*]]), $
```
bendudson: This `0.D` probably not needed, though doesn't hurt

johnomotani:
Actually, I just noticed that `BoutMesh` tries to read `MXG` and `MYG` from the grid file before reading from options. So if Hypnotoad wrote `MYG = n_yguards` into the gridfile, then the check at l.200-210 should ensure that an exception is thrown if the grid would not be read correctly.

tools/tokamak_grids/gridgen/create_nonorthogonal.pro
```diff
@@ -1080,13 +1146,13 @@ FUNCTION create_nonorthogonal, F, R, Z, in_settings, critical=critical, $
     start_ri = (SMOOTH([start_ri[(np-s):(np-1)], start_ri, start_ri[0:(s-1)]], s))[s:(np-1+s)]
     start_zi = (SMOOTH([start_zi[(np-s):(np-1)], start_zi, start_zi[0:(s-1)]], s))[s:(np-1+s)]
     
-    FOR i=0, np-1 DO BEGIN
-      follow_gradient_nonorth, interp_data, R, Z, start_ri[i], start_zi[i], start_f, ri1, zi1
-      start_ri[i] = ri1
-      start_zi[i] = zi1
-    ENDFOR
+    ;FOR i=0, np-1 DO BEGIN
```
bendudson: Is this refinement not needed, or does it lead to incorrect results?

tools/tokamak_grids/gridgen/create_nonorthogonal.pro
```diff
@@ -1450,20 +1516,20 @@ FUNCTION create_nonorthogonal, F, R, Z, in_settings, critical=critical, $
       ;;   REPEAT BEGIN
```
bendudson: I guess all this commented out code can be removed?

tools/tokamak_grids/gridgen/leg_separatrix.pro
```diff
@@ -30,22 +30,22 @@ FUNCTION leg_separatrix, dctF, R, Z, xpt_ri, xpt_zi, $
...
+  xz = INTERPOLATE(z,xpt_zi, /DOUBLE)
   
-  drdi = INTERPOLATE(DERIV(R), xpt_ri)
-  dzdi = INTERPOLATE(DERIV(Z), xpt_zi)
+  drdi = INTERPOLATE(DERIV(R), xpt_ri, /DOUBLE)
+  dzdi = INTERPOLATE(DERIV(Z), xpt_zi, /DOUBLE)
 
   ; Use finite-differencing
-  ;di = 2.
+  ;di = 2.DD
```
bendudson: DD? Anyway this commented-out code can go I think.

tools/tokamak_grids/gridgen/create_nonorthogonal.pro
johnomotani: I thought it was not needed, because I switched on refinement in `grid_region_nonorth`. Looking again, they're not quite the same thing, because the refinement I switched on is of the line after interpolating onto the poloidal grid, whereas this was for the line that is input to the interpolation. I guess more refinement should never hurt and might help, so I'll put back this stage of refinement (possibly at the beginning of `grid_region_nonorth` if that reduces code duplication.

johnomotani:
With the last commit I pushed, Hypnotoad writes `MYG` to grid files. I've checked that this makes BOUT-4.2.2 throw an exception where with grid files generated before that commit (no `MYG`) they could be read incorrectly (if a strange value of `MYG` was chosen). So Hypnotoad grid files should now be backwards-safe, although only backwards compatible for single null or if `y_boundary_guards=0`.

johnomotani:
@bendudson: In ca936c6 I turned back on the grid refinement that I commented out before, but moved into `grid_region_nonorth` so it is used in every case. The grids before and after look similar and sensible (tested on a MAST disconnected-double-null case), but the grid points move by up to ~2cm. Large changes in metrics, etc. are mostly just around the X-point where it's not surprising that small differences in position would cause large changes in value. So I think the update is OK, presumably the new grids are slightly 'better' (I wish we had a way to measure what 'better' means...).

PR #1648, opened 2019-03-21
Make transforming Field3D inputs from field aligned optional

johnomotani:
The previous behaviour was to transform 3d input expressions from field-aligned coordinates for compatibility with older versions. Make this optional: can be switched off by setting  `input:transform_from_field_aligned` option to false in `BOUT.inp`.

### DIFF
--- a/src/field/field_factory.cxx
+++ b/src/field/field_factory.cxx
@@ -53,7 +53,7 @@ FieldFactory::FieldFactory(Mesh* localmesh, Options* opt)
   // Set options
   // Note: don't use 'options' here because 'options' is a 'const Options*'
   // pointer, so this would fail if the "input" section is not present.
-  Options nonconst_options{opt == nullptr ? Options::root() : *opt};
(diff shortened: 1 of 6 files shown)


src/field/field_factory.cxx
```diff
@@ -50,6 +50,13 @@ FieldFactory::FieldFactory(Mesh* localmesh, Options* opt)
     : fieldmesh(localmesh == nullptr ? bout::globals::mesh : localmesh),
       options(opt == nullptr ? Options::getRoot() : opt) {
 
+  // Set options
+  // Note: don't use 'options' here because 'options' is a 'const Options*'
+  // pointer, so this would fail if the "input" section is not present.
+  Options nonconst_options{opt == nullptr ? Options::root() : *opt};
```
ZedThree: I think this should be an `Options&` to avoid making a copy.

It's a pity we have to do this, but I can't see a nicer alternative

ZedThree:
Please could you add `WithQuietOutput` to `Vector3DTest::ApplyBoundaryString`?

PR #1650, opened 2019-03-24
Add name argument to Options::isSection to test for presence of subsection

johnomotani:
Also add unit test for `Options::isSection()`

### DIFF
--- a/include/options.hxx
+++ b/include/options.hxx
@@ -545,9 +545,7 @@ public:
   bool isValue() const {
     return is_value;
   }
-  bool isSection() const {
-    return is_section;
-  }
+  bool isSection(const std::string name = "") const;
  private:
   
   /// The source label given to default values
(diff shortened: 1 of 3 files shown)


tests/unit/sys/test_options.cxx
```diff
@@ -32,6 +32,20 @@ TEST_F(OptionsTest, IsSetDefault) {
   ASSERT_FALSE(options.isSet("default_value"));
 }
 
+TEST_F(OptionsTest, IsSection) {
+  Options options;
+
+  // make sure options is initialized as a section
+  options["testkey"] = 1.;
+
+  ASSERT_TRUE(options.isSection(""));
```
bendudson: Probably should also check that calling it without arguments has the same effect.

include/options.hxx
```diff
@@ -545,9 +545,7 @@ public:
   bool isValue() const {
     return is_value;
   }
-  bool isSection() const {
-    return is_section;
-  }
+  bool isSection(const std::string name = "") const;
```
ZedThree: You probably want to take `std::string` by `const&`, unless you're going to take a copy in the function anyway

PR #1653, opened 2019-03-26
Options type and doc attributes

bendudson:
Adds some code to:

1. Record the type of an option when it is used

2. Enable users to record a documentation string as an attribute. This is done using a method `doc`, to allow this to be put inline where the option is used e.g. 

```
n0 = options["n0"].doc("Background density in cubic m").withDefault(1e19);
```
or
```
compressible = options["compressible"]
                       .doc("Compressible ExB term in density equation")
                       .withDefault(false);
```
The `blob2d` example now saves into `BOUT.settings`
```
[model]
b0 = 0.35		# type: BoutReal, doc: Value of magnetic field strength
boussinesq = true		# type: bool, doc: Use Boussinesq approximation in vorticity
compressible = false		# type: bool, doc: Compressible ExB term in density equation
d_n = 1e-6		# type: int, doc: Density diffusion coefficient
d_vort = 1e-6		# type: int, doc: Viscous diffusion coefficient
e = 1.602e-19		# type: BoutReal
```


ZedThree:
Thanks @bendudson, I think this is a pretty useful feature!

Please could you add some tests :)

include/options.hxx
```diff
@@ -388,7 +420,7 @@ public:
                   << ")" << std::endl;
       return def;
     }
-    T val = as<T>();
+    T val = as<T>(bout::utils::meshFromValue(def));
```
ZedThree: It looks like the signature for `as` is still `as<T>(Mesh* UNUSED(mesh) = nullptr)` -- should it now be using `mesh`?

manual/sphinx/user_docs/bout_options.rst
```diff
@@ -619,6 +648,87 @@ has a method::
 
 This is currently quite rudimentary and needs improving.
 
+Reading and writing to NetCDF
+-----------------------------
+
+If NetCDF4 support is enabled, then the ``OptionsNetCDF`` class
+provides an experimental way to read and write options.
```
ZedThree: Should we put it in a namespace like `bout::experimental::`? We could then feel free to change the API without deprecating things first

include/options.hxx
bendudson: The `<Field2D>` and `<Field3D>` specialisations do use `mesh`, but the generic version does not. Really it would be nice to just have an optional `T`, meaning "make one like this": This would make it easier to handle things like staggering location or other field attributes that we might want to pass through.

Is there a nice way to do this? Perhaps something like
```
as<T>(const T& similar_to = {})
```
Can a default initialiser work for arbitrary type?

manual/sphinx/user_docs/bout_options.rst
bendudson: Yes that sounds like a good plan

include/options.hxx
ZedThree: Oops, didn't spot that! Thanks!

Yes, I think you want:

```cpp
as<T>(T&& similar_to = {})
```

This should avoid copies when passing values but also allow rvalues: https://godbolt.org/z/zPczXB

ZedThree: We've also now got `emptyFrom` for fields that might be useful here

bendudson: Great thanks! Any objection to changing this API to the above `T&& similar_to = {}`? I don't think this functionality is used anywhere apart from tests and examples (yet).

ZedThree: Nope, fine by me!

bendudson: Great! I think with that change I could remove the `meshFromValue` hack, and the `emptyFrom` function might come in handy. Thanks!

src/invert/laplacexy/laplacexy.cxx
```diff
@@ -227,21 +226,22 @@ LaplaceXY::LaplaceXY(Mesh *m, Options *opt, const CELL_LOC loc)
...
-    BoutReal rtol, atol, dtol;
-    int maxits; ///< Maximum iterations
-    
-    OPTION(opt, rtol, 1e-5);     // Relative tolerance 
-    OPTION(opt, atol, 1e-10);    // Absolute tolerance
-    OPTION(opt, dtol, 1e3);      // Diverged threshold
-    OPTION(opt, maxits, 100000); // Maximum iterations
-    
+
+    BoutReal rtol = (*opt)["rtol"].doc("Relative tolerance").withDefault(1e-5);
```
ZedThree: Mostly as a _style_ thing, I would go with `const` when we can, as then it's clear this is the only place that the variable is touched -- everywhere else must only be reading it.

I'd also be tempted by `const auto`, as the type is (usually) clear from the right, although this doesn't work so well with `std::string`, unless you explicitly use `withDefault<std::string>("value")`.

Just suggestions for the future, rather than asking you to change all these places again! :smile:

include/options.hxx
bendudson: Quick semi-related question: In a couple of places I now have something like:
```
Field3D result {emptyFrom(similar_to)}; // Make a field with correct metadata
result = scalar_value; // Fill with the scalar value
return result;
```
would it be nicer to have something like
```
return emptyFrom(similar_to).fill(scalar_value);
```
or 
```
return fill(emptyFrom(similar_to), scalar_value);
```
?

johnomotani: We made a `zeroFrom` function as well as `emptyFrom` but decided at the time not to add anything for non-zero values because we didn't have a place where that was needed. Looks like that's changed now! I quite like the idea of having a `.fill(scalar_value)` method, it goes nicely with the method-chaining approach to changing properties (e.g. `emptyFrom(f).setLocation(location)`), and if we add it, we could get rid of `zeroFrom` :+1:

ZedThree: I am conflicted -- the method (first) is definitely easier to read, but the free function is cleaner. `fill` doesn't need to know anything about the field's internals, and is simple enough to implement as a template that works for basically anything.

As a third option, what about something like
```cpp
return filledFrom(similar_to, scalar_value);
```
?

For this particular case, you could even do the (less readable):
```cpp
return emptyFrom(similar_to) = scalar_value;
```
I'm not sure I like that though! :smile:

bendudson: Perhaps `filledFrom` would be good. I am also conflicted about the method vs free function, but this might be the cleanest way.

include/options.hxx
```diff
@@ -662,12 +666,12 @@ template<> void Options::assign<>(Tensor<BoutReal> val, const std::string source
...
-template <> BoutReal Options::as<BoutReal>(Mesh* mesh) const;
-template <> bool Options::as<bool>(Mesh* mesh) const;
-template <> Field2D Options::as<Field2D>(Mesh* mesh) const;
-template <> Field3D Options::as<Field3D>(Mesh* mesh) const;
+template <> std::string Options::as<std::string>(std::string&& similar_to) const;
+template <> int Options::as<int>(int&& similar_to) const;
+template <> BoutReal Options::as<BoutReal>(BoutReal&& similar_to) const;
+template <> bool Options::as<bool>(bool&& similar_to) const;
+template <> Field2D Options::as<Field2D>(Field2D&& similar_to) const;
+template <> Field3D Options::as<Field3D>(Field3D&& similar_to) const;
```
ZedThree: I think I led you astray with the suggestion of `T&&`, as I was thinking you would need to modify `similar_to` inside `as`, for some reason. This should just be `const T&` (or by value for the simple types).

include/options.hxx
```diff
@@ -420,7 +424,7 @@ public:
                   << ")" << std::endl;
       return def;
     }
-    T val = as<T>(bout::utils::meshFromValue(def));
+    T val = as<T>(std::forward<T>(def));
```
ZedThree: `def` is taken by value, so you can just pass it straight to `as`

bendudson: Passing it straight through caused problems when passing in strings; I got compiler errors about converting to `&&`. Possibly converting literal char strings to std::string, then passing. Probably changing to `const &` will fix this.

bendudson: Compiler error
```
../../include/options.hxx:427:7: error: no matching function for call to ‘Options::as<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >(std::__cxx11::basic_string<char>&)’
     T val = as<T>(def);
       ^~~
../../include/options.hxx:359:5: note: candidate: ‘template<class T> T Options::as(T&&) const’
   T as(T&& UNUSED(similar_to) = {}) const {
     ^~
../../include/options.hxx:359:5: note:   template argument deduction/substitution failed:
../../include/options.hxx:427:19: note:   cannot convert ‘def’ (type ‘std::__cxx11::basic_string<char>’) to type ‘std::__cxx11::basic_string<char>&&’
     T val = as<T>(def);
```

include/options.hxx
```diff
@@ -54,6 +56,23 @@ class Options;
 #include <utility>
 #include <cmath>
 
+namespace bout {
+namespace utils {
+
+  /// Get a mesh pointer from the given object.
+  /// In most cases this is nullptr
+  template <typename T>
+  Mesh* meshFromValue(const T& UNUSED(value)) {
```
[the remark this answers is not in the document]
bendudson: Ah no. Sorry I thought I'd deleted that.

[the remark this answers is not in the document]
bendudson: gone now. Thanks!

ZedThree:
Thanks @bendudson! I really like the new `Options` API, it's letting us do some very cool and useful things! :tada:

PR #1655, opened 2019-03-27
Generate FieldPerp arithmetic operators with jinja

johnomotani:
Adds `FieldPerp` operators to `gen_fieldops.py` and `gen_fieldops.jinja`. Also adds unit tests for a few `FieldPerp` arithmetic operators that I think were previously not implemented.

`test-code-style` complained about passing-by-value `const BoutReal` arguments. Previously, the declarations in the headers had no `const` for `BoutReal` arguments, but the implementations in `generated_fieldops.cxx` did have `const`. Don't know if this had any effect? Have now removed the `const` from the implementations as well for consistency.


tests/unit/field/test_fieldperp.cxx
```diff
@@ -825,6 +825,18 @@ TEST_F(FieldPerpTest, AddFieldPerpField2D) {
   EXPECT_TRUE(IsFieldEqual(c, 3.0));
 }
 
+TEST_F(FieldPerpTest, AddField2DFieldPerp) {
```
bendudson: Thanks for this, and all the new tests! It would be good to test the cases with `BoutReal` as one of the arguments, as there is some special handling in these cases (e.g. getting `localmesh`).

dschwoerer:
The const (for value) is an implementation detail, it does not change the interface.
The idea for the interface is that it does not matter to the caller, and thus can (and for consistency should) be dropped.

In the implementation, it however ensures that the value is not accidentally changed - it has the same effect as const for any local variable - it has no impact for the caller.

As this seems confusing - should we drop the no-const-value from the test-code-style?

tests/unit/field/test_fieldperp.cxx
johnomotani: The tests were pretty complete already; the tests for `BoutReal` operators are already there :-) The only ones I had to add were for operators that were previously not implemented.

bendudson: Ah great! Thanks @johnomotani

johnomotani:
Thanks @dschwoerer! Makes sense now. I'll revert the commit removing const-ness from `BoutReal` arguments.

PR #1658, opened 2019-03-31
Counters for timers, makes them correct with multiple Timer objects

johnomotani:
Previously, if multiple Timer objects exist at the same time with the same label, the timings might not be correct. The timer should run as long as any of the objects are in scope. This commit makes this happen by adding a counter to the `timer_info` struct, and only stopping the timer when the counter is 0.

Note the new unit test fails without this bugfix. I had to tighten the tolerances in the `TimerTest unit test a bit to make it fail though`.

I noticed this looking at the `LaplaceNaulin` solver, where `Timer timer("invert")` is created in `LaplaceNaulin` and also in the sub-solver (usually the cyclic solver). The fraction of time spent in Laplacian inversion looked suspiciously low. It increased from ~25% to about ~55% of runtime with these changes, in an electromagnetic case with 2 Laplacian solves per rhs evaluation.

### DIFF
--- a/include/bout/sys/timer.hxx
+++ b/include/bout/sys/timer.hxx
@@ -86,7 +86,7 @@ private:
     seconds time;                   ///< Total time
     bool running;                   ///< Is the timer currently running?
     clock_type::time_point started; ///< Start time
-    unsigned int counter{0};        ///< Number of Timer objects associated with this
+    unsigned int counter;           ///< Number of Timer objects associated with this
                                     ///  timer_info
   };
 
--- a/include/bout/sys/timer.hxx
+++ b/include/bout/sys/timer.hxx
@@ -86,6 +86,8 @@ private:
     seconds time;                   ///< Total time
     bool running;                   ///< Is the timer currently running?
     clock_type::time_point started; ///< Start time
+    uint counter{0};                ///< Number of Timer objects associated with this
+                                    ///  timer_info
   };
 
   /// Store of existing timing info objects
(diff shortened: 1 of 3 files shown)


include/bout/sys/timer.hxx
```diff
@@ -86,6 +86,8 @@ private:
     seconds time;                   ///< Total time
     bool running;                   ///< Is the timer currently running?
     clock_type::time_point started; ///< Start time
+    uint counter{0};                ///< Number of Timer objects associated with this
```
ZedThree: To be really picky, `uint` is not a standard type -- please could you change this to `unsigned int`? :)

johnomotani:
@ZedThree tests are failing because the `timer_info` constructor in `getInfo` only has three arguments
https://github.com/boutproject/BOUT-dev/blob/4341315155720569c8f2de11943480b419366b81/src/sys/timer.cxx#L32-L40
I'd tried to initialize the counter with a default value
https://github.com/boutproject/BOUT-dev/blob/4341315155720569c8f2de11943480b419366b81/include/bout/sys/timer.hxx#L85-L91
I guess the trouble is the mixed methods? Would it be better to give default values to all the fields, or to remove the default value from `counter` and pass it to the constructor instead?

ZedThree:
This turns out to be a problem in C++11 that was fixed in C++14. Giving a member of a struct a default initialisation makes the struct no longer an "aggregate", which means it no longer has the "sensible" constructor that we're using in `Timer::getInfo`. In C++14, this would work as expected.

The easiest solution is to remove the default value for `timer_info::counter` and add it to the constructor in `Timer::getInfo`. Otherwise we need to add default values and the appropriate constructors to `timer_info` (or upgrade to C++14!)

PR #1659, opened 2019-03-31
Under-relaxation for LaplaceNaulin

johnomotani:
Under-relaxation can make convergence of the iteration in LaplaceNaulin more robust and/or faster. This commit introduces under-relaxation:
- if the error increases on an iteration, reduce underrelax_factor by a factor of 0.9 and restart the iteration.
- initial value of underrelax_factor can be set with an option (defaults to 1., corresponding to no under-relaxation).

The mean number of resets with the underrelaxation factor reduced is saved to the dump file so that the user can monitor whether the initial under-relaxation factor should be reduced.

Thanks to Sarah Newton for telling me about under-relaxation in SOLPS's iteration, which inspired this PR!

### DIFF
--- a/src/invert/laplace/impls/naulin/naulin_laplace.cxx
+++ b/src/invert/laplace/impls/naulin/naulin_laplace.cxx
@@ -4,11 +4,14 @@
  * \brief Iterative solver to handle non-constant-in-z coefficients
  *
  * Scheme suggested by Volker Naulin: solve
- * Delp2(phi[i+1]) + DC(A/D)*phi[i+1] = rhs(phi[i]) + DC(A/D)*phi[i]
+ * Delp2(phi[i+1]) + 1/DC(C1*D)*Grad_perp(DC(C2))*Grad_perp(phi[i+1]) + DC(A/D)*phi[i+1]
+ *   = rhs(phi[i]) + 1/DC(C1*D)*Grad_perp(DC(C2))*Grad_perp(phi[i]) + DC(A/D)*phi[i]
  * using standard FFT-based solver, iterating to include other terms by
  * evaluating them on rhs using phi from previous iteration.
- * DC part (i.e. Field2D part) of A/D is kept in the FFT inversion so that all
- * Neumann boundary conditions can be used at least when DC(A/D)!=0.
+ * DC part (i.e. Field2D part) of C1*D, C2 and A/D is kept in the FFT inversion
+ * to improve convergence by including as much as possible in the direct solve
+ * and so that all Neumann boundary conditions can be used at least when
+ * DC(A/D)!=0.
  *
  * CHANGELOG
  * =========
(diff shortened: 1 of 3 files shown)


JosephThomasParker:
Thanks John!  Just so it's said explicitly, under-relaxation corresponds to adding an extra term in the Laplace inversion (line 99 in `naulin_laplace.cxx`):
```
phiNext = inverse_laplace( alpha * bCurr + ( 1 - alpha ) * bPrev )
```
where `alpha` is the under-relaxation factor. This is a bit like weighting-down the current right-hand side (in favour of the previous right-hand side) when you think the last iteration may have gone wrong.  That equation can also be approximated as 
```
Laplace( phiNext ) = bCurr + (1 - alpha) d(bCurr)/d(iteration)
``` 
which may or may not be meaningful. :stuck_out_tongue: 

Do you have a handle on how changing the under-relaxation factor changes things?  As I understand our chat, decreasing alpha decreases the number of iterations required, but decreasing it too far prevents convergence.  However, this code seems to decrease alpha when the last iteration increases the error, which is seems like the opposite of what we want.  Is that right?

Also: over-relaxation. Here `alpha` is asserted to be between 0 and 1.  We could have over-relaxation if alpha > 1.  Any idea if that might be helpful?  Is it just that under-relaxation works for this particular system?

johnomotani:
@JosephThomasParker I don't have a solid basis for this, but my guess is that decreasing `alpha` towards zero makes convergence more robust. Making `alpha` too small means it takes more iterations to converge because you only update the solution by a very small amount each time, but it's going in the right direction (I have no proof that it's actually going in the right direction) so will get there eventually: that's the motivation for always decreasing `alpha` if the iteration starts to diverge.

The case I was testing was STORM with magnetic potential `A_parallel` evolution included. It actually has two Laplacian inversions in, one for electrostatic potential `phi` with the form
```
n*Delp2(phi) + Grad_perp(n)*Grad_perp(phi) = vort   (1)
```
and another for `psi=A_parallel` with the form
```
-Delp2(psi) + beta*n*psi = rhs
```
In both `n` is a `Field3D` so we can't use the direct FFT solvers.
What I observe is that the `phi` solve converges with about 8 iterations without under-relaxation. The `psi` solver diverges with `underrelax_factor=1.`; converges in about 50 iterations for `underrelax_factor=0.9`; 17 iterations for `underrelax_factor=0.7`; 25 iterations for `underrelax_factor=0.5`.

I haven't tried over-relaxation. Slide 16 of this http://www.bakker.org/dartmouth06/engs150/05-solv.ppt from this forum thread https://www.cfd-online.com/Forums/fluent/44367-what-under-relaxation-factor.html says:
> Here U is the relaxation factor:
> * U < 1 is underrelaxation. This may slow down speed of convergence but increases the stability of the calculation, i.e. it decreases the possibility of divergence or oscillations in the solutions.
> * U = 1 corresponds to no relaxation. One uses the predicted value of the variable.
> * U > 1 is overrelaxation. It can sometimes be used to accelerate convergence but will decrease the stability of the calculation.

Which seems to support the idea that over-relaxation may sometimes be helpful.
OpenFoam uses under-relaxation but does not mention over-relaxation https://www.openfoam.com/documentation/user-guide/fvSolution.php#x24-990006.3.2

johnomotani:
Trying to over-relax the `phi` solve, which was already converging OK, still converged but more slowly, `overrelax_factor=1.1` took 10 iterations, rather than 8 for `overrelax_factor=1`.

johnomotani:
Added `WIP` label, because I think there may be a much neater way to structure the loop, using one or two lambda functions to update `b` and `x` instead of copy-pasting code...

src/invert/laplace/impls/naulin/naulin_laplace.cxx
```diff
@@ -211,46 +242,100 @@ const Field3D LaplaceNaulin::solve(const Field3D &rhs, const Field3D &x0) {
...
+    if (count>maxits) {
+      throw BoutException("LaplaceNaulin error: Not converged within maxits=%i iterations.", maxits);
+    }
+
+    output<<underrelax_factor<<" "<<count<<" "<<error_abs<<" "<<error_rel<<endl;
+
+    while (error_abs > last_error) {
+      // Iteration seems to be diverging... try underrelaxing and restart
+      underrelax_factor *= .9;
+      underrelax_count++;
```
ZedThree: Minor, but `++underrelax_count`

The postfix version increments and returns the old value -- we don't care about the return value, so use prefix instead

PR #1660, opened 2019-04-01
Unit tests and various small refactorings of Solver

ZedThree:
Fixes:

- `constraint` for `Vector`s had a copy-paste error in adding the individual fields
- using `constraint` led to a bad `delete` on `Solver::VarStr::MMS_err`

This are both still in `master`: I can pull them out though.

Changes:

- Add ~50% unit test coverage of `Solver`, and 100% coverage of `SolverFactory`
- Make `Solver::VarStr::MMS_err` `unique_ptr`
- Give sensible default values to `Solver` members
- Use `constexpr auto` instead of `#define` for solver names
- Rename some `Solver` and `Monitor` private members for readability
- Pull out some helper functions from various `Solver` internals
   - Partly for readability, partly to simplify code
- Don't use global `Mesh` in `Solver::getLocalN` -- now gets the correct number of points if solving with fields on different meshes
- Clang-format
- More consistent and expanded docstrings

Still missing tests for:
- Anything to do with `PhysicsModels`
- Anything that touches `Datafile`
- Any of the save/load loop vars stuff

### DIFF
--- a/examples/subsampling/README.md
+++ b/examples/subsampling/README.md
@@ -9,24 +9,22 @@ Therefore the fields, which are solved for, and any field that is
 added with `SAVE_REPEAT()` will be written out every second timeunit.
 
 In the code another monitor is added with an output timestep of 0.01.
-This Monitor uses a separat datafile, (via class `SimpleDatafile`, see monitor.cxx
-line 12 ff) to write the data. The options for the simple datafile are
-read from the `[probes]` section in the input.
-As netcdf supports several unlimited dimension, it is also possible to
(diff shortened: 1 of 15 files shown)


bendudson:
My current plan is to get `OptionsNetCDF` to the point where it could replace `DataFile` entirely, so have not put any effort into improving or testing `DataFile` for a while. For a while I think there will be two systems, then perhaps with 5.0 we can break some things like remove `DataFile` and replace `Output` and logging things with {fmt}.

PR #1667, opened 2019-04-04
Fix dereferencing null Coordinates in FieldFactory

ZedThree:
A combination of recent changes has resulted in next breaking on one of the unit tests

This was due to a unit test setting the mesh Coordinates to `nullptr` and `FieldFactory` attempting to call `field->getCoordinates()->getParallelTransform()` which segfaulted

Fix is to check `Coordinates` is not null before trying to use it, and also change the test to not do the inverse transform on initial_profile

### DIFF
--- a/src/mesh/coordinates.cxx
+++ b/src/mesh/coordinates.cxx
@@ -857,7 +857,9 @@ void Coordinates::setParallelTransform(Options* options) {
 }
 
 ParallelTransform& Coordinates::getParallelTransform() {
-  // Return a reference to the ParallelTransform object
+  if (transform == nullptr) {
+    setParallelTransform(Options::getRoot());
+  }
   return *transform;
 }
 
--- a/src/mesh/coordinates.cxx
+++ b/src/mesh/coordinates.cxx
@@ -856,13 +856,6 @@ void Coordinates::setParallelTransform(Options* options) {
   }
 }
 
-ParallelTransform& Coordinates::getParallelTransform() {
-  if (transform == nullptr) {
-    throw BoutException(_("Coordinates instance does not have parallel transform set"));
-  }
-  return *transform;
-}
-
 /*******************************************************************************
  * Operators
  *
--- a/src/mesh/coordinates.cxx
+++ b/src/mesh/coordinates.cxx
@@ -858,11 +858,10 @@ void Coordinates::setParallelTransform(Options* options) {
 
 ParallelTransform& Coordinates::getParallelTransform() {
   if (transform == nullptr) {
-    setParallelTransform(Options::getRoot());
+    throw BoutException(_("Coordinates instance does not have parallel transform set"));
   }
   return *transform;
 }
-
 
 /*******************************************************************************
  * Operators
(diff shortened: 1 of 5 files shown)


dschwoerer:
Should this go into next?

Already over 1k commits in next ...

ZedThree:
Thanks! Was in a rush

tests/unit/field/test_field_factory.cxx
```diff
@@ -587,6 +587,22 @@ TYPED_TEST(FieldFactoryCreationTest, CreateOnMesh) {
...
 
+TYPED_TEST(FieldFactoryCreationTest, CreateOnMeshWithoutCoordinates) {
+  constexpr auto nx = int{1};
+  constexpr auto ny = int{1};
+  constexpr auto nz = int{1};
+
+  FakeMesh localmesh{nx, ny, nz};
+  localmesh.setCoordinates(nullptr);
+  localmesh.createDefaultRegions();
+  localmesh.setCoordinates(nullptr);
```
johnomotani: Are both of the `setCoordinates(nullptr)` calls needed here?

src/mesh/coordinates.cxx
```diff
@@ -857,7 +857,9 @@ void Coordinates::setParallelTransform(Options* options) {
 }
 
 ParallelTransform& Coordinates::getParallelTransform() {
-  // Return a reference to the ParallelTransform object
+  if (transform == nullptr) {
+    setParallelTransform(Options::getRoot());
```
johnomotani: Is this necessary and useful? `setParallelTransform` is called in the `Coordinates` constructor, except the one where all the member fields are passed in as arguments which is only used in the unit tests. If this is needed for the case when that constructor is used, could we add the `setParallelTransform(Options::getRoot())` in that constructor instead*? I worry a bit about having this conditional because I don't think the `ParallelTransform` should normally be created with `Options::getRoot()` - it should use the options that were used to create the `Mesh` that owns the `Coordinates` that own the `ParallelTransform`. So I'm worried that this might mask a bug where somehow the `ParallelTransform` didn't get set when it should have been, and make that harder to track down because the `ParallelTransform` gets set to something 'wrong' rather than being a `nullptr` that makes it obvious there's an error.

--
\* is `Options::getRoot()` always available in the unit tests? I think I remember having issues with it not being, in which case it could be a bad idea to add this call to that constructor.

ZedThree: That is all entirely reasonable -- I will change this to throw instead.

`Options::getRoot()`/`Options::root()` are always available -- they just might not contain anything useful. If they are used in a test, the test/fixture should call `Options::cleanup()` on teardown.

src/mesh/coordinates.cxx
```diff
@@ -857,11 +857,12 @@ void Coordinates::setParallelTransform(Options* options) {
 }
 
 ParallelTransform& Coordinates::getParallelTransform() {
-  // Return a reference to the ParallelTransform object
+  if (transform == nullptr) {
+    throw BoutException(_("Coordinates instance does not have parallel transform set"));
```
johnomotani: Is there a case for making this an `ASSERT1` to make `getParallelTransform()` more inline-able in fully optimized builds? Probably doesn't matter, because this `if` really shouldn't ever be a significant overhead as `getParallelTransform` shouldn't ever be in an inner loop...

ZedThree: Given how simple this function is, I've moved it into the header so that it could be inlined.

PR #1669, opened 2019-04-06
Method for Laplace solvers to say whether they use 3D coefficients

johnomotani:
The FFT-based Laplace solvers use the DC component of the coefficients if they are passed `Field3D` coefficients. When running non-Boussinesq simualtions that require `Field3D` coefficients, I'd like to check if we have run with an appropriate Laplacian solver. The method added here makes this check much cleaner - at the moment I have to get the `laplace:type` option and check that.

I'd actually prefer the `Laplace::setCoefA(Field3D val)` methods to throw by default, instead of silently taking `DC(val)` (unless over-ridden by an implementation that can use `Field3D` coefficients). That would be a backward-incompatible change though: does it count as a 'bugfix' or is it a change to the interface that should wait for BOUT-5?

### DIFF
--- a/include/invert_laplace.hxx
+++ b/include/invert_laplace.hxx
@@ -173,6 +173,9 @@ public:
   virtual void setGlobalFlags(int f) { global_flags = f; }
   virtual void setInnerBoundaryFlags(int f) { inner_boundary_flags = f; }
   virtual void setOuterBoundaryFlags(int f) { outer_boundary_flags = f; }
+
+  /// Does this solver use Field3D coefficients (true) or only their DC component (false)
+  virtual bool uses3DCoefs() const { return false; }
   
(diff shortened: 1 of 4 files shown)


johnomotani:
In favour of changing `setCoefA(Field3D)` to throw: anyone who knows they want the old behaviour and is re-compiling their code against an updated version of BOUT can very easily change `setCoefA(n);` -> `setCoefA(DC(n));` or even
```
if (laplaceSolver->uses3DCoefs()) {
  laplaceSolver->setCoefA(n);
} else {
  laplaceSolver->setCoefA(DC(n));
}
```
On the other hand, anyone who's surprised by the exception probably needs to be alerted to the fact that they're only using DC components of their coefficients!

PR #1673, opened 2019-04-10
Runge-Kutta-Legendre stabilised explicit method

bendudson:
Uses Strang splitting (2nd order) to combine:

 - RK3-SSP for advection (3rd order)
 - Runge-Kutta-Legendre for diffusion (2nd order). These schemes uses multiple stages to increase stability, rather than accuracy; this is always 2nd order, but their stable timestep for diffusion problems increases as the square of the number of stages. The number of stages is an input option, can be arbitrarily large.

See https://doi.org/10.1016/j.jcp.2013.08.021

To try it out, set `solver:type=splitrk`. Use options `solver:timestep` to set the internal timestep, and `solver:nstages` to set the number of stages in the RKL method. 

Explicit method, adaptive time stepping by default, checking accuracy a specified number of timesteps.  

Tested by eye with a few examples, passes the unit and MMS tests. So far not particularly useful for real cases, but probably good enough to include.

Note that the `MMS/time` test has been modified to test both the convective and diffusive parts. This shows that Karniadakis converges at 1st order. I think we should remove that method, as there are better options (such as this).

### DIFF
--- a/examples/IMEX/diffusion-nl/diffusion-nl.cxx
+++ b/examples/IMEX/diffusion-nl/diffusion-nl.cxx
@@ -23,7 +23,7 @@ class DiffusionNL : public PhysicsModel {
 protected:
   int init(bool) {
     // Get the input parameter alpha
-    auto opt = Options::root();
+    auto& opt = Options::root();
     alpha = opt["alpha"].withDefault(2.5);
 
     // Specify that the operator is split
--- a/examples/IMEX/diffusion-nl/diffusion-nl.cxx
+++ b/examples/IMEX/diffusion-nl/diffusion-nl.cxx
@@ -21,7 +21,7 @@
 
 class DiffusionNL : public PhysicsModel {
 protected:
-  int init(bool restarting) {
+  int init(bool) {
     // Get the input parameter alpha
     auto opt = Options::root();
     alpha = opt["alpha"].withDefault(2.5);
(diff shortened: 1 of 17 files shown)


d7919:
Looks interesting!

examples/IMEX/diffusion-nl/diffusion-nl.cxx
```diff
@@ -21,7 +21,7 @@
 
 class DiffusionNL : public PhysicsModel {
 protected:
-  int init(bool restarting) {
+  int init(bool) {
     // Get the input parameter alpha
     auto opt = Options::root();
```
ZedThree: `auto&`? Might not matter here, I guess

bendudson: Ah yes, good catch! Good to get into that habit, so that docs etc. get saved.

include/bout/solver.hxx
```diff
@@ -459,6 +459,9 @@ private:
   /// Diffusive part (if split operator)
   rhsfunc phys_diff{nullptr};
 
+  /// Should unsplit physics models be treated as diffusive?
+  bool unsplit_diffusive{true};
```
ZedThree: How do you feel about `is_unsplit_diffusive`, or even `is_nonsplit_model_diffusive`?

bendudson: yes happy with either of those. Maybe `is_nonsplit_model_diffusive` is clearer?

Issue #1675, opened 2019-04-11
example/invertable_operator doesn't compile


dschwoerer:
```
Trying to compile ../../../examples/invertable_operator
  Compiling  invertable_operator.cxx
invertable_operator.cxx: In member function ‘virtual int HW::init(bool)’:
invertable_operator.cxx:87:14: error: ‘class bout::inversion::InvertableOperator<Field3D>’ has no member named ‘setOperatorFunction’
     mySolver.setOperatorFunction(mm);
              ^~~~~~~~~~~~~~~~~~~
invertable_operator.cxx:88:14: error: ‘class bout::inversion::InvertableOperator<Field3D>’ has no member named ‘setup’
     mySolver.setup();
              ^~~~~
invertable_operator.cxx:75:17: warning: unused parameter ‘restart’ [-Wunused-parameter]
   int init(bool restart) {
            ~~~~~^~~~~~~
invertable_operator.cxx: In member function ‘virtual int HW::rhs(BoutReal)’:
invertable_operator.cxx:106:28: error: ‘class bout::inversion::InvertableOperator<Field3D>’ has no member named ‘invert’
     solutionInv = mySolver.invert(n, 0.0);
                            ^~~~~~
invertable_operator.cxx:107:14: error: ‘class bout::inversion::InvertableOperator<Field3D>’ has no member named ‘reportTime’
     mySolver.reportTime();
              ^~~~~~~~~~
invertable_operator.cxx:112:32: error: ‘class bout::inversion::InvertableOperator<Field3D>’ has no member named ‘invert’
         solutionInv = mySolver.invert(n, solutionInv);
                                ^~~~~~
invertable_operator.cxx:121:26: error: ‘class bout::inversion::InvertableOperator<Field3D>’ has no member named ‘verify’
     auto pass = mySolver.verify(n) == 0 ? "False" : "True";
                          ^~~~~~
invertable_operator.cxx:139:14: error: ‘class bout::inversion::InvertableOperator<Field3D>’ has no member named ‘reportTime’
     mySolver.reportTime();
              ^~~~~~~~~~
invertable_operator.cxx:100:20: warning: unused parameter ‘time’ [-Wunused-parameter]
   int rhs(BoutReal time) {
           ~~~~~~~~~^~~~
make: *** [../../make.config:308: invertable_operator.o] Error 1
invertable_operator failed
```

d7919:
[...]

PR #1678, opened 2019-04-11
Fix examples

dschwoerer:
Various fixes in examples:
 * fci-wave
 * fci-wave-logn
 * make-config ($PATH was wrong)
 * #1675 check for petsc

Fixes for tests:
 * test-include : string header was missing

The fixes are rather simple, so a single PR ...


examples/fci-wave-logn/fci-wave.cxx
```diff
@@ -17,7 +17,7 @@ class FCIwave : public PhysicsModel {
     Field3D f_B = f / Bxyz;
     
     f_B.splitYupYdown();
-    mesh->getParallelTransform().integrateYUpDown(f_B);
+    mesh->getParallelTransform().integrateYupYdown(f_B);
 
     // integrateYUpDown replaces all yup/down points, so the boundary conditions
```
bendudson: Small point: `integrateYUpDown` in comment probably should be updated too

examples/fci-wave/fci-wave.cxx
```diff
@@ -18,7 +18,7 @@ class FCIwave : public PhysicsModel {
     Field3D f_B = f / Bxyz;
 
     f_B.splitYupYdown();
-    mesh->getParallelTransform().integrateYUpDown(f_B);
+    mesh->getParallelTransform().integrateYupYdown(f_B);
 
     // integrateYUpDown replaces all yup/down points, so the boundary conditions
```
bendudson: As above

examples/fci-wave-logn/fci-wave.cxx
johnomotani: While you're changing it, should probably be `integrateParallelSlices` rather than `integrateYupYdown` now, since #1635.

dschwoerer: gcc recommended `integrateYupYdown` :smile: 
Unfortunately I don't know whether we can teach gcc to show a different function ...

dschwoerer:
Should be fixed now :+1:

PR #1686, opened 2019-04-15
Bugfix: bracket operator slowdown

johnomotani:
There was a significant slowdown in the Arakawa implementation of the bracket operators. `Field3D::getRegion()` was used where a 2d region was needed (because of an explicit inner loop over `z`). This was a regression since 4.2.2; results were still correct, but calculation was repeated unnecessarily `Nz` times. It actually caused some of my simulations to take about 1.5 times longer.

I've fixed the issue by adding a `Field3D::getRegion2D` method, because I thought it looked nicer than using something like `result.getMesh()->getRegion2D()`, but we'd only use this method in a few places (I think I've included all the ones I would change in this PR). @ZedThree @bendudson @d7919 if you think the extra method's not the way to go, I'm happy to redo the PR and just change the 2 lines in `difops.cxx`.

### DIFF
--- a/include/field3d.hxx
+++ b/include/field3d.hxx
@@ -316,6 +316,11 @@ class Field3D : public Field, public FieldData {
   /// 
   const Region<Ind3D>& getRegion(REGION region) const;  
   const Region<Ind3D>& getRegion(const std::string &region_name) const;
+
+  /// Return a Region<Ind2D> reference to use to iterate over the x- and
+  /// y-indices of this field
+  const Region<Ind2D>& getRegion2D(REGION region) const;
+  const Region<Ind2D>& getRegion2D(const std::string &region_name) const;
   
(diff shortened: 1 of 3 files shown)


d7919:
Sorry, that was my fault -- good spot! I'm not sure about the `Field3D::getRegion2D` method yet, I'd have to think about it some more -- the general idea of asking for things from the field rather than the mesh is good so I think this is probably a good change but I always worry when there's methods for 3d that don't exist for 2d (potentially limits templates).

johnomotani:
> I always worry when there's methods for 3d that don't exist for 2d (potentially limits templates).

We could add `Field2D::getRegion2D`, either now or if/when it's needed?

dschwoerer:
I first thought it wouldn't make sense for Field2D, as you would only want to use that if you manually iterate over `z`, but you could just iterate over 1 point in z, and thus it would probably make sense to add it now, to keep the interface as surprise free as possible ...

Would it make sense for `FieldPerp`? To iterate only over the `x` dimension?

PR #1688, opened 2019-04-15
Fix elm pb example

bendudson:
Compiles and runs, but would like to make sure that it produces a similar result to the BOUT++ 3.1 version. The test case should produce a growth rate which can be compared to the value from ELITE.

Lots of reformatting, to put it into a `PhysicsModel` class. Best viewed with whitespace diff off.

### DIFF
--- a/examples/elm-pb/data/BOUT.inp
+++ b/examples/elm-pb/data/BOUT.inp
@@ -9,11 +9,8 @@ TIMESTEP = 1       # time between outputs
 archive = 20        # Archive restart files after this number of outputs
 wall_limit = 1.55  # wall time limit (in hours)
 
-ShiftXderivs = true # use shifted radial derivatives?
-TwistShift = true   # use twist-shift condition?
-
-MZ = 16             # number of points in z direction (2^n + 1)
 ZPERIOD = 15        # Fraction of a torus to simulate
+MZ = 16             # Number of points in Z
 
 grid = "cbm18_dens8.grid_nx68ny64.nc"  # Grid file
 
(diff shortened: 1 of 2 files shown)


dschwoerer:
I can assert that this does not give the same result as v3.1

The v3.1 does segfault while trying to write an int.

After adding some debug messages, and adding some missing return statements in datafile, it doesn't crash anymore, but the monitor signals to quit -- without any error ...

dschwoerer:
Same/similar seg fault with v3.0 ...

Compiler is gcc (GCC) 8.3.1 20190223 (Red Hat 8.3.1-2)

dschwoerer:
v3.1
early exit with `Monitor signalled to quit. Returning`

gcc (Ubuntu 5.4.0-6ubuntu1~16.04.11) 5.4.0 20160609

Not sure when the last time that example actually worked ...

bendudson:
IIRC this is/was a known bug, caused by adding a variable twice to the dump output. Probably the last time it worked was shortly after the last paper to be published using it. 2012? Since then there have been many derivatives and variations on it, but mainly in personal repos or at NERSC.

bendudson:
I think this is good enough to go in for now. It's not tested yet, but it runs and produces similar results (~10%) to previously. I don't know when I will have time to properly test this, and it's better than what's in the repo at the moment.

ZedThree:
LGTM -- do you mind if I just make the three laplacian solvers `unique_ptr`?

ZedThree:
@bendudson ok with 7f93806 - fccfc18?

PR #1691, opened 2019-04-16
Use asynchronous communications in LaplaceMultigrid

johnomotani:
Allows overlapping of sends and receives. Gave a modest speed-up in a test case where I forced all the domain-decomposition to be in the x-direction so I was doing Laplace solves on 2 nodes (96 processors) with a 192x16x128 grid: the run-time reduced from 86s (88.8% in inversion) to 74s (87.3% in inversion) with this update using intel/intelmpi compilers, and from 104s (92.4% inversion) to 100s (91.9% inversion) with gcc/openmpi.

### DIFF
--- a/src/invert/laplace/impls/multigrid/multigrid_alg.cxx
+++ b/src/invert/laplace/impls/multigrid/multigrid_alg.cxx
@@ -623,7 +623,7 @@ void MultigridAlg::communications(BoutReal* x, int level) {
   MAYBE_UNUSED(int ierr);
 
   if(zNP > 1) {
-    MPI_Request requests[4];
+    MPI_Request requests[] = {MPI_REQUEST_NULL, MPI_REQUEST_NULL, MPI_REQUEST_NULL, MPI_REQUEST_NULL};
     MPI_Datatype xvector;
 
     // Create datatype to hold z-direction guard cells, which are not continuous in memory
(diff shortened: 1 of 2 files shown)


bendudson:
Thanks @johnomotani !
1. Is it ok to `MPI_Isend` and `MPI_Irecv` to/from a processor which doesn't exist e.g. at the boundaries
2. Is there any benefit to calling `Irecv`s before `Isend`s or does it not make any difference?

johnomotani:
@bendudson:
1. Not sure, I copied the structure from the previous MPI_Sendrecv calls. Will check tomorrow. The integrated test passes, and a STORM test gives correct results, so I'm hopeful it's OK.
2. That was a mistake, I hadn't thought about it. I remember now the normal way is Irecv before Isend and internet suggests this has better performance.

johnomotani:
At the moment `LaplaceMultigrid` is communicating the x-boundary guard cells as if x was periodic. This should be unnecessary, and removing sends/receives from the inner boundary of processor `0` and outer boundary of processor `xNP-1` makes no difference to the test results (I printed out the maximum absolute errors from `test-multigrid-laplace/runtest` and they did not change at all) - I think those cells must be overwritten or ignored where the boundary conditions are used.

johnomotani:
While looking at `LaplaceMultigrid` it seems there were a lot of unnecessary communications being done. Arrays were being communicated at the end of one method and then again at the beginning of the next one. I've removed the communications at the start of methods in the last commit, going to check now whether my STORM test cases still pass, and if it speeds them up at all...

johnomotani:
My tests pass after removing the redundant communications, and it does speed things up a bit.

d7919:
> Thanks @johnomotani !
> 
> 1. Is it ok to `MPI_Isend` and `MPI_Irecv` to/from a processor which doesn't exist e.g. at the boundaries
> 2. Is there any benefit to calling `Irecv`s before `Isend`s or does it not make any difference?

Just a comment on 2. -- it does seem common wisdom that the order makes a difference but I've never actually noticed a performance benefit when I've looked in detail at this on other codes. It may well depend what sort of communication regime we're in (as well as which mpi implementation we have) but I'd say any performance motivation for the order should be secondary to whatever order is clearest/makes more sense (I've not looked at the code so this isn't a comment on that). The studies I've done in the past were also on machines we're trying to overlap work and communication didn't work very well, so it's possible things have changed since I last looked.

d7919:
I should also say going asynchronous is a good idea -- given that the communication patterns are presumably the same each time it may also be work considering setting up persistent communications (this is something that has been contemplated before but we've never got around to actually trying it, as far as I'm aware).

PR #1693, opened 2019-04-23
Move Gridfile::get implementations for Field3D/Field2D to .cxx

johnomotani:
Intel compilers don't instantiate the getField templates when they are called in the .hxx file rather than the .cxx. This commit moves the implementations from .hxx to .cxx to work around this.

PS for anyone trying to compile `next` with intel compilers, I've found I have to change `-std=c++11` to `-std=c++14` in `BOUT_FLAGS`.

### DIFF
--- a/include/bout/griddata.hxx
+++ b/include/bout/griddata.hxx
@@ -89,12 +89,8 @@ public:
   bool get(Mesh *m, int &ival, const std::string &name) override; ///< Get an integer
   bool get(Mesh *m, BoutReal &rval,
            const std::string &name) override; ///< Get a BoutReal number
-  bool get(Mesh *m, Field2D &var, const std::string &name, BoutReal def = 0.0) override {
-    return getField(m, var, name, def);
-  }
-  bool get(Mesh *m, Field3D &var, const std::string &name, BoutReal def = 0.0) override {
-    return getField(m, var, name, def);
-  }
+  bool get(Mesh *m, Field2D &var, const std::string &name, BoutReal def = 0.0) override;
+  bool get(Mesh *m, Field3D &var, const std::string &name, BoutReal def = 0.0) override;
 
   bool get(Mesh *m, std::vector<int> &var, const std::string &name, int len, int offset = 0,
            GridDataSource::Direction dir = GridDataSource::X) override;
--- a/src/mesh/data/gridfromfile.cxx
+++ b/src/mesh/data/gridfromfile.cxx
@@ -176,6 +176,13 @@ bool GridFile::get(Mesh *UNUSED(m), BoutReal &rval, const std::string &name) {
  * Successfully reads Field2D if the variable in the file is 0-D or 2-D.
  * Successfully reads Field3D if the variable in the file is 0-D, 2-D or 3-D.
  */
+bool GridFile::get(Mesh *m, Field2D &var, const std::string &name, BoutReal def) {
+  return getField(m, var, name, def);
+}
+bool GridFile::get(Mesh *m, Field3D &var, const std::string &name, BoutReal def) {
+  return getField(m, var, name, def);
+}
+
 template<typename T>
 bool GridFile::getField(Mesh* m, T& var, const std::string& name, BoutReal def) {
   static_assert(std::is_base_of<Field2D, T>::value || std::is_base_of<Field3D, T>::value,


ZedThree:
This is a bit surprising, on both points! What version of Intel are you using? Do you get an error message (on either point)? And do you have a test case?

ZedThree:
For reference, I've just tried with a few Intel versions. 17.0.1 doesn't work due to some compiler bugs. 17.0.4, 18.0.3 and 19.0.1 all work. I think we might technically support as low as 14.0, but I don't have access to any versions before 17.

If/when we move to C++14, we'll support Intel 17+.

johnomotani:
The `-std=c++14` thing is from `mpark/variant`. It says https://github.com/mpark/variant that GCC, CLANG, Xcode and Visual Studio are tested, but not Intel compilers. Using Intel 17.0.4 or 18.0.5 compilers I get a stack of errors like
```
In file included from ../../include/bout/sys/variant.hxx(24),
                 from ../../include/options.hxx(46),
                 from ../../include/bout/deriv_store.hxx(42),
                 from ../../include/bout/mesh.hxx(49),
                 from field.cxx(33):
/marconi_work/FUA33_SOLBOUT3/jomotani/BOUT-intel-minimal-check/externalpackages/mpark.variant/include/mpark/variant.hpp(447): warning #3801: deduced return types are a C++14 feature
          inline static constexpr auto &&get_alt(V &&v, in_place_index_t<0>) {
```
and
```
In file included from ../../include/bout/sys/variant.hxx(24),
                 from ../../include/options.hxx(46),
                 from ../../include/bout/deriv_store.hxx(42),
                 from ../../include/bout/mesh.hxx(49),
                 from field.cxx(33):
/marconi_work/FUA33_SOLBOUT3/jomotani/BOUT-intel-minimal-check/externalpackages/mpark.variant/include/mpark/variant.hpp(523): error: expected an identifier
            inline static constexpr DECLTYPE_AUTO invoke(Visitor &&visitor,
                                    ^

In file included from ../../include/bout/sys/variant.hxx(24),
                 from ../../include/options.hxx(46),
                 from ../../include/bout/deriv_store.hxx(42),
                 from ../../include/bout/mesh.hxx(49),
                 from field.cxx(33):
/marconi_work/FUA33_SOLBOUT3/jomotani/BOUT-intel-minimal-check/externalpackages/mpark.variant/include/mpark/variant.hpp(523): error: "auto" is not allowed here
            inline static constexpr DECLTYPE_AUTO invoke(Visitor &&visitor,
                                    ^

In file included from ../../include/bout/sys/variant.hxx(24),
                 from ../../include/options.hxx(46),
                 from ../../include/bout/deriv_store.hxx(42),
                 from ../../include/bout/mesh.hxx(49),
                 from field.cxx(33):
/marconi_work/FUA33_SOLBOUT3/jomotani/BOUT-intel-minimal-check/externalpackages/mpark.variant/include/mpark/variant.hpp(523): error #303: explicit type is missing ("int" assumed)
            inline static constexpr DECLTYPE_AUTO invoke(Visitor &&visitor,
                                    ^

In file included from ../../include/bout/sys/variant.hxx(24),
                 from ../../include/options.hxx(46),
                 from ../../include/bout/deriv_store.hxx(42),
                 from ../../include/bout/mesh.hxx(49),
                 from field.cxx(33):
/marconi_work/FUA33_SOLBOUT3/jomotani/BOUT-intel-minimal-check/externalpackages/mpark.variant/include/mpark/variant.hpp(523): error: expected a ";"
            inline static constexpr DECLTYPE_AUTO invoke(Visitor &&visitor,
```
complete output: [intel-compile-errors.txt](https://github.com/boutproject/BOUT-dev/files/3111542/intel-compile-errors.txt).
These all go away using `-std=c++14`.

On the `Gridfile` errors, this program
```
#include <bout.hxx>
#include <bout/griddata.hxx>

int main(int argc, char** argv) {
  BoutInitialise(argc, argv);
  Field3D test;
  GridFile f(data_format("test.nc"), "test.nc");
  f.get(mesh, test, "test");
  BoutFinalise();
  return 0;
}
```
fails to compile with either 17.0.4 or 18.0.5. I'd configured with `EXTRA_LIBS="-L$NETCDF_LIB" ./configure --without-hdf5 --enable-optimize=3 --enable-checks=no --enable-shared --with-sundials=/marconi_work/FUA33_SOLBOUT3/jomotani/pkg/sundials_inst_intel/`. The error is:
```
  Compiling  test-intel-gridfile.cxx
icpc: command line warning #10006: ignoring unknown option '-Wnull-dereference'
icpc: command line warning #10006: ignoring unknown option '-Wno-cast-function-type'
  Linking test-intel-gridfile
test-intel-gridfile.o: In function `GridFile::get(Mesh*, Field3D&, std::string const&, double)':
/marconi_work/FUA33_SOLBOUT3/jomotani/BOUT-intel-minimal-check/examples/test-intel-gridfile/../../include/bout/griddata.hxx:96: undefined reference to `bool GridFile::getField<Field3D>(Mesh*, Field3D&, std::string const&, double)'
make: *** [test-intel-gridfile] Error 1
```
strangely using `mesh->get(test, "test");` is fine even though `Mesh::get(Field3D,std::string)` calls `GridFile::get(Mesh*, Field3D&, std::string const&, double)`. I also don't get the error if I compile in debug mode. I guess must be some over-aggressive inlining?

d7919:
From variant.hpp it seems there are two branches of what `DECLTYPE_AUTO` might mean based on `MPARK_RETURN_TYPE_DEDUCTION`.  This is defined based on

````
#if defined(__cpp_return_type_deduction) || defined(_MSC_VER)
#define MPARK_RETURN_TYPE_DEDUCTION
#endif
````

in `config.hpp`. The first should only be defined for c++ 14 onwards and the second, I believe, should only be relevant if compiling on windows. This would suggest that this variable shouldn't be defined, but I suspect it might be for intel. It should be possible to check this by adding something like

````
#ifdef MPARK_RETURN_TYPE_DEDUCTION
#error "The thing is defined."
#endif
````

PR #1695, opened 2019-04-26
 Convert Options::isSection() argument to lower case 

johnomotani:
Option/section names are stored in lower case, so need to convert 'std::string name' argument to lower case before searching.

Also add more tests for case sensitivity/insensitivity in options unit tests to avoid this kind of error.

### DIFF
--- a/src/sys/options.cxx
+++ b/src/sys/options.cxx
@@ -126,7 +126,7 @@ bool Options::isSection(const std::string& name) const {
   }
 
   // Is there a child section?
-  auto it = children.find(name);
+  auto it = children.find(lowercase(name));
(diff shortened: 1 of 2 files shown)


johnomotani:
PS I vote for making options names/sections case sensitive (maybe in version 5?).

dschwoerer:
> PS I vote for making options names/sections case sensitive (maybe in version 5?).

I am not sure this is worth it:
 * It breaks code
 * It is confusing: `MXG` vs `mxg` - `nx` or `NX`?

Would also the arguments become case sensitive?

If you want to pursue this, we should probably discuss this in a separate issue ...

PR #1696, opened 2019-04-26
Experimental CMake support

ZedThree:
This is a proposal for eventually switching to CMake (possibly in v5). This also includes some minor fixes to the main library that I'll pull out into separate PRs

Not fully complete, but should be pretty usable. Current features:
- Most configure-time switches (`--enable_warnings/debug/optimize` can be handled by CMake itself)
- Most optional dependencies (missing support for ~~HDF5~~, pnetcdf, MUMPS, ~~SLEPc~~)
- Unit tests
- A few integrated tests (plus a macro to easily add more)
- Demonstration of use with an example (blob2d)

Use:

```bash
mkdir build && cd build
cmake ..
make
ctest
```

~~On OpenSUSE, I also have to pass `-DMPIEXEC_EXECUTABLE=$(which mpirun)` to detect MPI properly :(~~
Now clever enough to do this itself!

The easiest way to use with an actual physics model is to install BOUT++:
```bash
mkdir build && cd build
cmake .. -DCMAKE_INSTALL_PREFIX=/path/to/install
make && make install
# Now to build blob2d
cd ../examples/blob2d
mdir build && cd build
cmake .. -Dbout++_DIR=/path/to/install/lib/cmake/bout++
make
# Note that this makes the blob2d executable in a `build` directory
cp -r ../delta_0.25 .
mpirun -n 4 ./blob2d -d delta_0.25
```

A more complex CMake configure line looks like:
```
cmake .. \
  -DCMAKE_BUILD_TYPE=Debug \
  -DUSE_PETSC=on -DPETSC_DIR=/path/to/petsc \
  -DUSE_SUNDIALS=on -DSUNDIALS_ROOT=/path/to/sundials \
  -DNetCDF_ROOT=/path/to/netcdf \
  -DENABLE_OPENMP=ON \
  -DENABLE_COLOR=false \
  -DCHECK=2 \
  -DCMAKE_INSTALL_PREFIX=/path/to/install
```

`ccmake` can also be used for a text-based interface, or `cmake-gui` for a graphical one.

---

There are some pros and cons to switching to CMake:

### Cons:
- We already have a working build system!
- Debugging CMake is harder -- some things like `find_library` don't actually tell you what paths they're looking in
- Not installed everywhere by default
- Not a very nice language

### Pros:
- Our current build system has some creaky corners: 
    - if you change a header, you may need to `make clean` before recompiling -- CMake can track this for us and recompile just what's needed
    - no support for out-of-source builds -- CMake does this out of the box
- Easy to add dependencies that also use CMake (e.g. adding fmtlib is two lines!)
- Very easy to install: `pip install --user cmake`
- Easy to run static analysis (`-DCMAKE_CXX_CLANG_TIDY` to run `clang-tidy` at the same time as building)
- Nicer language than m4!
- As close to a standard C++ build system as there is


ZedThree:
- Added a few more integrated tests
- Added support for SLEPc
- Worked out the magic flag for Cray systems: `-DCMAKE_SYSTEM_NAME=CrayLinuxEnvironment` tells CMake to use the `PrgEnv-*`

Works on Archer with GNU: Intel compiler crashes, and Cray borks on stuff like

```
CC-1660 crayc++: ERROR File = /home/e607/e607/phtdotp/BOUT-dev/src/field/generated_fieldops.cxx, Line = 1795
  The for loop in an OpenMP parallel region contains an illegal conditional expression.
    BOUT_FOR(index, result.getRegion("RGN_ALL")) { result[index] = lhs / rhs[index]; }
```

Easy fix, but I'm not sure we should need to.

d7919:
Is there a simple way to install `ccmake` as well? It seems the pip install approach doesn't provide this (or I'm doing it wrong).

d7919:
On a fresh clone I found an initial `cmake` gave the following error

~~~~
CMake Error at CMakeLists.txt:21 (add_subdirectory):
  The source directory

    /...../BOUT_CMAKE/externalpackages/mpark.variant

  does not contain a CMakeLists.txt file.
~~~~

Should I manually checkout the submodules first before using `cmake`?

johnomotani:
for `ccmake` I'm using the standard repo, but it's a separate package to `cmake`. For me `sudo aptitude install cmake-curses-gui` installed both.

d7919:
> for `ccmake` I'm using the standard repo, but it's a separate package to `cmake`. For me `sudo aptitude install cmake-curses-gui` installed both.

Thanks, unfortunately I'm on a machine that I don't have admin rights for so have to rely on things like pip that I can easily install into local directories or building things myself.

ZedThree:
- Ensures git submodules are up-to-date (beware that CMake may automatically clobber any changes you make to submodules without committing)
- Fixes issue with PVODE by moving some files around
- Fixes issues with SUNDIALS not being found correctly
- Fixes SUNDIALS not being built on Travis
- Requires CMake >= 3.9

I've rewritten some things to ensure we don't require CMake > 3.10, as 3.10 is the latest version in the Archer modules. Although it is possible to install more recent versions via `pip`, I thought it best to not be too bleeding edge.

Also, Score-P is a faff and doesn't really play very nice with CMake. The best way of using it is:

```
$ SCOREP_WRAPPER=off CC=scorep-mpicc CXX=scorep-mpicxx cmake .. <usual CMake options>
```

I think there is a better way, I just spent way too long trying to make it work without success.

ZedThree:
This is ready to go in, albeit still as experimental for now.

Missing features:
- code coverage
- pnetCDF support
- MUMPS support
- targets for building/running just unit/integrated/MMS tests (although running a subset of tests can be done manually via `ctest`)
- target for making release tarball
- other utility targets that I've forgotten

Most integrated and MMS tests are missing, but can be added later.

johnomotani:
> pnetCDF support

I wouldn't trust the pnetCDF interface to work correctly anyway... Should we disable pnetCDF until it's checked thoroughly, or at least print some warnings if it's used?

> MUMPS support

Not sure the stand-alone MUMPS interface works. I started looking at updating/debugging it a while ago, but it wasn't simple (I forget why). As MUMPS can be called through PETSc anyway, are we planning to remove the stand-alone MUMPS (maybe deprecate it for the 4.3 release?)?

d7919:
There's also the ability to turn off PVODE (not super important but I like to disable it so I don't accidentally use it instead of cvode or others).

ZedThree:
@johnomotani Probably wise, on both fronts

@d7919 Yep, there's now `USE_PVODE`, defaults to `ON`.

PR #1699, opened 2019-04-29
Fieldperp I/O

johnomotani:
I wanted to be able to add a `FieldPerp` to restart files, and ended up making quite a few additions. The general idea is that a `FieldPerp` is written with a global y-index attribute, and is only read/written from processors that contain that global index.

Summary of changes:
* Read attributes of `Field`s from files if they are present (otherwise leave set to default values) [needed this to get back `yindex` values, but should be generally useful]
    * Functions to get `CELL_LOC`, `YDirectionType` or `ZDirectionType` from a `std::string` to turn the strings stored in the file back into `enum`s
* `FieldPerp` support in `Datafile`, `DataFormat`, `Ncxx4`, `Ncformat` and `H5format`
    * had to add some special `read_perp`, etc., functions to work around having no y-dimension. This seems a bit hack-y, but I couldn't think of a better way in the current structure. Suggest we aim for nicer/cleaner `FieldPerp` support in `Options` as that replaces `Datafile`...
* `FieldPerp` support in `GridFromFile` and `GridFromOptions`
    * needed to add `FieldPerp` to `test-io` and `test-io_hdf5`
* `FieldPerp` support in `collect.py`
    * again needed for the tests
* New tests `test-restart-io` and `test-restart-io_hdf5`. These generate a restart file from Python code, which is read in to 'restart' a simulation and then written out again. They test a different set of reading routines than `test-io` and `test-io_hdf5`, which read input through the `GridFile` interface. The new tests also test the `restart.redistribute` Python function.
* `FieldPerp` support in `restart.redistribute` from `restart.py`
    * used in new test cases
* [unrelated, but noticed while going through `GridFromOptions`] `GridFromOptions::get(vector<int>)` is not implemented. Previously it created a scalar `int` and then filled the whole vector with that value. I think it's better to throw an exception if this method is ever called, to signal that it needs implementing to behave as expected.
   * could easily revert or put in a separate PR if people prefer - I don't think it's a bug that's likely to affect anyone though!


johnomotani:
`WIP` because I don't think I've got the python tools and tests quite right yet...
I think I'll want to force-push to fix the test so only one binary file ends up in the history - @ZedThree is that OK? Sorry, I should've checked better before making the PR to avoid this.

johnomotani:
I've added more tests of the FieldPerp I/O now, and worked out some bugs with the Python routines. I think this is ready to be reviewed now. The tests pass, and the code works for my STORM test-case too.

include/field.hxx
```diff
@@ -136,6 +136,14 @@ class Field {
     return directions.z;
   }
 
+  /// Setters for *DirectionType
+  void setDirectionY(YDirectionType y_type) {
+    directions.y = y_type;
+  }
+  void setDirectionZ(ZDirectionType z_type) {
+    directions.z = z_type;
```
ZedThree: Can we just make `directions` public? Currently the getters and setters are trivial, so there's no invariant we need to preserve. Are we likely to add one in the future such that these might be less trivial?

manual/sphinx/developer_docs/file_io.rst
```diff
@@ -144,22 +151,116 @@ The important bits of the DataFormat interface are::
...
+global y-index in the attribute `yindex_global`. The intention is that a `FieldPerp` being
+saved should be a globally well-defined object, e.g. a set of values at one divertor
+target boundary, that will only be saved from processors holding that global y-index.
+Actually, the C++ I/O code should work fine even if a `FieldPerp` object is defined with
+different y-indices on different processors, but Python routines like `collect` and
+`restart.redistribute` will fail because they find inconsistent `yindex_global` values.
+
+`FieldPerp` objects should be added to the output files on every processor even if they
+are not allocated or used, otherwise `collect` cannot find the variable in the first
+output file to get its dimensions.
```
ZedThree: Sorry, I don't quite understand this -- it seems inconsistent with the previous paragraph? If all processors write a `FieldPerp`, doesn't that mean that `yindex_global` will be different in each dump file, which will break `collect`? Or am I missing something?

src/fileio/datafile.cxx
```diff
@@ -621,6 +636,58 @@ void Datafile::add(Field3D &f, const char *name, bool save_repeat) {
...
+  d.covar = false;
+
+  fperp_arr.push_back(d);
+
+  if (writable) {
+    // Otherwise will add variables when Datafile is opened for writing/appending
+    if (openclose) {
+      // Open the file
+      int MYPE;
+      MPI_Comm_rank(BoutComm::get(), &MYPE);
```
ZedThree: Could just be `BoutComm::rank()`

include/field.hxx
johnomotani: This does seem weird. Originally I think I'd hoped not to need the setters at all, but now that seems inconvenient at least. I think the setters in `Field` should go: there are setters in `Field3D`/`Field2D`/`FieldPerp` that are used for method-chaining like `auto result{emptyFrom(f).setDirectionY(YDirectionType::Aligned)}`. At the moment they call `Field::setDirectionY`, etc., but there's no need for them to. I think at one point we had a general `Direction` type, and the setters were checking that it was a valid y-direction, but that's gone now. I just noticed that `Field2D` is missing the setters, which it should probably have in case we want them in a template function at some point. I suggest I make a separate PR to tidy up `setDirectionY` and `setDirectionZ`.

manual/sphinx/developer_docs/file_io.rst
johnomotani: The idea is that there'll be a set of processors (connected in the x-direction) which all have a valid `FieldPerp` that they will write, all with the same `yindex_global`. All the other processors will write a `FieldPerp`, but with an invalid (i.e. sufficiently negative) `yindex_global`. Even invalid `FieldPerp`s in the dump files will have the right `bout_type` and dimensions, which is what `collect` needs from the variable in the first file.

I agree that second paragraph is not clear, I'll try and re-write. The idea is that you could write a FieldPerp from all processors (e.g. might be useful for some diagnostic/debugging purposes), but `collect` won't know what to do then (do you want `NXPE` `FieldPerp`s? if not which y-index do you want?). On the other hand, `collect` does need the `FieldPerp` variable to exist (even if invalid) in the first dump file, otherwise we'd have to search all dump files before we could say that a variable's missing. I guess that's possible, but would require more changes (and potentially be slow?).

src/sys/bout_types.cxx
```diff
@@ -13,6 +14,17 @@ const std::string& safeAt(const std::map<T, std::string>& mymap, T t) {
   return found->second;
 }
 
+template <typename T>
+const T& safeAtReverse(const std::map<std::string, T>& mymap, std::string s) {
```
ZedThree: You can drop the "reverse" -- it makes it sound like it's going to do something backwards! It's just an overload of `safeAt`

manual/sphinx/developer_docs/file_io.rst
ZedThree: Ah, ok, so there should be two sets of `FieldPerps`: those on processes with identical `yindex_global`, and those with invalid values for `yindex_global`?

ZedThree:
Thanks @johnomotani ! It's a big change for what should be a simple feature, but I think that highlights that the IO stuff needs a lot of TLC. Looks to be pretty well covered by tests, so it's probably right!

Issue #1701, opened 2019-04-30
"x + y + z + 3" as int


dschwoerer:
If the expression `x + y + z + 3` is parsed as an int - I thought it should fail, because the location is not known.

However, the current behaviour, is to return 3, and set all variables to zero.

Do we want to keep the current behaviour?

bendudson:
Yes, the default is to assume x=y=z=0 when evaluating scalars e.g. https://github.com/boutproject/BOUT-dev/blob/next/src/sys/options.cxx#L219 
Some possibilities would be to:
 - Have a generator property, something like `constant`, which is true for numbers etc. but false for `x`, `y`,`z`,`t`. When creating a generator the dependencies are known, so this property could be tracked. Some overhead in creation, and more boilerplate code in the generators
- A simpler way would be to just evaluate the value at a few different locations, and compare the result. This has a small runtime overhead, but would be quite simple to implement.

dschwoerer:
I am currently rewriting part of this, which would make it quite easy to throw if x is used in a const expression - I just wasn't sure whether we want to keep the old behaviour, or whether it is fine to have a breaking change ...

I will implement it now to keep it as is, this way we will be able to first enable warnings, and later throw ...

ZedThree:
We have something in now that deals with this, and we can deal with `x, y, z` at staggered locations. Except... I think we're still missing some bits. `FieldFactor::create*D` takes a `CELL_LOC` argument, but `GridFromOptions::get` doesn't pass the location argument.

If I change it to do so, this ends up causing problems with `getAtLoc` in `coordinates.cxx` due to the interpolation. I'm not 100% sure what's going on yet.

johnomotani:
Ah, I never thought about creating the metrics from an input file when I was writing the staggered coordinates stuff! I agree passing location to `Mesh::get()` is the way to go. It should be easy to modify hypnotoad2 to include CELL_LOCATION attributes to its outputs, so `Mesh::get()` could check the locations of `dx_ylow`, etc. (if they are given in the grid file), as well as evaluate expressions at correct locations.

PR #1705, opened 2019-05-02
FieldGenerate refactor

dschwoerer:
Rewrite of the generate, to allow to implement #1131 in a cleaner fashion.

@bendudson writes: This is now probably ready for another review, and potential merging. It uses a `Context` object to store the position `x`, `y`, `z` and time `t` variables, and other variables which can be set in the calling C++ code or in more complex expressions in the input file. This adds functionality by allowing data from physics models to be used in input expressions. 

The `Context` class is namespaced, so it's `bout::generator::Context`. 

Along with this there are a few small(ish) additions to the BOUT.inp syntax, including:
- Multi-line expressions, using the python convention that unclosed brackets continue to the next line
- Sub-expressions using `[a=.., b=...]{expression}` syntax. This also enables a kind of function call, since these variables (`a` and `b` here) are set in the `Context`. 
- Recursive expressions, which must be explicitly enabled in the input.
- A limited form of loop

These features are documented, and come with tests.

~~Currently the wave-1d test fails, but I am not sure why it fails.
wave-1d-y works fine, and the attached patch 'fixes' also wave-1d - but I am worried as I do not understand why it fails.~~

~~[wave-1d.patch.gz](https://github.com/boutproject/BOUT-dev/files/3137519/wave-1d.patch.gz)~~


bendudson:
Thanks @dschwoerer I think the parameters together in `Position` is a good way to go. It might be nice to be able to pass in additional data or parameters, so that values can be generated based on evolving variables. Expressions for viscosity which depends on the local shearing rate was something I wanted to do recently, for example. 

Given the lack of keywords in C++, would some kind of operator chaining be too gross? so we could have `.mesh()` and `.location()` as sort-of keywords, rather than positional arguments.
```
Ind3D ind;
Field3D shear = ...; // Evolving in time

value = generate(Position(ind).time(t).param("shear", shear[ind]));
```
then in the input expression could be something like 
```
viscosity = 1 + 0.1 * {shear}
```
where this `{}` operator (or some other notation) would look up "shear" in the parameters. Some way is needed to indicate to the parser (and anyone reading it) that this is a parameter to be determined at evaluation time, rather than a symbol to be looked up. Of course this is one more step towards turing completeness, sentience, world domination, ...

dschwoerer:
In general key words are fine, however I think it only makes sense for optional arguments.
I think most of the library should pass in the mesh and the indices, cause otherwise I cannot implement the `real()` modifier to position.
(Basically) all constructors in the library need to provide the mesh, the indices and the time, such that user code can benefit from that - we might want to provides different constructors to users ...

Besides that, I think string arguments for the shear example would be nicer:
`pos.set("shear", shear[ind])` as it would keep the interface cleaner ...

Though, I am slightly worried about using this in production code, due to performance - otherwise we start wanting to jit compile this, if speaking about sentience :-)

bendudson:
I suspect many of the `WARNING: Accessing zero value` messages may be caused by the default copy constructor; the `Position` objects are passed by value into the `generate` functions. 

I think I would prefer the "x" symbol to always mean the same thing, rather than depending on some other switches. The real coordinates could be implemented as a different symbol (`rx`), or another parameter if we go with a dictionary-like approach for passing values into expressions e.g. `{realx}`. I am working on a proof of concept and will make a PR soon...

dschwoerer:
The exessive message seem to be coming mostly from the unit tests, I think a lot of default constructor `Position()` is used there, and in part the behaviour of #1701 is tested.

`[ RUN      ] FieldFactoryCreationTest/0.CreateZ` + the staggered are accessing `z` in field2D - which is imho invalid.

My plan was to allow something like `real(sin(x)cos(y)-cos(z)*exp(t))` - but I have no strong preference - `realx`/`rx` would also work for me ...

src/mesh/position.cxx
```diff
@@ -0,0 +1,97 @@
...
+  case ONE:
+    if (sx) {
+      return fac * 0.5 * (msh->GlobalX(ix) + msh->GlobalX(ix - 1));
+    } else {
+      return fac * msh->GlobalX(ix);
+    }
+  case VALUE:
+    return _x;
+  case INVALID:
+    printf("WARNING: Accessing zero value\n");
```
ZedThree: `printf` -> `output_error`

Should this instead throw an exception? What does `INVALID` really imply?

src/mesh/position.cxx
```diff
@@ -0,0 +1,97 @@
...
+    // staggering info in boundary condition needs to be ignored
+    sy = true;
+    if (bndry->by < 0) {
+      iy += 1;
+    }
+  }
+};
+
+BoutReal Position::x() {
+  BoutReal fac = 1;
```
ZedThree: Maybe consider:

```cpp
// Assuming the enum in Position becomes an enum class
const BoutReal normalisation{fx == Flag::TWOPI ? TWOPI : 1.};
```

Benefits:
1. `const` so the reader knows they only ever need to read this one line to know what the value is at any later point
2. descriptive name -- `fac` requires more thinking about than `normalisation`
3. removes the body of the `TWOPI` case

src/mesh/position.cxx
```diff
@@ -0,0 +1,97 @@
...
+BoutReal Position::z() {
+  BoutReal fac = 2 * PI;
+  switch (fz) {
+  case ONE:
+    fac = 1;
+    BOUT_FALL_THROUGH;
+  case TWOPI:
+  case DEFAULT:
+    if (sz) {
+      return fac * (iz - 0.5) / static_cast<BoutReal>(msh->LocalNz);
```
ZedThree: Maybe consider moving the numerator into a temporary variable:

```cpp
const BoutReal grid_point{z_staggered ? iz - 0.5 : iz};
return normalisation * grid_point / mesh->LocalNz;
```

Now the numerator is easy to read, as everything has a name, and we've managed to remove the duplication in the return statements.

bendudson:
Ah yes I see that might be quite nice too. I worry a little that it may redefine symbols non-locally e.g. if I have
```
value = x + sin(y)
input = real( value + z )
```
are the `x` and `y` in `value` now interpreted as real space or index space? Personally I think having different symbols might be clearer. Putting them in a different namespace (e.g. `{...}`) reduces chances for confusion and could allow things like 'ix' for integer index space, 'rx' for real space and others if needed.

Thanks for doing this; I think passing in an object like this is much nicer than `x, y, z, t` separately.

include/bout/sys/position.hxx
```diff
@@ -0,0 +1,70 @@
...
+  virtual int getIx() {
+    return ix;
+  };
+  virtual int getIy() {
+    return iy;
+  };
+  virtual int getIz() {
+    return iz;
+  }
+  void setX(BoutReal val){
```
ZedThree: Maybe make these return `this` as a reference, so that we can do method chaining and put initialisation all on one line. This may mean we can then make some local uses of `Position` `const`

ZedThree:
Thanks @dschwoerer, this is looking good! Feels like this is the right direction.

A more general point: I don't think we should settle on the exact way of switching between the different coordinate systems just yet (i.e. via some globalish switch or different variable names). I can think of pros and cons for each approach, so we're going to want to actually field-trial them and really find out what does and doesn't work.
So even when this goes in, we should consider the inputs experimental

include/bout/sys/position.hxx
```diff
@@ -0,0 +1,70 @@
...
+# define BOUT_FALL_THROUGH __attribute__ ((fallthrough));
+#else
+# define BOUT_FALL_THROUGH
+#endif
+
+
+class BoundaryRegion;
+class Mesh;
+
+class Position{
```
ZedThree: `Position` is a super common class name, best stick this in a `bout` namespace -- and possibly one deeper, although I'm not sure what that should be yet

include/bout/sys/position.hxx
```diff
@@ -0,0 +1,70 @@
...
+    fz = INVALID;
+  };
+  Position(Ind3D i, CELL_LOC loc, Mesh * msh, BoutReal _t)
+    : Position(i.x(), i.y(), i.z(), loc, msh, _t){};
+  Position(int ix, int iy, int iz, CELL_LOC loc, Mesh * msh, BoutReal _t)
+    : ix(ix), iy(iy), iz(iz), _t(_t), msh(msh), sx(loc == CELL_XLOW), sy(loc == CELL_YLOW),
+      sz(loc == CELL_ZLOW){};
+  Position(): ix(-1), iy(-1), iz(-1), _t(0), msh(nullptr), _x(0), _y(0), _z(0), fx(INVALID), fy(INVALID), fz(INVALID){};
+  Position(const BoundaryRegion* bndry, int iz, CELL_LOC loc, BoutReal t, Mesh * msh);
+  Position(const BoundaryRegion* bndry, CELL_LOC loc, BoutReal t, Mesh * msh): Position(bndry, 0, loc, t, msh) {};
```
ZedThree: `bndry` presumably can't be `nullptr`, so take by reference (unless it can be `nullptr`!)

Also `bndry` -> `boundary`

src/sys/expressionparser.cxx
```diff
@@ -77,12 +83,100 @@ class FieldT : public FieldGenerator {
...
   std::string str() const override { return std::string("t"); }
 };
+
+class FieldParam : public FieldGenerator {
+public:
+  FieldParam(const std::string name) : name(name) {}
+  double generate(const Context& ctx) override {
+    return ctx.get(name); // Get a parameter
+  }
+  std::string str() const override { return std::string("{") + name + std::string("}"); }
```
ZedThree: For reference, we can now stick `using namespace std::string_literals;` at the top of the file, and then we get to use `""s` string literals:

```cpp
  std::string str() const override { return "{"s + name + "}"s; }
```

(not saying it's actually necessary to use though!)

src/sys/expressionparser.cxx
```diff
@@ -149,6 +250,41 @@ FieldGeneratorPtr ExpressionParser::parseIdentifierExpr(LexInfo& lex) const {
   string name = lowercase(lex.curident);
   lex.nextToken();
 
+  // sum(symbol, count, expr)
+  // e.g. sum(i, 10, {i}^2) -> 0 + 1^2 + 2^2 + 3^2 + ... + 9^2 
+  if (name == "sum") {
+    if (lex.curtok != '(') {
+      throw ParseException("Expecting ')' after 'sum' in 'sum(symbol, count, expr)'");
```
ZedThree: Should this be `"Expecting '(' after 'sum'...`?

src/sys/expressionparser.cxx
```diff
@@ -149,6 +250,41 @@ FieldGeneratorPtr ExpressionParser::parseIdentifierExpr(LexInfo& lex) const {
   string name = lowercase(lex.curident);
   lex.nextToken();
 
+  // sum(symbol, count, expr)
+  // e.g. sum(i, 10, {i}^2) -> 0 + 1^2 + 2^2 + 3^2 + ... + 9^2 
+  if (name == "sum") {
```
ZedThree: Just curious as to why `sum` needs hardcoding in the parser? I assume there's a reason, but it's not immediately obvious to me!

src/sys/expressionparser.cxx
bendudson: Thanks, good catch!

src/sys/expressionparser.cxx
bendudson: This is because arguments to functions are currently only `FieldGenerator` objects, rather than strings. This could be done, but would probably be more complex than just hardcoding this `sum` special case. Unfortunately BOUT.inp is not a particularly elegant language...

src/field/field_factory.cxx
```diff
@@ -57,30 +92,37 @@ FieldFactory::FieldFactory(Mesh* localmesh, Options* opt)
   transform_from_field_aligned
     = nonconst_options["input"]["transform_from_field_aligned"].withDefault(true);
 
+  // Convert using stoi rather than Options, or a FieldFactory is used to parse
+  // the string, leading to infinite loop.
+  max_recursion_depth = std::stoi(nonconst_options["input"]["max_recursion_depth"]
+                                  .doc("Maximum recursion depth allowed in expressions. 0 = no "
+                                       "recursion; -1 = unlimited")
+                                  .withDefault<std::string>("0"));
```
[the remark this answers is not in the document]
bendudson: Thanks; this now catches `std::exception` and throws `BoutException`. Thinking about this, perhaps it should be `ParseException`?

include/bout/sys/generator_context.hxx
```diff
@@ -0,0 +1,85 @@
...
+  /// This should be changed in a future release.
+  Context() { parameters["x"] = parameters["y"] = parameters["z"] = parameters["t"] = 0.0; };
+
+  /// The location on the boundary
+  Context(const BoundaryRegion* bndry, int iz, CELL_LOC loc, BoutReal t, Mesh* msh);
+  Context(const BoundaryRegion* bndry, CELL_LOC loc, BoutReal t, Mesh* msh)
+      : Context(bndry, 0, loc, t, msh){};
+
+  Context(const Context& other)
+    : localmesh(other.localmesh), parameters(other.parameters) {}
```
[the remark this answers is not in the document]
bendudson: Thanks, this is now done.

src/sys/expressionparser.cxx
bendudson: Excellent! Yes, I've now replaced all the `std::string("...")` in this file with `"..."s`

src/field/field_factory.cxx
```diff
@@ -43,6 +45,39 @@ FieldGeneratorPtr generator(BoutReal* ptr) {
...
+    /// depth_limit sets the maximum iteration depth. Set to < 0 for no limit
+    FieldIndirect(std::string name, int depth_limit = 0) : name(name), depth_limit(depth_limit) {}
+
+    /// Set the target, to be called when generator is called
+    void setTarget(FieldGeneratorPtr fieldgen) { target = fieldgen; }
+    
+    double generate(const Context& ctx) override {
+      if (depth_counter == depth_limit) {
+        throw BoutException("Calling %s to recursion depth %d exceeds maximum %d\n",
+                            name.c_str(), depth_counter, depth_limit);
```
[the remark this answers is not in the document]
bendudson: This is done now, with formatting strings changed to use `{fmt}`

include/bout/sys/generator_context.hxx
```diff
@@ -0,0 +1,85 @@
...
+  Mesh* getMesh() const {
+    ASSERT1(localmesh)
+    return localmesh;
+  }
+
+private:
+  Mesh *localmesh{nullptr}; ///< The mesh on which the position is defined
+  
+  /// Contains user-set values which can be set and retrieved
+  std::map<std::string, BoutReal> parameters;
```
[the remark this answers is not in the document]
bendudson: Thanks! This is done

src/field/field_factory.cxx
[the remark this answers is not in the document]
ZedThree: I think you might have accidentally clobbered it, it's back to old `%` style!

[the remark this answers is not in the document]
bendudson: Aargh! Thanks @ZedThree I think I've fixed it now.

PR #1706, opened 2019-05-03
Parse params

bendudson:
Adds the capability to put parameters in expressions, which are looked up when the value is generated. This allows values generated at run-time to be combined using expressions in the input file.

### DIFF
--- a/include/bout/sys/expressionparser.hxx
+++ b/include/bout/sys/expressionparser.hxx
@@ -129,7 +129,7 @@ private:
 
     LexInfo(const std::string& input, const std::string& reserved_chars = "");
 
-    /// Current token. -1 for number, -2 for string, 0 for "end of input"
+    /// Current token. -1 for number, -2 for symbol, -3 for {string}, 0 for "end of input"
     signed char curtok = 0;
     double curval; ///< Value if a number
     std::string curident;       ///< Identifier, variable or function name
--- a/include/bout/sys/position.hxx
+++ b/include/bout/sys/position.hxx
@@ -3,68 +3,91 @@
 #include "bout/region.hxx"
 
 #if defined(__cplusplus) && __cplusplus >= 201703L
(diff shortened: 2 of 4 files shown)


dschwoerer:
I tried passing by const reference, as that makes more sense - even if `real()` as modifier is chosen, it is easy enough to take a non-const copy, and pass that on. However, that does not silence the warnings ...

This patch makes the output acceptable:
```
--- a/tests/unit/quiet.sh
+++ b/tests/unit/quiet.sh
@@ -4,13 +4,23 @@ set -o pipefail
 for cmd in $@
 do
     $cmd | awk '{
+    if (match($0,"WARNING: Accessing zero value")) {
+      samecount = samecount + 1
+    } else {
+      if (samecount) {
+         print running
+         print samecount, "WARNING: Accessing zero value"
+        samecount = 0
+      }
+
     if (match($0,"^$")) {
         empty=1
         #print "empty", $0
-        if (running)
+        if (running) {
             out=1
-        else
+        } else {
             out=0
+        }
         
     } else {
         if (match($0,"\\[ *RUN *\\]")) {
@@ -41,6 +51,7 @@ do
         }
         empty=0
     }
+    }
     if (out)
         print $0
 }' || exit
```

include/bout/sys/expressionparser.hxx
```diff
@@ -129,7 +129,7 @@ private:
 
     LexInfo(const std::string& input, const std::string& reserved_chars = "");
 
-    /// Current token. -1 for number, -2 for string, 0 for "end of input"
+    /// Current token. -1 for number, -2 for symbol, -3 for {string}, 0 for "end of input"
```
ZedThree: Idea for future PR: make this `variant<signed char, enum{Number, Symbol, String, EOF}>`?

ZedThree:
With this change, `Position` feels like a bit of a misnomer. Maybe something like `Context`, or `GeneratorContext` would be more appropriate?

PR #1707, opened 2019-05-05
Fixes for FV::Div_a_Laplace_perp

bendudson:
- Extrapolation of metrics into the boundary meant that Dirichlet boundaries  applied to coefficients did not have the desired effect. Changed so that the diffusion coefficients are interpolated using simple midpoint.

- A copy-paste error using up rather than down coefficients.


ZedThree:
Do we have some tests for the FV stuff?

bendudson:
There are some convergence tests in https://github.com/boutproject/BOUT-dev/tree/master/examples/finite-volume but these issues were mainly to do with boundary conditions, which may be more tricky to test. More testing would be good...

PR #1708, opened 2019-05-06
Move pylib to separate repo and add as submodule

### LINKED ISSUE #1347: Python tools should be packaged properly
Currently all the python tools in pylib are stored in a subdirectory of the main BOUT repository. This causes a few problems:

1) The pylib is not packaged, so it can't be installed directly by python package managers such as pip and conda. The only way to import the scripts is to add them directly to your pythonpath, but this is not encouraged python practice. 

2) Not being able to use package managers properly makes tracking dependencies harder. With a proper `setup.py` file we can issue warnings about deprecated python versions, for example.

3) Any commits to BOUT++ python tools will be mixed in with commits to the main C++ codebase, which makes the git history less clear.

4) Users who want to fork just the core code or just the analysis tools have to fork both in one go, which is wasteful.

### Solution:

The best way to improve this would be to completely separate the python tools into their own repository, which is stored under [boutproject](https://github.com/boutproject), similar to the way [CHERAB](https://github.com/cherab) does it.

This repo could then by structured properly as a python package, with an informative `setup.py`, `__init__.py`'s, and a directory structure following the [recommended patterns](https://docs.python-guide.org/writing/structure/). 

Migrating all the python files to a new repo would be a breaking change in the sense that user's `$PYTHONPATH`s would no longer point to the right place, but it would not be hard to fix.

If we wanted then this package could then be uploaded to PYPI or Conda, or we could also just show people how to use pip or conda to install from their cloned source files directly. (or let them keep just adding the root directory to their pythonpaths)

Ideally the tools for each language (MATLAB etc.) would be placed in their own repo too, but it's most important to do it for python because it should be structured as a separate pure python package.

loeiten:
Refactoring `pylib` to it's own repo

This closes #1347

### DIFF
--- a/include/bout/mesh.hxx
+++ b/include/bout/mesh.hxx
@@ -276,6 +276,7 @@ class Mesh {
   /// @param[in] buffer A buffer of data to send
   /// @param[in] size   The length of \p buffer
   /// @param[in] tag    A label, must be the same at receive
+  [[gnu::deprecated("This experimental functionality will be removed in 5.0")]]
   virtual MPI_Request sendToProc(int xproc, int yproc, BoutReal *buffer, int size, int tag) = 0;
 
   /// Low-level communication routine
@@ -288,6 +289,7 @@ class Mesh {
   /// @param[inout] buffer  The buffer to fill with data. Must already be allocated of length \p size
   /// @param[in] size  The length of \p buffer
   /// @param[in] tag   A label, must be the same as send
+  [[gnu::deprecated("This experimental functionality will be removed in 5.0")]]
   virtual comm_handle receiveFromProc(int xproc, int yproc, BoutReal *buffer, int size, int tag) = 0;
   
   virtual int getNXPE() = 0; ///< The number of processors in the X direction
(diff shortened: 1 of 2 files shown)


dschwoerer:
Is pip able to build the boutcore module?

loeiten:
I'm not really sure what the boutcore really does, so it's a bit hard to tell. I guess, in theory, it should be doable, and that how hard it is depends on how much dependencies it has on the BOUT++ repo. In general, pip can build c++ binaries

ZedThree:
Just the coverage tests failing, I think because the PR is from a fork, so the CodeCov (or Travis?) tokens aren't correct.

A few things we need to think about:
- How does/should this interact with [xBOUT](https://github.com/boutproject/xBOUT)?
    - Once xBOUT is mature, we could just switch submodules?
- Should boutcore be its own separate repo?
    - I feel probably yes? Would make it easier to version
- Should we export the history into the new repo?
    - Also probably yes?

I'm prepping the 4.3 release this week, so we'll hold off on this till after then and mark it for 5.0.

dschwoerer:
boutcore provides a python wrapper for the C++ library, and thus makes most of the BOUT++ functions available in python.

It needs to link with BOUT++, and thus needs to know the compiler flags used for BOUT++.
Currently it assumes that it is in `$BOUT_TOP/tools/pylib` - so it should keep working as a submodule, but probably not in pip.

I think it makes sense to treat is differently to the rest of the pylib. Keeping it in the same repo as BOUT++ would have the advantage to have strong guarantees that the BOUT++ library is of the appropriate version, but besides that, it could also be it's own repository.

PR #1448, opened 2018-12-12
Start to introduce zstart/zend and change loops

d7919:
This builds on #1423 and aims to replace/update loops over z to use zstart and zend rather than hard-coding 0 to LocalNz.

The most effective approach to this is to replace nested loops with BOUT_FOR region loops where possible as this allows the index to be offset in z using the logic in SpecificInd rather than custom logic for each loop. This should therefore allow an easier transition from in memory periodicity to guard-cell/communication based periodicity.

As this is likely to touch a lot of code it would be helpful to have feedback as the PR evolves.

PR #1713, opened 2019-05-08
Start to introduce zstart/zend and change loops (updated)

ZedThree:
#1448 + next

The merge commit had a lot of conflicts, hence new PR.

I think maybe we should undo the changes to `ShiftedMetric` -- at least for now, we probably want to transform all of Z.


dschwoerer:
What is the best way to move this forward?
Is it worth trying to resolve the conflicts, or is it easier to start again?

Also, should this be split in several PRs, to make it easier (and thus faster) to review?

ZedThree:
We have a new RSE starting next week, this will be on their to-do list, but it might be awhile before they actually get to it.

I haven't looked at how bad the conflicts are, but I suspect that will determine how we actually proceed.

Multiple PRs might be better, but the hope with the new RSE is that we'll have a bit more bandwidth to review PRs and get them merged faster.

bendudson:
Can this be closed @ZedThree ? I suspect it's superseded by your recent Z parallelisation work.

ZedThree:
Yep, I rebased some commits and reimplemented some others, forgot to close this!

PR #1716, opened 2019-05-09
Remove deprecated invert_laplace from gyro_average

ZedThree:
Just does the "dumb" thing and essentially inlines `invert_laplace`. See #1453 for a possible nicer solution.

- Also make `INVERT_*` flags `constexpr` (didn't actually end up using this, but could be useful)
- Don't return `const` Fields from `Laplacian*::solve` or gyro-averages

There is an integrated test, `test-gyro`, that tests a couple of these routines, but it's just a regression test.

Anybody fancy writing down an analytic solution we could test all of them with?

### DIFF
--- a/src/physics/gyro_average.cxx
+++ b/src/physics/gyro_average.cxx
@@ -29,6 +29,7 @@
 
 #include <bout/mesh.hxx>
 #include <globals.hxx>
+#include <bout/sys/timer.hxx>
 #include <difops.hxx>
 #include <gyro_average.hxx>
 #include <invert_laplace.hxx>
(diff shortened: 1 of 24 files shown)


src/physics/gyro_average.cxx
```diff
@@ -28,84 +28,118 @@
...
-  Field2D d = -rho*rho;
-  
+Field3D gyroPade0(const Field3D& f, BoutReal rho, int flags) {
+  const Field2D a = 1.0;
+  const Field2D d = -rho * rho;
+
   // Invert, leaving boundaries unchanged
-  return invert_laplace(f, flags, &a, nullptr, &d);
+
+  Timer timer("invert");
```
d7919: Does this not double count the invert time here?

src/physics/gyro_average.cxx
```diff
@@ -28,84 +28,118 @@
...
   // Have to use Z average of rho for efficient inversion
   return gyroPade0(f, DC(rho), flags);
 }
 
-const Field3D gyroPade1(const Field3D &f, BoutReal rho, int flags) {
-  Field2D a = 1.0;
-  Field2D d = -0.5*rho*rho;
-  
+Field3D gyroPade1(const Field3D& f, BoutReal rho, int flags) {
+  const Field2D a = 1.0;
```
d7919: Can we just pass 1 to `setCoefA` like we do for `setCoefC`?

src/physics/gyro_average.cxx
```diff
@@ -28,84 +28,118 @@
...
+Field3D gyroPade0(const Field3D& f, BoutReal rho, int flags) {
+  const Field2D a = 1.0;
+  const Field2D d = -rho * rho;
+
   // Invert, leaving boundaries unchanged
-  return invert_laplace(f, flags, &a, nullptr, &d);
+
+  Timer timer("invert");
+
+  auto* lap = Laplacian::defaultInstance();
```
d7919: Can/should we allow an options section to be passed through should we want to use a non-default Laplacian setup?

PR #1719, opened 2019-05-16
More robust staggered Jacobian and RGN_NOCORNERS

johnomotani:
Fabio found that when calculating the Jacobian `J` at `CELL_YLOW` it is more robust to interpolate `J` from the cell-centre `J`, rather than calculating `J` from the interpolated metric coefficients; calculating from interpolated metric coefficients resulted in a negative `J` at ylow near the X-point, while interpolating `J` itself gave a result that was always positive. This PR also treats `Bxy` similarly for consistency, as `Bxy` is calculated from `J`.

Also add a `RGN_NOCORNERS` to the default regions created in `Mesh::createDefaultRegions()`, as well as `RGN_XGUARDS`, `RGN_YGUARDS` and `RGN_ZGUARDS` which are used to build `RGN_NOCORNERS`. `RGN_NOCORNERS` includes all the x- and y-guard cells (internal and boundary) except the corner guard-cells. This is useful for checking in the `Coordinates` constructors as `interpolateAndExtrapolate` sets the corner cells to `BoutNaN`, but it's still useful to check the guard cells.
Re-name the region created in `invertable_operator.hxx` from `RGN_NOCORNERS` to `RGN_WITHBNDRIES` as this region excludes internal guard cells, and only includes the boundary guard-cells.

Added some utility functions `checkFinite(f)` and `checkPositive(f)` that do similar things to `finite(f)` and `min(f) > 0.` but throw exceptions instead of returning a bool. The exception messages include the indices of the point where the check failed - added `toString` functions for `Ind3D`, `Ind2D` and `IndPerp` to support this.

The changes to the Jacobian calculation mean that the `Coordinates::jacobian()` method uses the `GridDataSource` from `Mesh` (to check whether it's necessary to extrapolate into boundary guard cells). This broke the tests as `source` was `nullptr` for the mesh in `FakeMeshFixture` - add a `FakeGridDataSource` to `test_extras.hxx` to get around this. `FakeGridDataSource` does not actually supply data, the `get(...)` methods all just return `false`, but gets around the segfault in `jacobian()`. It might be nice at some point to use `FakeGridDataSource` to actually supply `g11`, etc. so that we could unit-test the `Coordinates` constructors, but that's beyond the scope of this PR...

PR #1720, opened 2019-05-16
Templated Field functions and deprecate REGION enum arguments

johnomotani:
Adding `"RGN_NOCORNERS"` in #1719 pushed me to want to be able to use `const std::string& region` arguments everywhere instead of `REGION region`. I think that was the long-term plan anyhow. I've moved most of the functions `abs()`, `sin()`, etc. into `field.hxx` as template functions (using `bout::utils::EnableIfField` to make sure they don't clash with functions of non-`Field` variables - was very glad when I found that in the comments at the top of `field.hxx`! :smile:). There are deprecated overloads with `REGION region` argument for all the changed functions.

This PR reduces the total number of lines of code by a reasonable amount already, and I guess it should be at least a few hundred more once the deprecated forms are deleted. It is a bit big though - might be a good plan to leave off merging until after 4.3, although I think it's ready to go :crossed_fingers:


include/bout/index_derivs_interface.hxx
```diff
@@ -312,17 +312,17 @@ T VDDY(const T& vel, const T& f, CELL_LOC outloc = CELL_DEFAULT,
...
                                                                     method, region);
   } else {
-    const T f_aligned = toFieldAligned(f, RGN_NOX);
-    const T vel_aligned = toFieldAligned(vel, RGN_NOX);
+    const T f_aligned = toFieldAligned(f, "RGN_NOX");
+    const T vel_aligned = toFieldAligned(vel, "RGN_NOX");
     T result = flowDerivative<T, DIRECTION::Y, DERIV::Upwind>(vel_aligned, f_aligned,
                                                               outloc, method, region);
-    return fromFieldAligned(result, region);
+    return fromFieldAligned(result, toString(region));
```
ZedThree: Is `toString` needed here?

include/derivs.hxx
```diff
@@ -50,12 +90,9 @@
...
 ///                    If not given, defaults to RGN_NOBNDRY
-const Field3D DDX(const Field3D& f, CELL_LOC outloc = CELL_DEFAULT,
-                  const std::string& method = "DEFAULT", REGION region = RGN_NOBNDRY);
-inline const Field3D DDX(const Field3D& f, CELL_LOC outloc, DIFF_METHOD method,
-                         REGION region = RGN_NOBNDRY) {
-  return DDX(f, outloc, toString(method), region);
-};
+const Field3D DDX(const Field3D& f, CELL_LOC outloc = CELL_DEFAULT, const std::string&
+    method = "DEFAULT", const std::string& region = "RGN_NOBNDRY");
+DERIV_FUNC_REGION_ENUM_TO_STRING(DDX, Field3D)
```
ZedThree: As we're touching this might, may as well drop the `const` from the return value (similarly in the macro)

ZedThree:
I think you've missed some uses of the deprecated `toFieldAligned` in `fv_ops.hxx` btw

We probably want this in 4.3 so we can remove all the deprecated overloads in 5.0. I'm not sure we can deprecate the `REGION` enum completely right now, as we're using it all over the place in these deprecated functions! We might have to remove these in 5.0 then deprecate `REGION` and then remove that in 5.1

include/field.hxx
```diff
@@ -239,15 +243,15 @@ inline bool areFieldsCompatible(const Field& field1, const Field& field2) {
 
 /// Return an empty shell field of some type derived from Field, with metadata
 /// copied and a data array that is allocated but not initialised.
-template<typename T>
+template<typename T, typename = bout::utils::EnableIfField<T>>
```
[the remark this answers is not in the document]
johnomotani: :+1: I'll update this PR when we've merged #1730.

include/derivs.hxx
johnomotani: Do we want to drop the `const`s in `derivs.cxx` as well?

johnomotani: Ignore that, compiler error tells me we do drop those `const`s too...

include/field.hxx
```diff
@@ -266,7 +270,389 @@ inline T filledFrom(const T& f, BoutReal fill_value) {
...
+}
+template<typename T, typename = bout::utils::EnableIfField<T>>
+[[gnu::deprecated("Please use fromFieldAligned(const T& f, "
+    "const std::string& region = \"RGN_ALL\") instead")]]
+inline T fromFieldAligned(const T& f, REGION region) {
+  return fromFieldAligned(f, toString(region));
+}
+
+template<typename T, typename = bout::utils::EnableIfField<T>>
+inline BoutReal min(const T& f, bool allpe = false, const std::string& rgn = "RGN_NOBNDRY") {
```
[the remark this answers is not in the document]
johnomotani: I think I've updated this in the appropriate places now. @ZedThree could you check it looks sensible?

include/derivs.hxx
johnomotani: I guess we want to do the same in `coordinates.hxx`/`coordinates.cxx` too? I've made that update in the last commit.

johnomotani:
I've updated the deprecated function calls in tests and examples, and fixed a few other compiler warnings while I was there.

include/bout/coordinates.hxx
```diff
@@ -139,101 +139,96 @@ public:
...
   ///////////////////////////////////////////////////////////
 
-  const Field2D DDX(const Field2D& f, CELL_LOC outloc = CELL_DEFAULT,
-                    const std::string& method = "DEFAULT", REGION region = RGN_NOBNDRY);
-  const Field2D DDX(const Field2D& f, CELL_LOC outloc, DIFF_METHOD method,
-                    REGION region = RGN_NOBNDRY) {
+  Field2D DDX(const Field2D& f, CELL_LOC outloc = CELL_DEFAULT,
+      const std::string& method = "DEFAULT", const std::string& region = "RGN_NOBNDRY");
+  Field2D DDX(const Field2D& f, CELL_LOC outloc, DIFF_METHOD method,
+      const std::string& region = "RGN_NOBNDRY") {
```
ZedThree: We still need to provide the `REGION` overloads, just deprecated, before we remove them after 4.3

include/bout/index_derivs_interface.hxx
```diff
@@ -198,71 +198,71 @@ T D4DX4(const T& f, CELL_LOC outloc = CELL_DEFAULT, const std::string& method =
...
     ASSERT1(f.getDirectionY() == YDirectionType::Standard);
     return standardDerivative<T, DIRECTION::YOrthogonal, DERIV::Standard>(f, outloc,
                                                                           method, region);
   } else {
-    const T f_aligned = toFieldAligned(f, RGN_NOX);
+    const T f_aligned = toFieldAligned(f, "RGN_NOX");
     T result = standardDerivative<T, DIRECTION::Y, DERIV::Standard>(f_aligned, outloc,
                                                                     method, region);
-    return fromFieldAligned(result, region);
+    return fromFieldAligned(result, toString(region));
```
ZedThree: `region` is already a `string` here :) Similarly in the next few functions

ZedThree:
PR test is failing because next was failing. Restarted now that next is fixed, though I can't remember if Travis will use the updated next. We'll find out!

Also, for reference, this removes `REGION` from the derivative factory stuff entirely. It wasn't immediately obvious how one might keep a deprecated `REGION` version around at the same time as the `string` without massive code duplication.

PR #1721, opened 2019-05-17
fix nc-format

dschwoerer:
Makes it compile again :+1: 

Unit tests fail though, because, if I run them as root, I can safe the pid to `/` without issues ...
(failing code:`  EXPECT_THROW(bout::experimental::savePIDtoFile("/", 2), BoutException);`)

### DIFF
--- a/src/fileio/impls/netcdf/nc_format.cxx
+++ b/src/fileio/impls/netcdf/nc_format.cxx
@@ -489,10 +489,13 @@ bool NcFormat::addVarFieldPerp(const string &name, bool repeat) {
   if (!(var = dataFile->get_var(name.c_str()))) {
     // Variable not in file, so add it.
     auto nc_float_type = lowPrecision ? ncFloat : ncDouble;
-    if (repeat)
-      var = dataFile->add_var(name.c_str(), nc_float_type, 3, {tDim, xDim, zDim});
-    else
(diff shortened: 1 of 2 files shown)


dschwoerer:
`test-io` seg-faults - and this seems to fix it - however I am unsure if we should check all return values of `values()` for `nullptr`s ...
```
@@ -1272,12 +1275,14 @@ bool NcFormat::getAttribute(const std::string &varname, const std::string &attrn
       throw BoutException("Variable '%s' not in NetCDF file", varname.c_str());
     }
 
-    NcAtt* varAtt;
-    if (!(varAtt = var->get_att(attrname.c_str())))
+    NcAtt* varAtt  = var->get_att(attrname.c_str());
+    if (varAtt == nullptr)
       return false;
 
-    value = varAtt->values()->as_int(0);
-
+    auto values = varAtt->values();
+    if (values == nullptr)
+      return false;
+    value = values->as_int(0);
     return true;
   }
 }
```

ZedThree:
Thanks! We're going to deprecate the legacy interface in 4.3 and remove it in 5.0 (maybe just move it out into a separate, unmaintained, repo), but it's nice to make sure it works for now.

Also, I'm ok with saying we don't support running as root. If a user wants to do so, then on their heads be it.

dschwoerer:
Could we maybe update the instructions for ubuntu to use the non-legacy interface, which does not segfault?
It would be great if somebody using ubuntu could fix that ...

PR #1723, opened 2019-05-21
Remove Laplace3D

ZedThree:
I think this was an experimental feature that never really made it? @bshanahan Is this needed for the 3D metrics?


johnomotani:
I think `Laplace3D` can go. When @JosephThomasParker and I have looked at a 3d solver including the y-derivatives in `Laplace_perp`, it seemed better to put it as part of the `Laplacian` interface anyway. It fits well, overrides `Laplacian::solve(Field3D)` and throws an exception for `Laplacian::solve(FieldPerp)`. I think the only possible extra thing for `Laplace3D` was allowing the possibility of solving for `Laplace` not just `Laplace_perp`. I'm not sure we want/need to do that, but if we ever do, it could just be another coefficient added to `Laplacian` as an extension to a 3d solver, rather than a whole separate interface. So I don't think we'd be restricting ourselves to put any 3d solver as a sub-class of `Laplacian`.

bshanahan:
@ZedThree Not used in 3D metrics -- I mostly just use LaplaceXZ.

ZedThree:
Thanks both! Ok, looks like removing it is the right thing to do then

PR #1725, opened 2019-05-21
Bugfix: use region in Coordinates::calcCovariant/calcContravariant

ZedThree:
Extracted from #1696 

I can't remember exactly how I was triggering this bug, but basically `interpolateAndExtrapolate` in `Coordinates` sets the corner values of the metric components to `NaN`. In `calcCovariant/calcContravariant`, we loop over the whole domain and call `invert3x3` which obviously has `NaN`s in the corners now.

Somehow, the tests on Travis pass fine -- I think it's some combination of config options that's missing and that causes the issue.

Also, these should use region loops anyway.

@johnomotani I've tagged you as you're basically the de-facto owner of `Coordinates` at the minute :smile:


src/mesh/coordinates.cxx
```diff
@@ -313,13 +313,13 @@ Coordinates::Coordinates(Mesh* mesh, Options* options)
       output_warn.write("Not all covariant components of metric tensor found. "
                         "Calculating all from the contravariant tensor\n");
       /// Calculate contravariant metric components if not found
-      if (calcCovariant()) {
+      if (calcCovariant("RGN_NOBNDRY")) {
```
johnomotani: I think we want `RGN_NOCORNERS` from #1719 here: `interpolateAndExtrapolate` won't always fill the boundary cells - it doesn't extrapolate if the boundary cells were read from the grid file - so then we need to call invert3x3 in the boundary cells (but not the corners).

include/bout/coordinates.hxx
```diff
@@ -114,8 +114,10 @@ public:
   /// Calculate differential geometry quantities from the metric tensor
   int geometry(bool recalculate_staggered = true,
       bool force_interpolate_from_centre = false);
-  int calcCovariant(); ///< Inverts contravatiant metric to get covariant
-  int calcContravariant(); ///< Invert covariant metric to get contravariant
+  /// Invert contravatiant metric to get covariant components
+  int calcCovariant(const std::string& region = "RGN_ALL");
+  /// Invert covariant metric to get contravariant components
+  int calcContravariant(const std::string& region = "RGN_ALL");
```
johnomotani: If we use `RGN_NOCORNERS` maybe we should make that the default argument here.

src/mesh/coordinates.cxx
```diff
@@ -900,32 +900,27 @@ int Coordinates::calcCovariant() {
...
-      a(0, 1) = a(1, 0) = g12(jx, jy);
-      a(1, 2) = a(2, 1) = g23(jx, jy);
-      a(0, 2) = a(2, 0) = g13(jx, jy);
-
-      // invert
-      if (invert3x3(a)) {
-        output_error.write("\tERROR: metric tensor is singular at (%d, %d)\n", jx, jy);
-        return 1;
-      }
+  BOUT_FOR_SERIAL(i, g11.getRegion(region)) {
```
johnomotani: Is there a reason for this to be `BOUT_FOR_SERIAL` rather than `BOUT_FOR`? Not performance critical, since this loop is only needed for initialization - i'm just curious.

ZedThree: Yes, there's a `return` (if `invert3x3` fails) which is not allowed inside an OpenMP region

include/bout/coordinates.hxx
ZedThree: I'm leaving this as `RGN_ALL` as it keeps the previous behaviour and I think it's the least surprising as a default. Within the `Coordinates` ctors, we know we're setting the corners to NaN, but we don't know that in the general case.

ZedThree:
I've had to make a couple of seemingly unrelated change to make `test-laplace` pass with `--enable-sigfpe` due to one floating point exception in BOUT++ and one in NetCDF: 

- Don't do non-uniform correction in Laplacian
  - Corners of geometry variables are explicitly set to NaNs, and
    non-uniform correction involves derivative of dx in y-guards
- Don't convert to floats in output in test-laplace
  - NetCDF does a comparison when converting from double -> float to
    check for out-of-range errors. This comparison signals with NaNs
- Loosen tolerance in test-laplace
  - Benchmark fields are floats, but tolerance is smaller than
    single-precision! (this was my fault, d'oh!)

Incidentally, I came across two other thing for concern:

1. The interfaces for `DataFormat::write` (and `write_rec`) take `lx, ly, lz` which all default to `0`. When we write `Field2D`s we don't pass `lz`, and then loop over `lx * ly * lz` to remove NaNs and truncate large values if writing floats, which obviously don't get done for `Field2D`s.
2. We remove NaNs and truncate large values if writing floats in the actual field data!

We shouldn't be changing the actual field data when writing out, but also we probably don't want to actually truncate or remove NaNs anyway. Not doing so would also solve the first problem.
Something for the great DataFile rewrite!

johnomotani:
In `test-laplace` it might be better to keep the non-uniform correction, but remove the `include_yguards = true` option.

`include_yguards = true` solves in the y-boundary guard cells (it is not well named) - I think we should deprecate the option, and we should at least print some warnings if it is switched on, because the user will have to mess around with the corner-guard-cell values of `C2` and `dx` if they want to use it. The option exists to support the old 'free' boundary conditions that solved with one-sided differences in the boundary cells, but I don't think we support those any more (`inner_boundary_deriv_func` and `outer_boundary_deriv_func` stuff has been removed). We should probably remove the `BoundaryFree` class as well - it looks like it doesn't do anything any more.

ZedThree:
`test-laplace` doesn't set any of the geometry terms, so it uses `dx = 1.0`, so the non-uniform correction isn't needed here.

I tried setting `include_yguards = false`, but the problem is in the guard corners so it unfortunately has no effect. In order to turn on `non_uniform` we need to actually not include the y-guards (or put correct values in the geometry corners), which is currently not possible.

Agreed we should deprecate or remove `include_yguards`.

johnomotani:
@ZedThree setting `include_yguards = false` with `nonuniform = true` on this branch (with `--enable-sigfpe` set) lets the test pass for me now, while it fails with both `include_yguards = true` and `nonuniform = true`. Maybe you fixed something since you tested `include_yguards = false`? Even though `nonuniform=false` is correct for this test, I'd be in favour of having it set `true` for the test to catch exactly these kind of errors - especially as `nonuniform = true` is the default and usually needed to be correct.

ZedThree:
Ah, my bad. I think at that point I hadn't realised that there were two separate issues with the NaNs. Fixed, hopefully ready to go in now

PR #1728, opened 2019-05-22
Pybind experiment

d7919:
For discussion.

Adds pybind11 as a submodule and provides an example of how to use this to use python within a C++ example. The commit history also includes a hacked example of using this as a part of the field generators so that we could use python as a part of the field function generation. This has been reverted in the final version as it was very much a hack.


dschwoerer:
Needs to be adopted for cmake.

PR #1730, opened 2019-05-22
Add bout::utils::is_Field and variants

ZedThree:
`bout::utils::EnableIfField` is now implemented in terms of `bout::utils::is_Field`. There are also variants for `is_Field2D`, `is_Field3D` and `is_FieldPerp`, along with the corresponding versions of `EnableIfField*`

Also add static tests for these traits

All these traits are now in `include/bout/traits.hxx`, which is completely stand alone as it forward-declares the various `Field`s.  It might also make sense to move the recently merged `function_traits` to the same header. Thoughts?

The documentation for `EnableIfField` now also includes some guidance on when it is useful.

### DIFF
--- a/tests/unit/include/bout/test_traits.cxx
+++ b/tests/unit/include/bout/test_traits.cxx
@@ -1,0 +1,129 @@
+#include "gtest/gtest.h"
+
+#include "field.hxx"
+#include "field2d.hxx"
+#include "field3d.hxx"
+#include "fieldperp.hxx"
+#include "bout/traits.hxx"
+
+#include <type_traits>
+
+TEST(BoutTraitsTest, IsField) {
+  using namespace bout::utils;
+  static_assert(is_Field<Field>::value, "is_Field<Field> should be true");
+  static_assert(is_Field<Field2D>::value, "is_Field<Field2D> should be true");
+  static_assert(is_Field<Field3D>::value, "is_Field<Field3D> should be true");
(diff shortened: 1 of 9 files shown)


tests/unit/include/bout/test_traits.cxx
```diff
@@ -0,0 +1,129 @@
...
+                                                      std::declval<Field2D>()))>::value,
+      "EnableIfField should return Field2D for two Field2Ds");
+  static_assert(
+      std::is_same<Field3D, decltype(example_function(std::declval<Field2D>(),
+                                                      std::declval<Field3D>()))>::value,
+      "EnableIfField should return Field3D for a Field2D and a Field3D");
+  static_assert(
+      std::is_same<Field3D, decltype(example_function(std::declval<Field3D>(),
+                                                      std::declval<Field2D>()))>::value,
+      "EnableIfField should return Field3D for a Field2D and a Field3D");
```
johnomotani: Is there a reason not to check (`Field3D`,`Field3D`)->`Field3D` here? i.e.
```
  static_assert(
      std::is_same<Field3D, decltype(example_function(std::declval<Field3D>(),
                                                      std::declval<Field3D>()))>::value,
      "EnableIfField should return Field3D for a Field3D and a Field3D");
```

ZedThree: Nope, I just got distracted writing the other tests! The syntax here is pretty disgusting and it took me awhile to get it right

PR #1736, opened 2019-05-23
Output ParallelTransform variables

### LINKED ISSUE #426: paralleltransform and shiftOutput should be saved in output files by default
This information is important for interpreting output files (was e.g. the shifted metric procedure used)

Implementation of this requires capability to write string to net-cdf files (see #222   )

johnomotani:
Add a method outputVars(Datafile& file) to ParallelTransform, which is called from Coordinates::outputVars(Datafile& file) so that the ParallelTransform implementations can output variables to the dump files. Now ShiftedMetric saves zShift.

Resolves #426.

@ZedThree if you want to add `FCITransform::outputVars` feel free to push it here.

### DIFF
--- a/include/bout/paralleltransform.hxx
+++ b/include/bout/paralleltransform.hxx
@@ -56,6 +56,9 @@ public:
 
   virtual bool canToFromFieldAligned() = 0;
 
+  /// Output variables used by a ParallelTransform instance to the dump files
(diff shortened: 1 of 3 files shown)


ZedThree:
The FCI transform does calculate some things that might be interesting to dump to file, but they currently aren't stored, so I'm not sure how best to sort that out.

Issue #1665, opened 2019-04-04
Crashes in unit tests due to mpark variant


dschwoerer:
I tried to run the unit tests with mpark variant packaged by fedora (1.3.0) - and after fixing that the unit tests checked out mpark, I got illegal instruction error when running the unit tests.

With 1.4.0 I get a segmentation fault.

The default version works fine.

My question is, should the other versions work as well?

ZedThree:
Works for me with mpark commits 2931971 (1.3), 4988879 (1.4) and 0b488da (current master).
I've also tried the single header version from the 1.3 release.

So, yes, it should work with any version -- actually, just tried e1ddb34, (1.1.0) and that's the earliest that worked for me, so anything v1.1.0+ should work.

Can you check against the github version, or check if fedora are doing something weird to the source files?

dschwoerer:
I have tried both the version in fedora, and 1.4 from upstream, as well as 1.3 from upstream, directly checking it out in the sub-module.

I tried to reproduce it now, with current next, but it worked without issues, so it seems whatever caused this, was fixed in next ...

PR #1738, opened 2019-05-24
ensure make clean cleans tests

dschwoerer:
Otherwise `make clean check` does not force a rebuild of the tests,
which can cause confusion, as it hides or causes bugs. (potentially #1665 )

Further as we are printing before cleaning, it should be 'cleaning'
rather then 'cleaned'.


ZedThree:
Something else is needed too -- this does now run `make clean` in `tests/integrated`, but that appears to be a no-op. Running `make clean` in an individual test directory does clean it though

dschwoerer:
should be fixed now, but not sure it is needed, if we switch to cmake ...

PR #1740, opened 2019-05-24
Move boutdata and boututils to separate repo and add as submodule

### LINKED ISSUE #1347: Python tools should be packaged properly

loeiten:
This is the continuation of #1708, with the following differences:

1. The base branch now follows the [latest commit `BOUT-dev/next`](https://github.com/boutproject/BOUT-dev/commit/50e3b46b2bea4a8c207aeb7bfe9b42b89baa7969)
2. Only `boutdata` and `boututils` have been converted to submodules

Although #1708 was tagged with 5.0 one can consider if it is better update the submodules as updates are made to [`boututils`](https://github.com/boutproject/boututils) and [`boutdata`](https://github.com/boutproject/boutdata) and replace the submodules for pip installs at 5.0 instead. In this way, I think duplicated work is minimised, and the risk of diverging repos are kept to a minimum.

This closes #1347


loeiten:
This fails as the submodules now have an additional directory layer, meaning that what was `tools/pylib/boutdata` now becomes `tools/pylib/boutdata/boutdata`.

As the submodules are packaged, I suggest to remove them and add a pip install in `.travis.yml` and update the text after `configure` from "Add this to python path" to "Install these packages". Does this sound like a good idea @ZedThree and @TomNicholas?

dschwoerer:
Possible workaround: move the submodules somewhere else, and provide compat symlinks.

loeiten:
That is a possible workaround :). However, would we still like to have these files in the repo as they now can be obtained as packages?

loeiten:
Closed due to https://github.com/boutproject/BOUT-dev/issues/1347#issuecomment-602747944

PR #1741, opened 2019-05-26
Correct calculation of Bxy when interpolating from CELL_CENTRE

johnomotani:
Error was a typo introduced in #1719.

Thanks to Fabio Riva for spotting!


ZedThree:
It looks like we hit this in the unit tests but don't actually test it properly. We should fix that!

PR #1744, opened 2019-06-03
Laplacian: FFT include dfdz term, plus manual updates

johnomotani:
This PR fixes a bug that shouldn't affect anyone (thanks @friva000 for pointing out to me why!): the FFT solvers previously neglected the `1./C1*g13*DDX(C2)*DDZ(f)` term. We should almost always solve the laplacian in x-z orthogonal coordinates where `g13=0`, so this bug shouldn't be important.

The main point of this PR was to update the manual to note that the FFT solvers now support separate `C1` and `C2` coefficients, not just `C=C1=C2` as before. I came across the 'bug' above in the process. I've also updated the manual to mention the `multigrid` and `cyclic` Laplace-solvers, and to note that the `invert_laplace` functions are deprecated.

### DIFF
--- a/manual/sphinx/user_docs/coordinates.rst
+++ b/manual/sphinx/user_docs/coordinates.rst
@@ -1,5 +1,7 @@
 .. default-role:: math
 
+.. _sec-field-aligned-coordinates:
+
 =========================
 Field-aligned coordinates
 =========================
@@ -14,8 +16,6 @@ Introduction
 
 This manual covers the field-aligned coordinate system used in many
 BOUT++ tokamak models, and useful derivations and expressions.
-
-.. _sec:coordinates:
 
 Orthogonal toroidal coordinates
 ===============================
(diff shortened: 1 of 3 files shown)


ZedThree:
LGTM, thanks!

I've built this branch on RTD: https://bout-dev.readthedocs.io/en/laplacian-fft-dfdz/user_docs/laplacian.html

@johnomotani Does that look like what you were expecting? It looks like there's a couple of places where `c1` and `c2` need subscripts

While you're touching this file, would you mind if I pushed a couple of fixes for grammar errors I've just spotted?

johnomotani:
@ZedThree you're right, there were some typos with c1 and c2. I think I've fixed them now. Feel free to push grammar fixes! :+1:

ZedThree:
PR test has completed successfully, just not getting reported

PR #1746, opened 2019-06-05
Pin packages in requirements.txt for Sphinx docs

ZedThree:
The latest version of `breathe`, 4.13, is incompatible with the
version of `sphinx` on Readthedocs


ZedThree:
Although possibly `breathe` has been recently updated to pull in Sphinx 2, and rtd then uses that -- but rtd does not officially support Sphinx 2 yet. Seems to work about a quarter of the time :man_shrugging:

PR #1748, opened 2019-06-12
SNB nonlocal heat flux model

bendudson:
Nonlocal (kinetic) corrections to the Spitzer-Harm collisional heat flux, which become important when the mean free path becomes a small (~1%) fraction of the temperature scale length. Implemented in SI units, with an example and manual section.

Includes examples and manual, some tests. The `conduction-snb` examples show similar results to previous work on SNB and comparison to kinetic codes, and include some reference data which is put on the same plot. An integrated test `test-snb` runs some simple sanity and symmetry checks. These could be unit tests, but require MPI for now.


include/bout/snb.hxx
```diff
@@ -0,0 +1,63 @@
...
+  /// Ne: Electron density in m^-3
+  ///
+  /// Div_Q_SH_out : An optional output field to store the Spitzer-Harm heat flux
+  ///
+  /// Returns the divergence of heat flux in units of eV per cubic meter per second
+  /// -> multiply by e=1.602e-19 to get Watts per cubic meter.
+  Field3D divHeatFlux(const Field3D &Te, const Field3D &Ne, Field3D *Div_Q_SH_out = nullptr);
+  
+private:
+  InvertPar *invertpar; ///< Parallel inversion of tridiagonal matrices
```
ZedThree: This could be `std::unique_ptr<InvertPar> invertpar{nullptr}`, with the assignment above becoming:

```cpp
  invertpar = std::unique_ptr<InvertPar>{InvertPar::Create()};
```

We can then default the destructor and move constructor and move-assignment operator:

```cpp
  HeatFluxSNB(HeatFluxSNB&&) = default;
  HeatFluxSNB& operator=(HeatFluxSNB&&) = default;
  ~HeatFluxSNB() = default;
```

We can't get away with just not mentioning them at all as we've got a default constructor. Using a `unique_ptr` also means we can't have a copy constructor, unless we do something like: https://stackoverflow.com/a/43263477/2043465

ZedThree:
I've put `HeatFluxSNB` into the top-level `bout::` namespace.

PR #1749, opened 2019-06-13
Bug fix, test and tidy SLEPc solver

ZedThree:
- Fix bug in SLEPc solver (3ff8db9): use `VecGetArrayRead` instead of `VecGetArray`
- Fix eigen-box example (default value `dump_on_restart` causing extra point in `t_array`, which was getting confused for half an eigenvalue)
- Various improvements to the eigen-box visualisation script (pep8, graph titles, etc.)
- New test for SLEPc solver: check eigenvalue of `f * exp(t)` is `0 + 1i`
- Use new `Options` interface in SLEPc
- Various small improvements to SLEPc code, including clang-format

One for the wishlist: really separate eigen solvers from time solvers. Putting the eigenvalues in `t_array`, means they can only be accessed via the dump file, and not from with the library.

### DIFF
--- a/src/solver/impls/slepc/slepc.cxx
+++ b/src/solver/impls/slepc/slepc.cxx
@@ -23,94 +23,93 @@
  *
  **************************************************************************/
 
-
 #ifdef BOUT_HAS_SLEPC
 
 #include "slepc.hxx"
-
+#include <boutcomm.hxx>
 #include <globals.hxx>
+#include <interpolation.hxx>
+#include <msg_stack.hxx>
+#include <output.hxx>
 
 #include <cstdlib>
 
-#include <interpolation.hxx> // Cell interpolation
-#include <msg_stack.hxx>
-#include <output.hxx>
-#include <boutcomm.hxx>
-
 std::string formatEig(BoutReal reEig, BoutReal imEig);
 
-//The callback function for the shell matrix-multiply operation
-//A simple wrapper around the SlepcSolver advanceStep routine
-PetscErrorCode advanceStepWrapper(Mat matOperator, Vec inData, Vec outData){
+// The callback function for the shell matrix-multiply operation
+// A simple wrapper around the SlepcSolver advanceStep routine
+PetscErrorCode advanceStepWrapper(Mat matOperator, Vec inData, Vec outData) {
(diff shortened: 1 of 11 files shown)


src/solver/impls/slepc/slepc.cxx
```diff
@@ -23,202 +23,209 @@
...
-  options->get("nEig",nEig,0);//0 means keep the current value, i.e. autoset
+  nEig = options_ref["nEig"]
+             .doc("Number of eigenvalues to compute. 0 means keep the current value, "
+                  "i.e. autoset")
+             .withDefault(0);
 
-  options->get("tol",tol,1.0e-6);// Tolerance --> Note this is on SLEPc eig not BOUT
-  options->get("maxIt",maxIt,PETSC_DECIDE);
+  tol = options_ref["tol"].doc("SLEPc tolerance").withDefault(1.0e-6);
+  maxIt = options_ref["maxIt"].doc("Maximum iterations").withDefault(PETSC_DEFAULT);
```
d7919: I notice this has changed from `PETSC_DECIDE` to `PETSC_DEFAULT` is this intentional (and backwards compatible)?

src/solver/impls/slepc/slepc.cxx
```diff
@@ -336,437 +336,428 @@ void SlepcSolver::vecToFields(Vec &inVec){
       2. ?
    */
 
-  //Get pointer to data
-  PetscScalar *point;
-  VecGetArray(inVec,&point);
+  // Get pointer to data
+  const BoutReal* point;
+  VecGetArrayRead(inVec, &point);
```
d7919: Is this backwards compatible?

src/solver/impls/slepc/slepc.cxx
ZedThree: Yes -- this is what the docs say to use now, although in practice either is acceptable (had to check the source code to actually find out). Backwards compatible to 2006, which is much earlier than we support.

src/solver/impls/slepc/slepc.cxx
ZedThree: I'm pretty sure, yes -- we do the same thing in the PETSc solver

PR #1751, opened 2019-06-17
Build pdf manual

johnomotani:
Building a pdf version of the manual was broken because of some LaTeX errors. At least one of these errors also stopped an equation appearing in the html version (in the 'Differential geometry' section of https://bout-dev.readthedocs.io/en/latest/user_docs/coordinates.html, after 'which can only be true if ∇ui⋅ej=δij i.e. if' there should be an equation which is not rendered at the moment).

With this PR, running `make pdf` 3 or 4 times in the `manual` directory produces a pdf. I don't know if there's a fix for needing to run `make pdf` multiple times; if not I think we should update `manual/README.md` to note that it (may be) necessary.

I still get some LaTeX errors that seem to be to do with Chinese characters, like
```
! Package inputenc Error: Unicode char 於 (U+65BC)
(inputenc)                not set up for use with LaTeX.

See the inputenc package documentation for explanation.
Type  H <return>  for immediate help.
 ...                                              
                                                  
l.2059 ...}{代碼於 Nov  1 2018 17:41:02 编译}
```
but that may well be to do with my LaTeX installation/fonts rather than being an actual error. The manual still builds if I keep hitting 'return' at the `?` prompt for these errors. If this is to do with Chinese characters, would be nice to either skip them automatically if not installed, or have some instructions in `manual/README.md` (a) on how to fix the error (e.g. install the needed fonts?) and (b) that hitting enter is a work-around.

### DIFF
--- a/manual/sphinx/conf.py
+++ b/manual/sphinx/conf.py
@@ -210,7 +210,21 @@ latex_elements = {
 
     # Additional stuff for the LaTeX preamble.
     #
-    # 'preamble': '',
+    'preamble': r'''
+\usepackage[utf8]{inputenc}
+\usepackage[T1]{fontenc}
+\makeatletter
+\def\UTFviii@defined#1{%
+  \ifx#1\relax
+      ?%
+  \else\expandafter
+    #1%
+  \fi
+}
+\makeatother
+
+
+''',
 
     # Latex figure (float) alignment
     #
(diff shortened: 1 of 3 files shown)


bendudson:
Thanks @johnomotani ! I've added a "fix" for the unicode errors, so the manual now compiles without stopping (at least here).

ZedThree:
Thanks @johnomotani @bendudson ! I've checked the PDF builds on readthedocs

Issue #1753, opened 2019-06-20
Inconsistent warnings and defaults for ixsep etc


JosephThomasParker:
When the input file doesn't specify `ixsep1`, the output gives the contradictory warnings:
```
Variable 'ixseps1' not in mesh options. Setting to 0
        WARNING: Separatrix location 'ixseps1' not found. Setting to 12
```
(where `NxGlobal = 12`), and similarly for `ixseps2`, the `jyseps` and `ny_inner`.  This is because of how the checking is done in `include/bout/boutmesh.cxx`:
```
  // separatrix location
  if (Mesh::get(ixseps1, "ixseps1")) {
    ixseps1 = GlobalNx;
    output_warn.write(
        "\tWARNING: Separatrix location 'ixseps1' not found. Setting to %d\n", ixseps1);
  }
```
Here `Mesh::get` issues the first warning and sets the default to 0, then boutmesh overrides this.

I'm not sure if `Mesh::get` is used elsewhere, but perhaps the solution is to add an optional  "suppress_output" argument (default: false) to `Mesh::get` so that it doesn't write output when we're really only using it to see if the variable is set.

bendudson:
Would another possibility be to add an `isSet` method, like `Options` has, to test if a variable is present before trying to get its value? 
Possibility number three is to pass a default value to the call to `Mesh::get``

ZedThree:
Is there a possibility to replace `Mesh::get` and/or `GridSource` entirely with `Options` or `OptionsNetCDF`? I think `GridFile` can handle reading 3D variables stored as FFTs which `OptionsNetCDF` can't do (yet), so it's definitely not currently a drop-in replacement. Might be worth investigating though?

bendudson:
Yes, I think the plan is to replace `DataFIle` with `Options`, reading using `OptionsNetCDF` etc. then a grid would be created by just passing in an Options node. I'd like to add support for HDF5 or other format, and then replace `DataFile`.

JosephThomasParker:
> Would another possibility be to add an `isSet` method, like `Options` has, to test if a variable is present before trying to get its value?
> Possibility number three is to pass a default value to the call to `Mesh::get`

Number two works, but I think number three ends up being neater.  It looks like `Mesh::get` already accepted default values for Fields.

ZedThree:
Fixed in #1754, although indentation of the message is a bit higgledy-piggledy

PR #1754, opened 2019-06-20
Add default value to Mesh::get for ints

JosephThomasParker:
Addresses issue #1753


ZedThree:
You're just missing the equivalent change in `gridfromfile.cxx`

src/mesh/mesh.cxx
```diff
@@ -58,11 +58,14 @@ int Mesh::get(std::string &sval, const std::string &name) {
...
 
 /// Get an integer
-int Mesh::get(int &ival, const std::string &name) {
+int Mesh::get(int &ival, const std::string &name, int def) {
   TRACE("Mesh::get(ival, %s)", name.c_str());
 
-  if (source == nullptr or !source->get(this, ival, name))
+  if (source == nullptr or !source->get(this, ival, name, def)) {
+    // set ival to default in source==nullptr too:
+    ival = def;
```
JosephThomasParker: This is ugly: if `source == nullptr` then the default doesn't get set by `get`... so we set it silently here.  Without this, four of the integrated tests fail.

ZedThree: Could split it into two separate conditionals:

```cpp
if (source == nullptr) {
   output_warn << "No source; setting default"; // but better, obvs
   ival = def;
   return 1;
}
return source->get(this, ival, name, def);
```

ZedThree:
Should we also add default-defaults for the other overloads that are missing them?

JosephThomasParker:
Yes, if you're fine with the silent default-defaults.

ZedThree:
The `Mesh::get` overloads for `string`, `BoutReal`, and `Vector2D/3D` could also do with defaults.

src/mesh/data/gridfromfile.cxx
```diff
@@ -134,7 +134,7 @@ bool GridFile::get(Mesh *UNUSED(m), std::string &sval, const std::string &name)
  *   Boolean. True on success.
  * 
  */
-bool GridFile::get(Mesh *UNUSED(m), int &ival, const std::string &name) {
+bool GridFile::get(Mesh *UNUSED(m), int &ival, const std::string &name, int UNUSED(def)) {
```
ZedThree: We should be able to use `def` to set the default here

src/mesh/mesh.cxx
ZedThree: `GridDataSource::get` returns `true` on success, whereas `Mesh::get` returns false. So this just needed `return !source->get(...);`

ZedThree:
I thought this was a nice simplification of the implementation of the `GridFromOptions::get` routines:

https://github.com/boutproject/BOUT-dev/blob/8bac5c5640cca5d6fc76a668f95b4d054927e079/src/mesh/data/gridfromoptions.cxx#L17-L21

We want to know if the source actually had the variable, or if we've ended up using the default. So save the result of `hasVar` in order to return it. Then just let `options` do the work of either getting us the value or the default.

This, rather unfortunately, leads to some rather convoluted bugs. Take this bit of code, for example:

https://github.com/boutproject/BOUT-dev/blob/e0310e96404f808e52444868807696fdf6e2c3e9/src/mesh/impls/bout/boutmesh.cxx#L130-L137

If the mesh source is a file, then if `MXG` isn't in the file, we read from the `options` instead, with a default of `2`. We also save the value of `2` in `options`, annotating that it was a default value.

Now, if the mesh source is an `Options` and `MXG` isn't in it, we instead record `0` in `options`, as this is the default-default. `Mesh::get` still returns `true` because it was a default, so we still go into if-body, where we get a new default of `2`, which we actually use. _However_, this doesn't overwrite the previous default of `0` we recorded in `options`. So when we write out `BOUT.settings` at the end of the simulation, it gets `MXG = 0`. This breaks the `test-command-args`, because it uses the `BOUT.settings` from one run as the input for another, and currently `MXG = 0` is not valid (without periodic X).

I think it needs to go back to being something like:

```cpp
 bool GridFromOptions::get(Mesh* UNUSED(m), int& ival, const std::string& name, int def) { 
   if (!hasVar(name)) { 
     output_warn.write("Variable '%s' not in mesh options. Setting to %d\n", name.c_str(), 
                       def); 
     ival = def; 
     return false; 
   } 
  
   ival = (*options)[name]; // No default!
   return true; 
 } 
```

But this is important to bear in mind if we replace `GridDataSource` completely with something `Options`-based.

PR #1755, opened 2019-06-20
More clang-tidy fixes

ZedThree:
Lots of changes, but most of these commits are pretty boring mechanical changes from clang-tidy. A few interesting ones:

- 10a488e: `LaplaceXY::roundInt(BoutReal)` was only used in `LaplaceXY::globalIndex` and is easily replaced with `std::round`
- d1f1241: `LaplacyXY::indexXY` was being implicitly converted to `int`, when `globalIndex` should've been used
  - `indexXY` could almost be replaced with `Matrix<int>` except that it needs to be communicated!
- 7028cd0: tries to make as many initialisation-related things as "default" as possible

### DIFF
--- a/include/datafile.hxx
+++ b/include/datafile.hxx
@@ -98,8 +98,8 @@ class Datafile {
   bool init_missing; // Initialise missing variables?
   bool shiftOutput{false}; // Do we want to write out in shifted space?
   bool shiftInput{false};  // Read in shifted space?
-  int flushFrequencyCounter{
-      0};                // Counter used in determining when next openclose required
+  // Counter used in determining when next openclose required
+  int flushFrequencyCounter{0};
   int flushFrequency{1}; // How many write calls do we want between openclose
 
   std::unique_ptr<DataFormat> file;
--- a/include/datafile.hxx
+++ b/include/datafile.hxx
@@ -89,16 +89,17 @@ class Datafile {
  private:
   Mesh* mesh;
   bool parallel{false}; // Use parallel formats?
-  bool flush{true};        // Flush after every write?
-  bool guards{true};       // Write guard cells?
+  bool flush{true};     // Flush after every write?
+  bool guards{true};    // Write guard cells?
   bool floats{false};   // Low precision?
-  bool openclose{true};    // Open and close file for each write
+  bool openclose{true}; // Open and close file for each write
   int Lx,Ly,Lz; // The sizes in the x-, y- and z-directions of the arrays to be written
   bool enabled{true}; // Enable / Disable writing
   bool init_missing; // Initialise missing variables?
   bool shiftOutput{false}; // Do we want to write out in shifted space?
-  bool shiftInput{false}; // Read in shifted space?
-  int flushFrequencyCounter{0}; // Counter used in determining when next openclose required
+  bool shiftInput{false};  // Read in shifted space?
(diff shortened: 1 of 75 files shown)


src/sys/boutexception.cxx
```diff
@@ -101,7 +101,7 @@ void BoutException::makeBacktrace() {
   {                                                                                      \
     buflen = 0;                                                                          \
     buffer = nullptr;                                                                    \
-    if (s == nullptr) {                                                                  \
+    if ((s) == nullptr) {                                                                \
```
johnomotani: what do the extra brackets do here?

ZedThree: In general, macro arguments should be surrounded by brackets as otherwise you may get unintended behaviour when you pass expressions, e.g.:

```cpp
#define SQ(a) (a * a)

SQ(3); // 9
SQ(2 + 1); // -> (2 + 1 * 2 + 1) -> 5
```

Another reason to avoid macros where we can!

In this particular case, I'm not sure there's anything valid you could pass as `s` that would break without the brackets

src/mesh/impls/bout/boutmesh.hxx
```diff
@@ -40,70 +40,70 @@ class BoutMesh : public Mesh {
...
   ///
   /// @param[in] buffer  A buffer to put the data in. Must already be allocated of length \p size
   /// @param[in] size    The number of BoutReals to receive and put in \p buffer
   /// @param[in] tag     A label for the communication. Must be the same as sent
-  comm_handle irecvXIn(BoutReal *buffer, int size, int tag);
+  comm_handle irecvXIn(BoutReal *buffer, int size, int tag) override ;
 
-  MPI_Comm getXcomm(int UNUSED(jy)) const {return comm_x; } ///< Return communicator containing all processors in X
-  MPI_Comm getYcomm(int jx) const; ///< Return communicator containing all processors in Y
+  MPI_Comm getXcomm(int UNUSED(jy)) const override  {return comm_x; } ///< Return communicator containing all processors in X
```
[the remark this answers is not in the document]
ZedThree: Oops, good spot! Just not had `clang-format` run over this line

ZedThree:
Thanks @johnomotani ! I did `git clang-format next --extensions hxx,cxx` to run `clang-format` over all the changes in this PR. d6ddd28 is just whitespace changes.

include/datafile.hxx
```diff
@@ -89,16 +89,17 @@ class Datafile {
...
+  bool openclose{true}; // Open and close file for each write
   int Lx,Ly,Lz; // The sizes in the x-, y- and z-directions of the arrays to be written
   bool enabled{true}; // Enable / Disable writing
   bool init_missing; // Initialise missing variables?
   bool shiftOutput{false}; // Do we want to write out in shifted space?
-  bool shiftInput{false}; // Read in shifted space?
-  int flushFrequencyCounter{0}; // Counter used in determining when next openclose required
+  bool shiftInput{false};  // Read in shifted space?
+  int flushFrequencyCounter{
+      0};                // Counter used in determining when next openclose required
```
johnomotani: don't like clang-format's choice here

ZedThree: Agreed, it's hard to convince clang-format to handle situations like this nicely. I usually end up putting the comment on the line before.

PR #1760, opened 2019-07-01
Add mesh options for extrapolate_{x,y}

bendudson:
Allows extrapolation of metric components into boundary cells to be disabled using an input option, as these can create negative values for `g11`. The simulation then crashes when `checkPositive` is called with negative `g11` in the boundary cells.

Not sure what should happen in `interpolateAndExtrapolate` when `extrapolate_x` or `extrapolate_y` are false.

(also not sure why `3rdparty/variant` has appeared... will try to remove)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,3 +7,6 @@
 [submodule "mpark.variant"]
 	path = externalpackages/mpark.variant
 	url = https://github.com/mpark/variant.git
+[submodule "3rdparty/variant"]
+	path = 3rdparty/variant
+	url = https://github.com/mpark/variant.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,6 +7,3 @@
 [submodule "mpark.variant"]
 	path = externalpackages/mpark.variant
 	url = https://github.com/mpark/variant.git
-[submodule "3rdparty/variant"]
-	path = 3rdparty/variant
-	url = https://github.com/mpark/variant.git
--- a/src/mesh/coordinates.cxx
+++ b/src/mesh/coordinates.cxx
@@ -210,17 +210,17 @@ Coordinates::Coordinates(Mesh* mesh, Options* options)
   // 'interpolateAndExtrapolate' to set them. Ensures that derivatives are
   // smooth at all the boundaries.
 
-  const bool extrapolate_x = not mesh->sourceHasXBoundaryGuards();
-  const bool extrapolate_y = not mesh->sourceHasYBoundaryGuards();
+  const bool extrapolate_x = (*options)["extrapolate_x"].withDefault(not mesh->sourceHasXBoundaryGuards());
+  const bool extrapolate_y = (*options)["extrapolate_y"].withDefault(not mesh->sourceHasYBoundaryGuards());
 
   if (extrapolate_x) {
     output_warn.write(_("WARNING: extrapolating input mesh quantities into x-boundary "
-          "cells\n"));
+          "cells. Set option extrapolate_x=false to disable this.\n"));
   }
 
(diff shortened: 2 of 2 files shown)


johnomotani:
When `extrapolate_x` or `extrapolate_y` are false, `interpolateAndExtrapolate` doesn't change the x- or y-boundary guard cells. The effect should be that whatever `mesh->get` put in the boundary cells is left there (this applies when `interpolateAndExtrapolate` is called on a field read from the grid file, not one being interpolated from another location - `extrapolate_x` and `extrapolate_y` arguments should always be true when actually interpolating, otherwise the guard cells are not set at all).

I think `mesh->get` fills the boundary guard cells with the nearest grid-cell value if the boundary cells are not present in the grid file. If I'm remembering right, we had problems with that behaviour because the gradients of the metric components were not continuous at the y-boundaries, which made `G3` (which contains `DDY(J * g23)`) have a big jump in the last grid cell. Making the metric components smooth at the boundaries was the motivation for introducing the extrapolation.

Hypnotoad does now have the feature to add y-boundary guard cells, although it does have problems if extending the bout-grid makes it run into the bounding-box of the psi-grid...

ZedThree:
@johnomotani To be clear, are you saying this is the wrong thing to do, and the problem is in the grid generation?

johnomotani:
Not necessarily _wrong_, because I don't like extrapolation very much, and nearest-grid-value might be more robust sometimes. My personal preference though would be not to add more options to grid loading if there's another way around the issue. I think generating a nice grid file, that includes the boundary cells, is the nicest solution. I've had problems sometimes generating grid-files with hypnotoad with y-boundary guard cells as I said above - to cover those cases, we could provide a Python script that converts a hypnotoad grid file with no y-boundary guard cells into one that has them, with a choice of extrapolation methods, and possibly some work-arounds for negative values (e.g. if they are only in the second guard cell they won't affect unstaggered, 2nd order simulations, so you could just set them to any positive value). I think the script should be straightforward for single-null equilibria; it's a bit more of a faff for double-null because you need to insert extra cells into the middle of the grid but not too bad. I'd offer to write it, but I'm super busy the next couple of weeks...

friva000:
I agree with what John said, but I'd like to add something more: there is an ideal way of doing things, and there is an actual way of doing things. In my opinion, the ideal way would be to generate the metric components on a very fine grid (that does not depend on nx and ny), compute all the integrals and derivatives on that grid (e.g., G1, G2, G3, but also quantities like zShift), and then interpolate on the grid selected by the user (i.e., with the given nx and ny). If the user is selecting also a shifted grid, interpolate from the fine grid also to the shifted one. This is my ideal solution for two reasons. First, we don't require guard cells for the metric components, since all the derivatives are already computed. Second, the metric component do not depend on nx and ny (I noticed in the past that, for example, the number of turns that a field line does in the toroidal angle for a turn in the poloidal angle, i.e., q, depends on nx: for two different nx with the same equidisk file, we can obtain noticeably different metric components). Then there is the actual way of doing things: the user can (i) generate the guard cells from hypnotoad (not always possible because of the actual boundaries in the equidisk file); (ii) extrapolate from the domain into the guard cells with a third order scheme; and (iii) extrapolate from the domain with the nearest grid point scheme. The best solution is (i), because it ensures that the metric components are defined consistently in the guard cells. However, this is not always possible, in which case the best solution is the one that works...

ZedThree:
Thanks @johnomotani and @friva000 for clarifying. I agree that the best way to do this is via the grid generator, maybe even have it calculate the derivatives and staggered coordinates already?

I'm going to merge this anyway, because it is nice to have explicit options to turn things off.

PR #1762, opened 2019-07-02
Always use boundary cells from grid file if present

johnomotani:
Suggest this update to #1760 so that we never ignore boundary cell values from the grid file if they are present.

### DIFF
--- a/src/mesh/coordinates.cxx
+++ b/src/mesh/coordinates.cxx
@@ -210,17 +210,27 @@ Coordinates::Coordinates(Mesh* mesh, Options* options)
   // 'interpolateAndExtrapolate' to set them. Ensures that derivatives are
   // smooth at all the boundaries.
 
-  const bool extrapolate_x = (*options)["extrapolate_x"].withDefault(not mesh->sourceHasXBoundaryGuards());
-  const bool extrapolate_y = (*options)["extrapolate_y"].withDefault(not mesh->sourceHasYBoundaryGuards());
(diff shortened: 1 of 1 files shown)


src/mesh/coordinates.cxx
```diff
@@ -210,17 +210,27 @@ Coordinates::Coordinates(Mesh* mesh, Options* options)
...
+  bool extrapolate_x = not mesh->sourceHasXBoundaryGuards();
+  bool extrapolate_y = not mesh->sourceHasYBoundaryGuards();
+
+  // Allow extrapolation to be replaced by nearest-grid point value according to option
+  if (extrapolate_x) {
+    extrapolate_x = (*options)["extrapolate_x"].withDefault(true);
+  }
+  if (extrapolate_y) {
+    extrapolate_y = (*options)["extrapolate_y"].withDefault(true);
+  }
```
bendudson: I don't see the benefit of this rather than the version I wrote. I am biased though

bendudson: Is this to prevent someone from extrapolating if the guard cells are present?

johnomotani: Sorry, I was not thinking straight when I made this. I was thinking the nearest-grid-cell thing would override the grid file values if `extrapolate_y=false` was set explicitly, but that isn't true - when `extrapolate_y=false` the grid file values are used if present, and nearest-grid-cell value if they are not. So there is no benefit to this PR. Closing now...

PR #1764, opened 2019-07-04
More expressions "improvements"

bendudson:
Built on top of PR #1750 , this adds some features for making complex expressions in the input file easier to read / organise.

# Local Scope
Adds syntax to define local scope. Underneath this inserts a `FieldGenerator` which modifies the `Context` object. One use is to simplify expressions using local variables, for efficiency and readability e.g. ensuring that a value is positive:
```
positive_value = H(expression) * expression
```
where the expression would be evaluated twice. This can be replaced by:
```
positive_value = [value = expression]( H({value}) * {value} )
```
This mechanism can also be used to implement simple functions
```
mycosh = 0.5 * (exp({arg}) + exp(-{arg}))
```
which can be called by defining a new Context:
```
result = [arg = x*2](mycosh)
```
(a sort of dynamic scope system)

so we could write:
```
floor = H({arg}) * {arg}
positive_value = [arg = expression](floor)
```

# `where` function
Adds a `where(test, gt0, lt0)` function to `FieldFactory`. This enables more clear swiching than the H(x) Heaviside function.  Like the C++ functions in BOUT++, one of two expressions is evaluated,
depending on the sign of the first expression.

# Multi-line expressions
Enable multi-line expressions in input options, since long expressions can become hard to read in the options file if all on one line. This adopts the Python approach, where if there are unbalanced brackets ('()' or '[]') then the expression continues on the next line.

In the BOUT.inp file we can therefore write

```
value = (something
         + more
         - things)
```

or define complex expression using a Context scope:
```
result = [a = expression,
          b = something](
            {a} + {b})
```

# `sum` function for looping
Add a `sum` function enables a simple kind of loop, but one which is guaranteed to finish. Iterates a given symbol from 0 to count-1, evaluating and summing a given expression. Example:
```
sum(i, 4, {i}^2)  # => 0^2 + 1^2 + 2^2 + 3^3
```

# Recursive expressions
By default recursive expressions are not allowed in the input options. These can be enabled by setting `input:max_recursion_depth != 0` e.g.:
```
[input]
max_recursion_depth = 10  # 0 = none, -1 = unlimited
```
By putting a limit on the depth we avoid the halting problem at least, and expressions should (eventually) terminate, preferably before they overflow the stack.

As is traditional, I present to you the Fibonacci sequence:
```
fib = where({n} - 2.5,
            [n={n}-1](fib) + [n={n}-2](fib),
            1)
```
so if `n` = 1 or 2 then `fib` = 1, but if `n` = 3 or above then the recursion fun begins. 

And another ill-considered language is born. Sorry, I think it was inevitable.

### DIFF
--- a/src/sys/expressionparser.cxx
+++ b/src/sys/expressionparser.cxx
@@ -135,6 +135,47 @@ private:
   FieldGeneratorPtr expr;  ///< The expression to evaluate in the new context
 };
 
+  /// Sum expressions in a loop, given symbol, count and expression
+  class FieldSum : public FieldGenerator {
+  public:
+    /// Loop a symbol counter SYM from 0 to count-1
+    /// The count is calculated by evaluating COUNTEXPR, which must be a non-negative integer
+    /// Each iteration the expression EXPR is evaluated and the results summed.
+    FieldSum(const std::string& sym, FieldGeneratorPtr countexpr, FieldGeneratorPtr expr)
+        : sym(sym), countexpr(countexpr), expr(expr) {}
+
+    double generate(const Context& ctx) override {
+      // Get the count by evaluating the count expression
+      BoutReal countval = countexpr->generate(ctx);
+      int count = ROUND(countval);
+      
+      // Check that the count is a non-negaitve integer
+      if (fabs(countval - static_cast<BoutReal>(count)) > 1e-4) {
+        throw BoutException("Count %e is not an integer in sum expression", countval);
+      }
+      if (count < 0) {
+        throw BoutException("Negative count %d in sum expression", count);
+      }
+
+      BoutReal result {0.0};
+      Context new_context{ctx}; // Make a copy, so the counter value can be set
+      for (int i = 0; i < count; i++) {
+        // Evaluate the expression, setting the given symbol to the loop counter
+        new_context.set(sym, i);
+        result += expr->generate(new_context);
+      }
+      return result;
+    }
+    
+    std::string str() const override {
+      return std::string("sum(") + sym + std::string(",") + countexpr->str()
+             + std::string(",") + expr->str() + std::string(")");
+    }
+  private:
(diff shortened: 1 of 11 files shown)


src/sys/expressionparser.cxx
```diff
@@ -98,6 +98,43 @@ class FieldParam : public FieldGenerator {
 private:
   std::string name; // The name of the parameter to look up
 };
+
+/// Define a new context to evaluate an expression in
+/// This is essentially a dynamic scope mechanism
+class FieldContext : public FieldGenerator {
+public:
+  using variable_list = std::list<std::pair<string, FieldGeneratorPtr>>;
```
ZedThree: You probably want `std::vector` over `std::list`

src/sys/options/options_ini.cxx
```diff
@@ -107,6 +106,39 @@ void OptionINI::read(Options *options, const string &filename) {
         string key, value;
         // Get a key = value pair
         parse(buffer, key, value);
+
+        // Ensure that brackets '()' and '[]'are balanced
+
+        // Count net number of opening and closing brackets
+        auto count_brackets = [](const string& input) {
```
ZedThree: Another way of writing this, using std algorithms:

```cpp
auto count_brackets = [](const std::string& input) {
  return std::count_if(begin(input), end(input), [](const char& ch) { return ch == '(' or ch == '['; }) 
       - std::count_if(begin(input), end(input), [](const char& ch) { return ch == ')' or ch == ']'; });
}
```

bendudson: Thanks yes that would work. It traverses the string twice, but it is nicer in many ways. I'll change to this.

src/sys/expressionparser.cxx
bendudson: Done! Thanks

Issue #1637, opened 2019-03-11
Issue attempting to convert SOLT3D from bout-dev 3 to 4


ADimits:
In my attempt to convert Maxim Umansky's SOLT3D code from bout-dev 3 to 4, after running all of the sources through the bout_3to4.py script, and manually fixing a few things (including e.g., MZ=2^n+1 to NZ=MC=2^n, I get the output below (running on CORI at NERSC).

Before attempting to debug further, I was wondering whether (1) this is familiar and (2) whether the BOUT, Z length comment is likely to be related to the exception.

Thanks.

-------
.
.
.
        Option 2fluid:suppress_dc = 0 (default)
        Option mesh:paralleltransform = identity (data_drbm/data_1e0/BOUT.inp)
        Option hthe0 = 0.159155 (fluxtube_drbm.grd.nc)
    ****NOTE: input from BOUT, Z length needs to be divided by 1.108832e+02
Error encountered
Rank 3 [Mon Mar 11 11:23:51 2019] [c0-0c0s6n1] application called MPI_Abort(comm=0x84000002, 1) - process 3
====== Exception thrown ======
Field3D: Trying to set off-centre location on non-staggered grid
         Did you mean to enable staggered grids?

-------

johnomotani:
I suspect this is because we moved the `StaggerGrids` option from the top level to the `[mesh]` section. I don't think the `bout_3to4.py` script attempts to update input files. Somewhere near the top of your `BOUT.inp` file should be a line like
```
staggergrids = true
```
If you move it down under `[mesh]` that should fix this error.

ADimits:
Hi John,

Thanks for taking a stab. Maxim’s input files all appear have

#StaggerGrids = true

(i.e., commented), which I assumed implied that staggeredgrids is intended to default to be false. (if this is the boutdev v3 default.)

[I have tried to set up the build to work with Totalview, but have not really succeeded with NERSC’s compiler wrappers – but expect to eventually figure that out.]

Thanks again.

Andris

--

johnomotani:
Ah, OK. Sorry, we've made quite a few changes to the staggered grids support. In BOUT++ v4, with `staggergrids=false` it is an error to call, e.g., `setLocation(CELL_YLOW)`, whereas previously it would have been ignored. So either comment out the `setLocation` calls in the C++ physics model code, or wrap them in a conditional, e.g. instead of
```
Vpar.setLocation(CELL_YLOW);
```
put
```
[...]

PR #1765, opened 2019-07-09
Fix y-boundary cells of phi in conducting-wall-mode example

johnomotani:
Fixes a problem raised in #1637.

Note: as the example is currently set up, it doesn't actually use the boundary values changed by this PR, as they are multiplied by a coefficient that has been set to zero (issue was only when those boundary values were NaN). However, there are commented-out coefficients for the boundary conditions that would use the 'initial guess' of the boundary values of `phi`. I'm not clear what either boundary condition is doing exactly - although I haven't spent time trying to understand - but maybe it's useful to make it possible to run the commented-out alternative.

### DIFF
--- a/examples/conducting-wall-mode/README.md
+++ b/examples/conducting-wall-mode/README.md
@@ -1,0 +1,10 @@
+Conducting Wall Mode
+====================
+
+Simplified model for instability driven by sheath boundary condition
+and electron temperature gradient in a straight magnetic field.
+
+Created by B.Friedman to go with his [slides for a hands-on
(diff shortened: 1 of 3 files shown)


bendudson:
Is this still work in progress?

johnomotani:
I'd tagged as `WIP` in case the improvements @ADimits was working on for https://github.com/boutproject/BOUT-dev/issues/1637#issuecomment-509666879 would be ready to go in instead. Probably makes sense to merge this for 4.3 and better things can wait for the next release.

PR #1766, opened 2019-07-12
Fixes for Python DataFile class

johnomotani:
Two small changes:
* A bug-fix for writing `BoutArray` variables to netCDF files with the DataFile_netCDF class: previously the `bout_type` attribute was mistakenly ignored (variable `t` had been set to the data-type, not the variable type), so `bout_type` was always inferred from array shape
* Adds support for `bout_type == 'ArrayX'`, for 1d arrays that vary just in the x-direction. This is useful for handling the `ShiftAngle` variable that is needed in grid files.

@ZedThree @loeiten @bendudson Since we created separate repos for the Python tools, is there a process for bugfixes? Should they go to the stand-alone repos too, or will there be a merge of any bugfixes when we deprecate the use of tools from the `BOUT-dev` repo?

### DIFF
--- a/tools/pylib/boututils/datafile.py
+++ b/tools/pylib/boututils/datafile.py
@@ -478,6 +478,7 @@ class DataFile_netCDF(DataFile):
             ('x', 'y', 'z'): "Field3D",
             ('x', 'y'): "Field2D",
             ('x', 'z'): "FieldPerp",
+            ('x'): "ArrayX",
             (): "scalar",
         }
 
@@ -492,6 +493,7 @@ class DataFile_netCDF(DataFile):
             "Field3D": ('x', 'y', 'z'),
             "Field2D": ('x', 'y'),
             "FieldPerp": ('x', 'z'),
+            "ArrayX": ('x'),
             "scalar": (),
         }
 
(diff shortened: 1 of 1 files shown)


loeiten:
@johnomotani I can see that the master branch of `boututils` is out of synch with `v4.3.0-rc` (if not, I'd just copied the changes done in this PR to the `boututils` repo).

Therefore, to make life simple for ourselves and reduce the number of double commits, I suggest that we can keep the `boututils` repo up to synch with the `dev` branch of `BOUT-dev` until it's removed from `BOUT-dev`.  Does this make sense?

johnomotani:
@loeiten I think that makes sense. Maybe we could add some tags to the `boututils` repo - e.g. if the current master version is the same as the boututils in the v4.2.2 release of BOUT++ we could put a `v4.2.2` tag on it, while the `master` branch follows `BOUT-dev/next`?

loeiten:
Good idea @johnomotani. I've added some tags. If anyone is interested, I could let the `master` branches follow `BOUT-dev/next`. However, if there is no real need I will only update the `master` branches sporadically until the packages are removed from `BOUT-dev` (hopefully in `v4.3.0`)

PR #1770, opened 2019-07-15
Consistent global indices

johnomotani:
Introduces consistent set of methods: `getGlobalXIndex`, `getGlobalYIndex` and `getGlobalZIndex`. These replace `XGLOBAL` and `YGLOBAL` which are now deprecated. The new methods use a global cell numbering which includes all boundary cells. `XGLOBAL` and `YGLOBAL` were inconsistent as `XGLOBAL` included boundary cells, but `YGLOBAL` did not.

Note that `getGlobalYIndex` includes _all_ boundary cells - the boundary cells at the 'upper' divertor in a double-null geometry are included in the numbering.

### DIFF
--- a/examples/6field-simple/elm_6f.cxx
+++ b/examples/6field-simple/elm_6f.cxx
@@ -303,9 +303,9 @@ const Field2D N0tanh(BoutReal n0_height, BoutReal n0_ave, BoutReal n0_width,
       BoutReal xgrid_num = (Jxsep + 1.) / Grid_NX;
       // output.write("mgx = %e xgrid_num = %e\n", mgx);
       for (int jy = 0; jy < mesh->LocalNy; jy++) {
-        int globaly = mesh->YGLOBAL(jy);
+        int globaly = mesh->getGlobalYIndex(jy);
         // output.write("local y = %i;   global y: %i\n", jy, globaly);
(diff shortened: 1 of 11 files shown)


johnomotani:
Another question: is it worth having either another pair of methods, or a flag to `getGlobalXIndex`/`getGlobalYIndex` to get a global index _excluding_ boundary cells? In implementing this I left a private version of `YGLOBAL` in `BoutMesh` because `jyseps*` and `ny_inner` are defined without the boundary cells. There could be other times where it's more convenient to exclude them, and it's a pain to translate in general between indices with boundary cells and indices without, because you always have to allow for whether you're before or after the second divertor in double-null cases. Having both methods available would isolate that pain to a single place (plus the version without boundary cells is simpler anyway, so it wouldn't take much to add it).

Summary: I'm in favour of having a method to have global indices excluding boundary cells. Should it be a separate method or a flag to `getGlobalXIndex`/`getGlobalYIndex`?

ZedThree:
Should `XGLOBAL` just `return getGlobalXIndex`, and `YGLOBAL` `return getGlobalYIndexNoBoundaries`?

johnomotani:
@ZedThree sounds good to me. I've made the `getGlobalYIndexNoBoundaries` method public, added `getGlobalXIndexNoBoundaries` and `getGlobalZIndexNoBoundaries` methods for symmetry, and changed `XGLOBAL` and `YGLOBAL` to just return the new methods.

PR #1772, opened 2019-07-20
Construct Option with initializer_list

bendudson:
This enables initializer lists to be used to create custom options, to pass to solvers or other components. This is a slightly more compact way to create options trees than the existing method.

~~Unfortunately requires non-template constructors for the accepted types, but these are quite simple.~~

Includes tests and documentation.


include/options.hxx
```diff
@@ -164,6 +164,21 @@ public:
   Options(Options *parent_instance, std::string full_name)
       : parent_instance(parent_instance), full_name(std::move(full_name)){};
 
+  /// Initialise with a value
+  /// These enable Options to be constructed using initializer lists
+  /// @{
```
ZedThree: Can drop doxygen grouping comments surrounding this constructor now

include/options.hxx
```diff
@@ -164,6 +164,21 @@ public:
...
+  Options(T value) {
+    assign<T>(value);
+  }
+  /// @}
+  
+  /// Construct with a nested initializer list
+  /// This allows Options trees to be constructed, using a mix of types.
+  ///
+  /// Example:  { {"key1", 42}, {"key2", field} }
+  Options(std::initializer_list<std::pair<const std::string, const Options>> values);
```
ZedThree: Is `const` needed here? I think I can see why one might be needed on the `std::string`, as the map `value_type` is `pair<const Key, T>`

Note: I don't think it hurts anything to keep it!

ZedThree:
Codacy would like both these constructors to be `explicit`, which is probably fair, at least for `Options(T)`.

ZedThree:
Thanks @bendudson ! Very sorry this took such a long time to deal with! :grimacing:

PR #1777, opened 2019-07-31
File open error messages

johnomotani:
I just spent some time trying to figure out why a simulation wouldn't start - it turned out I'd accidentally left the option `append = true` set, which failed because the `BOUT.dmp.*` files did not exist yet. To help make this kind of error clearer, this PR updates the error messages to include the filename and whether we were trying to open for writing or appending. Also updates the `Ncxx4` implementation of `Dataformat` to catch exceptions thrown when opening a file and return `false` from the method instead, so that we use the error handling in `DataFile`. Now instead of a message like
```
terminate called after throwing an instance of 'netCDF::exceptions::NcException'
  what():  No such file or directory
file: ncFile.cpp  line:60
```
we'd get something like
```
Error encountered during initialisation: Datafile::open: Failed to open file data/BOUT.dmp.nc for appending!
terminate called after throwing an instance of 'BoutException'
  what():  Datafile::open: Failed to open file data/BOUT.dmp.nc for appending!
```


ZedThree:
Do you need to update this for the `bool` overload?

Also, I'm a bit in two minds about this. The new information about what operation we're trying to do is definitely useful, but we've now lost some information about the NetCDF error. Might be worth thinking about how to keep that info?

dschwoerer:
Could we change the interface to always return success, and always throw an error?
That would allow to add additional information in the error by catching and re-throwing.
Also it would be more C++ rather then C.
Haven't looked into it more, but if we move the throw part into the file - would that reduce the amount of new code? Or would it increase due to the duplication in the several interfaces?
Moving it would allow to catch and handle the different libraries easily ...

johnomotani:
> Do you need to update this for the bool overload?

Yes! good spot :+1: 


> Also, I'm a bit in two minds about this. The new information about what operation we're trying to do is definitely useful, but we've now lost some information about the NetCDF error. Might be worth thinking about how to keep that info?

Normally I'd agree, but in this case the netCDF exceptions seem to be incredibly uninformative - the only information I was getting was `what():  No such file or directory` - I don't think it's worth bothering about keeping that message. In principle netCDF might upgrade their exceptions, and then we might be losing information... I had the impression that the C++ interface is not maintained very actively, but just looked at the github page and noticed @ZedThree is contributing now so this may not be true!

johnomotani:
@dschwoerer I agree more C++ style error handling would be nice, but I think the file I/O is intended (?) to be replaced by @bendudson's Options-tree classes, so maybe not worth making major changes to `Datafile`/`Dataformat` now.

PR #1778, opened 2019-08-02
Scripts cleanup

dschwoerer:
Always calling python3 ensures that even if python still points to python2, all scripts work, even if numpy etc. is only installed for python3.


dschwoerer:
As they are runnable from the command line - should I add sym-links to the bin folder?
Or create them during the install step?

PR #1780, opened 2019-08-04
Correct manual - default for 'floats' is true for dump files too

johnomotani:
We changed the default for the 'floats' option to 'true' for dump files in #404.


ZedThree:
Thanks @johnomotani !

PR #1783, opened 2019-08-08
Fix some issues in hypnotoad

### LINKED ISSUE #1729: Hypnotoad: Bug in the computation of ShiftAngle
ShiftAngle is computed on line 306 of process_grid.pro as
`loop[xi] = f[xi,yi[N_ELEMENTS(yi)-1]] - f[xi,yi[0]]` i.e., it is computed as zShift(ix,y=yend)-zShift(ix,y=ystart), where ystart and yend are the first and the last points in the periodic (core) domain. However, ShiftAngle should be the integral from 0 to 2*pi of nu(psi,theta) [or from -pi to pi, or... nu is a periodic function in theta]. This means that in the computation of ShiftAngle there is an error of order dy, (ystart and yend are shifted by dy/2 from the boundaries of the domain).

This also affects the computation of zShift in setParallelTransform, since ShiftAngle is used to compute zShift in the y guard cells at the Branch cuts, which may be used to compute the field aligned fields at those point (useful for parallel derivatives and interpolation).

friva000:
There were a few issues with Hypnotoad, and this PR fixes them.
- dx was computed as psi(i+1)-psi(i), while now it is computed as psi(i+1/2)-psi(i-1/2)
- Fixes #1729
- Fixes some problems arising when the separatrix is exactly on a grid point

Question: lines 1375-1379 of create_grid.pro are used to handle the case when the private flux regions have the same number or radial points (or more) of the entire domain. Is it a case that can really take place?

### DIFF
--- a/tools/tokamak_grids/gridgen/process_grid.pro
+++ b/tools/tokamak_grids/gridgen/process_grid.pro
@@ -311,7 +311,7 @@ FUNCTION my_int_y, var, yaxis, mesh, loop=loop, nosmooth=nosmooth, simple=simple
       ;; Only set loop integral in closed (periodic) domains i.e. the
       ;; core. Otherwise it may be overwritten by values in a PF region
        
-      loop[xi] = f[xi,yi[N_ELEMENTS(yi)-1]] - f[xi,yi[0]]
+      loop[xi] = f[xi,yi[N_ELEMENTS(yi)-1]] - f[xi,yi[0]] - f[xi,yi[0]] + 0.5D*(yaxis[xi,yi[1]]-yaxis[xi,yi[0]])*(var[xi,yi[N_ELEMENTS(yi)-1]] + var[xi,yi[0]])
     ENDIF
   ENDREP UNTIL last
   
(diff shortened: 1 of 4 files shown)


tools/tokamak_grids/gridgen/process_grid.pro
```diff
@@ -311,7 +311,7 @@ FUNCTION my_int_y, var, yaxis, mesh, loop=loop, nosmooth=nosmooth, simple=simple
       ;; Only set loop integral in closed (periodic) domains i.e. the
       ;; core. Otherwise it may be overwritten by values in a PF region
        
-      loop[xi] = f[xi,yi[N_ELEMENTS(yi)-1]] - f[xi,yi[0]]
+      loop[xi] = f[xi,yi[N_ELEMENTS(yi)-1]] - f[xi,yi[0]] - f[xi,yi[0]] + 0.5D*(yaxis[xi,yi[1]]-yaxis[xi,yi[0]])*(var[xi,yi[N_ELEMENTS(yi)-1]] + var[xi,yi[0]])
```
johnomotani: Is `- f[ix,yi[0]]` supposed to be in there twice?

friva000: Sorry, my bad. I'll fix it.

PR #1784, opened 2019-08-13
Add macro for Scorep instrumentation of a user-defined code region

JosephThomasParker:
Add a macro to allow scorep instrumentation of a user-defined region of code.  The section of code
```
{
SCOREP_REGION("region name")
// stuff to time
}
```
appears as "region name" in the Scalasca profile.  AIUI, the existing `SCOREP0` macro can't be used to do this, as it (very usefully) times the whole function and automatic uses the function name as its display name.

### DIFF
--- a/include/bout/scorepwrapper.hxx
+++ b/include/bout/scorepwrapper.hxx
@@ -49,10 +49,10 @@
 
 /// Instrument a region with scorep
 #ifdef BOUT_HAS_SCOREP
-#define SCOREP_REGION(...)						\
+#define BOUT_SCOREP_REGION(...)						\
   SCOREP_USER_REGION(__VA_ARGS__, SCOREP_USER_REGION_TYPE_COMMON)
 #else
-#define SCOREP_REGION(...)
+#define BOUT_SCOREP_REGION(...)
 #endif
 
 #endif
--- a/include/bout/scorepwrapper.hxx
+++ b/include/bout/scorepwrapper.hxx
@@ -11,7 +11,7 @@
 #define SCOREPLVL 0
 #endif
 
-/// Instrument a region/function with scorep
+/// Instrument a function with scorep
 ///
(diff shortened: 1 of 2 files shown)


JosephThomasParker:
Yes, I went from the quick reference here:
https://apps.fz-juelich.de/scalasca/releases/scalasca/2.3/docs/QuickReference.pdf

ZedThree:
Can `SCOREP_USER_REGION` be used instead? Looks like it makes things a bit simpler.

http://scorepci.pages.jsc.fz-juelich.de/scorep-pipelines/docs/scorep-5.0/html/group__SCOREP__User.html

JosephThomasParker:
Thanks @ZedThree, that is simpler - updated!

JosephThomasParker:
Sorry, mission creep.  The approach documented doesn't build when BOUT is configured without scorep, so I've added a macro for the region that vanishes when there's no scorep support.

JosephThomasParker:
@d7919 - yes it works for me locally and on Archer. :slightly_smiling_face:

include/bout/scorepwrapper.hxx
```diff
@@ -47,4 +47,12 @@
 #define SCOREP3(...)
 #endif
 
+/// Instrument a region with scorep
+#ifdef BOUT_HAS_SCOREP
+#define SCOREP_REGION(...)						\
```
ZedThree: Apologies for bike-shedding, but it's plausible that `SCOREP_REGION` will clash with a future ScoreP macro/name. We might want to call this `BOUT_SCOREP_REGION` (or similar) instead.

JosephThomasParker: np - done.  

PS: I think it's only bike-shedding if I insist on `SCOREP_REGION_BOUT`, we get into a fight about it, people take sides, and consequently v4.3 doesn't get released.

Issue #1785, opened 2019-08-20
Flags not validated in LaplaceCyclic, etc.


johnomotani:
It would be nice to check that the flags passed by the user have been implemented. In particular for `LaplaceCyclic`, since that is the default Laplacian solver, but ideally for all the other implementations where the check is not already done.

The implementation could copy the one in `LaplaceMultigrid`
https://github.com/boutproject/BOUT-dev/blob/8640344cced1815676ab4ff14972036b0886e4e7/src/invert/laplace/impls/multigrid/multigrid_laplace.cxx#L79-L89
Would be straightforward if there is a volunteer: copy the code linked above into the `LaplaceCyclic()` constructor; for each flag in `invert_laplace.hxx` see if it's present in `cyclic_laplace.cxx`, and if it is add to the appropriate `implemented_global_flags` or `implemented_boundary_flags` variable.

PR #1786, opened 2019-08-20
Implement the INVERT_ZERO_DC flag in LaplaceCyclic

johnomotani:
I've labelled as a bugfix because previously the flag was ignored silently. The `INVERT_ZERO_DC` flag is useful when combining a `Laplacian` solver with a `LaplaceXY` solver which calculates the DC part of the potential.

It would be nice to check that the flags passed by the user have been implemented, but I don't have time to check all the flags now. See #1785.

### DIFF
--- a/src/invert/laplace/impls/cyclic/cyclic_laplace.cxx
+++ b/src/invert/laplace/impls/cyclic/cyclic_laplace.cxx
@@ -232,9 +232,19 @@ FieldPerp LaplaceCyclic::solve(const FieldPerp& rhs, const FieldPerp& x0) {
       auto k1d = Array<dcomplex>((localmesh->LocalNz) / 2 +
                                  1); // ZFFT routine expects input of this length
 
+      int startz = 0;
+      if(global_flags & INVERT_ZERO_DC) {
+        // No DC component
+        startz = 1;
+      }
+
       BOUT_OMP(for nowait)
       for (int ix = xs; ix <= xe; ix++) {
-        for (int kz = 0; kz < nmode; kz++)
+        if (startz == 1) {
+          k1d[0] = 0.;
+        }
+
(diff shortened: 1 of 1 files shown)


bendudson:
Thanks @johnomotani, just wondering why `startz` is an `int` rather than a `bool`.

johnomotani:
I'd used an `int` because when I was writing the first place I put it was `int kz = startz`, but `bool startz` would work. Would you prefer a `bool`?

src/invert/laplace/impls/cyclic/cyclic_laplace.cxx
```diff
@@ -232,9 +232,19 @@ FieldPerp LaplaceCyclic::solve(const FieldPerp& rhs, const FieldPerp& x0) {
       auto k1d = Array<dcomplex>((localmesh->LocalNz) / 2 +
                                  1); // ZFFT routine expects input of this length
 
+      bool zero_DC = false;
+      if(global_flags & INVERT_ZERO_DC) {
+        // No DC component
+        zero_DC = true;
```
ZedThree: Could do `const bool zero_DC = global_flags & INVERT_ZERO_DC;`

src/invert/laplace/impls/cyclic/cyclic_laplace.cxx
```diff
@@ -440,13 +446,23 @@ Field3D LaplaceCyclic::solve(const Field3D& rhs, const Field3D& x0) {
       auto k1d = Array<dcomplex>((localmesh->LocalNz) / 2 +
                                  1); // ZFFT routine expects input of this length
 
+      bool zero_DC = false;
```
ZedThree: Sorry, I meant to flag this one too!

PR #1789, opened 2019-08-22
LaplaceXY: finite difference option

johnomotani:
This PR adds a finite difference discretisation option for `LaplaceXY`, using the same discretisation as `A*Laplace_perp(f) + Grad_perp(A)*Grad_perp(A) + B*f` uses, instead of the original finite-volume flux-divergence discretisation. The new finite-difference implementation supports non-orthogonal grids with non-zero `g12`.

Also adds an integrated test for `LaplaceXY`, using the new discretisation. Note that the test is much faster using `pctype = hypre` instead of `pctype = shell` if hypre is available in PETSc.

This is a new feature, so I guess should go into `next` instead of `v4.3.0-rc`, although I wouldn't mind it going into 4.3...

I should probably flesh out the paragraph in the manual on the new option - I'll do that when this has been reviewed.

Edit: breaking change - deprecate `laplacexy:y_bndry_dirichlet` option in favour of `laplacexy:y_bndry`.


src/invert/laplacexy/laplacexy.cxx
```diff
@@ -28,15 +29,55 @@ static PetscErrorCode laplacePCapply(PC pc,Vec x,Vec y) {
...
+  instance_count++;
+  my_id = instance_count;
+
   if (opt == nullptr) {
     // If no options supplied, use default
     opt = &(Options::root()["laplacexy"]);
   }
   
+  finite_volume = (*opt)["finite_volume"].doc(
+      "Use finite volume rather than finite difference discritisation."
```
bendudson: discritisation -> discretisation  (sorry)

src/invert/laplacexy/laplacexy.cxx
```diff
@@ -214,6 +428,67 @@ LaplaceXY::LaplaceXY(Mesh *m, Options *opt, const CELL_LOC loc)
...
+    // Receive from outer
+    MPI_Irecv(&indexXY(localmesh->xend+1, localmesh->yend+1), 1, MPI_DOUBLE,
+        proc_xind+1, tag0 + proc_xind, xcomm, &requests[0]);
+  }
+  if (not localmesh->firstX()) {
+    // Need to send upper inner point in
+    MPI_Isend(&indexXY(localmesh->xstart, localmesh->yend+1), 1, MPI_DOUBLE,
+        proc_xind-1, tag0 + (proc_xind - 1), xcomm, &requests[1]);
+  }
+  MPI_Waitall(2, requests, statuses);
```
bendudson: Can the above set of 4 waits be merged, so have 8 requests and one `MPI_Waitall`?

manual/sphinx/user_docs/laplacian.rst
```diff
@@ -887,8 +887,33 @@ cell faces.
 
 Notes:
 
--  The ShiftXderivs option must be true for this to work, since it
-   assumes that :math:`g^{xz} = 0`
+-  The ``ShiftedMetric`` or ``FCITransform`` ParallelTransform must be used
+   (i.e. ``mesh:paralleltransform = shifted`` or
+   ``mesh:paralleltransform = fci``) for this to work, since it assumes that
```
ZedThree: Reminder to update this once #1896 goes in

include/bout/invert/laplacexy.hxx
```diff
@@ -129,20 +142,26 @@ private:
   PC pc;            ///< Preconditioner
 
   Mesh *localmesh;   ///< The mesh this operates on, provides metrics and communication
+
+  static int instance_count;
+  int my_id = 0;
```
ZedThree: This is to uniquely identify different instances of this type, right? Another option could be a public `std::string name`, and leave it to the user to identify? Or take the default value from the `Options` used to construct?

Just ideas for future polish, might be something we want to do more generally

src/invert/laplacexy/laplacexy.cxx
```diff
@@ -28,15 +29,55 @@ static PetscErrorCode laplacePCapply(PC pc,Vec x,Vec y) {
...
   
+  finite_volume = (*opt)["finite_volume"].doc(
+      "Use finite volume rather than finite difference discretisation."
+      ).withDefault(true);
+
+  ///////////////////////////////////////////////////
+  // Boundary condititions options
+  if (localmesh->periodicY(localmesh->xstart)) {
+    // Periodic in Y, so in the core
+    opt->get("core_bndry_dirichlet", x_inner_dirichlet, false);
```
ZedThree: Could use new `Options` API?

src/invert/laplacexy/laplacexy.cxx
```diff
@@ -28,15 +29,55 @@ static PetscErrorCode laplacePCapply(PC pc,Vec x,Vec y) {
...
+  } else {
+    // Non-periodic, so in the PF region
+    opt->get("pf_bndry_dirichlet", x_inner_dirichlet, true);
+  }
+  if ((*opt)["y_bndry_dirichlet"].isSet()) {
+    bool y_bndry_dirichlet;
+    opt->get("y_bndry_dirichlet", y_bndry_dirichlet, false);
```
ZedThree: Do we want to deprecate `y_bndry_dirichlet` in favour of `y_bndry`?

src/invert/laplacexy/laplacexy.cxx
```diff
@@ -105,96 +164,251 @@ LaplaceXY::LaplaceXY(Mesh *m, Options *opt, const CELL_LOC loc)
...
-  // X boundaries
-  if(localmesh->firstX()) {
-    // Lower X boundary
-    for(int y=localmesh->ystart;y<=localmesh->yend;y++) {
-      const int localIndex = globalIndex(localmesh->xstart - 1, y);
-      ASSERT1( (localIndex >= 0) && (localIndex < localN) );
-    
-      d_nnz[localIndex] = 2; // Diagonal sub-matrix
-      o_nnz[localIndex] = 0; // Off-diagonal sub-matrix
+  if (finite_volume) {
```
ZedThree: This whole bit is now a bit unwieldy. Worth pulling out into two helper functions? Then could just be

```cpp
if (finite_volume) {
   set_matrix_elements_for_finite_volume();
} else {
   set_matrix_elements_for_finite_difference();
}
```

src/invert/laplacexy/laplacexy.cxx
```diff
@@ -214,6 +428,66 @@ LaplaceXY::LaplaceXY(Mesh *m, Options *opt, const CELL_LOC loc)
...
+  int tag0 = localmesh->getYProcIndex()*localmesh->getNXPE();
+  MPI_Request requests[] = {MPI_REQUEST_NULL, MPI_REQUEST_NULL, MPI_REQUEST_NULL,
+                            MPI_REQUEST_NULL, MPI_REQUEST_NULL, MPI_REQUEST_NULL,
+                            MPI_REQUEST_NULL, MPI_REQUEST_NULL};
+  MPI_Status statuses[8];
+
+  // Get lower, inner corner
+  if (not localmesh->firstX()) {
+    // Receive from inner
+    MPI_Irecv(&indexXY(localmesh->xstart-1, localmesh->ystart-1), 1, MPI_DOUBLE,
```
ZedThree: Should these now be `localmesh->MPI_*`?

src/invert/laplacexy/laplacexy.cxx
```diff
@@ -314,98 +577,232 @@ void LaplaceXY::setCoefs(const Field2D &A, const Field2D &B) {
...
-      // Metrics on x-1/2 boundary
-      J = 0.5*(coords->J(x,y) + coords->J(x-1,y));
-      g11 = 0.5*(coords->g11(x,y) + coords->g11(x-1,y));
-      dx = 0.5*(coords->dx(x,y) + coords->dx(x-1,y));
-      Acoef = 0.5*(A(x,y) + A(x-1,y));
-      
-      val = Acoef * J * g11 / (coords->J(x,y) * dx * coords->dx(x,y));
-      xm = val;
-      c  -= val;
+  if (finite_volume) {
```
ZedThree: Another place for helper functions?

ZedThree:
Is this a candidate for using the new Petsc interface?

johnomotani:
> Is this a candidate for using the new Petsc interface?

That would be nice! I've been using this for TORPEX simulations since before the new interface existed...

src/invert/laplacexy/laplacexy.cxx
```diff
@@ -29,16 +30,49 @@ static PetscErrorCode laplacePCapply(PC pc,Vec x,Vec y) {
...
+  if (localmesh->periodicY(localmesh->xstart)) {
+    // Periodic in Y, so in the core
+    x_inner_dirichlet = (*opt)["core_bndry_dirichlet"].withDefault(false);
+  } else {
+    // Non-periodic, so in the PF region
+    x_inner_dirichlet = (*opt)["pf_bndry_dirichlet"].withDefault(true);
+  }
+  if ((*opt)["y_bndry_dirichlet"].isSet()) {
+    throw BoutException("y_bndry_dirichlet has been deprecated. Use y_bndry=dirichlet "
+                        "instead.");
```
ZedThree: Nice! A simple way to deal with deprecating options, I like it.

src/invert/laplacexy/laplacexy.cxx
```diff
@@ -29,16 +30,49 @@ static PetscErrorCode laplacePCapply(PC pc,Vec x,Vec y) {
...
+  }
+
+  // Check value of y_bndry is a supported option
+  if (not(
+    y_bndry == "dirichlet"
+    or y_bndry == "neumann"
+    or y_bndry == "free_o3")) {
+
+    throw BoutException("Unrecognized option '%s' for laplacexy:ybndry",
+        y_bndry.c_str());
```
ZedThree: Need to update this to `fmt` style:

```cpp
    throw BoutException("Unrecognized option '{}' for laplacexy:ybndry",
        y_bndry);
```

tests/integrated/test-laplacexy/runtest
```diff
@@ -0,0 +1,100 @@
...
+            'f:bndry_xin=dirichlet f:bndry_xout=dirichlet f:bndry_yup=free_o3 f:bndry_ydown=free_o3 b:function=.1',
+            'laplacexy:core_bndry_dirichlet=false laplacexy:pf_bndry_dirichlet=false laplacexy:y_bndry=neumann '
+            'f:bndry_xin=neumann f:bndry_xout=dirichlet f:bndry_yup=neumann f:bndry_ydown=neumann b:function=.1 laplacexy:pctype=hypre',
+            'laplacexy:core_bndry_dirichlet=false laplacexy:pf_bndry_dirichlet=false laplacexy:y_bndry=free_o3 '
+            'f:bndry_xin=neumann f:bndry_xout=dirichlet f:bndry_yup=free_o3 f:bndry_ydown=free_o3 b:function=.1 laplacexy:pctype=hypre',
+           ]
+
+print('Making LaplaceXY inversion test')
+shell_safe('make > make.log')
+
```
ZedThree: Please could you change this to

```python
build_and_log("Interpolation test")
```

`build_and_log` is in `boututils.run_wrapper`

It's a replacement for the two lines, except on Windows where it doesn't call `make`. (At some point I also need to make it work with non-make generators when using CMake on Linux too)

src/invert/laplacexy/laplacexy.cxx
johnomotani: Doing this does skip a 'deprecation warning' stage, although would be easy enough to stick one in in a similar style, but we're allowed to break stuff in 5.0! I should probably stick in a deprecation warning into 4.4 in #1924...

johnomotani:
> Please could you also add CMakeLists.txt for the new test-laplacexy?

I have added one, but the test as currently written might be a bit long for Travis... Is there a cmake equivalent of `#requires all_tests`? The test also checks a lot of different combinations of options at the moment, it might be worth making a shorter version that only tests a few?

ZedThree:
> > Please could you also add CMakeLists.txt for the new test-laplacexy?
> 
> I have added one, but the test as currently written might be a bit long for Travis... Is there a cmake equivalent of `#requires all_tests`? The test also checks a lot of different combinations of options at the moment, it might be worth making a shorter version that only tests a few?

Not yet, I will add one at some point though

src/invert/laplacexy/laplacexy.cxx
```diff
@@ -113,20 +164,187 @@ LaplaceXY::LaplaceXY(Mesh *m, Options *opt, const CELL_LOC loc)
...
+  // Convert indexXY from local index to global index
+  indexXY += Istart;
+  
+  // Now communicate to fill guard cells
+  localmesh->communicate(indexXY);
+
+  // Fill corner guard cells by getting the y-guard cells from the x-neighbours.
+  // This is an arbitrary choice, which makes a difference at the X-point, where taking
+  // the x-guard cells from the y-neighbours would give points in different regions.
+  auto xcomm = localmesh->getXcomm();
```
johnomotani: Can remove all this communication of corners now that #1885 has gone in. I'll push an update...

johnomotani:
I've removed some communication stuff that is now taken care of by the changes in #1885; updated to keep valid corner cells when returning the result; updated the indexing in the shell-preconditioner to account for the fact that the global indexing can be different for the finite-volume and finite-difference cases; tweaked the PETSc settings for faster convergence when using the shell-preconditioner; and updated various comments.

I checked through the rest of the PR as well, and it looks good to me now.

src/invert/laplacexy/laplacexy.cxx
```diff
@@ -1713,73 +1717,60 @@ void LaplaceXY::solveFiniteDifference(const Field2D& x0) {
...
   RangeIterator itdwn=localmesh->iterateBndryLowerY();
   if(!itdwn.isDone()) {
-    ind = globalIndex(itdwn.ind, localmesh->ystart-1);
     
     for(; !itdwn.isDone(); itdwn++) {
       // Should not go into corner cells, LaplaceXY stencil does not include them
       if (itdwn.ind < localmesh->xstart or itdwn.ind > localmesh->xend) {
         continue;
       }
+      const int ind = globalIndex(itdwn.ind, localmesh->ystart-1);
```
ZedThree: With this change, can the loop be rewritten to 

```cpp
for (RangeIterator it = localmesh->iterateBndryLowerY(); !it.isDone(); it++) {
```

to match the other loops?

ZedThree:
Seems to still be failing in the PR tests but not the push?

johnomotani:
I think there's a bug in the `Laplace` implementation from #1936. I'll make a fix...

PR #1791, opened 2019-08-29
Fix initial conditions not being transformed correctly

ZedThree:
The result field for creating initial conditions was always set to be un-aligned, which resulted in the transform being a no-op.

Still need to add a test to catch this in future, so setting PR as draft for now.

### DIFF
--- a/include/bout/index_derivs_interface.hxx
+++ b/include/bout/index_derivs_interface.hxx
@@ -239,7 +239,7 @@ T D4DY4(const T& f, CELL_LOC outloc = CELL_DEFAULT, const std::string& method = 
     return standardDerivative<T, DIRECTION::YOrthogonal, DERIV::StandardFourth>(
         f, outloc, method, region);
   } else {
-    const bool is_unaligned = (f.getDirectionY() == YDirectionType::Aligned);
+    const bool is_unaligned = (f.getDirectionY() == YDirectionType::Standard);
     const T f_aligned = is_unaligned ? toFieldAligned(f, "RGN_NOX") : f;
     T result = standardDerivative<T, DIRECTION::Y, DERIV::StandardFourth>(
         f_aligned, outloc, method, region);
--- a/include/bout/index_derivs_interface.hxx
+++ b/include/bout/index_derivs_interface.hxx
@@ -205,10 +205,11 @@ T DDY(const T& f, CELL_LOC outloc = CELL_DEFAULT, const std::string& method = "D
     return standardDerivative<T, DIRECTION::YOrthogonal, DERIV::Standard>(f, outloc,
                                                                           method, region);
   } else {
-    const T f_aligned = toFieldAligned(f, "RGN_NOX");
+    const bool is_unaligned = (f.getDirectionY() == YDirectionType::Aligned);
+    const T f_aligned = is_unaligned ? toFieldAligned(f, "RGN_NOX") : f;
     T result = standardDerivative<T, DIRECTION::Y, DERIV::Standard>(f_aligned, outloc,
                                                                     method, region);
-    return fromFieldAligned(result, region);
+    return is_unaligned ? fromFieldAligned(result, region) : result;
   }
 }
 
@@ -221,10 +222,11 @@ T D2DY2(const T& f, CELL_LOC outloc = CELL_DEFAULT, const std::string& method = 
     return standardDerivative<T, DIRECTION::YOrthogonal, DERIV::StandardSecond>(
         f, outloc, method, region);
   } else {
-    const T f_aligned = toFieldAligned(f, "RGN_NOX");
+    const bool is_unaligned = (f.getDirectionY() == YDirectionType::Aligned);
+    const T f_aligned = is_unaligned ? toFieldAligned(f, "RGN_NOX") : f;
     T result = standardDerivative<T, DIRECTION::Y, DERIV::StandardSecond>(
         f_aligned, outloc, method, region);
-    return fromFieldAligned(result, region);
+    return is_unaligned ? fromFieldAligned(result, region) : result;
   }
 }
 
@@ -237,10 +239,11 @@ T D4DY4(const T& f, CELL_LOC outloc = CELL_DEFAULT, const std::string& method = 
     return standardDerivative<T, DIRECTION::YOrthogonal, DERIV::StandardFourth>(
         f, outloc, method, region);
   } else {
-    const T f_aligned = toFieldAligned(f, "RGN_NOX");
+    const bool is_unaligned = (f.getDirectionY() == YDirectionType::Aligned);
+    const T f_aligned = is_unaligned ? toFieldAligned(f, "RGN_NOX") : f;
     T result = standardDerivative<T, DIRECTION::Y, DERIV::StandardFourth>(
         f_aligned, outloc, method, region);
-    return fromFieldAligned(result, region);
+    return is_unaligned ? fromFieldAligned(result, region) : result;
   }
 }
 
@@ -313,11 +316,15 @@ T VDDY(const T& vel, const T& f, CELL_LOC outloc = CELL_DEFAULT,
     return flowDerivative<T, DIRECTION::YOrthogonal, DERIV::Upwind>(vel, f, outloc,
                                                                     method, region);
   } else {
-    const T f_aligned = toFieldAligned(f, "RGN_NOX");
-    const T vel_aligned = toFieldAligned(vel, "RGN_NOX");
+    ASSERT2(f.getDirectionY() == vel.getDirectionY());
+    const bool are_unaligned = ((f.getDirectionY() == YDirectionType::Standard)
+                                and (vel.getDirectionY() == YDirectionType::Standard));
+
+    const T f_aligned = are_unaligned ? toFieldAligned(f, "RGN_NOX") : f;
+    const T vel_aligned = are_unaligned ? toFieldAligned(vel, "RGN_NOX") : vel;
     T result = flowDerivative<T, DIRECTION::Y, DERIV::Upwind>(vel_aligned, f_aligned,
                                                               outloc, method, region);
-    return fromFieldAligned(result, region);
+    return are_unaligned ? fromFieldAligned(result, region) : result;
   }
 }
 
@@ -333,11 +340,15 @@ T FDDY(const T& vel, const T& f, CELL_LOC outloc = CELL_DEFAULT,
     return flowDerivative<T, DIRECTION::YOrthogonal, DERIV::Flux>(vel, f, outloc, method,
                                                                   region);
   } else {
-    const T f_aligned = toFieldAligned(f, "RGN_NOX");
-    const T vel_aligned = toFieldAligned(vel, "RGN_NOX");
+    ASSERT2(f.getDirectionY() == vel.getDirectionY());
+    const bool are_unaligned = ((f.getDirectionY() == YDirectionType::Standard)
+                                and (vel.getDirectionY() == YDirectionType::Standard));
+
+    const T f_aligned = are_unaligned ? toFieldAligned(f, "RGN_NOX") : f;
+    const T vel_aligned = are_unaligned ? toFieldAligned(vel, "RGN_NOX") : vel;
     T result = flowDerivative<T, DIRECTION::Y, DERIV::Flux>(vel_aligned, f_aligned,
                                                             outloc, method, region);
-    return fromFieldAligned(result, region);
+    return are_unaligned ? fromFieldAligned(result, region) : result;
   }
 }
 
--- a/include/bout/index_derivs_interface.hxx
+++ b/include/bout/index_derivs_interface.hxx
@@ -205,7 +205,7 @@ T DDY(const T& f, CELL_LOC outloc = CELL_DEFAULT, const std::string& method = "D
     return standardDerivative<T, DIRECTION::YOrthogonal, DERIV::Standard>(f, outloc,
                                                                           method, region);
   } else {
-    const bool is_unaligned = (f.getDirectionY() == YDirectionType::Aligned);
+    const bool is_unaligned = (f.getDirectionY() == YDirectionType::Standard);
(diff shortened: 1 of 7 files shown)


ZedThree:
This now fixes a few related bugs to do with the direction tagging. There's a change I would like to make that perhaps needs a bit of discussion: in `ShiftedMetric::to/fromFieldAligned`, we now return a copy of the field if it's already in the result direction, effectively making it a no-op. I think we should change this to instead `ASSERT` that the direction is correct:

```cpp
Field3D ShiftedMetric::toFieldAligned(const Field3D& f, const std::string& region) {
  ASSERT2(f.getDirectionY() == YDirectionType::Standard);
  return shiftZ(f, toAlignedPhs, YDirectionType::Aligned, region);
}
```

3 of the 4 bugs fixed in this PR are because we were silently doing nothing when in fact the input field was in the wrong space, so I think we should make it an error to use `toFieldAligned` on something that's already field-aligned.

d7919:
> This now fixes a few related bugs to do with the direction tagging. There's a change I would like to make that perhaps needs a bit of discussion: in `ShiftedMetric::to/fromFieldAligned`, we now return a copy of the field if it's already in the result direction, effectively making it a no-op. I think we should change this to instead `ASSERT` that the direction is correct:
> 
> ```c++
> Field3D ShiftedMetric::toFieldAligned(const Field3D& f, const std::string& region) {
>   ASSERT2(f.getDirectionY() == YDirectionType::Standard);
>   return shiftZ(f, toAlignedPhs, YDirectionType::Aligned, region);
> }
> ```
> 
> 3 of the 4 bugs fixed in this PR are because we were silently doing nothing when in fact the input field was in the wrong space, so I think we should make it an error to use `toFieldAligned` on something that's already field-aligned.

Can you clarify what the wrong space was in those bugs?

bendudson:
Yes I think I agree that `toFieldAligned` implies that the user thinks the input is not field aligned. There many however be cases where the code author doesn't know if the function input is aligned or not. 

Perhaps we could have `ensureFieldAligned` which is a no-op if the field is already aligned, and `toFieldAligned` which checks and then calls `ensureFieldAligned`?

ZedThree:
Sure, I just mean we were essentially doing `toFieldAligned` on something already field-aligned, or `fromFieldAligned` on something non-field-aligned.

d7919:
I think part of the motivation for the tagging was to allow code to be written that doesn't care about the type of y a field is one, it will just handle things. As long as the tagging is correct it should work, @johnomotani may want to comment as I may be misremembering.

ZedThree:
Also, just to note that the problem isn't necessarily that the field was in the wrong space, but the call to `toFieldAligned` being a no-op hid the actual issue.

d7919:
> Also, just to note that the problem isn't necessarily that the field was in the wrong space, but the call to `toFieldAligned` being a no-op hid the actual issue.

I think this is the bit I'm struggling with at the moment -- if the field was in the correct space and the call was a no-op, so it didn't change the space, then surely the result was in the right space?

johnomotani:
I'm not against having `to/fromFieldAligned` require inputs to be such that they're not no-ops. It does change current functionality though: the reason for the current choices of checking in `toFromFieldAligned` is so that y-derivatives can work in as many cases as possible when using `ShiftedMetric`:
1. input is non-aligned, but has yup/ydown fields: go through yup/ydown branch
2. input is non-aligned, but does not have yup/ydown fields (maybe because it has not been communicated, but is calculated from fields with valid guard cells so that it's ok to take a derivative of it): we call `toFieldAligned`, and then go through the aligned branch
3. input can be field-aligned, then call `toFieldAligned` (but it's a no-op) and derivative can go through the aligned branch

So the behaviour of `toFieldAligned` is to allow either 2 or 3. If we want to change `toFieldAligned` as @ZedThree suggests, we'd either have to have extra special handling in the y-derivatives (which I would rather not have), or we disallow case 2 entirely. Disallowing case 2 seems like not a bad idea to me, because it's only there to catch a case when a user (possibly inadvertently) passed a field to a y-derivative which doesn't have yup/ydown but also isn't field aligned. So for example in STORM we only ever want to pass a field which is field-aligned to y-derivatives, and having stricter checking rather than an implicit `toFieldAligned` inside the derivative would be preferable. On the other hand, it would force some physics models to add extra code, and the extra code is not physics-related so (and I think this is the current position) we don't want to require it to be there, we just give options for users who want to optimize code (to minimize calls to `toFieldAligned`) or do custom things (e.g. sheath boundary conditions applied in the parallel, not poloidal, direction).
[Also, if I remember right, `fromFieldAligned` is always called on the result of the y-derivatives, being a no-op for a result from the yup/ydown branch, which maybe saves a line of code or so but isn't an obstacle to changing the behaviour.]

ZedThree:
> > Also, just to note that the problem isn't necessarily that the field was in the wrong space, but the call to `toFieldAligned` being a no-op hid the actual issue.
> 
> I think this is the bit I'm struggling with at the moment -- if the field was in the correct space and the call was a no-op, so it didn't change the space, then surely the result was in the right space?

For instance, the first commit in this PR: when we initialise a field using `FieldFactory::create3D` (e.g fields in input files), the default is to transform them `fromFieldAligned`. The bug was that we were creating a default `Field3D`, i.e. non-aligned, and the call to `fromFieldAligned` was a no-op -- but the intention was to do a transform. So the problem is not in `fromFieldAligned` but that the input field was not in the intended space.

Similarly in the bug in `test_yupydown`: `DDY_aligned` was creating a default `Field3D`, but the intention was to create a field-aligned one. Again, the bug is not in `fromFieldAligned` but in the input to it.

Lastly, the bug in the shifted metric test was a simple copy-paste error, with the wrong variable being passed to `fromFieldAligned`.

---

I think the direction tagging is overall a net benefit, but I think we've made changes that mean user code will now give different answers without an obvious cause. By ensuring that `to/fromFieldAligned` requires the input to not be in the result space already, we've now got somewhere we can catch the problem.

ZedThree:
> So the behaviour of `toFieldAligned` is to allow either 2 or 3. If we want to change `toFieldAligned` as @ZedThree suggests, we'd either have to have extra special handling in the y-derivatives (which I would rather not have), or we disallow case 2 entirely.

Be able to call `DDY(f)` on any `f` and have it Just Work is definitely what we want. But I think the change is just this:

```diff
     return standardDerivative<T, DIRECTION::YOrthogonal, DERIV::Standard>(f, outloc,
                                                                           method, region);
   } else {
-    const T f_aligned = toFieldAligned(f, "RGN_NOX");
+    const T f_aligned =
+        f.getDirectionY() == YDirection::Standard ? toFieldAligned(f, "RGN_NOX") : f;
     T result = standardDerivative<T, DIRECTION::Y, DERIV::Standard>(f_aligned, outloc,
                                                                     method, region);
     return fromFieldAligned(result, region);
```

(it does need to be done in a few places, however).

`fromFieldAligned` is only called in the non-parallel slice branch in all the derivatives.

Note that we still require fields to be have parallel slices already calculated before `DDY(f)` if FCI is being used. I think the only solution to that is another branch in `DDY` to check `canToFromFieldAlign`.

johnomotani:
@ZedThree that does make sense - it's actally a simple change! Now I totally agree we should change `to/fromFieldAligned` like you suggest.

I think the FCI case should be OK already - if a field without parallel slices is passed it will go to the field-aligned branch but call `FCITransform::toFieldAligned`, which will throw. Possibly the error message is a bit cryptic, but I'd suggest addressing that with a comment in the FCI section of the manual rather than changing code.

johnomotani:
I think always returning an un-aligned field is the behaviour I want... For example in STORM the evolved fields are all in the non-aligned space - I think of the non-aligned space as the 'default' and fields in the aligned space only exist as an 'optimization' or implementation detail. I think it's better to have for example
```
ddt(T) = Div_par(q_par_aligned);
```
where `T` is non-aligned, rather than requiring
```
ddt(T) = fromFieldAligned(Div_par(qpar_aligned));
```

Part of my motivation for this view is that I think I'm in favour of eventually putting the field-aligned version of a field as a member of the non-aligned field. We have operators that need to take both x- and y- derivatives (e.g. `Laplace_perp`). In STORM at the moment we need custom versions of those operators which take both non-aligned and aligned, which is fine for now, but would be nice to be neater in the long run, e.g.
```
ddt(n) += mu_n*storm_Laplace_perp(n, n_aligned);
```

bendudson:
I'm not so sure... If this took a field-aligned input and returned a non field-aligned output:
```
ddt(T) = Div_par(q_par_aligned);
```
then internally it's got to transform from field aligned, because calculating Div_par for a field-aligned grid is easy, but the result is field-aligned unless the transform is done. There would need to be some way to optimise the code to avoid the transform, especially for codes which keep everything field aligned by default (e.g. Hermes). Perhaps an optional argument which specifies the Y alignment? 

Personally I think the least surprising behaviour would be for the output to be in the same space as the input, but happy to be persuaded otherwise. There is the problem of what to do when there are multiple inputs...

johnomotani:
@bendudson 
> internally it's got to transform from field aligned

this is the current behaviour; sorry I should have commented that when I wrote before.

>  codes which keep everything field aligned by default (e.g. Hermes)

Isn't Hermes using `ShiftedMetric` with `yup/ydown` for parallel derivatives? I think 'everything field aligned' is `paralleltransform=identity` and isn't that a non-recommended option in non-slab geometry?

I agree that returning a result in the same space as the input is the least surprising option, but when using `ShiftedMetric` with `toFieldAligned` instead of `yup/ydown` it would mean more boiler-plate code. I'd argue that the primary use-case of `to/fromFieldAligned` is to give field-aligned inputs to y-derivative operators where the result is wanted in the non-aligned space (because if you wanted everything aligned you'd use `IdentityTransform`). I've no objection to having an argument to skip the `fromFieldAligned` if the result is wanted in field-aligned space - I can't see any problem with having more options - but I think the closest thing to having everything 'just work' is to transform the result to non-aligned space as the default.

ZedThree:
This is now ready for review.

It does change the y-derivatives to return the result in the same space as the input. I'm currently looking at what adding an option to change this behaviour looks like. I don't think it's a simple on/off switch, so it might look something like:

```cpp
enum class DDYResultSpace { Standard, Aligned, AsInput };

const Field3D Div_par(const Field3D& f, CELL_LOC outloc = CELL_DEFAULT,
                      const std::string& method = "DEFAULT",
                      DDYResultSpace result_space = DDYResultSpace::AsInput);
...

ddt(T) = Div_par(q_par_aligned, CELL_DEFAULT, "DEFAULT", DDYResultSpace::Standard);
```

which is a little verbose, and requires this extra argument to be added to every operator that might do a derivative in Y.

src/field/field_factory.cxx
```diff
@@ -231,6 +232,8 @@ Field3D FieldFactory::create3D(FieldGeneratorPtr gen, Mesh* localmesh, CELL_LOC
       // Transform from field aligned coordinates, to be compatible with
       // older BOUT++ inputs. This is not a particularly "nice" solution.
       result = fromFieldAligned(result, "RGN_ALL");
+    } else {
+      result.setDirectionY(YDirectionType::Standard);
```
johnomotani: I guess the purpose of this is to handle the default of `transform_from_field_aligned` being `true`, while this is not possible for some `ParallelTransform`s. But if the user has explicitly asked for `transform_from_field_aligned = true` with `FCITransform`, I'd expect an error because it's not possible. The nicest solution would be to default to `transform_from_field_aligned = false` but of course that's not backward-compatible; would the next-best thing be to have the default be `transform_from_field_aligned = coords->getParallelTransform().canToFromFieldAligned()`? Then in this `else` we could throw an exception because we'd only reach it if the user has asked for something impossible.

ZedThree: This turns out to be much more difficult that I first thought. The default for `transform_from_field_aligned` should be set in the constructor for `FieldFactory`. In order to call `canToFromFieldAligned()`, we need the `Coordinates`, but to make that we need the regions to be set up already.  
For `BoutMesh`, this is done at the _end_ of `BoutMesh::load`. Unfortunately, we may need `FieldFactory` at the _beginning_ in order to call `Mesh::get(nx, "nx")`.

So we've got a circular dependency if we want to do this.

I think I'd prefer to leave it for the time being and hopefully come with something clever to fix it in the future.

include/bout/index_derivs_interface.hxx
```diff
@@ -237,10 +239,11 @@ T D4DY4(const T& f, CELL_LOC outloc = CELL_DEFAULT, const std::string& method =
     return standardDerivative<T, DIRECTION::YOrthogonal, DERIV::StandardFourth>(
         f, outloc, method, region);
   } else {
-    const T f_aligned = toFieldAligned(f, "RGN_NOX");
+    const bool is_unaligned = (f.getDirectionY() == YDirectionType::Aligned);
```
johnomotani: Think this should be `YDirectionType::Standard`

tests/unit/field/test_field_factory.cxx
```diff
@@ -768,3 +744,147 @@ TEST_F(FieldFactoryTest, TanhhatArgs) {
...
+
+  const FieldPerp fromFieldAligned(const FieldPerp& f, const std::string&) override {
+    if (f.getDirectionY() != YDirectionType::Aligned) {
+      throw BoutException("Unaligned field passed to fromFieldAligned");
+    }
+    return -f;
+  }
+
+  const Field3D toFieldAligned(const Field3D& f, const std::string&) override { return -f; }
+  const FieldPerp toFieldAligned(const FieldPerp& f, const std::string&) override { return -f; }
```
johnomotani: `fromFieldAligned` checks that the input has `YDirectionType::Aligned` - should `toFieldAligned` similarly check that the input has `YDirectionType::Standard`?

include/bout/index_derivs_interface.hxx
ZedThree: Thanks, this was a particularly stupid typo from when I couldn't make up my mind which way round to do the condition!

PR #1793, opened 2019-09-04
Implement boundary conditions for CELL_ZLOW fields

johnomotani:
Previously, in several of the boundary conditions if the field was at CELL_ZLOW, the apply() method silently did nothing, leaving the boundary cells unset.

### DIFF
--- a/src/mesh/boundary_standard.cxx
+++ b/src/mesh/boundary_standard.cxx
@@ -141,7 +141,7 @@ void BoundaryDirichlet::apply(Field2D &f,BoutReal t) {
   // Check for staggered grids
   
   CELL_LOC loc = f.getLocation();
-  if(mesh->StaggerGrids && loc != CELL_CENTRE) {
+  if(mesh->StaggerGrids and (loc == CELL_XLOW or loc == CELL_YLOW)) {
     // Staggered. Need to apply slightly differently
     
     if( loc == CELL_XLOW ) {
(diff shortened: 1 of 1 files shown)


src/mesh/boundary_standard.cxx
```diff
@@ -481,6 +481,69 @@ void BoundaryDirichlet::apply(Field3D &f,BoutReal t) {
...
+    } else if (loc == CELL_ZLOW) {
+      // Shifted in Z
+
+      for(; !bndry->isDone(); bndry->next1d()) {
+        // Calculate the X and Y normalised values half-way between the guard cell and grid cell
+        BoutReal xnorm = 0.5*(   mesh->GlobalX(bndry->x)  // In the guard cell
+                                 + mesh->GlobalX(bndry->x - bndry->bx) ); // the grid cell
+
+        BoutReal ynorm = 0.5*(   mesh->GlobalY(bndry->y)  // In the guard cell
+                                 + mesh->GlobalY(bndry->y - bndry->by) ); // the grid cell
```
ZedThree: This was a bit surprising at first, but it's using the fact that we never have both `bx` and `by` non-zero, and so always does the correct thing. Is that right, @johnomotani ?

src/mesh/boundary_standard.cxx
```diff
@@ -2820,7 +2962,7 @@ void BoundaryFree_O2::apply(Field3D &f) {
   // Check for staggered grids
   
   CELL_LOC loc = f.getLocation();
-  if(mesh->StaggerGrids && loc != CELL_CENTRE) {
+  if(mesh->StaggerGrids and (loc == CELL_XLOW or loc == CELL_YLOW)) {
```
ZedThree: Is `CELL_ZLOW` incompatible with this boundary condition?

src/mesh/boundary_standard.cxx
johnomotani: @ZedThree that's right - and it wasn't me, all the boundary conditions use that pattern :wink:

src/mesh/boundary_standard.cxx
johnomotani: No, the reason for this condition is that for boundary conditions that don't use a value calculated from expressions, the 'unstaggered' version works for both `CELL_CENTRE` and `CELL_ZLOW`, so I wanted to drop into the `else` case.

PR #1794, opened 2019-09-20
 Allow NYPE to be given instead of NXPE in input files 

johnomotani:
Processor splitting could previously be set manually only by passing `NXPE`, with `NYPE=NPES/NXPE`. This commit adds the option to fix `NYPE` instead, with `NXPE=NPES/NYPE`. If neither are passed then the splitting is set automatically, as before.

I've added tests for `NXPE` and `NYPE` options to `tests/integrated/test-io`, but maybe someone can think of a better place to test?

### DIFF
--- a/manual/sphinx/user_docs/bout_options.rst
+++ b/manual/sphinx/user_docs/bout_options.rst
@@ -296,6 +296,13 @@ direction can be specified:
 
     NXPE = 1  # Set number of X processors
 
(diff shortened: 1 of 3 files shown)


johnomotani:
Not sure if this should go in v4.3 or not - if not, need to change the base branch of this PR.

ZedThree:
I'm ok with this in v4.3. Any other opinions?

PR #1795, opened 2019-09-24
Allow PETSc options to be passed from BOUT.inp

johnomotani:
This is a new feature, so shouldn't really go into v4.3, but it's quite important to me, and being forced to pass options to PETSc via the command line is a bit of a bug...

This PR allows setting arbitrary PETSc options in the `BOUT.inp` input file. Options can be set 'globally' in a `[petsc]` section (implemented in the first commit, if the rest of it is too much to add right away), or in subsections for specific solvers, e.g. `[laplacexy:petsc]`, so that if there are multiple PETSc solvers they can be passed different options.

Arbitrary options can be passed: take the command line option for PETSc; remove the initial hyphen; if the option is a flag that doesn't take a value, pass `true`. So for example command line options
```
-ksp_monitor -ksp_type gmres
```
could be replaced with
```
[petsc]
ksp_monitor = true
ksp_type = gmres
```

I started on this because I'm having difficulty with slow convergence of `LaplaceXY` for some simulations I'm doing. I'm using `hypre` as a preconditioner and wanted to try changing settings to see if that helped. In the PETSc documentation https://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/PC/PCHYPRE.html
> Apart from pc_hypre_type (for which there is PCHYPRESetType()), the many hypre options can ONLY be set via the options database (e.g. the command line or with PetscOptionsSetValue(), there are no functions to set them)

So I wanted a generic way to pass options to `PetscOptionsSetValue()`. Then I realised that it would not be too difficult to add options specific to a particular `PetscLib` object, and so to a particular solver object, by adding a unique prefix to the name of the PETSc option, using `KSPSetOptionsPrefix()`. I don't have a use for this yet, but it seems potentially a very useful feature if someone ever wants to use multiple PETSc-based solvers.

While I was adding a section to the manual, I noticed that the 'Invertable operator' section is missing from the index/table-of-contents, so there's a one-line commit adding that, which didn't seem worth a separate PR.


src/sys/petsclib.cxx
```diff
@@ -47,5 +77,42 @@ void PetscLib::cleanup() {
...
+          "values are allowed in the PETSc options, not subsections",
+          i.first.c_str(), options.str().c_str());
+    }
+    // Note, option names in the input file don't start with "-", but need to be passed
+    // to PETSc with "-" prepended
+    PetscErrorCode ierr;
+    if (lowercase(i.second) == "true") {
+      // PETSc flag with no value
+#if PETSC_VERSION_GE(3, 7, 0)
+      ierr = PetscOptionsSetValue(nullptr, ("-"+pass_options_prefix+i.first).c_str(),
```
d7919: Is it worth pulling out the `"-"+pass_options_prefix+i.first).c_str()` part into a named variable to avoid repetition in the different branches? We could then also use this in the throw message below so we know which variable caused the issue.

src/sys/petsclib.cxx
```diff
@@ -47,5 +77,42 @@ void PetscLib::cleanup() {
   count = 0; // ensure that finalise is not called again later
 }
 
+void PetscLib::setPetscOptions(Options& options, std::string pass_options_prefix) {
```
ZedThree: It doesn't look like we need a copy, so pass `pass_options_prefix` by `const&`.

Minor point, but I think you can also drop `pass_` from the name too -- it's not clear what that signifies here.

johnomotani: I wanted to give the method argument a different name from the class-member-variable `options_prefix`. Maybe the argument could just use a slightly different name, like `prefix`, or do you think it's better to use the same name for both?

src/sys/petsclib.cxx
```diff
@@ -13,7 +13,7 @@ int *PetscLib::pargc = nullptr;
 char ***PetscLib::pargv = nullptr;
 PetscLogEvent PetscLib::USER_EVENT = 0;
 
-PetscLib::PetscLib() {
+PetscLib::PetscLib(Options* opt) : options_prefix("") {
```
ZedThree: The default `string` constructor is the empty string, so we can either drop `options_prefix` from the initaliser list, or replace it with L38 perhaps?

src/sys/petsclib.cxx
```diff
@@ -22,6 +22,24 @@ PetscLib::PetscLib() {
...
+    // Use options specific to this PetscLib
+    // Pass options to PETSc's global options database, with a unique prefix, that will be
+    // passed to a KSP later.
+    // (PetscOptions type exists for non-global options, but apparently is only for user
+    // options, and cannot be passed to KSP, etc. Non-global options can be passed by
+    // defining a custom prefix for the options string, and then passing that to the KSP.)
+
+    options_prefix = "boutpetsclib" + std::to_string(count) + "_";
+
+    Options& options = (*opt)["petsc"];
```
ZedThree: Rather than take the "petsc" subsection from `opt`, why not just use `opt` itself, and pass in the correct subsection when constructing `PetscLib`? I'm not sure we're terribly consistent with this, but I think it's clearer that when we take an `Options` we will use that section itself, and not look for a particular subsection.

johnomotani: I've done it this way because if the `petsc` sub-section isn't in `opt` (see l.30), then `options_prefix` is left as `""` and the solver will use the global options from the `[petsc]` top-level section and/or the command line.

src/sys/petsclib.cxx
ZedThree: Ah, I see. Best not to shadow the member variable.

src/sys/petsclib.cxx
johnomotani: :+1: can drop it. Is it worth putting `= ""` in the header to be explicit instead of relying on the default constructor?

ZedThree: No, the default constructor is in the standard, so it's already explicit enough.

src/sys/petsclib.cxx
ZedThree: Sure, but if e.g. `LaplaceXY` passed `Options::root()["laplacexy"]["petsc"]` and L30 was

```c++
 if (opt != nullptr and opt->isSection()) {
```

that would be equivalent?

include/bout/petsclib.hxx
```diff
@@ -109,11 +119,13 @@ private:
 
 class PetscLib {
 public:
-  PetscLib() {}
+  PetscLib(Options* UNUSED(opt) = nullptr) {}
   ~PetscLib() {}
   
   static void setArgs(int &UNUSED(c), char** &UNUSED(v)) {}
   
+  void createKSPWithOptions(MPI_Comm& comm, KSP& ksp);
```
ZedThree: I know this is the PETSc interface, but it should maybe be `KSP createKSPWithOptions(const MPI_Comm& comm)` instead.

src/sys/petsclib.cxx
johnomotani: I guess that might work. In the case where `petsc` is not there, would your way create a new entry in the options tree? Maybe it doesn't matter if it did?

ZedThree:
Thanks @johnomotani ! I've come up against needing to set PETSc options before and only being able to set them on the command line was a bit limiting.

This sort of thing also just makes me wish for a proper C++ interface to PETSc.

include/bout/invertable_operator.hxx
```diff
@@ -134,7 +134,8 @@ public:
       : operatorFunction(func), preconditionerFunction(func),
         opt(optIn == nullptr ? optIn
```
d7919: Not related to your code but have we (probably I) got this ternary the wrong way around?

ZedThree: Oops, good spot!

include/bout/petsclib.hxx
```diff
@@ -47,6 +47,8 @@ class PetscLib;
 
 #ifdef BOUT_HAS_PETSC
 
+#include "options.hxx"
```
d7919: Could this just be a forward declaration? 

I'm not sure on our current policy regarding favouring forward declarations or not.

include/bout/petsclib.hxx
```diff
@@ -75,7 +77,10 @@ public:
    * The arguments will be passed to PetscInitialize()
    */ 
   static void setArgs(int &c, char** &v) { pargc = &c; pargv = &v;}
-  
+
+  /// Create a KSP linear solver that uses the options specific to this PetscLib
+  KSP createKSPWithOptions(MPI_Comm& comm);
```
d7919: Can we just pass the `BoutComm` instead? It's nice to avoid doing things with raw MPI where we can.

include/bout/petsclib.hxx
```diff
@@ -109,11 +119,13 @@ private:
 
 class PetscLib {
 public:
-  PetscLib() {}
+  PetscLib(Options* UNUSED(opt) = nullptr) {}
   ~PetscLib() {}
   
   static void setArgs(int &UNUSED(c), char** &UNUSED(v)) {}
   
+  KSP createKSPWithOptions(MPI_Comm& comm);
```
d7919: Is `KSP` going to be defined if we're building without petsc or do we need to do something like `define KSP = void*`?

Also without going into detail it looks like we could avoid needing two different definitions of `PetscLib` in this header if we just chopped out the implementations in the source file (which I guess is what happens for the createKsp routine, I've not gotten that far yet!)

include/bout/petsclib.hxx
```diff
@@ -109,11 +119,13 @@ private:
...
+
   static void cleanup() {}
 };
```
d7919: I thought, but might be wrong, that we tried to ensure the PetscLib was a singleton type object. I'm probably wrong but if I'm not then I think that might cause issues with having a top level `prefix_string`. 

As I wrote this I became less convinced that  it is a singleton so I'm probably just wrong.

ZedThree: It's referenced counted rather than a singleton. The last remaining instance will cleanup PETSc.

johnomotani: Before I started there was already a `static int count; ///< How many instances?`, so not a singleton (although I found it surprising that it wasn't! Did turn out to be convenient though...).

manual/sphinx/user_docs/petsc.rst
```diff
@@ -0,0 +1,40 @@
...
+
+in the input file you would put::
+
+    [petsc]
+    ksp_monitor = true
+    ksp_type = gmres
+
+
+.. [*] The object-specific options are passed to PETSc by creating an object-specific
+       prefix ``boutpetsclib#_``, where ``#`` is replaced with an integer counter,
```
d7919: Ah I see -- this sounds like it might be tricky to be sure of. Could we provide a way to set the specific prefix to use for a particular instance of something in the object constructor?

include/bout/petsclib.hxx
johnomotani: I'm not sure. As it is at the moment, I should have put an empty body into the no-PETSc branch of the header; I just missed it...

ZedThree: Good point, yes, we'll need a definition of `KSP` (although with `using KSP = void*`, but yes).

I think `class KSP;` might even be sufficient because nothing should be calling `createKSPWithOptions` without PETSc linked (that is, we should be `#ifdef`-ing the calls out)

manual/sphinx/user_docs/petsc.rst
johnomotani: I wondered about that, but wasn't sure if anyone would actually use it... I think it should be straightforward to do, we can add an optional argument to the constructor, so the user can set a name.

ZedThree: I don't think we need to do anything necessary here, we should just be able to use the usual BOUT++ mechanisms.

That is, instead of doing

```
./model -petsc_option
```

we can now do

```
./model laplacexy:petsc:petsc_option=value
```

More verbose, sure, but also more specific.

src/invert/laplacexy/laplacexy.cxx
```diff
@@ -29,7 +29,8 @@ static PetscErrorCode laplacePCapply(PC pc,Vec x,Vec y) {
 }
 
 LaplaceXY::LaplaceXY(Mesh *m, Options *opt, const CELL_LOC loc)
-    : localmesh(m==nullptr ? bout::globals::mesh : m), location(loc) {
+    : lib(opt==nullptr ? &(Options::root()["laplacexy"]["petsc"]) : &(*opt)["petsc"]),
```
d7919: Rather than storing a new petsc lib which has a specific options attached to it could we just store the ksp directly (and take an option prefix to use or similar)?

include/bout/petsclib.hxx
ZedThree: Just a note that the calling site will need `*BoutComm::getInstance()` to pass a reference to the singleton (as it returns a pointer). Also it can't be `const BoutComm&` because `BoutComm::get` is not `const` (and can't be).

include/bout/petsclib.hxx
ZedThree: Forward declaration should be ok here

src/sys/petsclib.cxx
```diff
@@ -22,6 +22,22 @@ PetscLib::PetscLib() {
...
+
+  if (opt != nullptr and opt->isSection()) {
+    // Use options specific to this PetscLib
+    // Pass options to PETSc's global options database, with a unique prefix, that will be
+    // passed to a KSP later.
+    // (PetscOptions type exists for non-global options, but apparently is only for user
+    // options, and cannot be passed to KSP, etc. Non-global options can be passed by
+    // defining a custom prefix for the options string, and then passing that to the KSP.)
+
+    options_prefix = "boutpetsclib" + std::to_string(count) + "_";
```
d7919: I suspect this could be non-unique. The counter gets incremented on construction and decremented on destruction so we could have something like

`
object someParentObject;
{
auto libA = make_a_lib(); //Make a lib instance in this scope  -- count is 1
SomeParentObject.lib = make_a_lib(); // Make a lib instance for something else -- count is 2
} // Leave the scope so libA is destroyed, count is now 1
auto libB = make_a_lib(); // count is 2 again

`

include/bout/petsclib.hxx
johnomotani: `BoutComm` only represents the global communicator. Methods like `Mesh::getXcomm` and `Mesh::getYcomm` return `MPI_Comm`, so I think getting rid of `MPI_Comm` here would need to be part of a bigger overhaul.

src/sys/petsclib.cxx
johnomotani: Good spot! I'd hope unlikely in practice, but it would be a pig to find if it ever happened. Would the easiest fix would be to just add another counter that only goes up?

d7919: Actually I've just realised `count` is of course static so has a single value across all instances at the time of access. The above concern still applies but there's a further point that we only care about the value of count at the point of creation as this is when the options will be registered, but therefore I don't think we should need to keep the instance of the lib as this really just holds the unique `options_prefix`, so we can instead just hold the `ksp` instance that has this prefix set _or_ hold the prefix and pass that to the ksp creation routine.

src/sys/petsclib.cxx
```diff
@@ -36,6 +52,22 @@ PetscLib::~PetscLib() {
   }
 }
 
+KSP PetscLib::createKSPWithOptions(MPI_Comm& comm) {
+  KSP ksp;
```
d7919: I find this name slightly confusing to start with -- I'm expecting this to set the actual options used, whereas it only sets the prefix such that future calls to set from options will find the right values. 

Would it make sense to pass the relevant section in here and then register the options with petsc at that point. We could either pass through the prefix or allow the section to provide a special prefix value?

Could we also call set from options here as well -- this would presumably then set the defaults, the external code may then override this and then call set from options again as it currently does in order to allow command line arguments to override the internal setup?

src/sys/petsclib.cxx
johnomotani: Is keeping instances of `PetscLib` a problem? It looks like the objects are very lightweight - basically just reference counting so that `PetscInitialize()` is called when the first `PetscLib` is created and `PetscFinalize()` is called when the last one is destroyed. If that's right, maybe the name is a bit misleading... It seems like a reasonable place to me to also hide away the `options_prefix` from the solvers that use PETSc, I think it probably simplifies the solver code a bit not to have to handle the prefix explicitly.

src/sys/petsclib.cxx
johnomotani: Good point, I didn't think that through. I'm not sure from the PETSc docs, but I think all the options-setting functions override anything set before, so you have to call `KSPSetFromOptions()` last to allow the 'options database' to set things at run-time. I don't think there's any benefit to calling it earlier.

I'd bundled the call to `KSPSetOptionsPrefix` in with `KSPCreate` without thinking too hard, just with the aim of not adding yet another function/method that each solver needs to call in the set-up. It might make more sense to instead put `KSPSetOptionsPrefix` and `KSPSetFromOptions` together in a method?

include/bout/petsclib.hxx
ZedThree: That's a fair point. Another thing for 5.0: remove raw `MPI_*` usage

src/invert/laplacexy/laplacexy.cxx
johnomotani: Each of the solvers does directly store a `KSP` already. I guess it could make sense to have some sort of wrapper for `KSP` which would handle the `PetscLib` instances internally (or replace the functionality); would be on the way to the C++ interface to PETSc that Peter wants :-) but on the other hand, the options-handling can be abstracted away from whether the solver is a `KSP`, `SNES`, time-solver, etc. Without totally re-doing our PETSc interface, the `PetscLib` design (with methods to either create (as I've implemented at present), or (alternatively) set options for, `KSP`, `SNES`, etc. instances) doesn't seem to me like a bad compromise for now.

ZedThree:
`InvertableOperator` already sets the options prefix. Is that going to interfere with this PR?

https://github.com/boutproject/BOUT-dev/blob/31650cb4ecc56654fa8a1b3d6f4763bb7de38e41/include/bout/invertable_operator.hxx#L365

johnomotani:
> InvertableOperator already sets the options prefix

Ah, hadn't noticed that. I guess one would override the other, depending which is called last, but I'd prefer to have a single mechanism, i.e. remove the prefix-setting from `InvertableOperator` now that it's more generally implemented. ?

johnomotani:
The checks are failing because `tests/integrated/test-invertable-operator` doesn't compile. `InvertableOperator` calls
```
ksp = createKSPWithOptions(BoutComm::get());
```
and the compiler complains
```
../../../include/bout/invertable_operator.hxx:342:51: error: no matching function for call to ‘PetscLib::createKSPWithOptions(MPI_Comm)’
     ksp = lib.createKSPWithOptions(BoutComm::get());
                                                   ^
In file included from ../../../include/bout/invertable_operator.hxx:50:0,
                 from invertable_operator.cxx:4:
../../../include/bout/petsclib.hxx:82:7: note: candidate: _p_KSP* PetscLib::createKSPWithOptions(ompi_communicator_t*&) <near match>
   KSP createKSPWithOptions(MPI_Comm& comm);
       ^
../../../include/bout/petsclib.hxx:82:7: note:   conversion of argument 1 would be ill-formed:
```
but `LaplaceXY` compiles OK, calling like
```
MPI_Comm comm = BoutComm::get();
ksp = createKSPWithOptions(comm);
```
If I change the function argument from `MPI_Comm&` to `MPI_Comm` the error goes away.

Does someone understand why?

src/sys/petsclib.cxx
johnomotani: Re: the method name - if we decide to keep the method as-is, would `createKSPWithPrefix` be better? (along with a much better docstring in the header to explain why it's called that)

ZedThree:
The compile error you were seeing is because you can't call a function that takes `T&` directly with a function call. This is because `T&` says you may modify the argument, but with a function result there's nothing you can modify. Using an intermediate variable works fine, because now there's something that can be modified.

There are three solutions: `T`, `const T&` and `T&&`. You almost certainly don't want the last one.

Technically, this is because `T&` binds to "non-const lvalues" and a function returns an "rvalue".

src/sys/petsclib.cxx
```diff
@@ -47,5 +78,42 @@ void PetscLib::cleanup() {
...
+    if (not i.second.isValue()) {
+      throw BoutException("Found subsection %s in %s when reading PETSc options - only "
+          "values are allowed in the PETSc options, not subsections",
+          i.first.c_str(), options.str().c_str());
+    }
+    // Note, option names in the input file don't start with "-", but need to be passed
+    // to PETSc with "-" prepended
+    PetscErrorCode ierr;
+    auto petsc_option_name = "-"+prefix+i.first;
+    if (lowercase(i.second) == "true") {
```
ZedThree: If `i.second == "false"`, this goes down the other branch, which is likely not what you intended. But from what I can tell, this branch isn't actually needed, it's just that passing `nullptr` is the same as passing `true`.

johnomotani: It was intended, but may well be a bad idea. I'm not sure the best way to handle this.
1. BOUT++ does handle an option with no value in `BOUT.inp` - it gives it the value `'true'` (I've made a mistake in the manual entry).
2. PETSc has flags that can be passed with no value, which should be set by passing `nullptr` as the last argument (the value) in `PetscOptionsSetValue()`.
3. I can't find an obvious value that can be passed instead of `nullptr`, e.g. passing any of `0`, `1`, `true`, `false`, `t`, `f`, `null`, `default` to `-ksp_monitor` results in a crash (apparently if there is an argument to `-ksp_monitor` it is taken to be a 'viewer' - `ascii` is valid).

There might be PETSc options that need to take the string `true` as an argument, but I haven't come across one yet. Maybe some kind of explicit null argument, e.g. `petsc_null`, would be safer, but taking `"true"` to mean 'pass no argument' lets us have input files like
```
[petsc]
ksp_monitor
options_view
```
work in the way you'd expect. I'd be in favour of leaving this code as it is, and updating the manual to warn that if you need to actually pass `"true"`, you'll need to pass it on the command line. (Alternatively, as a work-around, we could provide some argument like `petsc_true` and convert it to `true` before passing to PETSc?)

ZedThree: Right! Ok, this looks like the best way of handling it then. I had taken that when the PETSc manual says "passing NULL means `true`" that all options that don't take an argument are boolean toggles, but it seems that isn't the case.

We could still maybe remove the branch here by putting the branch on the argument:

```cpp
    // If the value is "true", then this _might_ indicate a PETSc option with an
    // optional value, which is not necessarily a bool. In which case, pass `nullptr`
    const char* value = lowercase(i.second) == "true" ? nullptr : i.second.as<std::string>().c_str();

#if PETSC_VERSION_GE(3, 7, 0)
    ierr = PetscOptionsSetValue(nullptr, petsc_option_name.c_str(), value);
#else
    ierr = PetscOptionsSetValue(petsc_option_name.c_str(), value);
#endif
```

bendudson: what what? why would passing NULL mean `true` in any sensible definition?

ZedThree: For clarity, it's really that you can pass `NULL` for options that don't take a value or take an optional value. In which case, it's equivalent to passing the default value, which is `"true"` for boolean flags.

src/sys/petsclib.cxx
```diff
@@ -47,5 +78,40 @@ void PetscLib::cleanup() {
...
+    auto petsc_option_name = "-"+prefix+i.first;
+    const char* value =
+      // "true" is the value given to an option with no value, when read from BOUT.inp
+      lowercase(i.second) == "true" ? nullptr :
+      // workaround to allow passing "true" to the petsc option
+      lowercase(i.second) == "petsc_true" ? "true" :
+      // allow this for symmetry with "petsc_true"
+      lowercase(i.second) == "petsc_false" ? "false" :
+      // default case: pass the value of the option
+      i.second.as<std::string>().c_str();
```
johnomotani: Is the nested-ternary a good idea here? It looks to me like the neatest way to write this, but it might look a bit cryptic. In https://stackoverflow.com/questions/18237432/how-to-rewrite-complicated-lines-of-c-code-nested-ternary-operator different people have opposite opinions on whether it's good or bad.

d7919: Would a switch be clearer?

d7919: I guess it removes the ability to have const on the declaration.

ZedThree: I'm just going to have a quick check, but I'm reasonably certain that we don't need a workaround to pass `"true"`, and it should be happy with `nullptr`.

This isn't too terrible. It is possible to have `const` with a `switch` using an immediately invoked lambda:

```cpp
const char* value = [i&]() {
   ...
  }();
```

but it would be even better if we could avoid needing the workaround here.

johnomotani: I tried putting a `switch` and found the argument to `switch` has to be an integer, so `switch` isn't an option here.

johnomotani: @ZedThree thanks - I got tired of trying to find in the PETSc docs a statement of all the things it might try to do with an option, so was trying to cover all bases.

d7919: > I tried putting a `switch` and found the argument to `switch` has to be an integer, so `switch` isn't an option here.

Ah yes, good point!

ZedThree: Ok I've had a play about with various options and I'm happy that no workaround is needed, and L97 & 99 can go.

I've also looked at the source code: https://www.mcs.anl.gov/petsc/petsc-current/src/sys/objects/options.c.html#PetscOptionsGetBool
and it does interpret `NULL` as truthy.

johnomotani: :+1: nice, thanks @ZedThree!

ZedThree:
@johnomotani This is very cool, and I want it in, but I'd be much happier putting it into `next` rather than v4.3. I think it's likely we're going to end up iterating a bit on the design once we've had some more experience with it in the code, and `next` is really the place for that.

include/bout/petsclib.hxx
```diff
@@ -61,7 +63,7 @@ public:
   /*!
    * Ensure that PETSc has been initialised
    */
-  PetscLib();
+  PetscLib(Options* opt = nullptr);
```
ZedThree, line 66:
```suggestion
  explicit PetscLib(Options* opt = nullptr);
```

include/bout/petsclib.hxx
```diff
@@ -107,13 +121,19 @@ private:
...
 #include "unused.hxx"
 
+// PETSc not available, so KSP not already defined. KSP should never be called, so forward
+// declaration OK here.
+class KSP;
+
 class PetscLib {
 public:
-  PetscLib() {}
+  PetscLib(Options* UNUSED(opt) = nullptr) {}
```
ZedThree, line 130:
```suggestion
  explicit PetscLib(Options* UNUSED(opt) = nullptr) {}
```

include/bout/petsclib.hxx
```diff
@@ -107,13 +121,19 @@ private:
...
   ~PetscLib() {}
   
   static void setArgs(int &UNUSED(c), char** &UNUSED(v)) {}
   
+  KSP setOptionsFromInputFile(KSP& ksp) {}
```
ZedThree, line 135:
```suggestion
  void setOptionsFromInputFile(KSP& ksp) {}
```

src/sys/petsclib.cxx
```diff
@@ -56,5 +87,37 @@ void PetscLib::cleanup() {
   }
 }
 
+void PetscLib::setPetscOptions(Options& options, const std::string& prefix) {
+  // Pass all options in the section to PETSc
+  for (auto& i : options.getChildren()) {
+    if (not i.second.isValue()) {
+      throw BoutException("Found subsection %s in %s when reading PETSc options - only "
+          "values are allowed in the PETSc options, not subsections",
+          i.first.c_str(), options.str().c_str());
```
ZedThree, line 96: Needs updating for `fmt`; run `bin/bout-v5-format-upgrader.py`

src/sys/petsclib.cxx
```diff
@@ -25,8 +27,25 @@ PetscLib::PetscLib() {
...
+
+  if (opt != nullptr and opt->isSection()) {
+    // Use options specific to this PetscLib
+    // Pass options to PETSc's global options database, with a unique prefix, that will be
+    // passed to a KSP later.
+    // (PetscOptions type exists for non-global options, but apparently is only for user
+    // options, and cannot be passed to KSP, etc. Non-global options can be passed by
+    // defining a custom prefix for the options string, and then passing that to the KSP.)
+
+    options_prefix = "boutpetsclib" + std::to_string(unique_id) + "_";
```
ZedThree, line 43: Could we somehow pass a `name` in `opt` down here, and only fallback to this if we don't have one? 
Could we even fallback to the section name? Then we could avoid the unique ID altogether perhaps?

`Options::str()` actually just returns the full name, including path from root, so this should be globally unique. This is probably more ergonomic, as then you should just be able to do something like `aparsolver:petsc:ksp_type=gmres` on the command line.

I could be missing something here!

src/sys/petsclib.cxx
```diff
@@ -56,5 +87,37 @@ void PetscLib::cleanup() {
...
+    // true so we should always be OK passing nullptr for null or "true".
+    const char* value =
+      lowercase(i.second) == "true" ? nullptr : i.second.as<std::string>().c_str();
+
+    PetscErrorCode ierr;
+#if PETSC_VERSION_GE(3, 7, 0)
+    ierr = PetscOptionsSetValue(nullptr, petsc_option_name.c_str(), value);
+#else
+// no PetscOptions as first argument
+    ierr = PetscOptionsSetValue(petsc_option_name.c_str(), value);
```
ZedThree, line 114:
```suggestion
#if PETSC_VERSION_GE(3, 7, 0)
    const auto ierr = PetscOptionsSetValue(nullptr, petsc_option_name.c_str(), value);
#else
// no PetscOptions as first argument
    const auto ierr = PetscOptionsSetValue(petsc_option_name.c_str(), value);
```

src/invert/laplacexy/laplacexy.cxx
ZedThree: Agreed, this is a reasonable compromise without needing to write a full C++ wrapper for PETSc

johnomotani:
> You might want to save the patch file to undo that change if we also want to back-port this to v4.4.

I don't think I'll have the time/energy to prepare a back-port myself, but I'll make the fmt changes last to make it easier to revert them.

src/sys/petsclib.cxx
johnomotani: I think using `opt->str()` is the best solution. If more than one solver gets its options from the same section, we might reset some options, but because they come from the same section, we'd only reset them to the same value so that should be fine.

While I was tidying up, I also moved getting the `"petsc"` subsection into `PetscLib` instead of getting it at the call-sites, which simplifies the rest of the code, and enforces a standardisation of the subsection naming.

manual/sphinx/user_docs/petsc.rst
ZedThree: This footnote can go now

johnomotani: :+1: good point. I'm testing out the current version now. Will update the manual before calling this PR ready to merge.

johnomotani:
Note for myself - things in the manual to update before merging:
- [ ] remove/update footnote about `prefix \`\`boutpetsclib#_\`\``
- [ ] note that passing command line options has no effect without prefix - recommend passing e.g. `laplace:petsc:ksp_type=gmres` instead

johnomotani:
The PR tests are failing in the "Coverage (minimal)" test with this compilation error
```
In file included from bout_test_main.cxx:5:0:
../../include/bout/petsclib.hxx:129:28: error: expected ‘)’ before ‘*’ token
   explicit PetscLib(Options* UNUSED(opt) = nullptr) {}
```
Anyone have any idea why?

ZedThree:
The forward declaration `class Options;` is inside `#if BOUT_HAS_PETSC`, so when we compile without PETSc, it's not seen.

johnomotani:
Thanks @ZedThree, that's fixed it! All tests pass now.

ZedThree:
Awesome, thanks @johnomotani !

PR #1796, opened 2019-09-25
Make CELL_LOC strongly typed

johnomotani:
Convert `CELL_LOC` to an `enum class` so that it is strongly typed. Use macros to allow continued use of `CELL_DEFAULT`, `CELL_CENTRE`, etc. without the prefix `CELL_LOC::CELL_DEFAULT`, etc.

I ran into a bug today that was (partially) due to accidentally passing a `CELL_LOC` as a function argument which should have been `Field3D`. I guess `enum` is implicitly convertible to `Field3D`, which is unfortunate. A strongly typed `enum class` is much nicer.

I couldn't find a better way to make it backward compatible than using macros `#define CELL_DEFAULT CELL_LOC::CELL_DEFAULT`, etc., to get around the namespace issue (C++20 apparently will allow a `using enum CELL_LOC` syntax). Anyone have a nicer idea for a work-around?

Changing this, I already found one bug (see second commit) that was hidden by implicit conversion of `CELL_CENTRE` to `bool`!


johnomotani:
PS, was re-defining `CELL_LOC` already on the list for v5.0? Even if this is merged, renaming and using scoping for v5.0 is probably sensible...

ZedThree:
It wasn't on the radar, but I think we should move all plain `enum`s to `enum class`es. It does feel like 5.0 thing though.

bendudson:
with the #define, does this actually break any code (apart from finding bugs)? If not then I'd be in favour of merging sooner than 5.0. A further change would be to properly namespace/scope it for 5.0.

johnomotani:
The intention is that it doesn't break any code... I hadn't tried compiling `boutcore` before though, and now am having troubles. Apparently `enum class` is tricky in Cython...

d7919:
That looks good, I'm strongly in favour of switching to enum class everywhere as soon as possible. Boutcore can work with them but may require some manual changes. I had to do this for some earlier enum classes in the region/Ind code

d7919:
Should we apply this change to all the other enums as well so that we're consistent? If there was a bug with one enum there may well be others!

johnomotani:
Thanks @d7919 for pointing me to the right place to fix the `enum class` handling in boutcore :+1:
I think it's working now...

johnomotani:
I converted all the other `enum`s to `enum class`. The changes should all be backward compatible, except for a couple that I think should only be used internally: `SOLVER_VAR_OP` and `BoundaryOpPar::ValueType`.

include/bout/griddata.hxx
```diff
@@ -65,11 +65,17 @@ public:
   virtual bool get(Mesh *m, Field3D &var, const std::string &name, BoutReal def = 0.0) = 0;
   virtual bool get(Mesh *m, FieldPerp &var, const std::string &name, BoutReal def = 0.0) = 0;
 
-  enum Direction { X = 1, Y = 2, Z = 3 };
+  enum class Direction { X = 1, Y = 2, Z = 3 };
+  // Define some aliases so GridDataSource::X, GridDataSource::Y and GridDataSource::Z can
+  // be used, for backward compatibility
+  static constexpr Direction X = Direction::X;
+  static constexpr Direction Y = Direction::Y;
+  static constexpr Direction Z = Direction::Z;
```
ZedThree: Very clever! :smile:

include/difops.hxx
```diff
@@ -314,16 +314,21 @@ const Field3D b0xGrad_dot_Grad(const Field3D &phi, const Field3D &A, CELL_LOC ou
...
+enum class BRACKET_METHOD {
+  standard = 0,        ///< Use b0xGrad_dot_Grad
+  simple = 1,     ///< Keep only terms in X-Z
+  arakawa = 2,    ///< Arakawa method in X-Z (optimised)
+  ctu = 3,        ///< Corner Transport Upwind (CTU)
                           /// method. Explicit method only,
                           /// needs the timestep from the
                           /// solver
-  BRACKET_ARAKAWA_OLD = 4 ///< Older version, for regression testing of optimised version.
+  arakawa_old = 4 ///< Older version, for regression testing of optimised version.
```
ZedThree: If we're touching all these, may as well also delete the explicit numbering too.

Also, very trivial, but please could you realign the comments? :smile:

include/boundary_region.hxx
```diff
@@ -15,13 +15,20 @@ namespace globals {
...
-               BNDRY_ALL=15,
-               BNDRY_PAR_FWD=16,   // Don't include parallel boundaries
-               BNDRY_PAR_BKWD=32};
+enum class BndryLoc {xin=1,
+                     xout=2,
+                     ydown=4,
+                     yup=8,
+                     all=15,
+                     par_fwd=16,   // Don't include parallel boundaries
+                     par_bkwd=32};
```
johnomotani: I just noticed the numbering for `BndryLoc` seems to be set up for some kind of binary-flag operations. An `enum class` probably doesn't work if we need to be able to pass an argument `3` and have it be interpreted as `xin + xout`. However, in `field_data.cxx` the `BndryLoc` is used directly as a key of a `std::map`, so I guess we don't currently support combining `BndryLoc`s. Do we want to? If so, should I revert the `BndryLoc` change?

bendudson: I think this was the original intention, but was never (?) used that way. It would be nice if there were a way to compose boundaries, but there is probably a better way than bit twiddling. 

Presumably the change would catch any code attempting to add `enum class`es, so if nothing turns up then I'm happy for this change to be kept. A better/different way of adding boundaries can be implemented later...

ZedThree: Looks like it's also used in `switch`es elsewhere too, and there's no bit-twiddling going on that I can see.

I would be worried about changing the numbers, in case someone is using the integer values to convert to the `enum`.

johnomotani: @ZedThree If anyone's passing `ints` and having them be implicitly converted, wouldn't the change to `enum class` break that use anyway? I doubt anyone's doing that, but if they are they'd have to change like `somefunction(2)` to `somefunction(<BndryLoc> 2)` even if we didn't get rid of the explicit values, and they might as well change to `BNDRY_XOUT` or `BndryLoc::xout` at the same time?

johnomotani: BTW, what does the `// Don't include parallel boundaries` comment mean? If `par_fwd` or `par_bkwd` is being set, isn't that including a 'parallel boundary'?

ZedThree:
I do wonder about the spelling of some of these. And `default` being a reserved keyword is pretty unfortunate, but we're getting into proper bike-shed territory.

I've checked all the examples compile, so this shouldn't cause anyone any problems.

PR #1800, opened 2019-10-08
Fix assignment instead of comparison in ASSERTs

ZedThree:
While trying to get BOUT++ compiled on Windows, I (read: the compiler) found a couple of dozen assignments in `ASSERT`s which should of course be comparisons.


bendudson:
Is it worth adding an `ASSERT_EQ` macro, to reduce instances of this? 

Perhaps (and this might be taking this too far) the macro could use `#` to convert the expression to a  string constant (as it does for the error message), and then check at compile time that this string doesn't contain `=`?
https://stackoverflow.com/questions/46307205/any-creative-ways-of-checking-if-a-string-contains-a-forbidden-character-at-comp

ZedThree:
Definitely worth thinking about. I'm pretty sure this is not the first time we've let them slip through.

PR #1801, opened 2019-10-09
Fix for Intel enabling -Wnull-dereference despite not accepting it

ZedThree:
Intel needs an extra flag when checking what are acceptable warning flags

There's still a _lot_ of warnings on Intel, but this one is super easy to fix.


d7919:
On a vaguely related note we should try to add `-h noomp` when not explicitly enabling openmp for the Cray compiler.

PR #1804, opened 2019-10-11
Remove some superfluous location checks in Laplacian

ZedThree:
Coefficient locations are checked when set, and checking again in `tridagCoefs` can cause >20% performance degradation.

Location checks retained in the overload called from `laplace_tridag_coefs` because this is called from `Coordinates::Delp2` which can set the location.

This and #1791 are the causes of the performance regression we saw from v4.2.2 -> v4.3.0-rc, and such are the last blockers for the release of v4.3.0!

### DIFF
--- a/src/invert/laplace/invert_laplace.cxx
+++ b/src/invert/laplace/invert_laplace.cxx
@@ -420,6 +420,10 @@ void Laplacian::tridagMatrix(dcomplex *avec, dcomplex *bvec, dcomplex *cvec,
                              const Field2D *d,
                              bool includeguards) {
 
+  // Better have either both or neither C coefficients
+  ASSERT3((c1coef == nullptr and c2coef == nullptr)
+          or (c1coef != nullptr and c2coef != nullptr))
+
   int xs = 0;            // xstart set to the start of x on this processor (including ghost points)
   int xe = localmesh->LocalNx-1;  // xend set to the end of x on this processor (including ghost points)
 
--- a/src/invert/laplace/invert_laplace.cxx
+++ b/src/invert/laplace/invert_laplace.cxx
@@ -299,12 +299,6 @@ void Laplacian::tridagCoefs(int jx, int jy, BoutReal kwave,
     localcoords = localmesh->getCoordinates(loc);
   }
 
-  ASSERT1(c1coef == nullptr || c1coef->getLocation() == loc);
-  ASSERT1(c2coef == nullptr || c2coef->getLocation() == loc);
-  ASSERT1( (c1coef == nullptr and c2coef == nullptr)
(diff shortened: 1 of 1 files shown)


bendudson:
Thanks Peter! Good sleuthing to find these!

ZedThree:
It looks like the location checks in `tridagCoefs(int jx, int jy, int jz, ...)` (which gets called via `Delp2`) could perhaps also be removed, but it doesn't look like it saves very much. It's technically `public` so people might be using it and then not get the benefit of the checking.

src/invert/laplace/invert_laplace.cxx
```diff
@@ -299,12 +299,6 @@ void Laplacian::tridagCoefs(int jx, int jy, BoutReal kwave,
     localcoords = localmesh->getCoordinates(loc);
   }
 
-  ASSERT1(c1coef == nullptr || c1coef->getLocation() == loc);
-  ASSERT1(c2coef == nullptr || c2coef->getLocation() == loc);
-  ASSERT1( (c1coef == nullptr and c2coef == nullptr)
-           or (c1coef != nullptr and c2coef != nullptr) );
```
johnomotani: Is it worth keeping the consistency check for C1 and C2 (maybe as `ASSERT3`)?

ZedThree: Maybe? This is a `protected` method, so isn't really callable from user code and we should be able to check library code is calling it correctly. Having it at `ASSERT3` might not hurt though.

johnomotani: My thinking is that checking here we have a single point to check, rather than relying on every different implementation of a Laplace solver, etc., to do the right thing. I think they are all correct by design at the moment (Laplace solvers initialise `Field2D`s for all the coefficients, and `Delp2` always passes `nullptr`), but maybe there will be a new implementation in future, and a check here might help catch bugs...

ZedThree: This is true :) Unfortunately, it's also called for each mesh point, many, many times, so I was hoping we could avoid any checks in here.

johnomotani: Despite `Delp2` using `tridagCoefs` (via `laplace_tridag_coefs`), this usage is deprecated (at least by comments in `invert_laplace.hxx`), so maybe it'd make sense to do the check in `tridagMatrix` - at least that gets the check out of the lowest-level loop.

ZedThree: :+1: sounds good

ZedThree: I think with some refactoring of `Laplacian` we could eventually remove the need for that check as well.

PR #1803, opened 2019-10-11
Algebraic multigrid

cmacmackin:
This is my work on the Algebraic multigrid solver for the fully-3D Laplace inversion. It is currently passing all unit tests but these only cover simple Cartesian metrics on a single processor. I am awaiting integration tests to try it in more complex scenarios. There is a little bit of simple optimisation still required, prior to merging, and possibly some tidying up hear and there.

There is also the implementation of an AMG solver for the 2D Laplace inversion, implemented by a previous developer. I have not been involved with this code and can not vouch for its behaviour.

I have implemented a new interface to PETSc (see issue #1771) which allows vectors and arrays to be set using the local BOUT++ index objects for that processor. The wrapper will automatically handle conversions to global indexing for PETSc. It also handles the interpolation weights needed when calculating along-field values in the y-direction. It has been extensively unit tested. The unit tests for the index conversion ability required some modest refactoring of the Mesh class, making some methods virtual which weren't previously and adding methods which wrap some MPI calls (so that they can be overridden for testing purposes). This allows the parallel behaviour of the indexing to be tested while running the unit tests in serial.

The way which @johnomotani  has changed the D2DXDY made my unit tests of the Laplace solver fail as the forward version of the Laplace operator I had defined could no longer work. See my email to John below:

> Hi John,
> 
> I got my Laplace solver to the point where it was passing all of my unit tests. I then decided to merge in the `next` branch. I've gotten everything compiling and unit tests are all passing, except the ones for the Laplace solver. These now all fail when I apply my forward Laplace operator due to NaNs occurring while taking D2DXDY (called from Coordinates::Laplace). This is a result of some changes which git-blame shows you made to how boundaries are handled. Previously the routine was
> ```C++
>const Field3D D2DXDY(const Field3D &f, CELL_LOC outloc, const std::string &method, REGION region) {
>    Field3D dfdy = DDY(f, outloc, method, RGN_NOY);
>    f.getMesh()->communicate(dfdy);
>    return DDX(dfdy, outloc, method, region);
>}
>```
> It is now
> ```C++
>Field3D D2DXDY(const Field3D& f, CELL_LOC outloc, const std::string& method,
>    const std::string& region, const std::string& dfdy_boundary_condition) {
>
>  // If staggering in x, take y-derivative at f's location.
>  const auto y_location =
>    (outloc == CELL_XLOW or f.getLocation() == CELL_XLOW) ? CELL_DEFAULT : outloc;
>
[...]

PR #1808, opened 2019-10-18
MPI wrapper class

cmacmackin:
I created a separate pull request for the refactoring of the MPI wrapper requested in PR #1803. I originally added some MPI wrapper methods to BoutMesh. These could be overridden in subclasses used for unit testing, allowing parallel-communication to be simulated on a single processor. @ZedThree suggested that they didn't really belong there and would be better off in their own class. As this requires quite a significant refactor to large portions of the code, I decided to make it a separate PR. However, this PR still contains the PETSc interface which I've submitted in PR #1805, as this is what the MPI wrapper was created for and demonstrates its use for testing. As such, it would be best that it is merged into `next` only after PR #1805.

I have created a new class called `MpiWrapper` and define a global instance of it `bout::globals::mpi`. Its member methods simply forward calls onto PETSc routines and have exactly the same interface. All MPI calls throughout the code now use the methods on the global object (or, in the case of meshes, on a pointer to the global object which I defined for convenience and brevity). The global `MpiWrapper` object is created and destroyed in `BoutInitialise()` and `BoutFinalise()`, respectively.


cmacmackin:
@ZedThree These changes are ready for review. The CI fails on the CMake build which now runs two tests which previously were not routine run: `test-petsc-laplace` and `test-petsc-laplace-MAST-grid`. It appears that these tests always fail including in the `next` branch itself, as reported in Issue #1850.

ZedThree:
Thanks @cmacmackin ! Looks like the simplest way to get this done, and now gives us the opportunity to mock out MPI calls in the tests, which I am unreasonably excited about! :tada:

I'm going to wait till #1859 goes in before merging this.

cmacmackin:
@ZedThree Unfortunately, I've merged this branch into #1859. I will be the first to admit that was not the best idea (I did it without thinking) but at this point it would be quite difficult to reverse it I think. So probably best if this does get merged in first.

d7919:
@cmacmackin at least on github it doesn't look like #1859 includes any of this, so perhaps you didn't push?

cmacmackin:
@d7919 That's right, the changes are still local to my machine. Should be pushing soon, once I iron out a few more things with integration tests.

Issue #1810, opened 2019-10-18
Parallel_transform and paralleltransform


d7919:
In v4.3.0 I seem to be getting an issue if I set `mesh:parallel_transform` to something other than identity. It seems this is a new option in v4.3.0-rc in addition to `mesh:paralleltransform`. 

I've probably just missed where this was introduced and the discussion related to this but I can currently get some quite unexpected behaviour if I the wrong variable by mistake, which is relatively easy given the similarity (also the `_` version needs `shiftedmetric` rather than `shifted`). 

Is this to do with loading the correct type of parallel transform from grid files?

ZedThree:
I believe the underscore version is from checking the grid was created with
the correct transform. Originally it was "coordinate_type" or similar, and
I suggested the change. I didn't anticipate that it would clash with the
other input parameter!

I'm not sure how best to proceed, but it likely needs renaming. There's an
issue here with it being set in hypnotoad too

johnomotani:
With the fix proposed in #1812, I'm not sure `parallel_transform` needs renaming, but if it does, then indeed it needs updating in Hypnotoad too.

ZedThree:
I think this is sufficiently fixed.

PR #1812, opened 2019-10-19
 Ensure parallel_transform is only read from grid files, not from options

johnomotani:
Possible fix for #1810:
`parallel_transform` is only used to check if a grid file is compatible with the chosen `ParallelTransform`. The grid generator (i.e. Hypnotoad) writes the `ParallelTransform` that the grid has been generated for to the variable `parallel_transform` to the grid file, which can then be checked when the `ParallelTransform` is created.
I don't think it's particularly useful to perform this check when the grid is created from options, because in that case the user is defining the whole geometry themself, so can be assumed to have chosen the correct, compatible value for `mesh:paralleltransform`.
So a possible fix for #1810, implemented here, is to add a method to check whether the grid is loaded from a file, and only check `parallel_transform` if it is. That way `parallel_transform` is never read from the options so no confusion can arise.

### DIFF
--- a/include/bout/mesh.hxx
+++ b/include/bout/mesh.hxx
@@ -217,7 +217,7 @@ class Mesh {
   int get(Vector3D &var, const std::string &name, BoutReal def=0.0);
 
   /// Test if input source was a grid file
-  bool isSourceFile() const;
+  bool isDataSourceGridFile() const;
 
   /// Wrapper for GridDataSource::hasVar
   bool sourceHasVar(const std::string &name);
--- a/include/bout/mesh.hxx
+++ b/include/bout/mesh.hxx
@@ -216,6 +216,9 @@ class Mesh {
   /// @returns zero always. 
   int get(Vector3D &var, const std::string &name, BoutReal def=0.0);
 
+  /// Test if input source was a grid file
+  bool isSourceFile() const;
+
   /// Wrapper for GridDataSource::hasVar
   bool sourceHasVar(const std::string &name);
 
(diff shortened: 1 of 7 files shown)


include/bout/mesh.hxx
```diff
@@ -216,6 +216,9 @@ class Mesh {
   /// @returns zero always. 
   int get(Vector3D &var, const std::string &name, BoutReal def=0.0);
 
+  /// Test if input source was a grid file
+  bool isSourceFile() const;
```
ZedThree: To be honest, I struggle to read this as anything other than "is `mesh` a 'source file'?", rather than "is the `mesh` source a file?".

On the other hand, I'm also struggling to think of a better name! `isSourceFromFile` maybe?

ZedThree:
I'm not overly keen on more things going into `Mesh`, but I really can't think of a better way of solving this.

include/bout/mesh.hxx
d7919: IsTheSourceAFile?

ZedThree: I quite like the latter. `mesh->isDataSourceGridFile()` reads pretty clearly to me.

johnomotani:
> I'm not overly keen on more things going into Mesh, but I really can't think of a better way of solving this.

Probably more of a v5 thing, but one longer term solution could be to use the `GridDataSource` directly, i.e. make it a public member of `Mesh`, or accessible with a getter. Then deprecate the use of `Mesh::get()`. The interface to `GridDataSource` looks pretty much OK to use already, the only thing I'd change would be to pass a pointer to the owning `Mesh` into `GridDataSource`'s constructor, to get rid of the `Mesh* m` arguments to `GridDataSource::get`.

A more radical, possibly better, solution could be to have an `Options` object that could read data from a grid file using the `OptionsNetCDF` interface, or from input file using the regular `Options`.

PR #1813, opened 2019-10-19
Make interp_to and finite-volume parallel operators return result in the same space as the input

johnomotani:
`interp_to` should follow the behaviour introduced for y-derivatives in #1791, and return the result in the same space (aligned/non-aligned) as the input.

Previous behaviour was a bug, following the changes to `toFieldAligned` in #1791, as `toFieldAligned` was always called on the input to `interp_to` when interpolating in the y-direction. This was missed by the tests, because `interp_to` was never called on field-aligned inputs: more unit tests added to cover this case.

Also updates the checks in `toFieldAligned`/`fromFieldAligned` in `ParallelTransformIdentity` to match the behaviour of `ShiftedMetric`, for consistency and so that the tests will pick up more errors.

### DIFF
--- a/include/bout/fv_ops.hxx
+++ b/include/bout/fv_ops.hxx
@@ -172,28 +172,34 @@ namespace FV {
   /// @param[in] v_in   The advection velocity.
   ///                   This will be interpolated to cell boundaries
   ///                   using linear interpolation
-  /// @param[in] wave_speed  Maximum wave speed in the system
+  /// @param[in] wave_speed_in  Local maximum speed of all waves in the system at each
+  //                            point in space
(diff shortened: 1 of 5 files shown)


johnomotani:
The finite-volume operators were missed in an earlier update of the behaviour of `toFieldAligned`/`fromFieldAligned` and parallel derivatives (to return field-aligned result if the input was field-aligned). I've added an update to this PR to fix.

johnomotani:
As discussed on Slack, there was a small bug in `FV::Div_par` where the `wave_speed` argument was not transformed to field-aligned, but should have been. Fixed by 19254ff.

Issue #1811, opened 2019-10-18
Inefficient derivatives with implicit Field2D conversion


d7919:
When using implicit conversion of a Field2D to a Field3D we will end up with a 3D field that does not have parallel slices and will lose the information that this data has trivially calculable parallel slices. This can have a performance impact, for example in `VDDY` we only provide `Field2D,Field2D` and `Field3D,Field3D` interfaces. Hence when we call `VDDY` on mixed fields, as is common in bracket operations, we end up converting the `Field2D` to `Field3D`. As this new 3D field doesn't have parallel slices then we end up going down the path that transforms both input fields to field aligned and then transforming back. These three transforms are an added cost that is not necessary.

I thought we might be able to avoid this by adding a `calcParallelSlices` call using an identity transform in the `Field3D(Field2D)` constructor but this may not work with FCI.

There may be other derivatives where this is an issue.

PR #1816, opened 2019-10-22
Performance fixes: ShiftedMetric and VDDY

ZedThree:
Two unrelated performance fixes for v4.3 compared to v4.2.2:

1. The new fancy `ShiftedMetric::calcParallelSlices` turns out to be significantly slower than the old method using pointers. The first commit here reverts back to using that method -- although still keeps the multiple parallel slices.
2. `VDDY` was not taking advantage of an optimisation that's in master: if we're not using staggered grids, then we don't need the parallel slices for the velocity. This means we only need to check if the input field has parallel slices.

Together, these two fixes can take us from being ~10% slower than v4.2.2 to ~5% faster per rhs (for a given model, on one core, etc.).

### DIFF
--- a/include/bout/index_derivs_interface.hxx
+++ b/include/bout/index_derivs_interface.hxx
@@ -308,9 +308,34 @@ template <typename T>
 T VDDY(const T& vel, const T& f, CELL_LOC outloc = CELL_DEFAULT,
        const std::string& method = "DEFAULT", const std::string& region = "RGN_NOBNDRY") {
   AUTO_TRACE();
-  const bool fHasParallelSlices = (f.hasParallelSlices());
-  const bool velHasParallelSlices = (vel.hasParallelSlices());
-  if (fHasParallelSlices && velHasParallelSlices) {
+
+  auto* localmesh = f.getMesh();
+  const CELL_LOC allowedStaggerLoc = localmesh->getAllowedStaggerLoc(DIRECTION::Y);
+  const CELL_LOC inloc = f.getLocation();
+  const CELL_LOC vloc = vel.getLocation();
+  if (outloc == CELL_DEFAULT) {
+    outloc = inloc;
+  }
+  const STAGGER stagger = localmesh->getStagger(vloc, inloc, outloc, allowedStaggerLoc);
+
+  // Some notes about the choice to be made here:
+  // - If we're not staggered, then we don't need the parallel slices
+  //   for `vel`, and we can use the `YOrthogonal` version if `f` has
+  //   parallel slices. Otherwise, we need to use the aligned version.
+  // - If we are staggered, we can only use the `YOrthogonal` version if
+  //   both `vel` and `f` have parallel slices.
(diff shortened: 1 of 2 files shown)


include/bout/index_derivs_interface.hxx
```diff
@@ -308,9 +308,34 @@ template <typename T>
...
+  //   parallel slices. Otherwise, we need to use the aligned version.
+  // - If we are staggered, we can only use the `YOrthogonal` version if
+  //   both `vel` and `f` have parallel slices.
+  // - If we are staggered and only `f` has parallel slices, then we
+  //   have another choice: calculate the parallel slices for `vel`
+  //   and use the `YOrthogonal` version; or throw away the slices for
+  //   `f` and use the aligned version. Which is cheaper depends on
+  //   how many parallel slices we're using. Our current estimate as
+  //   of October 2019 is if MYG > 1, it's cheaper to use the aligned
+  //   version.
```
d7919: This is also cheaper if neither field has slices for any myg.

include/bout/index_derivs_interface.hxx
```diff
@@ -308,9 +308,34 @@ template <typename T>
 T VDDY(const T& vel, const T& f, CELL_LOC outloc = CELL_DEFAULT,
        const std::string& method = "DEFAULT", const std::string& region = "RGN_NOBNDRY") {
   AUTO_TRACE();
-  const bool fHasParallelSlices = (f.hasParallelSlices());
-  const bool velHasParallelSlices = (vel.hasParallelSlices());
-  if (fHasParallelSlices && velHasParallelSlices) {
+
+  auto* localmesh = f.getMesh();
```
d7919: This is duplicated from the `flowDerivative` method I think, is it worth naming this in a function?

ZedThree: There's enough local variables that are needed outside the snippet that I didn't think it worth it. We'd have either calculate them outside the function again, or return them from the function

include/bout/index_derivs_interface.hxx
d7919: I'll just mention here as well that as we only have a `{3D,3D}` and `{2D,2D}` version of this routine it's not uncommon for us to get 2D->3D implicit conversion here -- these converted fields won't have parallel slices even though they're trivial to calculate for non-FCI. This relates to issue #1811. One solution there would be to but an identity transform in the converting constructor to ensure 3D created from 2D automatically gets sensible slices, but this wouldn't work for FCI etc. 

From timing VDDY etc. for different field combinations it seems that the calculation of the derivative can actually be cheaper than copying the data from the 2D into a 3D so the most cpu efficient option here may actually be to pass in Field3D versions of your 2D data.

include/bout/index_derivs_interface.hxx
d7919: Worth adding a comment to say what this should be consistent with (i.e. the code in the other routine)?

PR #1817, opened 2019-10-22
Add counters to Timers to avoid double-counting sections

ZedThree:
Reimplements some of #1658 in master

Reference counting Timers turns out to be necessary or nested timers
with the same name will end up double counting the section. Currently,
this results in the "Inv" timer report in the default monitor being
twice as large as it really is.

### DIFF
--- a/src/sys/timer.cxx
+++ b/src/sys/timer.cxx
@@ -5,20 +5,29 @@ using namespace std;
 
 Timer::Timer() {
   timing = getInfo("");
-  timing->started = MPI_Wtime();
-  timing->running = true;
+  if (timing->counter == 0) {
+    timing->started = MPI_Wtime();
+    timing->running = true;
+  }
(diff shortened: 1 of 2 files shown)


src/sys/timer.cxx
```diff
@@ -5,20 +5,29 @@ using namespace std;
...
+  timing->counter++;
 }
 
 Timer::~Timer() {
-  double finished = MPI_Wtime();
-  timing->running = false;
-  timing->time += finished - timing->started;
+  timing->counter--;
+  if (timing->counter == 0) {
+    double finished = MPI_Wtime();
```
d7919: Do we need to store this or can we just use `MPI_Wtime` directly below?

ZedThree: I was just adding the counters for this PR, as this method is changed again in 4.3

PR #1827, opened 2019-10-30
Bump version in next to 5.0.0-alpha

ZedThree:
To save confusion when comparing logs etc. from next and master, I'm going to start bumping the version in next immediately after release, and including an `-alpha` version tag.

It sounds like the consensus is we should have a long-lived `4.4` branch with backports. That will get versioned similarly.


CMakeLists.txt
```diff
@@ -6,9 +6,14 @@ else()
   cmake_policy(VERSION 3.12)
 endif()
 
+# CMake currently doesn't support proper semver
+# Set the version here, strip the prerelease identifier to use in `project`
+set(BOUT_FULL_VERSION 5.0.0-alpha)
+string(REPLACE "-alpha" "" BOUT_CMAKE_ACCEPTABLE_VERSION ${BOUT_FULL_VERSION})
```
bendudson: Is there a way to chop off everything after an `-` so we can have other labels if needed later?

ZedThree: Good point! Looks like `REGEX REPLACE` will do it

PR #1829, opened 2019-10-30
Include all boundary cells in GlobalNy

johnomotani:
The `GlobalNx`, `GlobalNy` variables include boundary cells, but previously `GlobalNy` only included boundary cells from one divertor. If a second divertor boundary is present, the guard cells from there should also be included. This change makes `GlobalNy` consistent with `getGlobalYIndex` in the sense that on the last processor `getGlobalYIndex(LocalNy) == GlobalNy` always.

The lines setting the global grid sizes are moved below those getting the separatrix/X-point locations just so that setting `GlobalNy` can use `jyseps2_1` and `jyseps1_2` - the other lines are not changed.

### DIFF
--- a/src/field/globalfield.cxx
+++ b/src/field/globalfield.cxx
@@ -243,7 +243,7 @@ int GlobalField2D::msg_len(int proc) const {
 
 GlobalField3D::GlobalField3D(Mesh *m, int proc)
   : GlobalField(m, proc, m->GlobalNx,
-                m->GlobalNy-mesh->numberOfYBoundaries()*2*m->ystart, m->LocalNz),
+                m->GlobalNy-m->numberOfYBoundaries()*2*m->ystart, m->LocalNz),
     data_valid(false) {
   
   if((proc < 0) || (proc >= npes))
--- a/src/field/globalfield.cxx
+++ b/src/field/globalfield.cxx
@@ -77,8 +77,9 @@ void GlobalField::proc_size(int proc, int *lx, int *ly, int *lz) const {
 ///////////////////////////////////////////////////////////////////////////////////////////
 
 GlobalField2D::GlobalField2D(Mesh *m, int proc)
-  : GlobalField(m, proc, m->GlobalNx, m->GlobalNy-m->numberOfYBoundaries()*m->ystart, 1),
-  data_valid(false) {
+  : GlobalField(m, proc, m->GlobalNx,
+                m->GlobalNy-m->numberOfYBoundaries()*2*m->ystart, 1),
+    data_valid(false) {
   
   if((proc < 0) || (proc >= npes))
     throw BoutException("Processor out of range");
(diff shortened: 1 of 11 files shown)


johnomotani:
The Travis tests are not completing because they time out when `test-initial` is trying to run on 3 processors (the 1 and 2 processor cases run and pass). The test only takes a few seconds on my laptop which only has 2 physical cores, and looks like it used to take ~30s on Travis (looking at the last commit on `master`). I also don't think this PR should particularly affect the test. Has Travis changed something??

ZedThree:
Not to my knowledge. The same tests work fine on other builds. I'll have a look on my machine now

ZedThree:
Works on my machine too, so must be something about travis. Will investigate further this afternoon

src/mesh/impls/bout/boutmesh.cxx
```diff
@@ -170,6 +162,17 @@ int BoutMesh::load() {
   Mesh::get(jyseps2_2, "jyseps2_2", ny - 1);
   Mesh::get(ny_inner, "ny_inner", jyseps2_1);
 
+  // Set global grid sizes
+  GlobalNx = nx;
```
ZedThree: This needs to be set before the calls to `Mesh::get(ixseps...)` on L157,158

src/mesh/impls/bout/boutmesh.cxx
```diff
@@ -170,6 +162,17 @@ int BoutMesh::load() {
...
+  GlobalNy = ny + 2 * MYG;
+  if (jyseps1_2 != jyseps2_1) {
+    GlobalNy += 2*MYG;
+  }
+  GlobalNz = nz;
+
+  if (2 * MXG >= nx)
```
ZedThree: I know this was already there, but it's identical to the check on L152, so let's delete it

src/field/globalfield.cxx
```diff
@@ -239,8 +241,10 @@ int GlobalField2D::msg_len(int proc) const {
 
 ///////////////////////////////////////////////////////////////////////////////////////////
 
-GlobalField3D::GlobalField3D(Mesh *m, int proc) : GlobalField(m, proc, m->GlobalNx, m->GlobalNy-2*m->ystart, m->LocalNz), 
-  data_valid(false) {
+GlobalField3D::GlobalField3D(Mesh *m, int proc)
+  : GlobalField(m, proc, m->GlobalNx,
+                m->GlobalNy-mesh->numberOfYBoundaries()*2*m->ystart, m->LocalNz),
```
ZedThree: I got a warning about `mesh` being uninitialised here -- I guess `m` is correct.

ZedThree:
Please can you add a section at the top of [`CHANGELOG.md`](https://github.com/boutproject/BOUT-dev/tree/next/CHANGELOG.md) like so:

```markdown
# Change Log

## [v5.0.0-alpha](https://github.com/boutproject/BOUT-dev/tree/next) 
[Full Changelog](https://github.com/boutproject/BOUT-dev/compare/v4.3.0...next)

### Breaking changes

- `BoutMesh::GlobalNy` now counts points from all
  y-boundaries. Previously it only counted points from one boundary
  [\#1829](https://github.com/boutproject/BOUT-dev/pull/1829)
```

PR #1832, opened 2019-11-04
Fix various warnings

ZedThree:
Fixes the following warnings on gcc:
- overloaded-virtual
- old-style-cast
- non-virtual-dtor
- pedantic

There are >300 warnings from our use of `int` instead of `size_t` in lots of places. I'm not a huge fan of unsigned types, but seeing as the standard library does use them, we might be better off switching to use them. There's probably a few places where it's not as simple as just changing the type, and we'd have to make other changes, but the majority of places are likely fine.

### DIFF
--- a/include/bout/field_visitor.hxx
+++ b/include/bout/field_visitor.hxx
@@ -22,7 +22,7 @@ public:
   virtual void accept(Field3D &f) = 0;
   virtual void accept(Vector2D &f) = 0;
   virtual void accept(Vector3D &f) = 0;
-  
+  virtual ~FieldVisitor() = default;
 };
 
 #endif // __FIELD_VISITOR_H__
--- a/include/field.hxx
+++ b/include/field.hxx
@@ -505,14 +505,14 @@ inline T pow(BoutReal lhs, const T& rhs, REGION rgn) {
 /// This loops over the entire domain, including guard/boundary cells by
 /// default (can be changed using the \p rgn argument).
 /// If CHECK >= 3 then the result will be checked for non-finite numbers
-FIELD_FUNC(sqrt, ::sqrt);
+FIELD_FUNC(sqrt, ::sqrt)
 
 /// Absolute value (modulus, |f|) of \p f over region \p rgn
 ///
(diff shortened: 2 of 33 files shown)


include/bout/field_visitor.hxx
```diff
@@ -22,7 +22,7 @@ public:
   virtual void accept(Field3D &f) = 0;
   virtual void accept(Vector2D &f) = 0;
   virtual void accept(Vector3D &f) = 0;
-  
+  virtual ~FieldVisitor() = default;
 };
```
bendudson: semicolon?

include/field.hxx
```diff
@@ -505,22 +505,22 @@ inline T pow(BoutReal lhs, const T& rhs, REGION rgn) {
 /// This loops over the entire domain, including guard/boundary cells by
 /// default (can be changed using the \p rgn argument).
 /// If CHECK >= 3 then the result will be checked for non-finite numbers
-FIELD_FUNC(sqrt, ::sqrt);
+FIELD_FUNC(sqrt, ::sqrt)
```
bendudson: I think I may have put in a lot of these semicolons after macro calls. I think it was due to the editor indentation; without the semicolon my editor at the time would think the next line was a continuation, and helpfully indent it. Is this an issue for clang formatting?

include/bout/field_visitor.hxx
ZedThree: Required here :)

bendudson: I am so confused...

include/field.hxx
ZedThree: Yeah, I seem to recall having problems here in the past, but both emacs and clang-format Do The Right Thing now, so I'm happy removing them.

include/bout/field_visitor.hxx
ZedThree: It's the end of a `class` declaration

bendudson: but the end of a template class doesn't need one?

Issue #1828, opened 2019-10-30
Issues with LaplacePDD


ZedThree:
Does anyone use LaplacePDD? I think there's at least one bug in it:

https://github.com/boutproject/BOUT-dev/blob/123e774f865d4c98d378b909b89accd7efd56dc3/src/invert/laplace/impls/pdd/pdd.cxx#L172-L174

`data.jy` should probably come before `kz`. I only noticed because I'm getting a warning about converting `kz * kwaveFactor` to an `int` which definitely smells wrong.

Secondly, `test-laplace` fails if I use PDD (and run it only with nproc=4), with max differences between 0.1 and 10. This requires the above fix, or else it throws out-of-bounds.

The header to PDD says it neglects "small" corrections and has a big warning:

> CHECK IF THIS ALGORITHM PRODUCES REASONABLE RESULTS FOR YOUR PROBLEM

This suggests it probably needs its own tests, and its assumptions stating up front. I guess `test-laplace` violates these assumptions somehow.

bendudson:
This was a test, the paper promised great things, but I don't think it ever worked. The bug probably didn't help, but I suspect it's just not a very good method. I vote for removing in 5

ZedThree:
Do we still want to remove LaplacePDD? We never "officially" deprecated it

bendudson:
Yes I vote for removing it. Joseph's methods are certainly better and probably faster

PR #1834, opened 2019-11-06
Remove deprecated Laplacian API

ZedThree:


src/invert/laplace/invert_laplace.cxx
```diff
@@ -71,45 +71,56 @@ Laplacian::Laplacian(Options *options, const CELL_LOC loc, Mesh *mesh_in)
...
+  if (maxmode < 0) {
+    maxmode = 0;
+  }
+  if (maxmode > ncz / 2) {
+    maxmode = ncz / 2;
+  }
+
+  low_mem = (*options)["low_mem"]
+                .doc("If true, reduce the amount of memory used")
+                .withDefault(false);
```
johnomotani: It looks like `low_mem` is only actually implemented now by `LaplacePDD`. If we remove that (#1828) we can remove the option too.

src/invert/laplace/invert_laplace.cxx
```diff
@@ -71,45 +71,56 @@ Laplacian::Laplacian(Options *options, const CELL_LOC loc, Mesh *mesh_in)
...
+  all_terms =
+      (*options)["all_terms"].doc("Include first derivative terms?").withDefault(true);
 
-  if (options->isSet("flags")) {
-    if ( options->isSet("global_flags") || options->isSet("inner_boundary_flags") || options->isSet("outer_boundary_flags") ) {
-      throw BoutException("Should not use old flags as well as new global_flags/inner_boundary_flags/outer_boundary_flags");
-    }
-    int flags = (*options)["flags"]
-                    .doc("Flags to control inner and outer boundaries.")
-                    .withDefault(0);
```
johnomotani: I agree with removing this option in v5, but it might be a good idea to put a deprecation warning on use of the `laplace:flags` option in v4.3.1/v4.4. I can't think of anything better than printing a message to `output_warn` here, which isn't very noticable, but better than nothing...

src/invert/laplace/invert_laplace.cxx
```diff
@@ -71,45 +71,56 @@ Laplacian::Laplacian(Options *options, const CELL_LOC loc, Mesh *mesh_in)
...
-    Laplacian::setFlags(flags);
-  }
-  else {
-    OPTION(options, global_flags, 0);
-    OPTION(options, inner_boundary_flags, 0);
-    OPTION(options, outer_boundary_flags, 0);
-  }
+  global_flags = (*options)["global_flags"].doc("Default flags").withDefault(0);
```
johnomotani: "Default flags" seems a bit vague, although I'm struggling to come up with something more descriptive - maybe something like "Flags to set Laplace solver options (apart from boundary conditions)"?

tests/integrated/test-laplace/test_laplace.cxx
```diff
@@ -12,74 +12,95 @@ int main(int argc, char **argv) {
...
 
-  Field3D flag0 = invert_laplace(input, 0);
-  Field3D flag3 = invert_laplace(input, 3);
+  auto lap = std::unique_ptr<Laplacian>{Laplacian::create()};
+
+  lap->setCoefA(0.0);
+  lap->setCoefC(1.0);
+  lap->setCoefD(1.0);
+  Field3D flag0 = lap->solve(input);
+  lap->setInnerBoundaryFlags(INVERT_DC_GRAD + INVERT_AC_GRAD);
```
johnomotani: It's a bit worrying that we don't test the outer-boundary-condition flags here. It'd be nice to at least set both inner- and outer-flags to the same thing. Then we'd have to update the benchmark data as well. Maybe there's a way to hack around that though... `input` and `a` are even in `(x-0.5)`, while `c` is odd, so if we duplicate each test, setting the outer boundary flags instead of the inner boundary flags then (remembering the `c` term is `1/C * Grad_perp(C)*Grad_perp(f)`, so it's sign doesn't flip even if `c`'s does) we'd have a new solution `f2(x-0.5,z)=f(-(x-0.5),z)`, where `f(x,z)` is the solution with the inner boundary condition set. So in `runtest` I think we'd just have to reverse the x-dimension of the stored solution (for the inner-boundary set case) to get the expected solution with outer boundary flags set.

src/invert/laplace/invert_laplace.cxx
ZedThree: Good idea! I think there's a few options we'd like to deprecate like this too, although none spring to mind right now.

src/invert/laplace/invert_laplace.cxx
ZedThree: Very true. I just took the docstring from the header.

tests/integrated/test-laplace/test_laplace.cxx
ZedThree: Are you ok if we hold off on fixing this now, and include it in whatever rewrite of the Laplacians we end up doing?

johnomotani: yep, makes sense to me.

Issue #1836, opened 2019-11-06
Problem using RGN_OUTER_X


AdamD94:
I might be misinterpreting how regions are intended to be used but I think the firstX() check below should be lastX(), maybe it's a copy-paste error from the previous region?
I think as it stands RGN_OUTER_X will always be empty.
Am I misinterpreting or should I open a pull request?

https://github.com/boutproject/BOUT-dev/blob/c52eba6b74a56765f456cfac2d2bd1988a05235f/src/mesh/impls/bout/boutmesh.cxx#L2341-L2356

d7919:
That certainly looks like a bug to me.

PR #1837, opened 2019-11-09
Fix typo in initialisation of rgn_outer_x

AdamD94:
Currently rgn_outer_x will contain the outer x guard cells of the first processors in x rather than the outer boundary cells from the last x processors unless the domain is periodic in x.
I don't think this is the intended behaviour so this pull request switches the firstX() check for lastX() when initialising the region.
Addresses #1836


bendudson:
Probably should go into master branch too

ZedThree:
Failing tests are just due to missing coverage token

ZedThree:
I'll cherry-pick into master now

Issue #1835, opened 2019-11-06
More user-friendly options for Laplacian solvers


johnomotani:
Using flags to set many of the options for Laplacian solvers is cryptic, and not consistent with the way options are handled everywhere else. Could we replace them with something nicer? For example:
```
// Global options
bool zero_dc = false; /// Zero the DC (constant in Z) component of the solution
bool start_new = false; /// Iterative method start from solution=0. Has no effect for direct solvers
kx_zero = false; /// Zero the kx=0, n = 0 component

// Boundary options
std::string inner_boundary_condition; /// possible values "dirichlet", "neumann", "laplacian" or "cylinder"
std::string inner_boundary_condition_dc; /// default to same as inner_boundary_condition, possible values "dirichlet", "neumann", "laplacian", "cylinder", "gradpar", "gradparinv"
std::string inner_set; /// "zero", "initial_guess" or "rhs" [no flag, INVERT_SET or INVERT_RHS respectively]
bool inner_one_point = false; /// Only use one boundary point
... same for outer_* [except no "cylinder" option, as this is only needed for the inner boundary that could be the axis of the cylinder]
```
* Don't need replacement for `INVERT_BOTH_BOUNDARY_ONE`, can be set with boundary options
* Not sure `INVERT_4th_ORDER` option is needed - the `serial_band` solver is always 4th order anyway, the PETSc solver made an attempt at a 4th order implementation, but that's experimental/untested and I suspect the boundary conditions may not be correct (also the PETSc solver ignores the flag and reads a `bool fourth_order` option already)
* `INVERT_KX_ZERO` is only supported by serial_tri and serial_band, maybe we can deprecate/remove?
* `INVERT_SYM` can be removed as it is not implemented anywhere (made redundant by re-definition of boundary conditions in 552c109511f68a2e03df26561f1e49487d8d60f2)

See also #91.

ZedThree:
I am absolutely for this. 

What's the best type for the `*_boundary_condition` options? Could users in principle define their own boundary conditions to be applied? If so, `string` makes sense. If they're things that needed to be implemented in the library, then an `enum class` might be better.

If lots of these options can be combined, bit flags might make sense, as opposed to taking lots of `bool`s for instance.

johnomotani:
I don't think there's a sensible way for users to define their own boundary conditions (without some major refactor that's beyond my imagination now anyway), so `enum class` probably makes most sense.

[...]

PR #1841, opened 2019-11-14
Macro for creating enum classes

johnomotani:
After the discussion on #1835 I was thinking about using an `enum class` as an option. The way we are using `enum class`es at the moment has some boiler-plate to define `toString` and `enumFromString` functions. Using some magic copied from `bout/macro_for_each.hxx`, it seems to be possible to replace that with a macro so you can write something like
```
BOUT_ENUM_CLASS(myenum, foo, bar);
```
and get an `enum class` with possible values `myenum::foo` and `myenum::bar`, along with `toString(myenum e)` and `myenumFromString(std::string s)` functions. I've also put in overloads for `Options::as<>()` and `ostream& operator<<` so that the `enum class` can be loaded from options, e.g.
```
myenum e = options["myenum_option"].withDefault(myenum::foo);
```

I'd also like to be able to assign a string to the `enum class`, but it doesn't seem to be possible to override `operator=` for an `enum class`.

Before I think about this any more: is this a useful feature, or an abuse of macros?

### DIFF
--- a/include/bout/bout_enum_class.hxx
+++ b/include/bout/bout_enum_class.hxx
@@ -53,53 +53,53 @@
 #define _ec_expand_10(_call, enumname, x, ...) \
   _call(enumname, x) _ec_expand_9(_call, enumname, __VA_ARGS__)
 
-#define _BOUT_ENUM_CLASS_MAP_ARGS(mac, enumname, ...)                              \
-  _GET_FOR_EACH_EXPANSION(__VA_ARGS__,                                             \
-                          _ec_expand_10, _ec_expand_9, _ec_expand_8, _ec_expand_7, \
-                          _ec_expand_6, _ec_expand_5, _ec_expand_4, _ec_expand_3,  \
-                          _ec_expand_2, _ec_expand_1)                              \
+#define BOUT_ENUM_CLASS_MAP_ARGS(mac, enumname, ...)                                   \
+  BOUT_GET_FOR_EACH_EXPANSION(__VA_ARGS__,                                             \
+                              _ec_expand_10, _ec_expand_9, _ec_expand_8, _ec_expand_7, \
+                              _ec_expand_6, _ec_expand_5, _ec_expand_4, _ec_expand_3,  \
+                              _ec_expand_2, _ec_expand_1)                              \
   (mac, enumname, __VA_ARGS__)
 
-#define _ENUM_CLASS_STR(enumname, val) {enumname::val, lowercase(#val)},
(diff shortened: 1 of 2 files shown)


ZedThree:
I think this is probably essential! Worthwhile looking over some other implementations of compiletime enum reflection:

- https://github.com/aantron/better-enums
- https://github.com/Neargye/magic_enum
- https://github.com/quicknir/wise_enum
- https://github.com/therocode/meta_enum
- https://github.com/BlackMATov/enum.hpp

Some of these are C++17 only, but could be backported depending on what they use; some make an actual `class` rather than an `enum`, which might be painful for the derivative template machinery.
We might be able to use one of those wholesale, or else adapt to our needs.

I've almost finished adding `fmt` to BOUT++, which could also be useful here. It has `fmt::basic_string_view` which is an implementation of C++17's `std::string_view` that some of the above implementations use.

bendudson:
Thanks @johnomotani I think this is both a wonderful abuse of macros, and a very useful feature. The overload of `Options::as` is very nice.

d7919:
Can the work in progress label be removed?

johnomotani:
Documentation of the new macro is definitely still needed.

@ZedThree - are there any useful changes to make right now from your earlier comment, or do you think it is easy enough to upgrade later?

include/bout/bout_enum_class.hxx
```diff
@@ -0,0 +1,105 @@
...
+    _BOUT_ENUM_CLASS_MAP_ARGS(_ENUM_CLASS_STR, enumname, __VA_ARGS__)      \
+  };                                                                       \
+  auto found = toString_map.find(e);                                       \
+  if (found == toString_map.end()) {                                       \
+    throw BoutException("Did not find enum %d", static_cast<int>(e));      \
+  }                                                                        \
+  return found->second;                                                    \
+}                                                                          \
+                                                                           \
+inline enumname _MAKE_FROMSTRING_NAME(enumname)(std::string s) {           \
```
ZedThree: `std::string s` -> `const std::string& s`

ZedThree:
Adding the fmt stuff can be done later once I work out how to do it.

One thing to change now though is that identifiers beginning with an underscore followed by a capital are reserved, so maybe change those -- I suggest just prefixing everything with `BOUT_`.

ZedThree:
Also, might need to use this trick to get `_BOUT_ENUM_CLASS_MAP_ARGS` working:
https://github.com/boutproject/BOUT-dev/pull/1874/commits/cfb7888b495aec5031ed3bbe80436c007191f782

But you'll need a Windows system to test it on, so that can wait

PR #1842, opened 2019-11-15
Standardise factories

ZedThree:
Sorry, this is a horrible PR to drop on a Friday afternoon! On the plus side, it's net negative lines!

Standardises almost all of the various type factories, with the exception of the boundaries and DataFormat, both of which require a fair bit of special handling, not to mention I'm hoping we'll get to rewrite them completely in the near future.

All type factories now return `std::unique_ptr<BaseType>`, with the exception of `Mesh::create`, because making `bout::globals::mesh` be a `unique_ptr` would break more things than I'm willing to fix just yet.

Standardising all the factories like this means it's now possible to register new _anything_ at runtime. This should allow users to experiment with their own `Solver`, `Laplacian` or even `Mesh`, without needing to recompile the framework.

I've provided `bin/bout-v5-factory-upgrader.py` which takes a list of files and provides fixes to use the new APIs. It spits out diffs and gives the option to fix each file automatically. It's possible to also dump the output to a patch file, or just shut up and fix everything.

Also adds command line options to just print out a list of available options for various factories. Useful to see what `Solver`s or `Laplacian`s are available. Less usefully, you can also see what `Mesh` is available (it's `"bout"`).


ZedThree:
I was also tempted to move all the factory classes into the corresponding base type and call them `Factory`. Then for a given type, you could always use `Type::Factory`. Might be much of a muchness though.

examples/2Dturbulence_multigrid/esel.cxx
```diff
@@ -14,7 +14,7 @@ class ESEL : public PhysicsModel {
   Field2D sigma_n, sigma_T, sigma_vort; // dissipation terms
   BoutReal zeta;                        // rho/R0
   BRACKET_METHOD bm;                    // Bracket method for advection terms
-  Laplacian* phiSolver;                 // Laplacian solver for vort -> phi
+  std::unique_ptr<Laplacian> phiSolver{nullptr};                 // Laplacian solver for vort -> phi
```
bendudson: Is there any value in trying to define a type here? e.g define `Laplacian::pointer` to be `std::unique_ptr<Laplacian>`?
It hides the fact that it's a unique pointer, which is probably bad, but allows us to change the pointer type later if needed.

examples/em-drift/2fluid.cxx
```diff
@@ -47,10 +47,10 @@ class EMdrift : public PhysicsModel {
   FieldGroup comms;
 
   // Inverts a Laplacian to get potential
-  Laplacian *phiSolver;
+  std::unique_ptr<Laplacian> phiSolver;
```
bendudson: Do these default initialise to `nullptr` like the others?

src/invert/laplacexz/laplacexz.cxx
```diff
@@ -1,29 +1,7 @@
...
-  if (strcasecmp(type.c_str(), "cyclic") == 0) {
-    return new LaplaceXZcyclic(m, options, loc);
-  } else if(strcasecmp(type.c_str(), "petsc") == 0) {
-    return new LaplaceXZpetsc(m, options, loc);
-  } else {
-    throw BoutException("Unknown LaplaceXZ solver type '%s'", type.c_str());
-  }
-  return nullptr;
-}
+// DO NOT REMOVE: ensures linker keeps all symbols in this TU
```
bendudson: nice!

examples/2Dturbulence_multigrid/esel.cxx
ZedThree: We have done this before for a couple of types, but I've since been convinced it's probably less helpful than one might think. There's a big difference between getting a `unique_ptr`, a `shared_ptr` and a raw pointer, and we probably shouldn't hide that.

ZedThree:
There is some documentation in the docstring for the `Factory`, including an example. I figure we won't be making new factories very often, but happy to stick a page in the RTD manual too.

examples/em-drift/2fluid.cxx
ZedThree: I believe so. I might have fixed this one manually while iterating on the upgrader tool

PR #1846, opened 2019-11-22
Adding an adaptive, arbitrary order, Adams-Bashforth solver

d7919:
Currently performs similarly to other adaptive explicit solvers such as
rkgeneric. Has the advantage that in certain cases we need the minimal
number of rhs calls per step possible.

Could possibly be modified to add adaptive order as well as adaptive
timestep. This may help improve performance in cases where solver
stability can be limiting.

Code may well need some restructuring but I'd thought I'd create the PR now so we can have the discussion here.

This is a candidate for backporting and as it is essentially just new files (and changes to makefiles) it was recommended that I branch from master, and hence my PR is currently targeting <s>master</s> v4.4.0-alpha. This should be easy to apply to other branches.


d7919:
The Newton-Cotes integration routines and very repetitive so I'd hope we could make the code a bit more generic and just pass in the parameters but my initial attempts at doing this didn't work when I wanted to make everything `constexpr`.

ZedThree:
I've pushed `v4.4.0-alpha`, which this can go into instead of master

d7919:
Changed target to v4.4.0-alpha.

I've also added an initial attempt at adaptive order - it could probably be more efficient but still seems to help for some test cases.

d7919:
Currently the adaptive order has unintended behaviour in that as soon as one internal timesteps flags that the order should be lowered we do not increase the order for the remainder of the output step. On fixing this I saw that total time actually increased, if I limited the maximum order to 2 (essentially disabling the adaptive order code) I got about the same time as with this "broken" version so I think the majority of the performance gain was essentially just coming from limiting the order to 2. With the fixed version I do still find a decent saving for the test case when using a maximum order of 5 and enabling adaptive order (72s vs 45s, the "broken" version is about 36s). Looking into this a bit more it seems that with the fixed adaptive order approach there are more failed internal steps, likely due to the smaller regions of stability as the order increases (and we allow more time at higher order in the fixed version). This explains some of the increase in runtime but not all of it (maybe 10-20% of it). The rest must come from an increase in the number of right hand side calls due to an average smaller dt. We could perhaps be more aggressive in how we pick the timestep but in practice this seems to generally make things worse.

We currently unconditionally increase the order up to the maximum order unless we've just decided to use the lower order method, perhaps we should be more conservative about this and just increase the order when the timestep gets too small with the current order (or similar).

d7919:
Testing on the interchange test case the current AB solver seems to fair much worse than rkgeneric with the cash-karp scheme. With AB the test case fails due to exceeding MXSTEP whilst rkgeneric passes the test with just 7 rhs per output time. One area of difference is in the definition of the error, for rkgeneric it's `err = Average<abs(accurate-approx)/(abs(accurate)+abs(approx)+atol)>`, whilst for AB it's `err = Max<abs(accurate-approx)>`. Changing the AB definition (Max->Avg and the internal calculation), tweaks where the AB case fails but it still consistently fails. Trying AB non-adaptive I can get the test case to complete but even with a timestep 1/100th of that used by rkgeneric the test still doesn't pass. Interestingly the reported test result does not depend on the order used if the timestep is small (1.0), but it does if the step is bigger (100.0, and interestingly the reported error is smaller).

d7919:
Ah, it seems there was a fairly significant bug -- the state vector was only being put into the fields at the end of output steps and inside the adaptive algorithm. This means that for adaptive runs the derivative used at the current time was actually lagging behind by half a time step and for non-adaptive runs it was lagging behind by a whole output step. Fixing this makes the interchange test pass and speeds up the drift test case (down to 28s, c.f. 64s for Cash-Karp) whilst maintaining accuracy. I guess this also highlights that our current tests aren't that sensitive to solvers doing exactly the correct thing.

d7919:
Wishlist for future work:

1. Adaptive order for "fixed" timestep
2. Adaptive order for failed internal timesteps (currently only consider changing order if internal timestep successful).
3. More user control of timestep changes
4. Tidy up the Newton-Cotes integration to reduce duplication.
5. More available diagnostics to track order changes, failed steps etc.
6. Explore adaptive adaptivity -- primary cost is typically in rhs, which we have to call once per internal timestep plus once more for each attempt at the internal timestep as a part of the error check. We may be able to get away with only calculating the error every other step or similar. Alternatively we may be able to find other ways of doing the error calculation (say interpolating the full step state array to the halfway point and comparing there instead).

ZedThree:
Sorry, I had already merged in `next` before I realised this was targeting v4.4

src/solver/impls/adams_bashforth/adams_bashforth.cxx
```diff
@@ -502,27 +533,7 @@ BoutReal AdamsBashforthSolver::take_step(const BoutReal timeIn, const BoutReal d
...
-
-  for (int j = 0; j < order; j++) {
-    const BoutReal factor = coefs[j];
-
-    BOUT_OMP(parallel for);
-    for (int i = 0; i < nlocal; i++) {
-      full_update[i] += history[j][i] * factor;
-    }
-  }
+  Array<BoutReal> full_update = AB_integrate(nlocal, timeIn + dt, times, history, order);
```
d7919, line 574: Does this involve a copy or is it likely to be "in-place"?

ZedThree: It should be in-place with the majority of compilers, and if not, it will be a cheap move instead

d7919:
Changes look good to me, thanks for tidying this.

ZedThree:
This is nice and self-contained and could be back-ported

PR #1847, opened 2019-11-22
Use {fmt} for all string formatting

ZedThree:
Another big PR late on a Friday, sorry!

We now use [fmt](https://fmt.dev) for all our string formatting, instead of the printf-style formatting. This affects calls to `Output`, `BoutException`/`ParseException`, `DataFile`, `OptionsReader`, and `MsgStack`/`TRACE`. Run `bin/bout-v5-format-upgrader.py` on your physics models to update them.

One big benefit of fmt is that `{}` will format any standard type correctly, and we can customise it to work with our types too. This PR uses the more specific `{:s}`, `{:e}` syntax because the upgrader tool preserves and corrects the existing specifier.

This will absolutely certainly cause conflicts with _everything_, but after merging `next` (either into this, or after this is merged, into your branch) running `bin/bout-v5-format-upgrader.py` on the affected files should fix things!
You may also need to run that tool on any files you've changed.

One issue is that `fmt` will accept, say, `%s` in a string, but won't format it, and will happily take more variables than format specifiers in strings. So we might need to be careful there.


configure.ac
```diff
@@ -1182,6 +1182,23 @@ AS_IF([test "$USE_NLS" = "yes"], [
...
+
+dnl If in a git repo, get submodule, unless BOUT_DONT_UPDATE_GIT_SUBMODULE is set
+AC_CHECK_FILE([.git], [
+  AS_IF([test "x$BOUT_DONT_UPDATE_GIT_SUBMODULE" == "x"], [
+    git submodule update --init externalpackages/fmt
+  ])
+])
+
+dnl Copy the one file we need to somewhere else
+AC_CONFIG_LINKS(src/fmt/format.cxx:externalpackages/fmt/src/format.cc)
```
dschwoerer: It would be nice if we would support compiling with the fmt library that is already installed on the system.

configure.ac
```diff
@@ -1182,6 +1182,23 @@ AS_IF([test "$USE_NLS" = "yes"], [
...
+
+dnl This include must come _first_
+EXTRA_INCS="-I\$(BOUT_TOP)/externalpackages/fmt/include $EXTRA_INCS"
```
dschwoerer: Similarly, it would be nice to have a way to opt out, instead of patching `configure.ac`

examples/6field-simple/elm_6f.cxx
```diff
@@ -290,7 +290,7 @@ const Field2D N0tanh(BoutReal n0_height, BoutReal n0_ave, BoutReal n0_width,
   mesh->get(Grid_NX, "nx");
   mesh->get(Jysep, "jyseps1_1");
   Grid_NXlimit = n0_bottom_x * Grid_NX;
-  output.write("Jysep1_1 = %i   Grid number = %e\n", int(Jysep), Grid_NX);
+  output.write("Jysep1_1 = {:d}   Grid number = {:e}\n", int(Jysep), Grid_NX);
```
dschwoerer: Could you split the PR into one with manual changes, and one with automatic changes?

I assume that would significantly simplifying this review.

configure.ac
ZedThree: We do, using CMake. Unfortunately, it was very difficult to do so with the makefile system, so I gave up.

ZedThree:
> So one can use %s?

Sorry, I wasn't clear: `%s` is not a valid format specifier for the new fmt-formatting. The problem is if a user passes a printf-style format string, we can't\* catch or warn about it. Hopefully the python tool included here will fix them all though.

There is a macro, `FMT_STRING` which does compile-time checking of format strings, otherwise it's runtime checking. I'm not sure how this interacts with internationalisation though.

---

\* Actually, we could, as fmt does have `fmt::sprintf`, but it's ugly and expensive. We'd need to format everything at least twice, once with the printf-style and once with the fmt-style, and then see which one we need to use.

examples/6field-simple/elm_6f.cxx
ZedThree: I don't think so, as the PR with the manual changes would be broken.

locale/fr/libbout.po
```diff
@@ -17,94 +17,684 @@ msgstr ""
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n > 1);\n"
 
-#: src/bout++.cxx:365
+#: ../src/mesh/impls/bout/boutmesh.cxx:347
```
bendudson: Thanks for updating the translations. Sorry I think the entries where `msgstr` is empty should be deleted. If they're deleted then I think gettext keeps the original.

bendudson: Ah sorry, my mistake. Some confusion in online fora, but the consensus seems to be that empty strings are counted as untranslated and so ignored.

locale/zh_CN/libbout.po
```diff
@@ -16,75 +16,679 @@ msgstr ""
...
+msgstr ""
+
+#: ../src/sys/options.cxx:216
+msgid "Couldn't get integer from option {:s} = '{:s}'"
+msgstr ""
+
+#: ../src/bout++.cxx:388
+#, fuzzy
+msgid "DataDir \"{:s}\" does not exist or is not accessible\n"
 msgstr "\"%s\" 不存在或不可访问\n"
```
bendudson: Can the .po files be searched-and-replaced to change %s -> {:s} etc. in `msgstr` entries?

bendudson:
Thanks @ZedThree ! I think this is a slightly painful change, but it has some really nice advantages and is a big improvement.

locale/zh_CN/libbout.po
bendudson: Done in PR #1866

ZedThree:
The failing tests in the push build are fixed in next, which is why the PR tests are passing.

PR #1851, opened 2019-11-22
Bugfix: $MPIRUN should include -np

dschwoerer:
Found this issue in the released 4.3.0

Is master the branch that will become 4.3.1?

edit: This matches the python utils behaviour now, where the default is also `mpirun -np`


bendudson:
Yes, I believe `master` will become 4.3.1. Thanks @dschwoerer

PR #1855, opened 2019-11-25
Address needs to be aligned

dschwoerer:
This value gets dereferenced by python, but without knowing it's true
type. Thus a SIGBUS is thrown on some architectures, namely arm. This
can be avoided by copying it to a larger type first, e.g. long, that
is sufficiently aligned, and thus can be safely dereferenced by
python.


bendudson:
Thanks! Presumably this should go into `next` too?

ZedThree:
There's a few bug fixes gone into master since 4.3 -- probably about time for 4.3.1 shortly

PR #1856, opened 2019-11-25
Try to prevent race condition while building.

dschwoerer:
This improves the current situation, as before it could happen that
building failed, because the .a was not yet created. Now it waits for
it to be created. Switching to atomic mv rather then to create it
inplace, ensures that if the file is seen, it is valid.
This does not address the case when the .a is recreated, and the
dependend file is checked before whether the .a has changed.

### DIFF
--- a/make.config.in
+++ b/make.config.in
@@ -242,11 +242,15 @@ endif
 
 ifeq ("$(TARGET)", "sub")
 LIB=$(MODULE_DIR)/$(SUB_NAME).a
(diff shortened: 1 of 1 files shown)


ZedThree:
I think I'm missing something here, what does this fix? There's a long standing issue where changing a source file, then rebuilding with `make -j` only rebuilds that object file and not the library, and requires another `make` call to do so. This doesn't seem to fix that case, so is this fixing something else?

dschwoerer:
Yes. A clean, parallel build could previously fail, because the .a file was created, but did not yet contain a full archive, but make would check only the presence of the file to determine whether it can link, which then failed because the a didn't contain what it should contain. I know this isn't a particular common issue ...

PR #1858, opened 2019-12-02
C wrapper functions (FFI)

bendudson:
This is an experiment in C bindings for BOUT++. The purpose is to inform the choice between:

- Manually writing and maintaining C bindings. These are easy to bind to from many other languages, but the interface itself is hard/impossible to generate automatically.
- C++ bindings can be generated partly automatically for e.g. python, and exist at the moment as `boutcore`, which is functional and useful in its present state. Some languages have interfaces to python which they can use.

The only reason for doing this the C way would be simplicity, so if this is not simple to do and  maintain then there is no point really. The main principles I tried to follow are:
- Keep the API as small as possible e.g. all fields are `Field3D`, and the only collection type is `Options`.
- Avoid global state or spooky action at a distance through references. The only exception is `Options`, because putting values into a collection is needed. The simulation state is therefore passed into the user function, rather than modifying user variables:
```
int rhs(Options* ddt, Options* state, BoutReal time)
```
- Make the interface as consistent as possible: (almost) every function returns an error code, and the return value is first in the argument list. The only exception is some `Options` functions, because getting root or a subsection/value with `[string]` always succeeds (except in cases of unrecoverable memory error, meteor strike etc.)
- Make the interface close(r) to how we would like the C++ interface to eventually be. The hope is that over time this helps make the glue code smaller rather than bigger...

Currently a simple physics model written in C solves df/dt = -f .

A side effect of all this object-oriented code in C, is that the result looks suspiciously like PETSc...


bendudson:
One thought, from here: https://stackoverflow.com/questions/54695030/exposing-a-c-library-via-c-bindings
If safety is a concern, and a little extra overhead and additional glue code acceptable, then handles could be passed to C rather than pointers. The objects would then be stored in a `map`, and C functions would operate on these handles. This can then eliminate possibilities for segfaults, which are no fun at all when developing an interface from e.g. an interactive session. 

For example, `Field3D_scale` is currently:
```
extern "C" int Field3D_scale(Field3D* x, BoutReal alpha) {
  CWRAP_START();
  (*x) = (*x) * alpha;
  CWRAP_END();
}
```
but could be
```
extern "C" int Field3D_scale(FieldHandle x, BoutReal alpha) {
  CWRAP_START();
  Field3D& x_field = field_store.at(x);
  x_field = x_field * alpha;
  CWRAP_END();
}
```
where `FieldHandle` could be an `int`, and `field_store` a `map<FieldHandle, Field3D>`.

PR #1860, opened 2019-12-04
Mesh requires Option to stay around.

dschwoerer:
Python will, without this patch, in some cases delete the option
that was used. Thus we need to save a copy, so that while the Mesh
stays around, so does the Option that was used to create it, and
thus option is not GC. Otherwise segfaults or worse happens.

### DIFF
--- a/tools/pylib/_boutcore_build/boutcore.pyx.in
+++ b/tools/pylib/_boutcore_build/boutcore.pyx.in
@@ -682,6 +682,7 @@ cdef class Mesh:
     cdef double isNormalised
     cdef FieldFactory factory
     cdef Coordinates _coords
+    cdef options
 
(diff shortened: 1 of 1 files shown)


dschwoerer:
```
#!/usr/bin/env python3                                                                                                                  
import boutcore
boutcore.init("")
opt = boutcore.Options.fromFile("data/BOUT.settings")
mesh = boutcore.Mesh(options=opt, section="mesh")
opt=None
f = boutcore.create3D("sin(z)", mesh)
sim = boutcore.DDZ(f)
sim.setLocation('YLOW')
```
This can reproduce the error. Output for me currently is that MPI complains about being initialised twice, but I have seen other errors caused by this in the past.

PR #1861, opened 2019-12-05
Make end inclusive for 3-element list index argument to collect

johnomotani:
When lists are passed to the `tind/xind/yind/zind` arguments of collect, for a 2-element list `[start, end]` the `end` index is interpreted as inclusive, as described in the docstring. 3-element lists `[start, end, step]` were not documented explicitly before, but should behave similarly. Previously 3-element lists treated `end` as exclusive.

### DIFF
--- a/tools/pylib/boutdata/collect.py
+++ b/tools/pylib/boutdata/collect.py
@@ -117,7 +117,7 @@ def _convert_to_nice_slice(r, N, name="range"):
         temp_slice = slice(r2[0], r2[1] + 1)
     elif len(r) == 3:
         # Convert 3 element list to slice object
-        temp_slice = slice(r[0],r[1],r[2])
+        temp_slice = slice(r[0], r[1] + 1, r[2])
     else:
         raise ValueError("Couldn't convert {} ('{}') to slice".format(name, r))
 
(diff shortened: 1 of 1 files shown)


dschwoerer:
Note that this change breaks current scripts, thus I oppose to it.

Further, the end should always be exclusive, and thus I would prefer fixing the two-element version instead. I didn't do so. because I didn't want to break other peoples scripts, but if we are anyway breaking things, we could break them towards the standard, and not the non-standard (albeit documented) behaviour.

I thus reject that this is a bugfix, and think it makes thinks worse. Before I could add a 1 to get the "correct" behaviour.

bendudson:
On balance I think I agree with @dschwoerer on this one. Perhaps a way forward would be to:
- Add new keywords (`tindex`, `xindex` etc?) with the standard behaviour
- In 4.x print a deprecation warning if `tind`, `xind` are used
- In 5.0 remove `tind`, `xind` and keep just the standard behaviour
I think this would avoid breaking existing scripts in non-obvious ways, while ending up where we want to be in terms of the standard behaviour.

johnomotani:
At the moment you can get 'standard' behaviour by passing a slice directly, e.g. `xind = slice(start, stop, step)` - if you pass a `slice`, then `collect` does not mess around with it, just uses it directly. So one possibility would be to deprecate use of `list` or other non-`slice` sequences for `tind,xind,yind,zind`. That would also (eventually) simplify the code in `collect.py` a fair bit.

However, given the documented behaviour at the moment, I do think the 3-element list form is wrong: it is awful that using `xind = [1, 42]` and `xind = [1, 42, 1]` will give different outputs. Since the 3-element form is (relatively) recently introduced (b735923dc2952e66428171488ec644d4312a507a), and undocumented, I think that's the one that has to change.

johnomotani:
We could just remove the 3-element list form, as it is undocumented and the functionality is available by passing a `slice`?

ZedThree:
Was this replaced by #1862?

johnomotani:
Replaced by #1862

PR #1862, opened 2019-12-05
Remove 3-element list indexers for collect()

johnomotani:
Previously there was an undocumented option to pass a 3-element list `[start, end, step]` to the `tind,xind,yind,zind` arguments of `collect()`. The 3-element list used an exclusive `end`, while the 2-element form uses an inclusive `end`, which is inconsistent and potentially confusing. `step` functionality is available by passing a Python `slice` to collect. Removing the 3-element list form is a safe solution: no existing scripts will silently change their behaviour, and the provided functionality is consistent.

Replaces #1861.

### DIFF
--- a/tools/pylib/boutdata/collect.py
+++ b/tools/pylib/boutdata/collect.py
@@ -116,7 +116,9 @@ def _convert_to_nice_slice(r, N, name="range"):
         # Lists uses inclusive end, we need exclusive end
         temp_slice = slice(r2[0], r2[1] + 1)
     else:
-        raise ValueError("Couldn't convert {} ('{}') to slice".format(name, r))
+        raise ValueError("Couldn't convert {} ('{}') to slice. Please pass a "
+                         "slice(start, stop, step) if you need to set a step."
+                         .format(name, r))
 
     # slice.indices converts None to actual values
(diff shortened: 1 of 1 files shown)


bendudson:
Thanks John! I think this is probably a good way forward. In future passing in a slice seems to be the easiest and most consistent way, so in 5.0 perhaps we also remove the 2-element list option?

johnomotani:
What branch should this merge into? I've got it set to `master` at the moment, because I was originally thinking of #1861 as a 'bugfix', but maybe it should go to `next` and `v4.4.0-alpha` instead?

ZedThree:
No strong preference, but probably 4.4

johnomotani:
Should the change go into `next` as well? If I understand right, we won't be merging changes from 4.4 to `next`. Is the plan to remove `tools/pylib` in favour of the separate `boututils` and `boutdata` repos so the change is not needed in `next`?

ZedThree:
My preferred plan is that we completely replace `boututils/boutdata` with `xBOUT` in `next` once it's sufficient for the library needs.

From a maintenance point of view, an immediate switch-over seems like the easiest thing, but it's not clear to me how clean that will be for users. Something to discuss next week!

Issue #1864, opened 2019-12-05
TimerTest.GetTime fails sometimes


dschwoerer:
On Fedora 31, this failed now 2 out of 3 times on a kvm host (AFAIK):
```
[ RUN      ] TimerTest.GetTime
sys/test_timer.cxx:26: Failure
The difference between timer.getTime() and elapsed.count() is 0.012817351000000001, which exceeds bout::testing::TimerTolerance, where
timer.getTime() evaluates to 0.0010988409999999999,
elapsed.count() evaluates to 0.013916192000000001, and
bout::testing::TimerTolerance evaluates to 0.00050000000000000001.
[  FAILED  ] TimerTest.GetTime (15 ms)
[----------] 10 tests from TimerTest (30 ms total)
```
and
```
[ RUN      ] TimerTest.GetTime
sys/test_timer.cxx:26: Failure
The difference between timer.getTime() and elapsed.count() is 0.0073374699999999996, which exceeds bout::testing::TimerTolerance, where
timer.getTime() evaluates to 0.0011076980000000001,
elapsed.count() evaluates to 0.0084451679999999994, and
bout::testing::TimerTolerance evaluates to 0.00050000000000000001.
[  FAILED  ] TimerTest.GetTime (8 ms)
[----------] 10 tests from TimerTest (25 ms total)
```
Is this expected that it fails sometimes?
Could we decrease the accuracy, or maybe switch to a relative error, so that if we run on a busy system we are less likely to encounter this?
Or is there something wrong?
NB: it failed once with mpich and once with openmpi.

dschwoerer:
The build succeeded now. Should I close this? It failed 2 out of 5 times ...

ZedThree:
This is unfortunate, but an "acceptable" failure of sorts. It's difficult to make this test robust because the the `sleep` functions have some jitter we can't control due to the OS, scheduler, etc.

One suggestion I've seen is to change the `sleep_for(1ms)` to `sleep_until(system_clock::now() + 1ms)`. I don't completely understand why this might be better, but might be worth a try?

dschwoerer:
Thanks for looking into it.
I hope servers aren't that busy in the future, will close this as won't fix

opoplawski:
I'm running into this fairly consistently now with Fedora 35/36 armv7hl builds.  E.g.: https://kojipkgs.fedoraproject.org//work/tasks/1459/73681459/build.log

ZedThree:
[...]

PR #1865, opened 2019-12-06
Evaluate all timings before call to macro.

dschwoerer:
On highly loaded systems, this test failed several times, as the differences where to high. By moving the evaluation of timer earlier within the function, we reduce the risk of interrupts or cache misses happening, that could introduce larger differences then the tolerance.

With this change the differnce between the two calls is around .1 µs, while the call to macro took over 13 µs - so hopefully this ensures that even on a busy vm this stays below the 500 µs we have as tolerance ...

Maybe this helps with #1864 - 6 out of 6 tries have worked so far ...

### DIFF
--- a/tests/unit/sys/test_timer.cxx
+++ b/tests/unit/sys/test_timer.cxx
@@ -14,12 +14,11 @@ constexpr auto sleep_length = ms(1.);
 } // namespace bout
 
 TEST(TimerTest, GetTime) {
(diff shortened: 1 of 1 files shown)


dschwoerer:
I am confused - I am not even using gettime ...
Maybe this was just a few random hickups? Or the swap of construction and call to `now()` did help?

PR #1867, opened 2019-12-06
Add BOUT_REVISION (git hash of compiled version) to dump files

johnomotani:
Helps provenance tracking of simulations.


ZedThree:
I'm happy for this to go into master. It's small and obvious enough.

johnomotani:
Now includes a small bugfix for `Datafile::setAttribute()`.

ZedThree:
I'm happy for this to go in now, but a suggestion for a future tweak is to replace the `#ifdef`s with a single `constexpr` value `bout_revision` which is either the revision if `REVISION` is `#defined`, or some sensible default value otherwise.

PR #1869, opened 2019-12-11
Basic lock for mpark.variant

dschwoerer:
Ensures that, if several instances try to download mpark.variant,
only one can succeed with the atomic mkdir, and thus only one
instance tries to download. The others are waiting for the lock
directory to vanish, and then resume with building.

### DIFF
--- a/make.config.in
+++ b/make.config.in
@@ -223,7 +223,10 @@ uninstall:
 MPARK_VARIANT_SENTINEL = $(MPARK_VARIANT_INCLUDE_PATH)/mpark/variant.hpp
 $(MPARK_VARIANT_SENTINEL):
 	@echo "Downloading mpark.variant"
-	git submodule update --init --recursive $(BOUT_TOP)/externalpackages/mpark.variant
+	@lock=$(BOUT_TOP)/externalpackages/.get.mpark ;ex=0; mkdir $$lock && ( \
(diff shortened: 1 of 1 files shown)


dschwoerer:
Can we get this merged, so a parallel build after download and configure succeeds?

ZedThree:
Sorry, need to get round to checking it works on other systems, will try and get to this afternoon. I suspect a cleaner version might be to just put an order-only dependency on the `%.o: %.cxx` recipes on the mpark sentinel

dschwoerer:
No worries, I have just checked also ubuntu 18.04, where it works as expected. Is there any other OS I should check?

I have been running
```
podman run --rm -it ubuntu /bin/bash
apt-get update
apt-get install git mpich libmpich-dev libfftw3-dev libnetcdf-dev libnetcdf-cxx-legacy-dev git g++ make
git clone https://github.com/boutproject/BOUT-dev.git
cd BOUT-dev/
git checkout mpark-lock
./configure && make -j 4
```

and for archlinux:
```
podman run --rm -it archlinux /bin/bash
pacman -Sy openmpi fftw netcdf-cxx git gcc make
git clone https://github.com/boutproject/BOUT-dev.git
cd BOUT-dev/
git checkout mpark-lock
./configure && make -j 4
```

PR #1871, opened 2019-12-11
Do not include if is released

dschwoerer:
Should be released, rather than release.

### DIFF
--- a/make.config.in
+++ b/make.config.in
@@ -84,6 +84,9 @@ BOUT_VERSION  = @PACKAGE_VERSION@
 
 OBJ    = $(SOURCEC:%.cxx=%.o)
 ifndef RELEASED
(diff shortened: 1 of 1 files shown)


ZedThree:
Where is `RELEASED` set? Is this something we need to be setting somewhere when creating tarballs?

dschwoerer:
[I currently set RELEASED after installing](https://src.fedoraproject.org/rpms/bout++/blob/master/f/bout++.spec#_376). Should I add this to the installer script?

ZedThree:
I'm not 100% sure I see the need for this conditional: if it's installed, when would a library rebuild get triggered anyway?

If we are going to have it, best that this variable is set in this repo rather than elsewhere. So yes, please could you add something to `/bin/bout-archive-helper.sh`? Or otherwise remove conditional and fix in your rpm script.

PR #1876, opened 2019-12-31
If variable is added to Datafile twice, do not write twice

johnomotani:
Writing a time-evolving variable twice can corrupt data files, so if the same variable is added twice with the same name, just skip adding the second time.

This seems to have been partially implemented already, as a variable added twice was not added to the netCDF file twice, but was added to the `BoutReal_arr`, etc., twice so was written twice, which I think caused an array overrun and corrupted the output files.


johnomotani:
Actually, it may not have been corrupting the files, just extending the `'t'` dimension unexpectedly. So all the variables that were added only once have lots of entries that were never written to, which made the output look very strange.

PR #1877, opened 2020-01-02
Implement toFieldAligned and fromFieldAligned for Vector3D

johnomotani:
We found a case where we needed `toFieldAligned()` and `fromFieldAligned()` for a `Vector3D`, so implemented them.

### DIFF
--- a/src/field/vector3d.cxx
+++ b/src/field/vector3d.cxx
@@ -603,6 +603,34 @@ const Field3D abs(const Vector3D &v, const std::string& region) {
   return sqrt(v*v, region);
 }
 
+const Vector3D toFieldAligned(const Vector3D& v, const std::string& region) {
+  Vector3D result;
+  result.setLocation(v.getLocation());
+  if (not v.covariant) {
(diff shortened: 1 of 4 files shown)


ZedThree:
Should we also add it for `Vector2D`?

Also, you can drop the `const` from the return type. We currently have it in lots of places, but we should probably remove it, as I don't think it's ever helpful. Looks like `clang-tidy` can even do so for us.

src/field/vector3d.cxx
```diff
@@ -603,6 +603,34 @@ const Field3D abs(const Vector3D &v, const std::string& region) {
   return sqrt(v*v, region);
 }
 
+const Vector3D toFieldAligned(const Vector3D& v, const std::string& region) {
+  Vector3D result;
+  result.setLocation(v.getLocation());
+  if (not v.covariant) {
+    result.toContravariant();
```
bendudson: We probably should have an `emptyFrom` and `zeroFrom` for vectors, like for fields.

PR #1882, opened 2020-01-09
Generate report for Timers

ZedThree:
Adds two methods to `Timer`:

- `getTotalTime` to see the total time elapsed for a timer (i.e. doesn't get reset with `resetTime`)
- `listAllInfo` generates a table reporting all the timers, like:

```
Timer name | Current time (s) | Total time (s) | Hits
-----------|------------------|----------------|-----
comms      | 0                | 0.028078329    | 546 
evolve_rho | 0.499829339      | 0.499829339    | 62  
invert     | 0                | 0.539184925    | 62  
io         | 0.010658816      | 0.090785536    | 57  
rhs        | 0                | 4.15958722     | 62  
run        | 0.014304425      | 4.32613615     | 1   
```

Adds an option `write_final_timings` which prints the above table at the end of a simulation.

I'm not so sure the "current time" column is particularly useful, probably a "mean time per hit" would be better. Also I'm not overly keen on the name `write_final_timings` at global scope, so suggestions welcome!


src/sys/timer.cxx
```diff
@@ -46,13 +53,61 @@ double Timer::getTime(const Timer::timer_info& info) {
...
+  return seconds{current_duration}.count();
+}
+
+void Timer::listAllInfo() {
+  const std::string headerName = "Timer name";
+  const std::string headerTime = "Current time (s)";
+  const std::string headerTotal = "Total time (s)";
+  const std::string headerHits = "Hits";
+
+  const auto longest_width = [](const std::string& header, const auto& op) {
```
d7919: A comment may help here -- for me it's not immediately obvious what each stage of this is doing, although I can guess.

ZedThree: Added some comments to explain what's going on

ZedThree:
I've changed the columns to drop the current time and add mean time per hit:

```
Timer name | Total time (s) | Hits | Mean time/hit (s)
-----------|----------------|------|------------------
comms      | 0.019139448    | 546  | 3.50539341e-05   
evolve_rho | 0.357001864    | 62   | 0.00575809458    
invert     | 0.376253795    | 62   | 0.0060686096     
io         | 0.055955419    | 57   | 0.000981674018   
rhs        | 2.97283263     | 62   | 0.0479489133     
run        | 3.07722771     | 1    | 3.07722771   
```

Maybe the switch should be `show_timing_report`?

bendudson:
Would sorting by one of the columns be simple to do? e.g. sort by total time might be the most useful most of the time

ZedThree:
I've changed the option to `time_report:show`, sticking it in its own section for a couple of reasons: I'd prefer not to add stuff to the global section if possible, and to not preclude future customisation options.

Also added a column with fraction of the largest time. Now looks like:

```
Timer name | Total time (s) | % of top | Hits | Mean time/hit (s)
-----------|----------------|----------|------|------------------
run        | 4.37208176     | 100.00%  | 1    | 4.37208176       
rhs        | 4.18987402     | 95.83%   | 62   | 0.0675786133     
invert     | 0.543508674    | 12.43%   | 62   | 0.00876626894    
evolve_rho | 0.504323351    | 11.54%   | 62   | 0.0081342476     
io         | 0.101402551    | 2.32%    | 57   | 0.00177899212    
comms      | 0.029098885    | 0.67%    | 546  | 5.32946612e-05   
```

PR #1885, opened 2020-01-10
Set corner cells consistently

johnomotani:
To properly support non-orthogonal grids (where the x- and y- directions are not orthogonal) we should properly fill the corner cells which are needed by mixed-derivative stencils (i.e. `D2DXDY`).

This PR addresses this by adding communication of corner guard cells, and updating boundary conditions to apply something to the corner boundary cells.

The communications are defined to make the corner cells consistent with what would happen if there was no processor boundary (in particular y-processor boundaries are required at X-points, but the separatrices can be anywhere in the x-domain, either on boundaries or not).

The corner boundary points are set by applying a y-boundary condition using x-boundary points as input. This is a somewhat arbitrary choice. It requires applying x-boundary conditions before y-boundary conditions (which happened to be the existing behaviour anyway). User-code boundary conditions (e.g. sheath boundary conditions) need to take care with the order. The y-boundary iterators have been updated to include the corner boundary points so user-implemented boundary conditions should not need any special handling of the corners, just care that x-boundary conditions have been applied before they are called.

There is an option `mesh:include_corner_cells` which can be set to `false` to revert to to the original behaviour, which is correct for orthogonal grids and might be slightly more optimized for that case.

~~WIP because tests of the new communications are still needed.~~

Possible backport candidate (maybe with the default of `include_corner_cells` set to `false`)? If we want to back-port the 3d Laplace solvers to v4.4 then this will be needed.


ZedThree:
Related and might be interesting, while researching multiblock libraries this week, I came across one library which extrapolated into the corners. Unfortunately, I can't remember which library! I'll try to dig it out.
Extrapolating is essentially equivalent to changing the stencil weights, I think, so might end up reducing the overall order.

src/mesh/impls/bout/boutmesh.cxx
```diff
@@ -1806,12 +2013,33 @@ void BoutMesh::set_connection(int ypos1, int ypos2, int xge, int xlt, bool ts) {
...
+      if (UDATA_XSPLIT <= 0) {
         UDATA_INDEST = -1;
+      }
 
       TS_up_out = ts;
       output_info.write("=> This processor sending out up\n");
     }
+
+    if (include_corner_cells) {
+      // Connect corners as if communicating all guard cells first in y, then in x
```
dschwoerer: I fail to see why you not do that?
That would keep the number of communications the same, only introduce slightly larger messages.

What you would do is first communicate in y as we do now. Then you communicate in x, but including the boundary cells. This way you need the name number of communications, which would mean less impact on the runtime, as the package size change only increases slightly.

johnomotani: The trouble is you'd have to wait for the y-communication to complete before you can do the x-communication; at the moment, and in the changes proposed here, all the communications can be sent at the same time, and you only have to call `wait` once. So while it's true that the amount of information passed would not change much, I'm under the impression that communications are limited by latency rather than bandwidth, so we don't want to wait for two sets of communication to complete rather than one.

dschwoerer:
I guess we should do some proper benchmarking ...

A simple way to do this, without any code changes should be switching
between async and sync send, where we already have a switch for.
If we are limited by latencies, the async version should make the
communication roughly 4x as fast, but as far as I remember this is not
the case, but I might be wrong.

It's been a while, but I seem to remember that the async switch didn't
have a significant impact.

src/mesh/impls/bout/boutmesh.cxx
dschwoerer: I guess we should do some proper benchmarking ...

A simple way to do this, without any code changes should be switching
between async and sync send, where we already have a switch for.
If we are limited by latencies, the async version should make the
communication roughly 4x as fast, but as far as I remember this is not
the case, but I might be wrong.

It's been a while, but I seem to remember that the async switch didn't
have a significant impact.

johnomotani: The switch is `async_send`, which using `MPI_Isend` instead of `MPI_Send`, but because MPI can copy the buffer and return from `MPI_Send` without needing to actually complete the communication, this may not make much difference https://stackoverflow.com/questions/10017301/mpi-blocking-vs-non-blocking. I just did a quick test and communicating over 4 nodes of Marconi, `async_send` made no noticable difference. However, the receives in BOUT++ are always asynchronous - I think that means that regardless of the `async_send` setting, the communications are overlapping. It would need more hacking around to force the x- and y-communications to really be consecutive to test this further - would be interesting to prove there's a difference, but I'm not going to pursue now.

johnomotani:
I've now added an integrated test, so this PR is ready for review.

johnomotani:
@dschwoerer did some testing (thank you! :tada:) and found that communicating in y, waiting, and communicating in x is as fast as sending in y and x concurrently and waiting for both after. Either way seems to be (slightly) faster than communicating the corners separately. I've checked this too on Marconi, and it seems to be true even when both x- and y-communications require communication between nodes, not just within a node. Apparently our communications are limited by bandwidth and not latency, which was a surprise to me... It is indeed much simpler to just do the y- and x-communications consecutively, so I'm re-writing to implement that way.

cmacmackin:
@johnomotani Your changes have not added the corner cells to the boundary regions created in `BoutMesh::addBoundaryRegions()` (e.g., `RGN_LOWER_Y`, `RGN_UPPER_Y`, `RGN_INNER_X`, `RGN_OUTER_X`). I guess it is not immediately obvious which of these regions they should be placed into. However, the way the PETSc interface is currently implemented (as of #1889 being accepted) requires corner cells to be included in a boundary region if they are to be included in the linear system to be solved.

johnomotani:
Thanks @cmacmackin, they should be added there too. The logic should be pretty much the same, I will add.

Comment for an eventual boundary conditions re-design - there are currently at least 3 places that are boundary-region-like: `BoundaryRegion` which is used by the built-in boundary conditions; `Mesh::iterateBndryLowerY()` and `Mesh::iterateBndryUpperY()` which are typically used by user-defined boundary conditions; and the regions created in `BoutMesh::addBoundaryRegions()`. It would be nice to unify them somehow...

johnomotani:
I've refactored now to implement corners by doing communications first in the y-direction then in the x-direction. Across 4 nodes on Marconi, communication seems to take about the same time as on current `next`, which is better than it was when communicating the corners separately, and the code is much simpler too.

I've also added the boundary regions @cmacmackin mentioned.

ZedThree:
> Apparently our communications are limited by bandwidth and not latency, which was a surprise to me

@johnomotani This is an interesting result! Any chance the data and generating code could be shared? Maybe stick it in `examples/performance`?

At some point, I will get around to making a `BOUT++ performance` repo, with all our test cases and harnesses, data, etc.

johnomotani:
@ZedThree I added the code in `examples/performance/communications` already. I was testing with the `data/BOUT.inp.smallblocks` case which for 4x48proc nodes on Marconi is only 2x2 grids per processor, and has the x-domain split across 2 nodes. I'll double check and upload some example log files.

johnomotani:
Here are some logs for simulations with 2x2 blocks on each processor:
* 12.7s from `next` [next.log](https://github.com/boutproject/BOUT-dev/files/4098806/next.log)
* 15.0s this branch [consistent-corner-cells.log](https://github.com/boutproject/BOUT-dev/files/4098808/consistent-corner-cells.log)
* 15.2s this branch with `include_corner_cells=false`, so communication patterns are the same as `next` and x- and y-communications are sent at the same time [consistent-corner-cells_turned-off.log](https://github.com/boutproject/BOUT-dev/files/4098809/consistent-corner-cells_turned-off.log)
* 18.5s from before refactoring, when this branch was communicating corner cells separately and sending them at the same time as the x- and y-buffers (58e7659) [communicate-corners.log](https://github.com/boutproject/BOUT-dev/files/4098810/communicate-corners.log)

Here are some with a larger grid (`192x128x128`, with `nxpe=96`, `nype=2`) (note that the root processor claims to spend less time in communications, but almost all the rest spend about 60% in communications for all cases below - I've uploaded all the logs in the tarballs):
* 28.3s `next` [next.log](https://github.com/boutproject/BOUT-dev/files/4098846/next.log) [nextlogs.tar.gz](https://github.com/boutproject/BOUT-dev/files/4098847/nextlogs.tar.gz)
* 29.0s this branch [consistent-corner-cells.log](https://github.com/boutproject/BOUT-dev/files/4098851/consistent-corner-cells.log) [consistent-corner-cellslog.tar.gz](https://github.com/boutproject/BOUT-dev/files/4098852/consistent-corner-cellslog.tar.gz)
* 29.1s before refactoring, communicating corners separately [communicate-corners.log](https://github.com/boutproject/BOUT-dev/files/4098857/communicate-corners.log) [communicate-cornerslog.tar.gz](https://github.com/boutproject/BOUT-dev/files/4098858/communicate-cornerslog.tar.gz)

cmacmackin:
@johnomotani Thanks, I can confirm that the regions are now behaving correctly. I have already pulled this work into #1803, where it seems to be working.

dschwoerer:
@johnomotani The test you send me behave strange, and behave better if you set the time derivative `ddt(f)=0` explicitly. Not sure why it didn't crash without setting the derivative, but without setting the derivative, some procs spend 9 times in rhs/solver compared to otherwise. This might have caused issues for timing ...

I haven't got the new results of Marconi before it went down ... so will post again tomorrow.

In your case there are several ones that spend 2 times as much in calc and also more time in solving. As such the timings are not that reliably, as all the other procs need to wait on the slow ones in comm.

johnomotani:
> In your case there are several ones that spend 2 times as much in calc and also more time in solving. As such the timings are not that reliably, as all the other procs need to wait on the slow ones in comm.

@dschwoerer has a good point. I've updated the performance test to have an `MPI_Barrier()` just before the call to `mesh->communicate()` and to time only that call - printed on the `just comms` line in the output.

With that update, I get (using a 192x128x128 grid on 4 nodes of Marconi with nxpe=48 so both x- and y-directions are split onto 2 nodes):
* `7.46 s` existing standard implementation (without consistent corner cells)
* `15.49 s` using consecutive x- then y-communications
* `13.43 s` using overlapping x-, y- and corner-communications (version from 5ae549f, before `Revert`s)

I still don't understand why the version with corner communications but everything overlapping is so much slower, but it seems to be faster than the x-then-y version (although not massively) - numbers above are from repeating each run twice and the timings of the communications seem fairly consistent. I propose to reset this PR to the previous, overlapping-communications version (and cherry-pick the last 4 commits which are still needed/useful). Any objections? As an argument against, the non-overlapping implementation is much simpler, and we don't usually spend much time in `mesh->communicate()` anyway.

src/mesh/coordinates.cxx
```diff
@@ -119,13 +119,15 @@ Field2D interpolateAndExtrapolate(const Field2D& f, CELL_LOC location,
...
   }
 
-  // Set corner guard cells
-  for (int i = 0; i < localmesh->xstart; i++) {
-    for (int j = 0; j < localmesh->ystart; j++) {
-      result(i, j) = BoutNaN;
-      result(i, localmesh->LocalNy - 1 - j) = BoutNaN;
-      result(localmesh->LocalNx - 1 - i, j) = BoutNaN;
-      result(localmesh->LocalNx - 1 - i, localmesh->LocalNy - 1 - j) = BoutNaN;
+  if (not localmesh->include_corner_cells) {
```
dschwoerer: What is the point?
If they are not included, they shouldn't be used.
If you think they are invalid, then you should set them to NaN in all cases.
These seems to me like this is causing unnecessary breakage.
Especially if this is generated from a expression, like `1.0`, there is absolutely nothing wrong with the corner cells, and setting them to NaN has caused issues for me several times in the past.

dschwoerer: My issue is that in the [coords3d](https://github.com/boutproject/BOUT-dev/tree/coord3d_merged_master) branch I removed the setting to NaN, so that dz is constant if it is generated from a constant expression. Another option would be to extrapolate them if we extrapolate in x and y, however that would be more work ...

ZedThree: Could wrap setting them in `#if CHECK > 0` (level to be debated)?

What do you do if dz is not generated from a constant expression? And do you need dz in the corners?

At some we're also going to add guards and boundaries for z, so this may get even more fun.

dschwoerer: Wrapping in `CHECK` is probably useful anyway, `> 0` is probably fine because it is only used in the init. However, that does not solve this for coords3d.

I check whether `dz` (or `zlength()`) is constant. Currently the check is for `RGN_ALL` as then no further checks are needed, what the actual used domain is, and the first element is used.

In the case of a non-constant `dz`, any of the routines that assume a constant `dz`/`zlength()` cannot be used, so `dz` doesn't need to be constant, and corner cells are no more important then in any other case.

It is probably best to check in all cases what domain is used, and only check constantness for this domain, rather then `RGN_ALL`, but `dz(0,0)` is so much easier then `dz[dz.getRegion(...).begin()[0]]`

To do things right, I think we should:
 * extrapolate into corners, if they are to be used, otherwise leave them. The motivation to not set them to NaN is that they might be generated from a grid or input, which already includes guards and corners. We also do not break boundaries, if they are not extrapolated.
 * use appropriate regions in coord3d

ZedThree: > It is probably best to check in all cases what domain is used, and only check constantness for this domain, rather then RGN_ALL, but dz(0,0) is so much easier then dz[dz.getRegion(...).begin()[0]]

Might be wise to wrap this up in something like `constant_dz(const string& region)` that throws if it's not constant over that region.

> extrapolate into corners, if they are to be used, otherwise leave them

Is that not what is done in this change? This flag doesn't say whether or not to use the corners, just whether or not they are communicated.

We could of course mandate that grids for v5 provide values in the boundaries, which removes all this extrapolation stuff. That might just be moving the problem up a level.

dschwoerer: > > It is probably best to check in all cases what domain is used, and only check constantness for this domain, rather then RGN_ALL, but dz(0,0) is so much easier then dz[dz.getRegion(...).begin()[0]]
> 
> Might be wise to wrap this up in something like `constant_dz(const string& region)` that throws if it's not constant over that region.

:+1: will do

> > extrapolate into corners, if they are to be used, otherwise leave them
> 
> Is that not what is done in this change? This flag doesn't say whether or not to use the corners, just whether or not they are communicated.

I do not see any corner extrapolation. Either they are set to NaN (if `! include_corner_cells`) and otherwise left as they are. With the methods from the non-uniform boundary stuff, it shouldn't be that hard to extend this into 2D/3D to allow extrapolation into corners.

> We could of course mandate that grids for v5 provide values in the boundaries, which removes all this extrapolation stuff. That might just be moving the problem up a level.

Certainly would be easier for the initialization. Also the extrapolation would only need to be done once, at grid creation time. Writing it as script might be easier, because we don't have to think at all possible cases at once.

Also with the extrapolation, I feel uncertain on whether we should include the grid spacing. For dz, dy, dz that is non-trivial, but might be more accurate for the other ones ...

johnomotani: Sorry, all the corner-cell handling is rather awful. I probably need more comments/documentation...

If `include_corner_cells == true`, we should be extrapolating into the corners. This PR modifies the boundary regions so that if `include_corner_cells == true` then the y-boundaries include the corner cells. As the y-boundaries are added after the x-boundaries, the `for (auto& bndry : localmesh->getBoundaries()) {` loop above should first extrapolate into the x-boundaries, then into the y-boundaries, including extrapolating into corner cells using values from x-boundary cells.

johnomotani: I'd be happy to set the corners to NaN only when we are extrapolating into some boundary cells. Otherwise there is correct data in the corner cells, which there is no reason to overwrite. Keeping it can occasionally be useful: for example when using staggered grids, there is a case when we want to invert a Laplacian at `Ny+1`, and the non-uniform grid correction breaks unless there is valid data in the corner cells.

examples/performance/communications/communications.cxx
```diff
@@ -0,0 +1,38 @@
...
+
+  int outputMonitor(BoutReal, int, int) {
+
+    output<<endl<<"just comms "<<Timer::getTime("pre-synchronised communications")<<endl;
+
+    return 0;
+  }
+};
+
+BOUTMAIN(CommsSpeed);
```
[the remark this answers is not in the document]
johnomotani: Yes, that is simpler :+1:

include/bout/mesh.hxx
```diff
@@ -514,6 +552,8 @@ class Mesh {
   //////////////////////////////////////////////////////////
   
   int GlobalNx, GlobalNy, GlobalNz; ///< Size of the global arrays. Note: can have holes
+  /// Size of the global arrays excluding boundary points.
+  int GlobalNxNoBoundaries, GlobalNyNoBoundaries, GlobalNzNoBoundaries;
```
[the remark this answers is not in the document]
johnomotani: Some way to get `GlobalNyNoBoundaries` is very useful, because the number of y-boundary cells can change depending on whether there is a second X-point or not.

The `*NoBoundaries` versions correspond to the `getGlobal*IndexNoBoundaries()` methods, so are useful for things like "am I within 8 grid points of the outer radial boundary?".

Is there a different interface for getting this kind of data that you'd rather have? If we could access `GlobalNx`, etc. through there we could also remove them from the public interface.
* Maybe something like `gnx = mesh->get("GlobalNx");`, except we need a different name because `Mesh` already has a `get` method? I guess this would be implemented with a `std::map` that we could initialise at the end of `BoutMesh::load()` - is the cost of a lookup in a small `map` worth worrying about at all (assuming we don't call this method inside tight loops)?

include/bout/mesh.hxx
```diff
@@ -960,6 +1000,10 @@ protected:
   /// Pointer to the global MPI wrapper, for convenience
   MpiWrapper* mpi = nullptr;
 
+public:
+  // Switch for communication of corner guard and boundary cells
+  const bool include_corner_cells;
```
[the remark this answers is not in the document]
johnomotani: I don't immediately see how to move the logic into `BoutMesh::send` - the `include_corner_cells` case needs to wait twice (in the current implementation, this may change if/when I revert back to the version with simultaneous communication of x-, y- and corner-guard cells).

On the other hand, I expect `include_corner_cells` might be wanted, at least for checks, in other places eventually. For example in `D2DXDY` we could have a simpler, possibly more consistent, implementation when we have consistent corner cells; we might even want to only allow `D2DXDY` to be called when `include_corner_cells == true`.

include/bout/mesh.hxx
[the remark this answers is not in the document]
ZedThree: I think just wrapping them in a getter and not providing a setter:

```cpp
class Mesh {
private:
  int global_nx_no_boundaries;
public:
  int Mesh::getGlobalNxNoBoundaries() { return global_nx_no_boundaries; }
};
```

There's a lot of stuff that are currently just public data members that should not be, for example `LocalN*`, `GlobalN*`, all the metric components in `Coordinates`, and so on.

It could be a lot of pain to convert everything to use getters, but it would be more correct.

Leave them as public members for now, otherwise it would be weirdly inconsistent, which is probably worse!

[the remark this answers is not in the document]
johnomotani: Just to note, the metric components should probably stay public. It's likely that users (e.g. STORM) want to normalise lengths, which means the metric components need to be rescaled in physics-model code.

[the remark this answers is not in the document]
ZedThree: Oh, for sure those also want a setter, but it should be a function so that we can then recalculate things that need to be consistent

PR #1886, opened 2020-01-14
Hypre interface

ZedThree:
WIP wrapper for HYPRE's `HYPRE_IJVector` and `HYPRE_IJMatrix` following the `PetcscVector/PetscMatrix`

Mostly complete, but I've only updated the CMake system to enable , as it was much easier and I'm lazy.

There's a couple of commits for incidental things that could go in other PRs:

- Rename some unit tests to be a little more robust -- these are "death tests" that check a process will `exit` correctly. At least in one case, I suspect a redesign of the code (`parseCommandLine`) would be best anyway
- Add `Mesh::getRegion<T>`. This is useful if you've only got a `Mesh` but want a `Region<T>` and don't know what `T` is yet (which is necessary for this PR!)

I've also moved the `GlobalIndexer` out of the PETSc interface so that this PR can use it too. It does raise the question of should this be hanging onto a `PetscLib`? I suspect it should really be moved into the owners of the `PetscMatrix`, etc.

Like the PETSc interface, when creating a `HypreVector` from a field, it sets values one-by-one. It might be possible to speed this up if necessary by, instead of using `BOUT_FOR`, directly looping over the region blocks and passing a pointer to the first element in that block:

```cpp
    for (const auto& block : region.getBlocks()) {
      std::vector<HYPRE_Int> indices;
      indices.reserve(block.second.ind - block.first.ind);
      for (auto index = block.first; index < block.second; ++index) {
        indices.emplace_back(static_cast<HYPRE_BigInt>(indexConverter->getGlobal(index)));
      }
      HYPRE_IJVectorSetValues(hypre_vector,
                              static_cast<HYPRE_Int>(block.second.ind - block.first.ind),
                              indices.data(), &f[block.first]);
    }
```

This works, but I don't know if it's actually any faster. I'm also not sure if the global index will ever be -1 for an index in a block. The blocks represent contiguous blocks in a region, so assuming the region only ever contains valid indices, this should be fine?


ZedThree:
Just to note that it currently clashes with the version of HYPRE in PETSc <= 3.11. Might work with the version in 3.12, not checked.

ZedThree:
I pulled out the non-related changes into separate PRs, as they can go in sooner than this.

If this is to go in, it will need some refactoring after the generalised Petsc interface and AMG go in.

ZedThree:
Closed in favour in the bigger HYPRE/GPU changes

PR #1887, opened 2020-01-15
Remove unused parameters to silence warnings in MsgStack

ZedThree:
As they're in the header and it's included everywhere, the warnings make non-debug builds _very_ noisy


d7919:
Should this be backported as well?

ZedThree:
No, sorry, this came in from the fmt PR. I didn't catch it at the time because I mostly just build in debug mode on my machine.

It's tempting to put `-Werror` on Travis to catch new warnings, but I fear that would be overly strict

Issue #1873, opened 2019-12-13
Generalising the PETSc interface


cmacmackin:
Following up on the work done for #1771 and #1805:

As discussed at the BOUT++ workshop, the choice of which grid cells to include in a PETSc vector/matrix is problem dependent. The PETSc interface currently just selects the cells which I needed for writing the 3D AMG Laplace solver (#1803). We need to consider how best to generalise the interface to work for any problem. This can be done using my new [`OperatorStencil` class](https://github.com/boutproject/BOUT-dev/blob/0a842b75d39d7a4debeb4b564883066eb81d7262/include/bout/operatorstencil.hxx). My current thoughts on how to do this are summarised below:

- Pass a stencil when creating a `GlobalIndexer` object
- Iterate over `RGN_NOBNDRY`, getting the stencil for each
    - If no stencil is available, that cell will be excluded from the PETSc representation of the field
    - If a stencil is available, any boundary cells which it includes will be placed in a list
- Iterate over the list of boundary cells, getting the stencil for each
    - If no stencil is available, this is an error (the `OperatorStencil` object is inconsistent)
    - If a stencil is available, add any boundary cells which are not in the first list to a second one
- Iterate over the second list and repeat until no new cells are added
- Create `Region<>` objects representing the boundaries and the domain as a whole
- Use the latter-most of these regions to iterate over and fill field objects with global indices
- While doing this, build up information on the sparsity pattern, to be saved as a component of the `GlobalIndexer`
- Add routines to get the `Region<>` objects from the `GlobalIndexer`, `PetscVector`, and `PetscMatrix` classes

The main issues to be resolved with this is how to avoid performing the rather expensive operation multiple times. There could no longer be a global instance of the `GlobalIndexer` which corresponded to the global mesh, as there would be different `GlobalIndexer`s for this mesh depending on the stencil being used. Probably the simplest approach would be to require the user to pass the a shared pointer to a `GlobalIndexer` to the constructors for `PetscVector`/`PetscMatrix`. There could also be a method to get the shared pointer from the vectors and matrices.

Thoughts @ZedThree @johnomotani @bendudson @xxu?

cmacmackin:
[...]

PR #1889, opened 2020-01-16
Petsc interface generalisation

cmacmackin:
As described in #1873, this generalises the PETSc interface so it can be used on a wider variety of problems. This PR contains two significant changes:

- the introduction of an OperatorStencil class which provides information on a matrix's sparsity pattern, needed for preallocating memory
- the OperatorStencil is further leveraged to determine which guard cells should be given unique global indices and included in PETSc vec objects.

These changes will eventually be merged as part of #1803, but as work is ongoing to apply the PETSc interface to other parts of the code I thought it would be best to get them merged into `next` ASAP.


include/bout/petsc_interface.hxx
```diff
@@ -42,64 +42,223 @@
...
 
   PetscLib lib;
   Mesh* fieldmesh;
 
   /// Fields containing the indices for each element (as reals)
-  Field3D indices3D;
-  Field2D indices2D;
-  FieldPerp indicesPerp;
+  T indices;
+  PetscInt globalStart, globalEnd;
```
ZedThree: Am I right in saying `globalEnd` is the inclusive end? Best to leave a comment if so

include/bout/region.hxx
```diff
@@ -605,8 +605,8 @@ public:
     return *this;
   }
 
-  /// Return a new region equivalent to *this but with indices contained
-  /// in mask Region removed
+  /// Updates the region so that indices contained in mask Region are
```
ZedThree: The previous comment was more correct: this returns a copy of `this` with `maskRegion` removed

include/bout/region.hxx
```diff
@@ -636,6 +636,38 @@ public:
     return *this; // To allow command chaining
   };
 
+  /// Updates the region to only include indices also contained in the
```
ZedThree: This returns a new value, rather than updating `this`

include/bout/petsc_interface.hxx
```diff
@@ -42,64 +42,223 @@
...
+	newIndices.clear();
+        for (const ind_type i : candidateIndices) {
+	  insertIndex(i, allIndices, newIndices);
+        }
+	candidateIndices = newIndices;
+      }
+      std::vector<ind_type> allIndicesVec(allIndices.begin(), allIndices.end());
+      allCandidate = Region<ind_type>(allIndicesVec);
+    } else {
+      regionAll = getRegionNobndry();
```
ZedThree: `regionAll` is unconditionally assigned to below (L120); is this value used?

include/bout/petsc_interface.hxx
```diff
@@ -211,6 +343,20 @@ public:
...
   }
 
+  PetscVector<T>& operator=(const T& f) {
+    BOUT_FOR_SERIAL(i, indexConverter->getRegionAll()) {
+      const PetscInt ind = indexConverter->getGlobal(i);
+      if (ind != -1) {
+        // TODO: consider how VecSetValues() could be used where there
+        // are continuous stretches of field data which should be
+        // copied into the vector.
+        VecSetValue(*vector, ind, f[i], INSERT_VALUES);
```
ZedThree: I tried this for the Hypre wrapper and it worked. No idea on performance:

```cpp
    for (const auto& block : region.getBlocks()) {
      std::vector<HYPRE_Int> indices;
      indices.reserve(block.second.ind - block.first.ind);
      for (auto index = block.first; index < block.second; ++index) {
        indices.emplace_back(static_cast<HYPRE_BigInt>(indexConverter->getGlobal(index)));
       }
      HYPRE_IJVectorSetValues(hypre_vector,
                              static_cast<HYPRE_Int>(block.second.ind - block.first.ind),
                              indices.data(), &f[block.first]);
     }
```

cmacmackin: Yes, I was thinking something along these lines could work. I'd prefer to leave it for later, potentially after some benchmarking however.

cmacmackin:
Thanks for the comments. Some of these are duplicates of those on #1803, which I am in the process of addressing right now and will place on this branch once ready.

include/bout/mesh.hxx
```diff
@@ -379,7 +379,8 @@ class Mesh {
   MPI_Comm getXcomm() {return getXcomm(0);} ///< Return communicator containing all processors in X
   virtual MPI_Comm getXcomm(int jy) const = 0; ///< Return X communicator
   virtual MPI_Comm getYcomm(int jx) const = 0; ///< Return Y communicator
-  
+  MpiWrapper* getMpi() { return mpi; } ///< Return pointer to the mesh's MPI Wrapper object
```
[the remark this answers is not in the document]
cmacmackin: Don't see a problem with this.

include/bout/operatorstencil.hxx
```diff
@@ -0,0 +1,241 @@
...
+
+template<class T>
+const inline T operator+(const T &lhs, const IndexOffset<T> &rhs) {
+  // If CHECKLEVEL >= 3 then SpecificInd<N>.zp() complains about
+  // negative values.
+  if (rhs.dz >= 0) {
+    return lhs.xp(rhs.dx).yp(rhs.dy).zp(rhs.dz);
+  } else {
+    return lhs.xp(rhs.dx).yp(rhs.dy).zm(-rhs.dz);
+  }
```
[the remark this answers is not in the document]
cmacmackin: Yes, I've fixed both of these. I missed the `offset` method.

include/bout/petsc_interface.hxx
```diff
@@ -42,64 +42,223 @@
...
-  /// Finish setting up the indexer, communicating indices across processes.
-  void initialise();
+  void initialiseTest() {}
+
+  /// Finish setting up the indexer, communicating indices across
+  /// processes and, if possible, calculating the sparsity pattern of
+  /// any matrices.
+  void initialise() {
+    fieldmesh->communicate(indices);
+    fieldmesh->communicate(indices);
```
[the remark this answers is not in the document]
cmacmackin: Will do, although the work of @johnomotani in #1885 will eliminate the need for this soon.

include/bout/petsc_interface.hxx
cmacmackin: Yes, that's right.

include/bout/region.hxx
cmacmackin: Apologies for the incorrect change.

include/bout/petsc_interface.hxx
cmacmackin: Thanks for catching that. It was a remnant of an older (incorrect) algorithm I was using.

include/bout/operatorstencil.hxx
```diff
@@ -0,0 +1,241 @@
...
+  const_iterator cend() const { return std::cend(stencils); }
+  reverse_iterator rbegin() { return std::rbegin(stencils); }
+  const_reverse_iterator rbegin() const { return std::rbegin(stencils); }
+  const_reverse_iterator crbegin() const { return std::crbegin(stencils); }
+  reverse_iterator rend() { return std::rend(stencils); }
+  const_reverse_iterator rend() const { return std::rend(stencils); }
+  const_reverse_iterator crend() const { return std::crend(stencils); }
+
+private:
+  std::vector<std::pair<stencil_test, stencil_part>> stencils = {};
```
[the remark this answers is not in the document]
cmacmackin: True, much tidier, plus a more succinct type-name.

cmacmackin:
@ZedThree I'm not sure if many more methods on the vector and matrix classes _should_ be made `const`, actually. While strictly speaking they _can_ be (as it is not the contents of the class itself which is modified) they still are affecting changes to the values stored in the backend PETSc object. Thus while the compiler wouldn't complain they are semantically non-`const`.

A question on style/convention when it comes to these issues: is it every appropriate to apply `const` to the function but not to the function result? Specifically, it would be possible to make the indexing methods `const` as they only return an accessor for an element of the vector/matrix. However, it makes no sense to declare these accessors as `const`, as their main purpose is for setting the value of elements. Given that the indexing would usually be used to return an accessor which would immedoately be used to set the value of an element, it strikes me that it would be misleading to declare these methods as `const`, even though strictly speaking they are. What would normally be done in this case?

include/bout/petsc_interface.hxx
ZedThree: Yes, that's wise

ZedThree:
You're right that's there's not much more on the vector/matrix classes that could be `const`, I was thinking of `GlobalIndexer`, apologies! Things like `getGlobal` and `isLocal` should be `const`. Less certain about e.g. `getNumDiagonal`, as it would need `numDiagonal` to be `mutable`. Probably ok if you view the latter as a cache variable. It should probably at least return `const&` though to avoid a copy.

Indeed, a `const` method means "logically `const`" as opposed to "physically `const`", so it's not really correct to make something `const` that returns a object that can be used to modify the state of the parent. When it comes to these things, we can add a `const` overload that returns a `const&` or `const*`. I think you already have this for `PetscVector::get`. Dropping the `const` from the return type only makes sense for simple types like `double` or `int` (really, value types) where they can do harm, as it were.

`Element` is really a reference type, and as you point out, should be `const`. However, for the indexing, I had the following overloads for the `HypreMatrix::operator()`:

```cpp
Element operator()(const ind_type& row, const ind_type& column);
BoutReal operator()(const ind_type& row, const ind_type& column) const;
```

where the latter just gets the value directly. You could return a `const Element`, but as the only thing you (probably?) want to do with it is to cast it to a `BoutReal`, I just skipped that part. Also there was some trickiness to creating a `const Element` to begin with for reasons. Also also, note the `const` on the arguments!

I think `PetscVector::toField` should be `const` and return just `T`. Other than that, you're right, there's not much else on vector/matrix to make `const`.

cmacmackin:
Yes, unfortunately `getNum[Off]Diagonal()` can not be `const` because it caches the result. I have switched it to returning a `const` referance however, and am not sure how I missed that earlier.

Currently I have methods to convert from an `Element` to a `BoutReal` but it probably makes more sense to return the latter directly from `operator()`. It also allows for the `const` overload to be simpler than the normal indexing method (at least in the case of matrices), as it doesn't need to worry at all about getting positions and weights.

Yes, I'd already spotted that `toField()` should be made `const`.

cmacmackin:
I believe the latest push should address all of your comments. I think it should pass the CI too, although we'll find out in an hour or so.

ZedThree:
Just to be clear on the `getNum[Off]Diagonal()`: it can be made `const` if `num[Off]Diagonal` is made `mutable`. This allows the member to be changed even in `const` methods: remember, this is ok as long as the method doesn't _logically_ change the instance, which is the case here because `num[Off]Diagonal` is just acting as a cache. Calling `getNum[Off]Diagonal` doesn't result in any observable changes to `GlobalIndexer`\*.

Also `toField` returns `const T`, when it could be `T` (`const` on returned value types is most likely ignored by compiler anyway).

Neither of these are blocking changes and can be fixed later, I just wanted to clarify my earlier points!


\* except via a timing side-channel, and we don't care about that here.

cmacmackin:
There are some issues with the OpenMP builds which I am currently addressing. More concerning are the failures with the `clang` build, which I am at a loss to explain.

ZedThree:
The clang build has some warnings about inconsistent or missing `virtual`/`override` -- it's worth addressing these, even if they don't fix the problem.

OpenMP I guess you've got sorted, but probably due to the gtest `ASSERT_EQ` inside `BOUT_FOR` -- just change to `BOUT_FOR_SERIAL`

cmacmackin:
I've fixed those issues but the errors with the clang run still persist. What is even stranger is that one of the OpenMP runs still breaks: the one built with CMake. The autoconf build works just fine, despite apparently using the same build-options as the CMake one. I have no idea what could be causing that problem.

cmacmackin:
The problem in the `clang` build proved to be quite subtle. Although I was caching the sparsity pattern calculated in the `GlobalIndexer` class, I actually recalculated it any time either `GlobalIndexer::getNumDiagonal()` or `GlobalIndexer::getNumOffDiagonal()` were called. This could invalidate pointers to data from any previous calls. While that didn't cause problems with GCC, it did with clang. I've now fixed the caching so cached values are actually used when available, without recalculating. That fixes the problem.

cmacmackin:
I don't fully understand what the issue was with the CMake/OpenMP case. However, it seems that IndexerTest/TestGetRegionBndry was not properly cleaning up after itself, resulting in PETSc not being properly initialised in future tests. The problem seems to have been related to a for-loop with the iterator being a reference to GlobalIndexer objects. Switching to using pointers got rid of the problem (when running locally, at least).

cmacmackin:
@ZedThree 
All tests are passing and all requested changes have been made. I believe it is now ready to be merged.

ZedThree:
Thanks @cmacmackin !

I suspect the issue with that test was something to do with making a copy of the `GlobalIndexer`s, but I can't quite see what the actual problem is. The fix is fine though, so I won't worry about it too much.

PR #1897, opened 2020-01-23
Algebraic Multigrid: ZInterpolation update

johnomotani:
This PR addresses the comment https://github.com/boutproject/BOUT-dev/pull/1803#issuecomment-572015750 about setting options for interpolation by splitting `HermiteSplineOnlyZ` (renamed as `ZHermiteSpline`) into a separate interface from the 2d interpolation classes. The split made more sense after the changes in #1896 (and had conflicts with that PR that would need resolving eventually anyway) so I've merged that here. We should perhaps hold off merging this until #1896 is approved, just in case major changes are needed.


johnomotani:
Note: on my computer the unit tests don't all pass, but the failures I get are the same as the ones I get on the current algebraic_multigrid branch, so I think they are unrelated.

johnomotani:
This has been made redundant by #1944 and #1946, we just need to merge `next` into #1803. Closing...

PR #1913, opened 2020-02-18
pkg-config petsc (master)

dschwoerer:
Allow to use pkg-config to detect petsc.
PETSc provides pkg-config files. Unlike the normal method, this is less dependent on finding the magic files of petsc in various locations, and can be compatible with the [FHS](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard)

### DIFF
--- a/cmake/FindPETSc.cmake
+++ b/cmake/FindPETSc.cmake
@@ -352,7 +352,7 @@ endif ()
 
 if (NOT PETSC_INCLUDES)
   include(FindPkgConfig)
-  pkg_check_module(PkgPETSC PETSc>3.4.0 petsc>3.4.0)
+  pkg_search_module(PkgPETSC PETSc>3.4.0 petsc>3.4.0)
   set (PETSC_LIBRARIES ${PkgPETSC_LINK_LIBRARIES} CACHE STRING "PETSc libraries" FORCE)
   set (PETSC_INCLUDES ${PkgPETSC_INCLUDE_DIRS} CACHE STRING "PETSc include path" FORCE)
   set (PETSC_EXECUTABLE_RUNS "not-needed")
--- a/cmake/FindPETSc.cmake
+++ b/cmake/FindPETSc.cmake
@@ -350,7 +350,7 @@ int main(int argc,char *argv[]) {
   mark_as_advanced (PETSC_INCLUDES PETSC_LIBRARIES PETSC_COMPILER PETSC_DEFINITIONS PETSC_MPIEXEC PETSC_EXECUTABLE_RUNS)
 endif ()
 
-if (NOT PETSC_FOUND)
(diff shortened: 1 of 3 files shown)


ZedThree:
I think the pkg-config package name is `PETSc`. On all the systems I've got access to, it's case sensitive.

Also this requires some more logic to deal with static libraries. As far as I'm aware, pkg-config doesn't actually tell you whether or not something needs linking statically, so we'll need to get the static linker flags too and check if they're needed.

I spent some time yesterday trying to get the CMake system to use pkg-config as a first guess, but I struggled to pass the required information to where it's needed. `pkg_check_modules` creates a target which can be used directly, but it's missing all the info about static libraries. Some of this can be added, but I didn't manage to get the library paths in there. Looking again, this might be `IMPORTED_LOCATION`, but I'm not sure if it takes a list.

Lastly, the PETSc.pc puts the necessary `-fopenmp` flag in `cflags_extra` if it was built with OpenMP, so that variable needs to be checked. With CMake, this will then need to `find_package(OpenMP)` and add that as a dependency.

I'm not really convinced that the amount of work needed to get this robustly working is worth the effort.

dschwoerer:
It seems they are called `petsc` on fedora. There this works, as it is not statically linked and also build without openmp.

In that case we could still merge it, as it improves the situation on fedora, but shouldn't cause issues otherwise, as it is only used as a fall-back if the detection with `PETSC_DIR/PETSC_ARCH` fails.

With configure I think I could just modify `make.config` after `./configure` without petsc, but with CMake I wouldn't know how to make it work on fedora ...

ZedThree:
I'm ok with it as a fall-back and not working with static libs, seeing as most (all?) distros install shared libs -- but please fix it to also check `PETSc` as that is much more common.

dschwoerer:
Changed, I will complain in fedora about the wrong file name :+1: 
I left it in as fall-back for now.

cmake/FindPETSc.cmake
```diff
@@ -350,6 +350,14 @@ int main(int argc,char *argv[]) {
   mark_as_advanced (PETSC_INCLUDES PETSC_LIBRARIES PETSC_COMPILER PETSC_DEFINITIONS PETSC_MPIEXEC PETSC_EXECUTABLE_RUNS)
 endif ()
 
+if (NOT PETSC_INCLUDES)
+  include(FindPkgConfig)
+  pkg_check_module(PkgPETSC PETSc>3.4.0 petsc>3.4.0)
```
ZedThree: Should be `pkg_search_module` to find first of list? `pkg_check_modules` needs all modules to be found

dschwoerer: Thanks for noticing, I thought only the missing s would be the different. Fixed now.

PR #1914, opened 2020-02-18
pkg-config petsc (next)

dschwoerer:
Allow to use pkg-config to detect petsc.
PETSc provides pkg-config files. Unlike the normal method, this is less dependent on finding the magic files of petsc in various locations, and can be compatible with the [FHS](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard)

#1913 for next


ZedThree:
Looks like this has pulled in a bunch of commits from elsewhere.

dschwoerer:
I merged master into next, as master contains e.g. a bug fix, that is not fixed in next:

5403eeb:
```
 Conflicts:
	src/fileio/datafile.cxx

Use master version, where we do not add the same variable twice.
```

dschwoerer:
Replaced by #1999

dschwoerer:
Something went wrong with #1999

The configure.ac master history list the pkg-config change:
https://github.com/boutproject/BOUT-dev/commits/master/configure.ac

However, the next one doesn't:
https://github.com/boutproject/BOUT-dev/commits/next/configure.ac

dschwoerer:
I had to cherry-pick the old commits. The CMake changes where still there, but the configure.ac changes where gone ...

dschwoerer:
@ZedThree Can you please merge this?
It was in, and was removed by accident.

If you would like me to clean the history, let me know.

PR #1918, opened 2020-02-21
Implement BOUT_FOR for FV::div_f_v

bshanahan:
Replaces traditional loops with a BOUT_FOR in the finite volume Div_f_v
**Also allows possibility for coordinate objects to be 3D (hence the i.k indices)


ZedThree:
Seeing as we're touching them anyway, please could you run `clang-format` over these functions?

I don't think we've currently got tests for these either? Would you be able to throw something basic together?

PR #1919, opened 2020-02-25
Add v5 upgrader scripts for XZInterpolation

ZedThree:
Two BOUT++ v4->v5 upgrader scripts for the XZInterpolation rename PR: one for the `Interpolation` types, one for the input files. Plus changes to `boutdata.data.BoutOptions` required for the latter.

1cdf2d1 is a PEP8 formatting commit, so touches a bunch of lines in `boutdata/data.py`

The type upgrader optionally uses the python libclang bindings. These allow more powerful/accurate refactorings, but need libclang and the python bindings installed. Best installed through your system package manager.

I had to change `BoutOptions` to keep track of comments and the original case of keys so that writing to file resulted in a minimal diff. Unfortunately, it still results in some changes even if no input options need renaming:

* nested sections are moved to be under their parent section, while preserving relative order
* empty sections are removed
* floating point numbers may have their format changed, although the value will not change
* consecutive blank lines will be reduced to a single blank line
* whitespace around equals signs will be changed to exactly one space
* trailing whitespace will be removed

Because of this, the input file upgrader first shows the diff between the original and the "canonical" version.

I consider "canonicalising" all the input files in the repo, but there's a lot and it's mostly whitespace changes.

### DIFF
--- a/bin/bout-v5-input-file-upgrader.py
+++ b/bin/bout-v5-input-file-upgrader.py
@@ -125,6 +125,8 @@ if __name__ == "__main__":
 
               * trailing whitespace will be removed
 
+              * comments will always use '#'
+
             Files that change in this way will have the "canonicalisation" patch
             presented first. If you choose not to apply this patch, the "upgrade
             fixer" patch will still include it."""
--- a/bin/bout-v5-input-file-upgrader.py
+++ b/bin/bout-v5-input-file-upgrader.py
@@ -12,7 +12,11 @@ from boututils.boutwarnings import AlwaysWarning
 # This should be a list of dicts, each containing exactly two keys:
 # "old" and "new". The values of these keys should be the old/new
 # names of input file values or sections
-REPLACEMENTS = []
+REPLACEMENTS = [
+    {"old": "mesh:paralleltransform", "new": "mesh:paralleltransform:type"},
+    {"old": "fci", "new": "mesh:paralleltransform"},
+    {"old": "interpolation", "new": "mesh:paralleltransform:xzinterpolation"},
+]
 
 
 def fix_replacements(replacements, options_file):
--- a/bin/bout-v5-input-file-upgrader.py
+++ b/bin/bout-v5-input-file-upgrader.py
@@ -42,7 +42,7 @@ def apply_fixes(replacements, options_file):
 
     modified = copy.deepcopy(options_file)
 
-    fix_replacements(replacements, options_file)
+    fix_replacements(replacements, modified)
 
     return modified
 
--- a/bin/bout-v5-input-file-upgrader.py
+++ b/bin/bout-v5-input-file-upgrader.py
@@ -1,0 +1,196 @@
+#!/usr/bin/env python3
+
+import argparse
+import copy
+import difflib
+import textwrap
+import warnings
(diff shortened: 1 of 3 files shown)


bin/bout-v5-input-file-upgrader.py
```diff
@@ -0,0 +1,200 @@
...
+            ) from e
+
+
+def apply_fixes(replacements, options_file):
+    """Apply all fixes in this module
+    """
+
+    modified = copy.deepcopy(options_file)
+
+    fix_replacements(replacements, options_file)
```
johnomotani: Should this apply `fix_replacements` to `modified` rather than `options_file`?

ZedThree: Good spot, thanks!

tools/pylib/boutdata/data.py
```diff
@@ -266,17 +511,41 @@ class BoutOptionsFile(BoutOptions):
...
+                    continue
+
+                comment_match = self.COMMENT_REGEX.search(line)
+                if comment_match is not None:
+                    line = line.split("#")[0].strip()
+                    comment_whitespace = comment_match.group(1)
+                    inline_comment = comment_match.group(2).strip()
+                else:
+                    inline_comment = None
                 startpos = line.find(";")
```
johnomotani: I never use `;` for comments in `BOUT.inp`, but I think we support `;` as equivalent to `#`. Is it possible to treat them both the same here?

ZedThree: I somehow just ignored that bit. It looks like there are no input files in the repo that actually use ';' as a comment.

Still, fixed!

bendudson: Probably use of `;` as comments can be deprecated / removed in 5? It was a simpler time of IDL where `;` was first used, but now is not.

ZedThree: Can also handle that in this script! Just always write out '#' as the initial comment character

johnomotani:
Probably it would be good to start a section in the manual listing the breaking changes in 4->5, and the updater scripts that can be used to fix code/input files. @ZedThree could you add to this PR?

johnomotani:
> Probably it would be good to start a section in the manual listing the breaking changes in 4->5, and the updater scripts that can be used to fix code/input files. @ZedThree could you add to this PR?

Actually, might be better to add the section to the manual in #1920 as that will probably get merged first. Then we could merge `next`/#1920 and add the scripts from here before #1896 is merged.

ZedThree:
There's a section in CHANGELOG.md that lists breaking changes. I'll just copy-paste that into the manual.

Might be an idea to set up a PR template to remind us of things like this? I certainly forgot about it!

johnomotani:
> There's a section in CHANGELOG.md that lists breaking changes. I'll just copy-paste that into the manual.

Can we just import `CHANGELOG.md` into the manual somehow, or link to it from an 'upgrading from v4 to v5' section? Would be easier to keep the list up to date in only one place!

ZedThree:
I think something like that should work. I'll do it in a separate PR though, and just note the breaking changes in CHANGELOG.md for now

PR #1920, opened 2020-02-25
Generate version header

ZedThree:
This solves a problem with the CMake system: everything gets rebuilt on every commit because `-DBOUT_VERSION*` are passed on the command line to every file, and CMake knows that if the preprocessor macros for a file change then the file needs to be rebuilt.

This is solved by adding `include/bout/version.hxx` which is generated by the build system from `include/bout/version.hxx.in`. This header contains `constexpr` variables in the `bout::version::` namespace, and replaces the preprocessor macros `BOUT_VERSION`, `REVISION`, `BOUT_VERSION_STRING` and `BOUT_VERSION_DOUBLE`.

It might be nice to mark those macros as deprecated somehow. Not found a nice portable way so far!

There's a little funkiness with the makefile build system: `BOUT_REVISION` (renamed from `REVISION`) is still passed to each file via the command line. I couldn't see a way round this, which is why `version.hxx.in` still has an `#ifndef` in it.

Also, there is a potential issue here: if in one cxx file you take the address of a `constexpr` variable defined in a header, you may/will get a different address than in another cxx file. C++17 solves this with `inline` variables.
Mostly we shouldn't be taking the address of `constexpr` variables, but unfortunately, this is how `Datafile::add` works. Basically, don't call `Datafile::add` on the same `bout::version` variable twice in different files, or it will throw. But you probably shouldn't be doing that anyway!

Lastly, there's a bunch of unrelated changes to `configure` because I think the gettext macro has been updated on my system. Happy to pull those changes out into a separate PR.


include/bout.hxx
```diff
@@ -52,8 +52,7 @@
 #include "where.hxx"
 #include "bout/mesh.hxx"
 #include "bout/solver.hxx"
-
-const BoutReal BOUT_VERSION = BOUT_VERSION_DOUBLE; ///< Version number
+#include "bout/version.hxx"
```
johnomotani: Do we need to include `bout/version.hxx` here? Won't that result in as many files being rebuilt as when the macro was in `bout.hxx` directly?

johnomotani:
I guess this is a breaking change. People probably not using `BOUT_VERSION`, etc. but if they are the names have changed. Needs adding to a `4to5` script?

include/bout.hxx
ZedThree: I thought that this header was only included in user code (and `bout++.cxx`), turns out I forgot it's also included in `physicsmodel.hxx` which is then included in `solver.hxx`, etc. 

We don't currently have a one-true-public-header that bundles everything needed for user code. I think we probably do want one, and this can then go in there.

I'll remove it for now.

ZedThree:
Yes, probably wise!

include/bout.hxx
ZedThree: So, I initially had the commit hash in a separate file, `revision.hxx`, and then I merged it with `version.hxx`.

However, the version info changes much less frequently than the commit. Maybe it does make more sense to have them in separate files? Currently only `bout++.cxx` uses the commit, whereas `physicsmodel.cxx` and a couple of tests use the version number.

johnomotani: I think that makes sense. Minimising the number of files to recompile after changing the commit hash would be nice :+1:

ZedThree: Ok, down to just `bout++.cxx` that needs recompiling on commit. CMake seems to still also rebuild all the tests when I think they only need relinking. I suspect it might be because it might not be tracking header includes across targets, but I might be doing something wrong.

ZedThree:
Possible gotcha: if you build with `./configure` and then with CMake, it will pick up the makefile system version of `include/bout/revision.hxx`, not the CMake generated one.
`make distclean` now cleans up the generated headers.

I'm now happy that this does what we want for CMake: on commit `revision.hxx` is regenerated, only `bout++.cxx` gets recompiled, and all the executables only get relinked and not rebuilt.

`bin/bout-v5-macro-upgrader.py` does a "best effort" attempt to convert macros to variables, add the necessary headers, and remove unneeded `#ifdef`s. It doesn't attempt to expand other macros or fix **#ifndef**s as this got tricky very quickly: a simple replacement might not be correct.

bin/bout-v5-macro-upgrader.py
```diff
@@ -0,0 +1,227 @@
...
+        )
+        modified = fix_ifdefs(replacement["old"], modified)
+        modified = fix_replacement(replacement["old"], replacement["new"], modified)
+
+    return modified
+
+
+def yes_or_no(question):
+    """Convert user input from yes/no variations to True/False
+
```
johnomotani: Being picky, but should this blank line be deleted?

ZedThree: Probably, but I'm going to pull out some these repeated functions into a `v5_upgrader` module, as I keep copy-pasting between the upgrader scripts

At some point, there will be a super "do everything" script that just calls all the `apply_fixes` functions

johnomotani:
> It doesn't attempt to expand other macros or fix #ifndefs as this got tricky very quickly: a simple replacement might not be correct.

Would it be worth adding some warnings to flag up any uses that weren't fixed? The compiler should give errors there anyway, but I think it would be friendlier if this script gave a warning.

ZedThree:
Good point. We're now not defining those macros, so `#ifndef` sections will suddenly get compiled. 
Maybe remove the `#ifndef` lines and also flag a warning?

ZedThree:
This now takes care of situations like the contrived:

```diff
-#ifndef BOUT_VERSION_STRING
-#ifdef REVISION
-  output_progress.write(_("BOUT++ version {:s}\n"), BUILDFLAG(REVISION));
-#else
-  output_progress.write(_("BOUT++ version {:s}\n"), "Unknown");
-#endif
-#else
-  output_progress.write(_("BOUT++ version {:s}\n"), BOUT_VERSION_STRING);
-#endif
-#ifdef REVISION
-  output_progress.write(_("Revision {:s}\n"), BUILDFLAG(REVISION));
-#endif
-#ifndef REVISION
-  output_progress.write(_("Revision {:s}\n"), "Unknown");
-#endif
+  output_progress.write(_("BOUT++ version {:s}\n"), bout::version::full);
+  output_progress.write(_("Revision {:s}\n"), BUILDFLAG(bout::version::revision));
```

I consider all these upgrader scripts as "best effort only" and somewhat dangerous. The output must be checked by the users, so there's no specific warning for this one. When I get them all into a single framework type thing, it'll have a big flashing warning.

PR #1922, opened 2020-02-27
Workaround for Petsc init/finialise not being idempotent

### LINKED ISSUE #1916: unit tests fail with mpich + petsc
When running with mpich + petsc I am encountering the error:
```
Fatal error in MPI_Add_error_class: Other MPI error, error stack:
MPI_Add_error_class(95): MPI_Add_error_class(errorclass=0x7ffff3fad958) failed
MPI_Add_error_class(74): No more user-defined error classes
[unset]: write_line error; fd=-1 buf=:cmd=abort exitcode=671687183
:
system msg for write_line failure : Bad file descriptor
[Inferior 1 (process 31862) exited with code 017]
```

The error happens as the PETSc library is initialised the 128th time.

I thus assume that this is an issue in MPICH, which does not support more than 128 custom error classes.

Can we get a work around?

ZedThree:
Fixes #1916

The `output.disable/enable` thing is a little bit gross. Could solve that by add `bool quiet` arguments to `PetscLib` constructor and `cleanup`, although it wouldn't be possible to do so to the destructor.

Checked with PETSc git master and MPICH 3.2

### DIFF
--- a/tests/unit/bout_test_main.cxx
+++ b/tests/unit/bout_test_main.cxx
@@ -2,7 +2,9 @@
 
 #include "gtest/gtest.h"
 #include "bout/array.hxx"
+#include "bout/petsclib.hxx"
 #include "fft.hxx"
+#include "output.hxx"
 
 GTEST_API_ int main(int argc, char** argv) {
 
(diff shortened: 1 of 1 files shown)


dschwoerer:
If `PetscLib` would store the bool, it would be only needed in the constructor.

Anyway, this fixes the issue, thanks for the workaround as well as reporting this on all the other levels :+1:

Issue #1923, opened 2020-02-28
error when running make for an example of lapd-drift


bellsis83:
HI all,,

I am trying to run an example of lapd-drift but i have got an error when run make.
The error is as below:

$ make
  Compiling  lapd_drift.cxx
lapd_drift.cxx: In member function ‘virtual int LAPDdrift::rhs(BoutReal)’:
lapd_drift.cxx:508:36: warning: ‘const Field3D Grad_par_LtoC(const Field3D&)’ is deprecated: Grad_par_LtoC is deprecated. Staggering is now supported in Grad_par. [-Wdeprecated-declarations]
       jpar = ((Tet*Grad_par_LtoC(ni)) - (Nit*Grad_par_LtoC(phi)))/(fmei*0.51*nu);
                                    ^
In file included from ../../include/bout.hxx:41:0,
                 from ../../include/bout/physicsmodel.hxx:40,
                 from lapd_drift.cxx:6:
../../include/difops.hxx:254:22: note: declared here
 inline const Field3D Grad_par_LtoC(const Field3D &var) {
                      ^~~~~~~~~~~~~
lapd_drift.cxx:508:63: warning: ‘const Field3D Grad_par_LtoC(const Field3D&)’ is deprecated: Grad_par_LtoC is deprecated. Staggering is now supported in Grad_par. [-Wdeprecated-declarations]
       jpar = ((Tet*Grad_par_LtoC(ni)) - (Nit*Grad_par_LtoC(phi)))/(fmei*0.51*nu);
                                                               ^
In file included from ../../include/bout.hxx:41:0,
                 from ../../include/bout/physicsmodel.hxx:40,
                 from lapd_drift.cxx:6:
../../include/difops.hxx:254:22: note: declared here
 inline const Field3D Grad_par_LtoC(const Field3D &var) {
                      ^~~~~~~~~~~~~
lapd_drift.cxx:543:38: warning: ‘const Field3D Grad_par_CtoL(const Field3D&)’ is deprecated: Grad_par_CtoL is deprecated. Staggering is now supported in Grad_par. [-Wdeprecated-declarations]
         ddt(ni) += Grad_par_CtoL(jpar); // Left hand differencing
                                      ^
In file included from ../../include/bout.hxx:41:0,
                 from ../../include/bout/physicsmodel.hxx:40,
                 from lapd_drift.cxx:6:
../../include/difops.hxx:226:22: note: declared here
[...]

PR #1925, opened 2020-02-28
Fixes for staggering in lapd-drift example

johnomotani:
Also replaces deprecated Grad_par_CtoL() and Grad_par_LtoC() operators.

### DIFF
--- a/examples/lapd-drift/data/BOUT.inp
+++ b/examples/lapd-drift/data/BOUT.inp
@@ -23,6 +23,9 @@ dump_format = "nc"   # Output format (PDB = "pdb", NetCDF="nc")
 
 non_uniform = true
 
+[mesh]
+StaggerGrids = true
+
 ##################################################
 # derivative methods
 
--- a/examples/lapd-drift/lapd_drift.cxx
+++ b/examples/lapd-drift/lapd_drift.cxx
@@ -505,7 +505,8 @@ protected:
     
     if (ZeroElMass) {
       // Set jpar,Ve,Ajpar neglecting the electron inertia term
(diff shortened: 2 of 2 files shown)


ZedThree:
> @ZedThree, @bendudson The staggered derivative operators in `lapd_drift.cxx` were being used the opposite way round to what I thought they meant. `Grad_par_CtoL()` was taking a flux (which I expected to be at `CELL_YLOW`) as argument, and returning a result for a scalar equation (which I expected to be at `CELL_CENTRE`), I thought it should be the other way around: i.e. `CtoL` meant 'argument at `C`' and 'result at `L`'. Doesn't really matter now that the `CtoL` and `LtoC` operators are deprecated anyway, but noting here in case issues come up converting old physics models.


_Originally posted by @johnomotani in https://github.com/boutproject/BOUT-dev/issues/1923#issuecomment-592649012_

I double checked the old implementation of the `LtoC/CtoL` operators, and they did work as you thought. A little bit worrying! Checking a couple of other old physics models, and they are mostly using them the correct way round. I think only `examples/tokamak-2fluid/2fluid.cxx` and `examples/elm-pb/elm-pb.cxx` are also backwards.

@johnomotani Please could you check those models too?

johnomotani:
> @johnomotani Please could you check those models too?
>
> EDIT: most of the examples that use the LtoC/CtoL operators don't actually call setLocation, so they're probably all a little bit dubious

I guess originally as long as the model was consistent with itself it didn't matter much. Having LtoC/CtoL 'back to front' would just correspond to having the scalars at `CELL_YLOW` and fluxes at `CELL_CENTRE`. I think now it would make sense to update all of them to the current conventions, ignoring which way round CtoL/LtoC were originally. I'll make a PR.

PR #1926, opened 2020-02-28
Fix segfaults from PvodeSolver

johnomotani:
`Solver::init()` sets `Solver::initialised=true`. The destructor `PvodeSolver::~PvodeSolver()` uses `initialised` to check if it should free memory. Previously, `Solver::init()` was called near the start of `PvodeSolver::init()`, but this meant that `Solver::initialised` was set to true before the `PvodeSolver` had completed initialisation. An error during the initialisation (which is not unlikely, as `CvodeMalloc` calls the rhs function!) could then cause a segfault as the destructor free'd memory that had not been fully allocated. Calling `Solver::init()` at the end of `PvodeSolver::init()` ensures `Solver::initialised` is not set until `PvodeSolver` really is fully initialised.

This fixes the segfault @johnomotani reported in #1923.

### DIFF
--- a/src/solver/impls/cvode/cvode.cxx
+++ b/src/solver/impls/cvode/cvode.cxx
@@ -115,7 +115,7 @@ CvodeSolver::CvodeSolver(Options* opts) : Solver(opts) {
 }
 
 CvodeSolver::~CvodeSolver() {
-  if (initialised) {
+  if (cvode_initialised) {
     N_VDestroy_Parallel(uvec);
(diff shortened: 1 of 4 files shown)


johnomotani:
Not sure how this worked when I checked it before... Turns out `Solver::init()` needs to be called before `Solver::getLocalN()`, which is needed in `PvodeSolver::init()`. I think `PvodeSolver` needs a separate `pvode_initialised` flag.

PR #1936, opened 2020-03-03
Add extra region option for mixed derivatives

ZedThree:
Cherry picked commits from #1803 + clang-format

Note: clang-format doesn't have an option to control spacing/line breaks around `[[attributes]]`, so they're a little funky.


include/derivs.hxx
```diff
@@ -672,26 +680,35 @@ inline Field3D D2DXDY(const Field3D& f, CELL_LOC outloc, DIFF_METHOD method,
 ///                    If not given, defaults to RGN_NOBNDRY
 /// @param[in] dfdy_boundary_condition Boundary condition to use to set the guard cells of
 ///                                    df/dy, before calculating the x-derivative.
+/// @param[in] dfdy_region Region in which to calculate df/dy. If an empty string
+/// (default)
```
johnomotani: indentation not right here

src/mesh/index_derivs.cxx
```diff
@@ -423,8 +423,13 @@ class FFTDerivativeType {
...
     // Only allow a whitelist of regions for now
-    ASSERT2(region == "RGN_ALL" || region == "RGN_NOBNDRY"
-            || region == "RGN_NOX" || region == "RGN_NOY");
+    ASSERT2(region == "RGN_ALL" || region == "RGN_NOBNDRY" || region == "RGN_NOX"
+            || region == "RGN_NOY" || region == "RGN_INNER_X" || region == "RGN_OUTER_X"
+            || region == "RGN_LOWER_INNER_Y" || region == "RGN_UPPER_INNER_Y"
+            || region == "RGN_LOWER_OUTER_Y" || region == "RGN_UPPER_OUTER_Y"
+            || region == "RGN_LOWER_Y" || region == "RGN_UPPER_Y" || region == "RGN_BNDRY"
+            || region == "RGN_WITH_XBNDRIES" || region == "RGN_WITH_YBNDRIES"
+            || region == "RGN_WITH_BNDRIES");
```
johnomotani: Is the whitelist here actually needed? Would `theMesh->getRegion2D(region)` not throw an exception anyway if the region was not allowed?

src/mesh/index_derivs.cxx
```diff
@@ -491,8 +496,13 @@ class FFT2ndDerivativeType {
...
```
johnomotani: ditto: whitlist?

src/mesh/index_derivs.cxx
ZedThree: Good point, I think that should be sufficient

ZedThree:
Should probably add a region argument to `Laplace_perpXY` at some point too

johnomotani:
> Should probably add a region argument to `Laplace_perpXY` at some point too

`Laplace_perpXY` gets around the issue addressed in this PR by (as I understand it) assuming x- and y-directions are orthogonal. If I remember right, all the finite-volume stuff requires x-y orthogonal. That means `Laplace_perpXY` doesn't need the extra region argument. Also, we don't provide region arguments for the operators in `difops.hxx`, only for the lower-level derivatives `DDY`, `D2DXDY`, etc., for odd corner cases where they are useful internally.

PR #1941, opened 2020-03-03
MsgStack and PetscLib thread safety

ZedThree:
I've not actually managed to make a test for `MsgStack` that fails without this change and passes with it. For example:

```cpp
TEST(MsgStackTest, ThreadSafety) {
  MsgStack msg_stack;

  const std::string expected_dump = "====== Back trace ======\n";

  BOUT_OMP(parallel for)
  for (int i = 0; i < 1000; i++) {
    int id = msg_stack.push("Message", i);
    msg_stack.pop(id);
  }
  auto dump = msg_stack.getDump();
  EXPECT_EQ(dump, expected_dump);
}
```

looks like something this change should fix, but doesn't.

@johnomotani Any chance you've got an example of something that was failing before?

We could use standard library `std::lock_guard` with a custom mutex here, that might also let us be thread safe for other types of threads, if anyone is ever likely to use them.

### DIFF
--- a/src/sys/msg_stack.cxx
+++ b/src/sys/msg_stack.cxx
@@ -32,7 +32,7 @@
 
 #if CHECK > 1
 int MsgStack::push(std::string message) {
-  BOUT_OMP(critical(MsgStack_push)) {
+  BOUT_OMP(critical(MsgStack)) {
     if (position >= stack.size()) {
       stack.push_back(std::move(message));
     } else {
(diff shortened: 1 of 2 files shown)


johnomotani:
> @johnomotani Any chance you've got an example of something that was failing before?

I don't think I ever had a test that was failing because of `MsgStack`, it was just that TSan was complaining about it.

PR #1944, opened 2020-03-04
Add ZInterpolation, ZHermiteSpline

ZedThree:
For interpolation in Z only

Cherry-picked, rebased, squashed and formatted from #1897 but into `next`


ZedThree:
I've made some minor changes:

- make more things `const`
- add a constructor for `BoutMask` that takes `Mesh*` in order to make the `ZInterpolation`/`ZHermiteSpline` constructors a bit cleaner
- `explicit` on new constructors
- simplified the `getWeights` method by using aggregate construction of the `vector<PositionAndWeights>`

PR #1955, opened 2020-03-08
Fix possible data race in MsgStack::push()

johnomotani:
Fix for issue noted by @d7919 here https://github.com/boutproject/BOUT-dev/pull/1803#discussion_r385240172

Would it be better with a `BOUT_OMP(single)` as suggested here https://github.com/boutproject/BOUT-dev/pull/1803#discussion_r385242185?

### DIFF
--- a/src/sys/msg_stack.cxx
+++ b/src/sys/msg_stack.cxx
@@ -53,20 +53,22 @@ int MsgStack::setPoint() {
 }
 
 void MsgStack::pop() {
-  if (position <= 0)
+  if (position <= 0) {
     return;
-  BOUT_OMP(atomic)
-  --position;
+  }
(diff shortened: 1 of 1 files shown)


ZedThree:
`single` means we'd only get the message stack from one thread, which implies they should all be doing the same thing and precludes different tasks.

We currently only apply OpenMP to loops, so this would be fine. I don't believe we have plans to do any other kind of task sharing, so I would be happy making it `single`.

ZedThree:
I think we also need `single` in `MsgStack::pop`

johnomotani:
It should be `single` in both overloads of `MsgStack::pop` right?

ZedThree:
Yep

PR #1957, opened 2020-03-09
Exclude corner cells in LaplaceXY

ZedThree:
Think this should've been in #1885 maybe. My bad, definitely a downside to breaking #1803 into separate PRs!

This is why half the PR tests are failing. Not sure why the other half are yet


ZedThree:
Lots of errors on Travis because they decided to handle quotes in environment variables differently for a couple of hours.

ZedThree:
Looks like I also broke test-yupdown

johnomotani:
> Looks like I also broke test-yupdown

I'm looking at that one. There's a small error just on the `ystart` points...

ZedThree:
Fixed, and also fixed and tidied up test-yupdown-weights while I was at it.

PR #1958, opened 2020-03-11
Minor fixes for elm-pb and python tools

### LINKED ISSUE #1071: Replacing Bunch in python tools
In a bunch (ha) of modules, we use the `bunch` module. This provides a dict-like data structure that has fields that can be accessed with the dot notation, as well as dict-style. Basically, a combination of `namedtuple` and `dict`.

Unfortunately, it doesn't have great support for Python 3, and looks to have been replaced by a few forked projects.

I'm in favour of just replacing it everywhere with plain `dict`s. This would be a breaking change though, so would probably have to wait awhile.

dschwoerer:
Resolves #1071


ZedThree:
Thanks! Is there supposed to be a `boututils/bunch.py` in this PR?

To be honest, I think there's a ton of stuff in pylib we can just delete in v5.

tools/pylib/boututils/moment_xyzt.py
```diff
@@ -30,8 +29,6 @@ def moment_xyzt( sig_xyzt, *args):#rms=None, dc=None, ac=None):
 #; -AC (DC subtracted out), i.e., a function of (x,y,z,t)
 #;-------------------------------------------------------------------
```
johnomotani: It wasn't changed in this PR, but since we're touching the methods here...
This `moment_xyzt` seems to be mis-named - it's actually using `(t,x,y,z)` (I guess this is due to a port from IDL to Python). It would be nice to change the names and comments so that they reflect the actual `(t,x,y,z)` behaviour.

tools/pylib/boututils/surface_average.py
```diff
@@ -35,37 +34,28 @@ def surface_average(var, g, area=None):
...
     if s == 4 :
         nx = np.shape(var)[1]
         ny = np.shape(var)[2]
         nt = np.shape(var)[0]
 
         result = np.zeros((nx,nt))
         for t in range (nt):
-
-            result[:,t] = surface_average(var[t,:,:,:], g, area=area)
+            result[:,t] = surface_average(var[t,:,:,:], grid, area=area)
```
johnomotani: I'd prefer to switch the result so it's `result[t,x]`.

tools/pylib/boututils/moment_xyzt.py
dschwoerer: Sorry, but I would rather remove them.
I think what the function does is to simply to put actually in a library ...

tools/pylib/boututils/surface_average.py
dschwoerer: I would rather remove it ... If you think this function is useful I'd do it though ...

tools/pylib/boututils/moment_xyzt.py
johnomotani: I don't have strong feelings about it. Given what @ZedThree said in the BUG meeting today about getting rid of stuff in the old pylib, I'd vote for leaving as is for now (get this PR merged), and remove stuff from pylib in a dedicated PR.

dschwoerer:
> Thanks! Is there supposed to be a `boututils/bunch.py` in this PR?

Yes. I had `git clean`ed it, but found it in a backup.

PR #1959, opened 2020-03-11
fix for running unit-test as root

dschwoerer:
This is the only test that fails if we run as root.
Would be nice to not having to create a user etc and still run the tests


ZedThree:
Thanks, good catch!

Ideally, we'd be able to mock out the file system for the unit tests, but then that would require us to be able to inject the mock class somehow.

PR #1960, opened 2020-03-11
Field3d copy parallel slices

ZedThree:
Copying Field3Ds should be cheap, so copy any parallel slices from the rhs

This modifies the implementation from #1803 by just copying the parallel slice vectors themselves, rather than copying the fields in them. This removes the need to keep track of a separate bool for whether or not to copy the parallel slices of parallel slices.

I think the bool was originally there to avoid an infinite recursion when creating parallel slices using the identity transform, which just copies the field into its own slices, relying on copy-on-write to make this cheap. This implementation makes an initial copy and ensures that that doesn't have parallel slices, and uses that copy to fill in the original's parallel slices. This should all be cheap thanks to copy-on-write.

Also adds a move assignment operator to Field3D to make this even cheaper by moving the parallel slice vectors.

### DIFF
--- a/include/field3d.hxx
+++ b/include/field3d.hxx
@@ -271,27 +271,23 @@ class Field3D : public Field, public FieldData {
   /// Return reference to yup field
   Field3D &yup(std::vector<Field3D>::size_type index = 0) {
     ASSERT2(index < yup_fields.size());
-    ASSERT3(!yup_fields[index].copy_yupdown_fields);
     return yup_fields[index];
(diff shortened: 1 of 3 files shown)


d7919:
Is there any data on the perfomance impact (if any)?

ZedThree:
I compared elm-pb on 8 cores, and the difference between this and next was about 10%, in favour of this branch, but the jitter was ~20%.

PR #1962, opened 2020-03-11
Add region argument to XZInterpolations

ZedThree:
Cherry-picked and formatted from #1803 + `const`-ification of local vars

### DIFF
--- a/include/interpolation_xz.hxx
+++ b/include/interpolation_xz.hxx
@@ -51,17 +51,22 @@ public:
   }
   virtual ~XZInterpolation() = default;
 
-  virtual void calcWeights(const Field3D &delta_x, const Field3D &delta_z) = 0;
-  virtual void calcWeights(const Field3D &delta_x, const Field3D &delta_z,
-                           const BoutMask &mask) = 0;
-
-  virtual Field3D interpolate(const Field3D &f) const = 0;
-  virtual Field3D interpolate(const Field3D &f, const Field3D &delta_x,
-                              const Field3D &delta_z) = 0;
-  virtual Field3D interpolate(const Field3D &f, const Field3D &delta_x,
-                              const Field3D &delta_z, const BoutMask &mask) = 0;
 
   void setMask(const BoutMask &mask) { skip_mask = mask; }
+  virtual void calcWeights(const Field3D& delta_x, const Field3D& delta_z,
+                           const std::string& region = "RGN_NOBNDRY") = 0;
+  virtual void calcWeights(const Field3D& delta_x, const Field3D& delta_z,
+                           const BoutMask& mask,
+                           const std::string& region = "RGN_NOBNDRY") = 0;
+
+  virtual Field3D interpolate(const Field3D& f,
+                              const std::string& region = "RGN_NOBNDRY") const = 0;
+  virtual Field3D interpolate(const Field3D& f, const Field3D& delta_x,
+                              const Field3D& delta_z,
+                              const std::string& region = "RGN_NOBNDRY") = 0;
+  virtual Field3D interpolate(const Field3D& f, const Field3D& delta_x,
+                              const Field3D& delta_z, const BoutMask& mask,
+                              const std::string& region = "RGN_NOBNDRY") = 0;
 
   // Interpolate using the field at (x,y+y_offset,z), rather than (x,y,z)
   int y_offset;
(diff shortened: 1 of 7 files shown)


johnomotani:
I think the test failure is due to a race condition in `MsgStack`. I can reproduce with 2 threads on my desktop. `BoutMask::operator()` is called inside a `BOUT_FOR` (in this case at`lagrange_4pt_xz.cxx:48`). `BoutMask::operator()` has a `TRACE()` which calls `MsgStack::push()`. If I change the `BOUT_OMP(single)` calls in `msg_stack.cxx` to `BOUT_OMP(critical(MsgStack))` then `fci_mms` runs fine. I don't understand why `BOUT_OMP(single)` was incorrect though...

ZedThree:
This is hanging with OpenMP in `MsgStack::push`, via the call to `BoutMask::operator()` in `XZLagrange4pt::calcWeights`: https://github.com/boutproject/BOUT-dev/blob/d99e318f251229edb3eff4fa7a382150e3c5c32d/src/mesh/interpolation/lagrange_4pt_xz.cxx#L43-L48

Rolling back the `MsgStack` changes to the `BOUT_OMP(critical(MsgStack))` form makes this pass.

The OpenMP (5.0) spec says:

> The single construct specifies that the associated structured block is executed by only one of the threads in the team (not necessarily the master thread), in the context of its implicit task. The other threads in the team, which do not execute the block, wait at an implicit barrier at the end of the single construct unless a nowait clause is specified.

And this pragma is inside a conditional. I added `nowait` to the `single` form and it mostly worked, but now I'm getting random `double free or corruption`.

An alternative is to replace the `BoutMask` with a region. However, this would then need to be combined with the `region` argument to `calcWeights`/`interpolate`, which might be expensive. Some kind of caching maybe?

johnomotani:
I added some `output<<` statements to see which thread is doing what. It looks like for me the `BOUT_FOR` in `XZLagrange4pt::calcWeights` is only using thread 0, I get no output from thread 1; the print statement there is `output<<"calcWeights, thread "<<omp_get_thread_num()<<endl;`. For comparison I put a similar `output<<"operator - thread "<<omp_get_thread_num()<<endl;` in `operator-(Field3D,Field3D)`, which did give output from both thread 0 and thread 1. Don't understand why this should be true, but I guess this is why `BOUT_OMP(single)` is hanging in `calcWeights()`.

It's not quite the same, but I guess `BOUT_OMP(single)` may fail inside `BOUT_FOR` generally because each thread does not necessarily do the same number of iterations, and any thread that does more iterations than another will get to the barrier in a `BOUT_OMP(single)` and hang.
@ZedThree when you used `nowait`, maybe the double free or corruption is because the second thread carries on to the `MsgStackItem` destructor (which calls `MsgStack::pop()`) and tries to run that before the first thread finishes the constructor (which calls `MsgStack::push()`)?

Preferring `BOUT_OMP(single)` to `BOUT_OMP(critical)` is for performance right? But all the `MsgStack` stuff is skipped for optimized builds anyway, so I'd think it's not a problem to use `critical`?

d7919:
> I added some `output<<` statements to see which thread is doing what. It looks like for me the `BOUT_FOR` in `XZLagrange4pt::calcWeights` is only using thread 0, I get no output from thread 1; the print statement there is `output<<"calcWeights, thread "<<omp_get_thread_num()<<endl;`. For comparison I put a similar `output<<"operator - thread "<<omp_get_thread_num()<<endl;` in `operator-(Field3D,Field3D)`, which did give output from both thread 0 and thread 1. Don't understand why this should be true, but I guess this is why `BOUT_OMP(single)` is hanging in `calcWeights()`.
> 
> It's not quite the same, but I guess `BOUT_OMP(single)` may fail inside `BOUT_FOR` generally because each thread does not necessarily do the same number of iterations, and any thread that does more iterations than another will get to the barrier in a `BOUT_OMP(single)` and hang.
> @ZedThree when you used `nowait`, maybe the double free or corruption is because the second thread carries on to the `MsgStackItem` destructor (which calls `MsgStack::pop()`) and tries to run that before the first thread finishes the constructor (which calls `MsgStack::push()`)?
> 
> Preferring `BOUT_OMP(single)` to `BOUT_OMP(critical)` is for performance right? But all the `MsgStack` stuff is skipped for optimized builds anyway, so I'd think it's not a problem to use `critical`?

After discussion with @ZedThree I think the _correct_ fix is to use an arena for the msg_stack so each thread gets their own stack. Currently with single or critical the produced msg_stack can be misleading for the case where different threads may take different paths. This could be as easy as replacing `std::vector<std::string>` with `Array<std::vector<std::string>>` and then adding `[0]` in a few places inside the msg stack header (possibly tidier ways).

ZedThree:
I think because it's getting to `MsgStack::push` via a conditional inside a `parallel` region, and I suspect that all the threads are expected to hit the `single` region, this is causing problems.

Having spoken to David about this, we've come to the conclusion that both `single` (or `master`) and `critical` are the wrong thing to do: `single` because you'll only get the stack from one thread, and `critical` because threads may pop out of order, etc.

He's just beaten me to posting a message here, but I'm now trying with `Array<std::vector<std::string>> stack{1}`

ZedThree:
5f1fe66 is a temporary fix to get this PR in. Essentially does an immediate return from `MsgStack::push/pop` if there's currently more than one thread. This gives us thread safety at the expense of not getting messages from inside OpenMP regions.

A better fix would be to use an arena, but just making `MsgStack::stack` an `Array<vector<string>>` doesn't work because the `Array` is created outside the OpenMP regions.
I don't think it's a complicated fix, but one for another PR, and should have some tests!

There's two other issues we've discovered with `MsgStack` due to this:

1. the index derivs machinery uses `MsgStack`, but the registration bits shouldn't as they happen at initialisation time, before `main`. This can happen before the global `msg_stack` gets created, which is just asking for trouble

2. the interpolations' `skip_mask` should probably be dropped entirely, and require the user to pass the correct `region`. I very much doubt this will affect user code, and it's basically only the FCI transform that would need updating.

PR #1966, opened 2020-03-18
Fix collect when given similar name

bendudson:
It's useful when doing interactive analysis to have a bit of flexibility / DWIM in the collect routine. This fixes a bug which broke this feature. Now works again.

Added a test to reproduce problem and fix.


dschwoerer:
I would personally prefer to not do this at all. I would prefer just giving a list of options that the user could have meant, including similar names (e.g. compare with [Levenshtein](https://en.wikipedia.org/wiki/Levenshtein_distance) )
While not as nice as otherwise in interactive use, it is much safer in scripts, if e.g. a simulation is changed and now saves `temp_log` rather then `temp`. A non-adopted script will try to read `temp`, but will get `temp_log` and give wrong answers ...
I know there is the `strict` keyword, but I would prefer to do the right thing ...

bendudson:
I agree that this functionality could potentially lead to nasty surprises, but it's such a pain to remember whether it's "n" or "Ne" or "ne" that I should be collecting. Probably this says something about consistency of variable names in models.
Since this is fixing a current bug, I think we should fix this in master. In `next` perhaps it would be best as @dschwoerer  suggests to apply a different fix, in which the possible corrections are suggested but an error still occurs.

dschwoerer:
Yes, fixing in master, and changing in next would be great. I will put in a PR for next.

PR #1968, opened 2020-03-19
Translations 4.3.1

dschwoerer:
Waiting if @bendudson has time to update fr and/or zh and @estaremp to update es


ZedThree:
I suspect not, better to get this in without the other languages than hold up 4.3.1

bendudson:
Thanks @dschwoerer sorry I probably won't get to this for a while

PR #1969, opened 2020-03-19
Install make.config with header files

dschwoerer:
Installing into datadir causes issues as make.config contains many
settings and is thus not interchangable between different setups. If
the FHS [1] is used, different installations would conflict. This
has caused issues for fedora, where different MPI installations
share @datadir@.

[1] https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard

### DIFF
--- a/make.config.in
+++ b/make.config.in
@@ -183,32 +183,31 @@ install: libfast
 	$(INSTALL_DATA)  tools/idllib/README $(DESTDIR)@datadir@/bout++/idllib/
 	$(INSTALL_DATA)  tools/pylib/boutdata/*.py $(DESTDIR)@datadir@/bout++/pylib/boutdata/
 	$(INSTALL_DATA)  tools/pylib/boututils/*.py $(DESTDIR)@datadir@/bout++/pylib/boututils/
-	$(INSTALL_DATA)  make.config $(DESTDIR)@datadir@/bout++/
+	$(INSTALL_DATA)  make.config $(INSTALL_INCLUDE_PATH)
 	for mo in $(MO_FILES); do $(MKDIR) $(DESTDIR)@localedir@/`dirname $$mo`; $(INSTALL_DATA) locale/$$mo $(DESTDIR)@localedir@/$$mo; done
 	$(POST_INSTALL)    # Post-install commands follow.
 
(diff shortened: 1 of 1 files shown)


ZedThree:
Did you mean to also remove the mpark variant stuff?

dschwoerer:
Would you consider merging?

No, was an accident. Should be fixed now :+1:

ZedThree:
It's not clear to me that `@includedir@` is definitely the correct place given that this isn't a header file, but as you point out, `@datadir@` is supposed to be architecture-independent. I'm not suggesting we copy PETSc, but they put their equivalent files under `lib/petsc/conf`. CMake and pkg-config both install their files under `@libdir@` for me.

The GNU installation standard suggests maybe `@sysconfdir@`, `${prefix}/etc`? https://www.gnu.org/prep/standards/html_node/Directory-Variables.html

Also, I'm currently on a Fedora machine, and there's a `/usr/lib64/openmpi/share` directory -- is this not `datadir`? Is Fedora doing something other than setting `prefix=/usr/lib64/openmpi`?

I'm a bit less sure about this than I was initially. To install multiple versions of BOUT++, presumably you'd need to set `prefix` differently. Why then would they end up installing `make.config` in the same place?

I think this is likely fine, I'd just like to understand why it's necessary!

dschwoerer:
Fedora is doing something different. For example header files are in `/usr/include/openmpi-x86_64/bout++/`
More documentation:
https://fedoraproject.org/wiki/Packaging:MPI#Packaging_of_MPI_compilers

We could use a different `@datadir@`, but that would result in duplicated `.mo` files, which isn't as nice. Also it violates the idea that datadir should be arch-independent. Some paths are arch-dependent, mostly due to multilib support.

`@sysconfdir@` would avoid the conflict, but that would suggest the user can edit those files. However, this will most likely brake the installation, so I don't think that will work. `@libdir@` might be better ...

ZedThree:
`@sysconfdir@` should be for read-only files according to that GNU document.

I don't have a strong opinion, as long as it works and is not obviously wrong.

dschwoerer:
I think read-only refers to the program. This is basically `/etc` where the user/sysadmin can edit config files.

PR #1970, opened 2020-03-19
Compat make.config for make 4.3

dschwoerer:
We are affected by the first of the non-compatible changes in [gnu make 4.3](https://fedoraproject.org/wiki/Changes/MAKE43)

Still waiting on local tests to finish.

Should I add a test on travis for fedora?

### DIFF
--- a/make.config.in
+++ b/make.config.in
@@ -131,8 +131,6 @@ $(DIRS):
 	@$(MAKE) -s --no-print-directory TARGET=$(TARGET) -C $@ $(TARGET)
 else
 MODULE_DIR?=$(shell pwd)
-## Workaround for different make versions
-H := \#
 $(DIRS):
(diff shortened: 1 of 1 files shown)


ZedThree:
Urgh, that's a pain. An alternative is to use `$(basename f)`. Personally I find the `${f##*/}` syntax not especially enlightening.

dschwoerer:
Switched to basename. You are right, that is much cleaner, and requieres less fidelling, as e.g. `a/b/c///` is now working as well :+1:

ZedThree:
Thanks for finding and fixing this!

dschwoerer:
Now BOUT++ successfully builds on several arches: https://koji.fedoraproject.org/koji/taskinfo?taskID=42619777

PR #1971, opened 2020-03-19
Fix documented requirements

### LINKED ISSUE #1863: build fails with gcc 4.8 on CentOS 7
The build fails on CentOS 7:
```
../../../../../include/fieldperp.hxx: At global scope:
../../../../../include/fieldperp.hxx:69:3: warning: unused parameter 'rhs' [-Wunused-parameter]
In file included from ../../../../../include/options.hxx:50:0,
                 from ../../../../../include/bout/deriv_store.hxx:42,
                 from ../../../../../include/bout/mesh.hxx:49,
                 from naulin_laplace.cxx:140:
../../../../../include/field3d.hxx: In function 'FieldPerp pow(const Field3D&, const FieldPerp&, REGION)':
../../../../../include/field3d.hxx:586:37: note: synthesized method 'FieldPerp::FieldPerp(FieldPerp&&)' first required here 
   return pow(lhs, rhs, toString(rgn));
                                     ^
In file included from ../../../../../include/difops.hxx:45:0,
                 from naulin_laplace.cxx:144:
../../../../../include/bout/solver.hxx: At global scope:
../../../../../include/bout/solver.hxx:345:39: error: invalid use of incomplete type 'class std::unique_ptr<T>'
     std::unique_ptr<T> MMS_err{nullptr}; /// Error for MMS
                                       ^
In file included from /usr/include/c++/4.8.2/memory:81:0,
                 from ../../../../../include/bout/array.hxx:31,
                 from ../../../../../include/utils.hxx:36,
                 from ../../../../../include/options.hxx:45,
                 from ../../../../../include/bout/deriv_store.hxx:42,
                 from ../../../../../include/bout/mesh.hxx:49,
                 from naulin_laplace.cxx:140:
/usr/include/c++/4.8.2/bits/unique_ptr.h:109:11: error: declaration of 'class std::unique_ptr<T>'
     class unique_ptr
           ^
make[5]: *** [naulin_laplace.o] Error 1
make[4]: *** [naulin] Error 2
make[4]: *** Waiting for unfinished jobs....
make[3]: *** [impls] Error 2
make[2]: *** [laplace] Error 2
make[1]: *** [invert] Error 2
make: *** [src] Error 2
```

I suspect that gcc 4.8 is to blame.
Do we intend to support 4.8?

How to reproduce:
```
docker run --rm -it centos:7 /bin/bash

yum install git gcc-c++ mpich-devel epel-release
yum install netcdf-cxx-devel netcdf-devel make fftw-devel
yum install $(EDITOR)
useradd build
su build

cd && git clone https://github.com/boutproject/BOUT-dev.git
cd ~/BOUT-dev/ && module load mpi && ./configure && make -j 4
```

ZedThree:
We need gcc >= 4.9 due to a bug in previous versions. FFTW is now an optional dependency.

Closes #1863

### DIFF
--- a/manual/sphinx/user_docs/installing.rst
+++ b/manual/sphinx/user_docs/installing.rst
@@ -126,7 +126,8 @@ The bare-minimum requirements for compiling and running BOUT++ are:
    MPICH ( `https://www.mpich.org/ <https://www.mpich.org/>`__) or
    LAM (`www.lam-mpi.org/ <www.lam-mpi.org/>`__)
    
-#. The NetCDF library ( `https://www.unidata.ucar.edu/downloads/netcdf <https://www.unidata.ucar.edu/downloads/netcdf>`__ )
+#. The NetCDF library (`https://www.unidata.ucar.edu/downloads/netcdf
+   <https://www.unidata.ucar.edu/downloads/netcdf>`__)
    
 The FFTW-3 library (`http://www.fftw.org/ <http://www.fftw.org/>`__)
 is also strongly recommended. Fourier transforms are used for some
@@ -136,7 +137,7 @@ FFTW-3, these options will not be available.
 
 .. note::
    Only GCC versions >= 4.9 are supported. This is due to a bug in
-   previous versions
+   previous versions.
 
 .. note::
    If you use an Intel compiler, you must also make sure that you have
--- a/manual/sphinx/user_docs/installing.rst
+++ b/manual/sphinx/user_docs/installing.rst
@@ -131,6 +131,10 @@ The bare-minimum requirements for compiling and running BOUT++ are:
 The FFTW-3 library ( `http://www.fftw.org/ <http://www.fftw.org/>`__ ) is also strongly recommended
 
 .. note::
+   Only GCC versions >= 4.9 are supported. This is due to a bug in
+   previous versions
+
+.. note::
    If you use an Intel compiler, you must also make sure that you have
    a version of GCC that supports C++11 (GCC 4.8+).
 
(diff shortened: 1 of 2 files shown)


manual/sphinx/user_docs/installing.rst
```diff
@@ -130,6 +130,10 @@ The bare-minimum requirements for compiling and running BOUT++ are:
    
 The FFTW-3 library ( `http://www.fftw.org/ <http://www.fftw.org/>`__ ) is also strongly recommended
 
+.. note::
+   Only GCC versions >= 4.9 are supported. This is due to a bug in
+   previous versions
+
 .. note::
    If you use an Intel compiler, you must also make sure that you have
    a version of GCC that supports C++11 (GCC 4.8+).
```
dschwoerer: probably update here as well

manual/sphinx/user_docs/installing.rst
```diff
@@ -130,6 +130,10 @@ The bare-minimum requirements for compiling and running BOUT++ are:
    
 The FFTW-3 library ( `http://www.fftw.org/ <http://www.fftw.org/>`__ ) is also strongly recommended
```
dschwoerer: Maybe make it clear that it is not requiered, but some features will be missing?
Though I am happy to keep as is

manual/sphinx/user_docs/installing.rst
ZedThree: I think this one is fine -- Intel uses the GCC standard library, but the bug is in the compiler front end, so it shouldn't affect us here.

manual/sphinx/user_docs/installing.rst
ZedThree: Thanks, fixed

PR #1972, opened 2020-03-19
Remove -Og from debug flags

### LINKED ISSUE #1879: Debugging optimization flags
While debugging a build configured with `./configure --enable-debug ...` with gdb, at a certain point in the program (but not at other points) I got `value has been optimized out` messages for a bunch of variables. Removing the `-Og` flag (by editing `make.config`), so that `-O0` was used fixed this problem. Compiler was gcc-4.6. I don't think this change is a 'fix' because [gcc's documentation](https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html) says
> Optimize debugging experience. -Og should be the optimization level of choice for the standard edit-compile-debug cycle, offering a reasonable level of optimization while maintaining fast compilation and a good debugging experience. It is a better choice than -O0 for producing debuggable code because some compiler passes that collect debug information are disabled at -O0.

so I'm just posting this in case other people run into the same issue.

ZedThree:
In gcc at least, this flag optimises too much, making it less useful
for debugging.

Fixes #1879

### DIFF
--- a/configure
+++ b/configure
@@ -5928,8 +5928,7 @@ if test "$enable_debug" != ""; then :
   { $as_echo "$as_me:${as_lineno-$LINENO}: Enabling all debug options" >&5
 $as_echo "$as_me: Enabling all debug options" >&6;}
   enable_checks="3"
-  # use -Og with available, otherwise fall back to -O0
-  OPT_FLAGS="-g -O0 -Og -fno-inline -hipa1"
+  OPT_FLAGS="-g -O0 -fno-inline -hipa1"
 
 else
 
--- a/configure.ac
+++ b/configure.ac
@@ -262,8 +262,7 @@ OPT_FLAGS=""
 AS_IF([test "$enable_debug" != ""], [
   AC_MSG_NOTICE([Enabling all debug options])
   enable_checks="3"
-  # use -Og with available, otherwise fall back to -O0
-  OPT_FLAGS="-g -O0 -Og -fno-inline -hipa1"
+  OPT_FLAGS="-g -O0 -fno-inline -hipa1"
 ], [
   AS_IF([test "x$enable_optimize" != "xno"], [
     AS_CASE(["$enable_optimize"],


ZedThree:
This is failing on clang, because it turns out our signal handler is non-standards compliant :disappointed: 

https://en.cppreference.com/w/cpp/utility/program/signal

> POSIX requires that signal is thread-safe, and specifies a list of async-signal-safe library functions that may be called from any signal handler.

> Signal handlers are expected to have C linkage and, in general, only use the features from the common subset of C and C++. It is implementation-defined if a function with C++ linkage can be used as a signal handler.

See also here for reasoning: https://stackoverflow.com/a/16891799/2043465

This excludes even stuff like `printf`. It seems that the correct way to do this is to have the signal handler set a `volatile sigatomic_t flag`, and have another function that is run periodically check this flag and handle it there. I have no idea where we would stick that polling function, given that the main purpose of this is to give us a backtrace in the event of a segfault.

I don't think there's any way to get the backtrace directly in the signal handler in a compliant fashion, given that it will necessarily involve memory allocation, which is apparently a no-no here.

On the other hand, this mostly works! Which is sort of surprising.

Options:

1. Close this PR. Downside is `-Og` does cause real problems for debugging, plus this behaviour might just come bite us in the bum later. Upside is it might not!
2. Stick `-Og` back in if using Clang and the signal handler. Downside: urgh, no.
3. Rewrite the signal handler to be compliant. Downside: time! Plus, while I think we can make this work for some signals, I'm not sure it's possible for all of them. Upside: it might work all the time.
4. Add `--disable-signal` for the Clang build on Travis, note in the docs that the signal handler is technically non-compliant. Downside: still kicking the problem down the road. Upside: we might be able to do this indefinitely.

Personally, my preference is for option 4, and maybe investigate option 3 longer term. @bendudson @dschwoerer any thoughts?

ZedThree:
Travis are having some issues reporting checks at the mo, but both builds have actually passed.

PR #1974, opened 2020-03-19
Conditionally show backtrace

dschwoerer:
Read the environment variable BOUT_SHOW_BACKTRACE.
If the variable is set, show the backtrace, otherwise just show the
message.
The allows to get the backtrace without having to attache gdb, or catch
the exception. The default behaviour is not changed.


dschwoerer:
I was debugging test-communications but it could also be quiet useful for the python interface. I think getenv should be sufficiently portable, otherwise it could be easily wrapped in an ifdef and detected at configure time ...

include/boutexception.hxx
```diff
@@ -28,6 +28,11 @@ public:
   ~BoutException() override;
 
   const char* what() const noexcept override {
+    if (getenv("BOUT_SHOW_BACKTRACE") != nullptr) {
```
ZedThree: This is portable, but should be `std::getenv` from `<cstdlib>`

https://en.cppreference.com/w/cpp/utility/program/getenv

ZedThree:
Neat, thanks!

Could the integrated test be a unit test instead? It turns out `setenv/unsetenv` aren't actually in the standard for some reason :disappointed: but it's possible to write a small wrapper/`#ifdef` something reasonable: https://github.com/googleapis/google-cloud-cpp/issues/100

Also interesting:
https://www.greenend.org.uk/rjk/tech/putenv.html

dschwoerer:
I am not sure a unit test is much better, as that would always be build, whereas integrated tests can be easily disabled by default. We are spending most time compiling, so I prefer having it separated.

Also `setenv`/`putenv` not being standard is another reason to not put it in a unit test.

ZedThree:
That is a downside of the unit tests currently, yes. But incremental builds should be pretty fast.

The benefits of having unit tests are having as much stuff as possible in the same framework, which allows easy filtering, reruns, etc, and can really benefit run time.
The other advantage is that stuff in the unit tests gets the coverage run, and means we don't need to come up with some way of working out which integrated tests can be stuck into the coverage run and how to do that in a nice way.

dschwoerer:
It might be a bit tricky, because we probably want to restore the state afterwards, because that might very well be used to debug failures in the unit tests ...

However, currently it does break the unit tests, if the env var is set -.-

dschwoerer:
I think in this case it might make sense to keep it outside of the unit-tests.
As `setenv` is not standard, and also might be leaking memory. I am hesitant to put it in the unit tests, as it might cause more issues, and having a separate test doesn't seem that bad to me.

ZedThree:
Alright, you've convinced me :) This is a useful feature, we'll just have to start adding some integrated tests to the coverage build!

At the moment, this test won't be built for CMake -- might be easier to convert `runtest` to a python script in order to use `run_wrapper.build_and_log`.

manual/sphinx/user_docs/advanced_install.rst
```diff
@@ -29,6 +29,10 @@ For (sometimes) more useful error messages, there is the
 ``--enable-track`` option. This keeps track of the names of variables
 and includes these in error messages.
 
+To get a backtrace, you can set the environment variable
+`BOUT_SHOW_BACKTRACE` in order for the exception to include the
```
ZedThree, line 33: This one always catches me too, rst needs double backticks:
```suggestion
``BOUT_SHOW_BACKTRACE`` in order for the exception to include the
```

tests/integrated/test-backtrace/makefile
```diff
@@ -0,0 +1,6 @@
+
+BOUT_TOP	= ../../..
```
ZedThree, line 2:
```suggestion
BOUT_TOP	?= ../../..
```

tests/integrated/test-backtrace/runtest
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+set -ex
+
+#requires all_tests
+
+e=1
+make || e=$?
```
ZedThree, line 7: This won't work if using CMake and a non-makefile build generator, so it might be easier to convert this to a python script

ZedThree (changes requested):

ZedThree:
I've moved the added code from `what()` to the `BoutException` constructor because it was returning a pointer to a temporary and so didn't work. Because `what()` is `const` and needs to return `const char*`, we need to have made the message already -- the only place to do this is the constructor.

I also converted the `runtest` to Python, so now it works when using CMake + Ninja.

ZedThree (approved):

ZedThree:
@dschwoerer Look ok to you? Still works for your use case?

PR #1975, opened 2020-03-19
Read staggered J and Bxy from grid if present (master)

### LINKED ISSUE #1883: Minor bug in coordinates for staggered grids
If dx_ylow is present in the grid file (similar problem for _xlow), BOUT++  read dx_ylow, dy_ylow and the metric components (at CELL_YLOW) from the grid file, but does not read J_ylow and Bxy_ylow. It only computes J and Bxy from the metric components at CELL_YLOW:  https://github.com/boutproject/BOUT-dev/blob/1ab0d0030ca971a62d128f1a844232949ff7ad3d/src/mesh/coordinates.cxx#L510 
This is not consistent with what it is done for CELL_CENTRE, where it first computes J and Bxy from the metric components:
https://github.com/boutproject/BOUT-dev/blob/1ab0d0030ca971a62d128f1a844232949ff7ad3d/src/mesh/coordinates.cxx#L339 then tries to read J and Bxy from the grid file:
https://github.com/boutproject/BOUT-dev/blob/1ab0d0030ca971a62d128f1a844232949ff7ad3d/src/mesh/coordinates.cxx#L344 and https://github.com/boutproject/BOUT-dev/blob/1ab0d0030ca971a62d128f1a844232949ff7ad3d/src/mesh/coordinates.cxx#L360 and if successful, print the difference between the computed J and Bxy, Jcalc and Bcal, and the ones in the grid files and store these values for computation.

johnomotani:
Previously staggered `Coordinates` calculated `J` and `Bxy` from the staggered metric components, even if they were present in the grid file (or `[mesh]` section of `BOUT.inp`). The input values should be preferred if they are available (as for the `CELL_CENTRE` version).

Fixes #1883.

### DIFF
--- a/src/mesh/coordinates.cxx
+++ b/src/mesh/coordinates.cxx
@@ -510,14 +510,15 @@ Coordinates::Coordinates(Mesh* mesh, Options* options, const CELL_LOC loc,
 
     /// Calculate Jacobian and Bxy
     if (jacobian()) {
-      throw BoutException("Error in jacobian call while constructing staggered Coordinates");
+      throw BoutException("Error in jacobian call while constructing staggered "
+                          "Coordinates");
     }
 
     // Attempt to read J from the grid file
     Field2D Jcalc = J;
     if (getAtLoc(mesh, J, "J", suffix, location)) {
       output_warn.write(
-          "\tWARNING: Jacobian 'J_{}' not found. Calculating from metric tensor\n",
+          "\tWARNING: Jacobian 'J_%s' not found. Calculating from metric tensor\n",
           suffix);
       J = Jcalc;
     } else {
@@ -533,7 +534,7 @@ Coordinates::Coordinates(Mesh* mesh, Options* options, const CELL_LOC loc,
     // Attempt to read Bxy from the grid file
     Field2D Bcalc = Bxy;
     if (getAtLoc(mesh, Bxy, "Bxy", suffix, location)) {
-      output_warn.write("\tWARNING: Magnitude of B field 'Bxy_{}' not found. Calculating "
+      output_warn.write("\tWARNING: Magnitude of B field 'Bxy_%s' not found. Calculating "
                         " from metric tensor\n", suffix);
       Bxy = Bcalc;
     } else {
--- a/src/mesh/coordinates.cxx
+++ b/src/mesh/coordinates.cxx
@@ -357,6 +357,13 @@ Coordinates::Coordinates(Mesh* mesh, Options* options)
     Bxy = sqrt(g_22) / J;
   }
 
+  // Check jacobian
+  bout::checkFinite(J, "J", "RGN_NOCORNERS");
+  bout::checkPositive(J, "J", "RGN_NOCORNERS");
+  if (min(abs(J)) < 1.0e-10) {
+    throw BoutException("\tERROR: Jacobian becomes very small\n");
+  }
+
   // Attempt to read Bxy from the grid file
   auto Bcalc = Bxy;
   if (mesh->get(Bxy, "Bxy")) {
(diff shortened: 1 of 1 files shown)


src/mesh/coordinates.cxx
```diff
@@ -508,7 +510,40 @@ Coordinates::Coordinates(Mesh* mesh, Options* options, const CELL_LOC loc,
 
     /// Calculate Jacobian and Bxy
     if (jacobian()) {
-      throw BoutException("Error in jacobian call while constructing staggered Coordinates");
+      throw BoutException("Error in jacobian call while constructing staggered "
+                          "Coordinates");
+    }
+
+    // Attempt to read J from the grid file
+    Field2D Jcalc = J;
```
dschwoerer: could use auto here :-)

src/mesh/coordinates.cxx
```diff
@@ -508,7 +510,40 @@ Coordinates::Coordinates(Mesh* mesh, Options* options, const CELL_LOC loc,
...
+
+      // Compare calculated and loaded values
+      output_warn.write("\tMaximum difference in J is %e\n", max(abs(J - Jcalc)));
+
+      // Re-evaluate Bxy using new J
+      Bxy = sqrt(g_22) / J;
+    }
+
+    // Attempt to read Bxy from the grid file
+    Field2D Bcalc = Bxy;
```
dschwoerer: could use auto here :-)

src/mesh/coordinates.cxx
```diff
@@ -508,7 +510,40 @@ Coordinates::Coordinates(Mesh* mesh, Options* options, const CELL_LOC loc,
...
+      output_warn.write("\tWARNING: Magnitude of B field 'Bxy_%s' not found. Calculating "
+                        " from metric tensor\n", suffix);
+      Bxy = Bcalc;
+    } else {
+      Bxy = interpolateAndExtrapolate(Bxy, location, extrapolate_x, extrapolate_y);
+
+      output_warn.write("\tMaximum difference in Bxy is %e\n", max(abs(Bxy - Bcalc)));
+      // Check Bxy
+      bout::checkFinite(Bxy, "Bxy" + suffix, "RGN_NOCORNERS");
+      bout::checkPositive(Bxy, "Bxy" + suffix, "RGN_NOCORNERS");
```
dschwoerer: shouldn't this be checked in all cases?
It would imho make sense to remove the check from `jacobian()` - as one might want to provide those values as the interpolation is causing issues.

johnomotani: Yes, that makes a lot of sense.

src/mesh/coordinates.cxx
```diff
@@ -1027,21 +1106,9 @@ int Coordinates::jacobian() {
   // deriving from extrapolated covariant metric components
   J = interpolateAndExtrapolate(J, location, extrapolate_x, extrapolate_y);
 
-  // Check jacobian
-  bout::checkFinite(J, "The Jacobian", "RGN_NOCORNERS");
-  bout::checkPositive(J, "The Jacobian", "RGN_NOCORNERS");
-  if (min(abs(J)) < 1.0e-10) {
-    throw BoutException("\tERROR: Jacobian becomes very small\n");
-  }
```
ZedThree: Did you mean to also remove this check?

johnomotani: I did - I moved it along with the others into the `Coordinates` constructors on the basis that it might fail for the re-calculated Jacobian but pass for the read-in one, so if we have read one in, there is no error.

ZedThree: My bad, I missed the new location!

PR #1976, opened 2020-03-19
Read staggered J and Bxy from grid if present

johnomotani:
Previously staggered `Coordinates` calculated `J` and `Bxy` from the staggered metric components, even if they were present in the grid file (or `[mesh]` section of `BOUT.inp`). The input values should be preferred if they are available (as for the `CELL_CENTRE` version).

I've made this as a fix for #1883. It's not an issue introduced by this PR, but really we should have a test for reading staggered fields from a grid. I don't have time to write one right now, but I think it should be fairly straightforward (@ZedThree?). The grid could be read from a `BOUT.inp`, would just need `g*_YLOW`, `J_YLOW`, etc. as well as the versions without a suffix.


ZedThree:
Might even be possible to write a unit test. I'll investigate

ZedThree:
This got merged in via #1999

Issue #1982, opened 2020-03-22
interpolateAndExtrapolate not working as expected


dschwoerer:
I had some issues with tests failing on coordinate_3d with next merged.

It seems the boundaries are not properly initialised. Specifically I was looking at `tests/integrated/test-coordinates-initialization` which failed with `J` not being positive. There it seems the boundaries (`J.fieldmesh->getBoundaries()`) only contain x boundaries, not y. Therefore only into the x boundaries is interpolated, not the y ones. I did not understand what could have broken this, until I realised this is the same for next.

This patch makes the error obvious:
```
diff --git a/src/mesh/coordinates.cxx b/src/mesh/coordinates.cxx
index a463cba32..c8a365f21 100644
--- a/src/mesh/coordinates.cxx
+++ b/src/mesh/coordinates.cxx
@@ -32,7 +32,16 @@ Field2D interpolateAndExtrapolate(const Field2D& f, CELL_LOC location,
     bool no_extra_interpolate = false) {
 
   Mesh* localmesh = f.getMesh();
-  Field2D result = interp_to(f, location, "RGN_NOBNDRY");
+  Field2D result_ = interp_to(f, location, "RGN_NOBNDRY");
+  result_.allocate();
+  Field2D result = result_;
+  result.allocate();
+  BOUT_FOR(i, result.getRegion("RGN_ALL")){
+    result[i]=BoutNaN;
+  }
+  BOUT_FOR(i, result.getRegion("RGN_NOBNDRY")){
+    result[i]=result_[i];
+  }
   // Ensure result's data is unique. Otherwise result might be a duplicate of
   // f (if no interpolation is needed, e.g. if interpolation is in the
   // z-direction); then f would be communicated. Since this function is used
@@ -40,7 +49,6 @@ Field2D interpolateAndExtrapolate(const Field2D& f, CELL_LOC location,
   // field lines (due to dependence on integrated shear), we don't want to
   // communicate f. We will sort out result's boundary guard cells below, but
   // not f's so we don't want to change f.
-  result.allocate();
   localmesh->communicate(result);
 
   // Extrapolate into boundaries (if requested) so that differential geometry
```
This ensure that after `interp_to` only the region `RGN_NOBNDRY` is valid, and the boundaries contain NaNs.

Am I missing something?
Is there a bug that there are no y boundaries?
Is there a bug in communicate? (the communicate test also fails)

johnomotani:
[...]

PR #1983, opened 2020-03-23
Use custom main instead of PhysicsModel in test-communications

johnomotani:
Avoids error from Pvode solver which checks whether anything is being evolved - partial fix for #1982. Also generally neater not to use `PhysicsModel` when we don't actually need time-evolution.

Also adds `test-communications` to the CMake build.

Most of the lines changed are just whitespace due to a change of indentation in updating `test-communications.cxx`.

### DIFF
--- a/tests/integrated/CMakeLists.txt
+++ b/tests/integrated/CMakeLists.txt
@@ -1,5 +1,6 @@
 add_subdirectory(test-attribs)
 add_subdirectory(test-command-args)
+add_subdirectory(test-communications)
 add_subdirectory(test-coordinates-initialization)
 add_subdirectory(test-cyclic)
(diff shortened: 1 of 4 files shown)


ZedThree:
Failing on Travis because the python f-strings require 3.6, and Ubuntu 16 is stuck on 3.5.

We could upgrade Travis to use Bionic (Ubuntu 18), which has python 3.6, but then we'd be dropping support for everyone else too. I'm not opposed to doing that, but we probably want some wider discussion.

Please could you change this to use `.format` instead?

ZedThree:
Thanks @johnomotani ! :+1:

ZedThree:
Tests have passed, just not reported correctly

Issue #1981, opened 2020-03-22
next test failures


dschwoerer:
One example fails to compile:
```
Trying to compile ../../../examples/laplacexy/alfven-wave
  Compiling  alfven.cxx
In file included from alfven.cxx:2:
../../../include/bout/invert/laplacexy.hxx:39:2: warning: #warning LaplaceXY requires PETSc. No LaplaceXY available [-Wcpp]
   39 | #warning LaplaceXY requires PETSc. No LaplaceXY available
      |  ^~~~~~~
alfven.cxx: In member function ‘virtual int Alfven::init(bool)’:
alfven.cxx:81:33: error: ‘mesh’ was not declared in this scope; did you mean ‘bout::globals::mesh’?
   81 |       laplacexy = new LaplaceXY(mesh);
      |                                 ^~~~
      |                                 bout::globals::mesh
In file included from ../../../include/field_data.hxx:46,
                 from ../../../include/field2d.hxx:35,
                 from ../../../include/options.hxx:49,
                 from ../../../include/bout/deriv_store.hxx:42,
                 from ../../../include/bout/mesh.hxx:49,
                 from ../../../include/bout/invert/laplacexy.hxx:41,
                 from alfven.cxx:2:
../../../include/boundary_region.hxx:13:16: note: ‘bout::globals::mesh’ declared here
   13 |   extern Mesh* mesh; ///< Global mesh
      |                ^~~~
alfven.cxx:89:37: error: ‘mesh’ was not declared in this scope; did you mean ‘bout::globals::mesh’?
   89 |       newSolver = LaplaceXZ::create(mesh);
      |                                     ^~~~
      |                                     bout::globals::mesh
In file included from ../../../include/field_data.hxx:46,
                 from ../../../include/field2d.hxx:35,
                 from ../../../include/options.hxx:49,
                 from ../../../include/bout/deriv_store.hxx:42,
                 from ../../../include/bout/mesh.hxx:49,
                 from ../../../include/bout/invert/laplacexy.hxx:41,
                 from alfven.cxx:2:
../../../include/boundary_region.hxx:13:16: note: ‘bout::globals::mesh’ declared here
   13 |   extern Mesh* mesh; ///< Global mesh
      |                ^~~~
alfven.cxx: In member function ‘virtual int Alfven::rhs(BoutReal)’:
alfven.cxx:108:5: error: ‘mesh’ was not declared in this scope; did you mean ‘bout::globals::mesh’?
[...]

PR #1985, opened 2020-03-23
fix test-laplace-petsc3d

dschwoerer:
fix for part of #1981

### DIFF
--- a/tests/integrated/test-laplace-petsc3d/runtest
+++ b/tests/integrated/test-laplace-petsc3d/runtest
@@ -18,7 +18,7 @@ build_and_log('Laplace 3D with PETSc')
 success = True
 for directory,nproc in test_directories:
     try:
-        command = 'test-laplace3d -d ' + directory + ' > ' + directory + '/output.log 2>&1'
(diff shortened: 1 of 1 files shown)


dschwoerer:
openmpi seems to also include `.` in `PATH`, however mpich doesn't.

ZedThree:
Might also be an idea to just remove the `try ... except`? It's probably more useful to have the actual output printed and not do the rest of the tests, especially on Travis

johnomotani:
> Might also be an idea to just remove the try ... except? It's probably more useful to have the actual output printed and not do the rest of the tests, especially on Travis

I agree. The `try...except` was useful for developing/debugging the solver, but now that the tests do all pass, and should continue to, it would be better to fail immediately if one of them fails.

dschwoerer:
Updated.

In that case [test-communications](https://github.com/boutproject/BOUT-dev/blob/next/tests/integrated/test-communications/runtest) could also be updated to fail early and not print the errors at the end, without context

johnomotani:
> In that case test-communications could also be updated to fail early and not print the errors at the end, without context

Done in #1983.

Issue #1473, opened 2018-12-19
Free 'communicate(Field3D)' function?


johnomotani:
In discussion under #1470, @d7919 suggested
> I wonder if it would be worth introducing a free function `communicate(Field3D &var) { var.getMesh()->communicate(var); return;}`

I like the idea. Could tidy up user code, reduce uses of global mesh pointer, and allow handling of fields with different meshes in the same call (i.e. if f and g have different fieldmeshes, communicate(f, g) could handle calling communicate from each mesh).

d7919:
Of course the downside is that if you're communicating lots of fields on the same mesh this approach of communicating one after the other would lead to lots of small messages rather than a single larger message. I think that's a price worth paying (as we would still provide the more efficient approach as well).

d7919:
If the communicate was done in a non-blocking manner it might actually be possible to loop over the passed fields, starting the communication and then waiting -- this might help overlap work and comms (although my guess is that this wouldn't work very well).

dschwoerer:
I didn't know there was a RFE for this. I implemented this [basically here](https://github.com/boutproject/BOUT-dev/blob/coord3d_merged2/src/mesh/coordinates.cxx#L355)

What header should this go into? Make mesh.hxx more expensive?

The version I implemented assumes all fields are on the same mesh (because I only needed this)
I think it should be easy for the user to call communicate for all meshes separately ...

johnomotani:
Maybe have an `ASSERT2` that the meshes of all fields are the same? I'd think that would be enough...

ZedThree:
> What header should this go into? Make mesh.hxx more expensive?

Maybe `physicsmodel.hxx`? Are we likely to use it much in the library? Quick search looks like it won't be much of a benefit to library code, I think we mostly already have a local `mesh*` for other reasons, and we rarely (ever?) `communicate` multiple fields in library code.

Also, could use some trick like: https://www.fluentcpp.com/2019/03/05/for_each_arg-applying-a-function-to-each-argument-of-a-function-in-cpp/
to just call `Mesh::communicate` on each argument. It's pretty gross without C++17 though.

PR #1986, opened 2020-03-23
Add PhysicsModel::mesh/dump members

dschwoerer:
Last issue in #1981 

EDIT (PH):

This is a step towards removing `bout::globals::mesh/dump`, although we're not there yet

- Add `PhysicsModel::mesh` and `PhysicsModel::dump` members that are initialised to the `bout::globals` variables
- Remove `using namespace bout::globals` from header
- Update tests/examples to be "globals" correct
- Fix some warnings from tests/examples (deprecation, sign-conversion)

Most user code shouldn't be affected by this change, as `mesh` will now resolve to `PhysicsModel::mesh` instead of `bout::globals::mesh`. However, free functions and legacy models (using `physics_init`, `physics_run`) will need to be fixed

Further work:
- Add a constructor to `PhysicsModel` to set the mesh and dumpfile
  - This will then need a delegating constructor added to all user models! It will hopefully be possible to automate this
- Add `communicate` free function
- Remove legacy models
  - Hopefully we can also automate upgrading to `PhysicsModel`


johnomotani:
This change shouldn't be needed - if it fixes compilation then something's gone wrong elsewhere.

`alfven.cxx` includes `bout/physicsmodel.hxx` which includes `bout.hxx` which has `using bout::globals`, so the statement should already be included (the logic being that anything that includes `physicsmodel.hxx` is 'user code' that shouldn't have to worry about the `bout::globals` namespace). One way that it could be failing to be included is if the macro `BOUT_NO_USING_NAMESPACE_BOUTGLOBALS` is defined for some reason - if it's defined then `bout.hxx` skips the `using bout::globals` statement (so that we can include `bout.hxx` or `physicsmodel.hxx` in the library code without including the `using` statement if necessary).

This isn't the best mechanism! What's happened is that `bout/invert/laplacexy.hxx` includes `bout/solver.hxx` which includes `bout/physicsmodel.hxx` (but without the `using bout::globals`), so when `alfven.cxx` comes to include `bout/physicsmodel.hxx`, it hits the include-guard of `bout.hxx` and skips the `using bout::globals`.

Is a solution to move the `using bout::globals` outside the `#idndef __BOUT_H__`? As long as it doesn't matter if `using bout::globals` is repeated, then that would be a solution.

dschwoerer:
Good spot :+1: 

However, it's not quiet as simple.
If we only do that, `bout.hxx` is still only included once, and thus the `using` isn't included.

I hope I have spotted all cases that include `bout.hxx` and would lead to the `using` being set.
Other workaround would be to change the order in `alfven.cxx`

ZedThree:
I think the `using namespace bout::globals` in `physicsmodel.hxx` was just temporary to avoid breaking user code in v4. Personally, I think the correct fix is now for user code to use either `bout::globals::mesh`, or `using namespace bout::globals` explicitly. Alternatively, to use `f.getMesh()` or suchlike where appropriate

dschwoerer:
Should I revert?
d686ea7 - just add to this example?

ZedThree:
That would be my preference, along with also removing the `using namespace bout::globals` from `physicsmodel.hxx` and fixing the remaining examples (but could be separate PR).

Does anyone else have opinions on this though?

dschwoerer:
`GRID_LOAD` requieres that `mesh` exists.

Is this acceptable?
Hardcode `bout::globals::mesh`?
Any better ideas?

`examples/finite-volume/diffusion`

bendudson:
Perhaps we could include a `Mesh` pointer as a `PhysicsModel` member: A physics model is constructed on a mesh (defaulting to global mesh for now). The `GRID_LOAD` macro would then use the `PhysicsModel` mesh pointer, and most user code wouldn't change.

ZedThree:
Sounds good. One thing to be wary of is if a `PhysicsModel` is constructed on the non-global `Mesh`, all the `Field`s and so on would need an explicit `mesh` parameter, but this would affect very few people I think.

We should probably update all the examples to have explicit `mesh` parameters where needed.

ZedThree:
Implementing #1473, at least for a single field, might make some of these changes a bit cleaner.

examples/2Dturbulence_multigrid/esel.cxx
```diff
@@ -66,7 +66,7 @@ class ESEL : public PhysicsModel {
       return 1;
     }
 
-    Coordinates* coord = mesh->getCoordinates();
+    Coordinates* coord = sigma_n.getMesh()->getCoordinates();
```
ZedThree: `sigma_n.getCoordinates()`?

examples/IMEX/drift-wave/test-drift.cxx
```diff
@@ -2,10 +2,11 @@
 #include <bout/physicsmodel.hxx>
 
 #include <invert_laplace.hxx>
+using bout::globals::mesh;
 
 class DriftWave : public PhysicsModel { 
 protected:
-  int init(bool) {
+  int init(bool UNUSED(restart)) {
```
ZedThree: Previous version was fine?

examples/em-drift/2fluid.cxx
```diff
@@ -209,7 +211,7 @@ class EMdrift : public PhysicsModel {
       output.write("=> Not evolving Apar\n");
       initial_profile("Ajpar", Ajpar);
       if (ZeroElMass) {
-        dump.add(Ajpar, "Ajpar", 1); // output calculated Ajpar
+        this->dump.add(Ajpar, "Ajpar", 1); // output calculated Ajpar
```
ZedThree:
```suggestion
        bout::globals::dump.add(Ajpar, "Ajpar", 1); // output calculated Ajpar
```

examples/orszag-tang/mhd.cxx
```diff
@@ -46,7 +46,7 @@ class MHD : public PhysicsModel {
     B.covariant = false; // evolve contravariant components
     bout_solve(B, "B");
 
-    Coordinates *coord = mesh->getCoordinates();
+    Coordinates *coord = B.x.getMesh()->getCoordinates();
```
ZedThree:
```suggestion
    Coordinates *coord = B.x.getCoordinates();
```

examples/performance/bracket/bracket.cxx
```diff
@@ -109,7 +110,7 @@ int main(int argc, char **argv) {
       time_output << "Case legend";
       time_output << "\n------------------------------------------------\n";
 
-      for (int i = 0; i < names.size(); i++) {
+      for (uint i = 0; i < names.size(); i++) {
```
ZedThree: `uint` is non-standard. `std::size_t`, or a type alias for `std::vector<std::string>::size_type` would be more appropriate

include/bout/physicsmodel.hxx
```diff
@@ -271,6 +271,9 @@ protected:
       return model->outputMonitor(simtime, iter, nout);
     }
 
+    Mesh * mesh = bout::globals::mesh;
+    Datafile & dump = bout::globals::dump;
```
ZedThree: These are in `PhysicsModelMonitor`, should they be in `PhysicsModel` itself?

ZedThree:
Thanks for going through and fixing the other warnings in the examples too! :+1:

ZedThree:
Think this is done now. I've updated the title and description to reflect what this PR actually does now

Might be cleaner to make the changes in two PRs?
1. Adding `mesh/dump` members to `PhysicsModel`
2. Removing `using namespace bout::globals` from header

Thoughts?

johnomotani:
@ZedThree what is the "best practice" for using `mesh` and `dump` in `PhysicsModels` after this PR goes in? Within a `PhysicsModel`'s methods, `mesh` and `dump` are still available and usable (being the same as `bout::globals::mesh` and `bout::globals::dump`, but it looks like in here a lot of the examples were modified (maybe before this change?) to avoid using `mesh` and `dump` (e.g. `T->getMesh()->communicate(T)` instead of `mesh->communicate(T)`).

src/physics/physicsmodel.cxx
```diff
@@ -34,7 +34,8 @@
 
 #include <bout/mesh.hxx>
 
-PhysicsModel::PhysicsModel() : modelMonitor(this) {
+PhysicsModel::PhysicsModel()
+    : mesh(bout::globals::mesh), dump(bout::globals::dump), modelMonitor(this) {
```
johnomotani, line 38: Should we provide a constructor that allows passing a `mesh` or `dump` that's different from the global one? E.g. with `nullptr` default values
```
PhysicsModel::PhysicsModel(Mesh* m, DataFile* d)
    : mesh(m == nullptr ? bout::globals::mesh : m),
      dump(d == nullptr ? *bout::globals::dump : *d),
      modelMonitor(this) {
```

ZedThree: I started to do that, but we'd need constructors on the actual user model for this to be useful:

```cpp
class MyModel : public PhysicsModel {
public:
  MyModel(Mesh* mesh, DataFile* dump) : PhysicsModel(mesh, dump) {}
...
};

// In BOUTMAIN():

// This needs to call MyModel ctor
auto model = bout::utils::make_unique<ModelClass>(mesh, dump);         \
```

We can add the constructor as you suggest on `PhysicsModel` now, we just can't call it in `BOUTMAIN` yet!

ZedThree:
> @ZedThree what is the "best practice" for using `mesh` and `dump` in `PhysicsModels` after this PR goes in? Within a `PhysicsModel`'s methods, `mesh` and `dump` are still available and usable (being the same as `bout::globals::mesh` and `bout::globals::dump`, but it looks like in here a lot of the examples were modified (maybe before this change?) to avoid using `mesh` and `dump` (e.g. `T->getMesh()->communicate(T)` instead of `mesh->communicate(T)`).

Sorry, I think this PR has got a little tangled! Might be best to pull out the separate changes into new PRs?

I think best practice will be to use plain `mesh`/`dump` in `PhysicsModel` methods, and `f->getMesh()` in free functions.

johnomotani:
> Sorry, I think this PR has got a little tangled! Might be best to pull out the separate changes into new PRs?

Agreed :+1:

src/physics/physicsmodel.cxx
johnomotani: Ah, I see. Maybe a change for another day then, especially since adding more arguments to the constructor won't break backward compatibility so no need to get it in early.

PR #1987, opened 2020-03-24
conditionally add -Og for non-gcc

dschwoerer:
basically version 2 from @ZedThree suggestions

Advantage over 1. it fixes the bug
Advantage over 3. it's implemented
Advantage over 4. it also works for users, not just travis

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -265,7 +265,7 @@ AS_IF([test "$enable_debug" != ""], [
   OPT_FLAGS="-g -O0 -fno-inline -hipa1"
   # Only for not gcc add -Og due to
   # https://github.com/boutproject/BOUT-dev/issues/1879
-  $MPICXX --version | grep gcc -q || OPT_FLAGS="$OPT_FLAGS -Og"
+  $MPICXX --version | grep g++ -q || OPT_FLAGS="$OPT_FLAGS -Og"
 ], [
   AS_IF([test "x$enable_optimize" != "xno"], [
     AS_CASE(["$enable_optimize"],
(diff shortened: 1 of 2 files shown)


ZedThree:
The reason I'm never overly keen on solutions like this is that it can be very tricky to get this sort of detection robust and portable. On three systems I've got access to, `g++ --version` returns something like `g++ (GCC) ...` (note caps), on others it's the OS name, `(Red Hat)`/`(Ubuntu ...)`/`(SUSE Linux)`.

It looks like the string `Free Software Foundation` does appear on all versions across all OSes that I have access to.

Note that still does leave the problem in place. I don't know _why_ `-Og` is necessary with clang.

configure.ac
```diff
@@ -263,6 +263,9 @@ AS_IF([test "$enable_debug" != ""], [
   AC_MSG_NOTICE([Enabling all debug options])
   enable_checks="3"
   OPT_FLAGS="-g -O0 -fno-inline -hipa1"
+  # Only for not gcc add -Og due to
+  # https://github.com/boutproject/BOUT-dev/issues/1879
+  $MPICXX --version | grep gcc -q || OPT_FLAGS="$OPT_FLAGS -Og"
```
ZedThree: Note that this looks for `gcc`, but should be looking for `g++`. This seems pretty robust across all the systems and variations of GCC I've got access to at least. The current form fails to detect GCC across any of them.

dschwoerer:
Sorry for pushing this broken patch. Changed to g++ now.

I'll try to look into why clang fails, and I agree that fixing it properly for clang would be best.
I'll open an issue as reminder.

ZedThree:
Thanks @dschwoerer !

I thought I'd dig into how CMake determines the compiler, and it basically needs to check a whole bunch of possible preprocessor macros.

> I'll try to look into why clang fails, and I agree that fixing it properly for clang would be best.
I'll open an issue as reminder.

We're in undefined behaviour territory I'm afraid.

PR #1999, opened 2020-03-27
Merge v4.3.1 into next

ZedThree:
I deliberately excluded the language translation files as there were a ton of conflicts. Probably easier to do that separately.


dschwoerer:
> I deliberately excluded the language translation files as there were a ton of conflicts. Probably easier to do that separately.

Fine with me. I don't think there was a lot of work in there, should be easier to do from scratch ...
Also I guess it's very difficult to work with a foreign language ...

PR #2002, opened 2020-03-30
remove checklib dependency from unit tests

dschwoerer:
This triggers unneeded rebuilds. It is not used anywhere else.

Should this go to master instead?

### DIFF
--- a/tests/unit/makefile
+++ b/tests/unit/makefile
@@ -91,7 +91,7 @@ TEST_OBJECTS = $(TEST_SOURCES:%.cxx=%.o)
 
 $(TEST_OBJECTS): | $(GTEST_SENTINEL)
 
-serial_tests: makefile $(BOUT_TOP)/make.config $(OBJ) $(LIB) checklib \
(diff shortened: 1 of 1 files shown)


ZedThree:
It's also used here: https://github.com/boutproject/BOUT-dev/blob/38eae3c7c80fcb65ce567221e1fb002a20180503/make.config.in#L293

`CHANGED` is: https://github.com/boutproject/BOUT-dev/blob/38eae3c7c80fcb65ce567221e1fb002a20180503/make.config.in#L98

That should be `.hxx` not `.h` :man_facepalming:

dschwoerer:
I noticed the `.h`  as well. But it doesn't work anyway if only the hxx are changed, it doesn't trigger a rebuild of affected files. Should I remove it?

That is why the physics model is always rebuild -.-
Are you happy with removing that as well, or is that to confusing for physics model?
If so, I think it should be a separate PR / this PR should be renamed so that it ends up in the changelog.

Issue #2003, opened 2020-04-02
test-restarting fails without netcdf


dschwoerer:
Seems to be broken:
```
====== Exception path ======
[bt] #10 ./test_restarting() [0x40ee1d]
_start at ??:?
[bt] #9 /usr//lib64/libc.so.6(__libc_start_main+0xf5) [0x2b7869aecb15]
__libc_start_main at ??:?
[bt] #8 ./test_restarting() [0x40b1be]
std::_MakeUniq<BoutMonitor>::__single_object std::make_unique<BoutMonitor>() at /marconi/prod/opt/compilers/gnu/7.3.0/none/include/c++/7.3.0/bits/unique_ptr.h:825
 (inlined by) main at /marconi/home/userexternal/dschwoer/soft/bout-dev/parallel-tests/tests/integrated/test-restarting/test_restarting.cxx:22
[bt] #7 ./test_restarting() [0x435118]
PhysicsModel::initialise(Solver*) at /marconi/home/userexternal/dschwoer/soft/bout-dev/parallel-tests/src/solver/../../include/bout/physicsmodel.hxx:85
 (inlined by) Solver::setModel(PhysicsModel*) at /marconi/home/userexternal/dschwoer/soft/bout-dev/parallel-tests/src/solver/solver.cxx:89
[bt] #6 ./test_restarting() [0x421aa2]
PhysicsModel::postInit(bool) at /marconi/home/userexternal/dschwoer/soft/bout-dev/parallel-tests/src/physics/physicsmodel.cxx:125
[bt] #5 ./test_restarting() [0x56cb54]
Datafile::read() at /marconi/home/userexternal/dschwoer/soft/bout-dev/parallel-tests/src/fileio/datafile.cxx:990
[bt] #4 ./test_restarting() [0x56687f]
Datafile::read_f2d(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, Field2D*, bool) at /marconi/home/userexternal/dschwoer/soft/bout-dev/parallel-tests
/src/fileio/datafile.cxx:1326
[bt] #3 ./test_restarting() [0x571a1a]
DataFormat::readFieldAttributes(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, Field&) at /marconi/home/userexternal/dschwoer/soft/bout-dev/parallel-
tests/src/fileio/dataformat.cxx:61
[bt] #2 ./test_restarting() [0x48a495]
[...]

PR #2000, opened 2020-03-28
Parallel tests

dschwoerer:
Resolves #1451 

It now works to run `make -j 4 check` to build and run tests in parallel.

There is still further improvements possible, e.g. the [jobserver](https://www.gnu.org/software/make/manual/html_node/POSIX-Jobserver.html) isn't available to `test-compile-examples` (which isn't parallel, but should be easy to fix)

There are some draw-backs:
Because other jobs are running, the info that the job started is only printed once the job is finished.

This could be worked around by adding a time-out. Maybe 30 to 120 minutes should be safe? 10 doesn't do, as compiling all examples takes more time.

Also the speed-up isn't awesome, because many tests already run in parallel, so there is nothing to gain.

jobs | time
------|-----------------
1 |real    5m27.523s
2 | real    4m19.255s
4 | real    3m44.568s
8 | real    2m25.634s
12 | real    2m11.765s
16 | real    1m48.893s
24 | real    2m22.265s

(`time make -j <n> check` on 12 cores, 24 threads, 6 GB ram)

PR #2004, opened 2020-04-02
Without netcdf

dschwoerer:
This adds support for `./configure --without-netcdf && make check`

I found #2003 during this, but other tests are now passing / being skipped.

Depends on #2000 to avoid merge conflicts.


tests/integrated/test-stopCheck-file/runtest
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env python3
 
-#requires: not travis
+# requires: not travis
+# h5py doesn't load the data as expected
```
ZedThree, line 4: Might be worth checking this is still the case

dschwoerer: This is with python3-h5py-2.10.0
If the data-length is 1, it is returned as a number, rather, then an array of length 1.

ZedThree: Fixed. It couldn't tell the difference between a time-dependent scalar with only one value, and an time-independent scalar

ZedThree:
Tests are passing, just not getting reported correctly, again.

The change to tools/pylib/boututils/datafile.py in cc61817 should maybe go into master and/or the boutdata repo too

PR #2008, opened 2020-04-03
Potential fix for laplacexz

dschwoerer:
Not quite sure it should be `section_name` or `"laplace"`

### DIFF
--- a/include/bout/invert/laplacexz.hxx
+++ b/include/bout/invert/laplacexz.hxx
@@ -51,6 +51,7 @@ public:
 
   ReturnType create(Mesh* mesh = nullptr, Options* options = nullptr,
                     CELL_LOC loc = CELL_CENTRE) {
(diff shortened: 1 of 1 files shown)


johnomotani:
I think the section name was already handled by a combination of `getType`, which  is implemented in the generic `Factory` template and gets the type from `section_name`, and the constructors handling `options == nullptr` to read the `laplacexz` section. I had a similar confusion somewhere before @ZedThree put me straight!

dschwoerer:
Without this, `test-laplacexz` is crashing on me.

Might well be that this should be fixed some where else, but it is not working without a fix ...

ZedThree:
Looks like the PETSc LaplaceXZ does check for `options == nullptr` but `cyclic` doesn't. Best fixed there, especially because `LaplaceXZcyclic` takes a default argument of `nullptr` for `options` anyway.

dschwoerer:
Replaced by #2025

PR #2010, opened 2020-04-03
Add `boutconfig` Python module with build configuration information

dschwoerer:
Allows to check the configure / cmake options in python

Part of 3D metrics. Thus not yet used in the tests.


dschwoerer:
It needs to not be in boututils/boutdata as that doesn't work with out-of-souce builds (cmake).
Forgot to run autoconf (again :disappointed:), should be fixed now.

A dict is probably cleaner :+1: 

I am aware of the issue, however, as a normal user, I guess only one BOUT++ version is used. Of course, as a developer, this is not the case.
A way around this would be providing module files. That would be the cleanest.
We anyway have cases where things are breaking, if you use the wrong pylib.
This is especially true with `boutcore`. So I think we have the issue independently of `boutconfig`.

I am not at all certain about the `--config` - than we need to think so much about forward/backward compatibility. Not sure it is worth it ...
Keep in mind you want to make sure, that if you call an old binary, that does not support that, that it shouldn't overwrite any data. From my experience with `boutcore`, that is not at all trivial ...

Requiring dump files or a binary also seems like it would make the interface more complicated ...

The use-case was tests, and in that case we anyway ensure that `$PYTHONPATH` is correct.

If this proposal is reject I can write a wrapper for `bout-config`, that would also solve my problem, but I thought this is nicer ...

CMakeLists.txt
```diff
@@ -799,6 +799,16 @@ if(PACKAGE_TESTS)
...
+option(ENABLE_METRIC_3D "Enable 3D metric support" OFF)
+if(ENABLE_METRIC_3D)
+  set(BOUT_METRIC_TYPE "3D")
+  target_compile_definitions(bout++
+    PUBLIC "COORDINATES_USE_3D")
+  target_link_libraries(bout++ PUBLIC ${CMAKE_DL_LIBS})
+else()
+  set(BOUT_METRIC_TYPE "2D")
+endif()
+set(BOUT_USE_METRIC_3D ${ENABLE_METRIC_3D})
```
ZedThree, line 811: Suspect you didn't mean to bring this in

ZedThree:
Sounds like another argument to move the preprocessor defines to a configure-time header, and provide some nice `constexpr` wrapper functions.

It'll be next week before I have time to do so, but I'll try and get it done next.

CMakeLists.txt
dschwoerer: Yes, shouldn't go in yet.

ZedThree:
Instead of generating the files from the config system, is it possible to just import the built/installed header using Cython? Would also be useful to import the version header

dschwoerer:
I think the hardest would be to ensure we find the correct header. Especially if files are installed or otherwise moved around.

If we suggest to export BOUT_TOP, I wouldn't mind to switch over, but for now I don't think that will help.

ZedThree (approved):

ZedThree:
I tidied this up a bit for CMake, removed the 3D metric-specific bits for now, and made sure the `configure` version gets deleted on `make distclean`

ZedThree (approved):

PR #2012, opened 2020-04-03
Provide better error if fields are not compatible

dschwoerer:
Helps with debugging ...


ZedThree:
Very nice, thanks @dschwoerer !

PR #2016, opened 2020-04-11
Fix bug in configure when using dash as shell

dschwoerer:
Fix for @loeiten's issues

Also adds test for dash, we had issues in the past ...


ZedThree:
:+1: Using dash in one of the jobs is very useful, thanks!

PR #2019, opened 2020-04-12
Add script for creating environment module files

dschwoerer:
This program, bin/add-mod-path.py provides
a convenient way to create modules for the environment module for
BOUT++ that sets PYTHONPATH, PATH and BOUT_TOP, so that switching
between different BOUT++ installations is easy.
bin/add-mod-path.py stores the loaded modules as prerequesit, so that you
don't accidentially mix e.g. mpi versions. Feel free to edit the
crated module file, if there are non-needed dependencies.

bin/add-mod-path.py can be used interactively, or the location where to
create the modfiles, the name and the BOUT++ directory can be
passed:
  bin/add-mod-path.py [modulepath [name [bout-top]]]

If modulepath is not passed, the script checks whether any of the
paths in $MODULEPATH are writeable, and asks which of them to
choose. If non are writeabel, and you can or want not to run as
root, you can extend this. Create a new folder and at it to your
`.bashrc` e.g. with:
```
 mkdir ~/.module
 echo 'export MODULEPATH=$HOME/.module:$MODULEPATH'>> ~/.bashrc
 source ~/.bashrc
```

name can by any name by which you will be later able to recognice
this specific bout installation, e.g. `3d-metrics`. You can later
load the module with `module load bout/name` e.g. `module load
bout/3d-metrics`.

bout-top is the root directory of the BOUT++ installation. Note that
it only works with in-source-builds.

Might help with #2010

### DIFF
--- a/bin/add-mod-path.py
+++ b/bin/add-mod-path.py
@@ -1,0 +1,128 @@
+#!/usr/bin/env python3
+
+import os
+import sys
+
+def error(*args):
+    print(*args)
+    sys.exit(1)
+
+def info():
+    print(f"""This program, {sys.argv[0]} provides
+a convenient way to create modules for the environment module for
+BOUT++ that sets PYTHONPATH, PATH and BOUT_TOP, so that switching
+between different BOUT++ installations is easy.
+{sys.argv[0]} stores the loaded modules as prerequesit, so that you
+don't accidentially mix e.g. mpi versions. Feel free to edit the
+crated module file, if there are non-needed dependencies.
+
+{sys.argv[0]} can be used interactively, or the location where to
+create the modfiles, the name and the BOUT++ directory can be
+passed:
+  {sys.argv[0]} [modulepath [name [bout-top]]]
+
+If modulepath is not passed, the script checks whether any of the
+paths in $MODULEPATH are writeable, and asks which of them to
+choose. If non are writeabel, and you can or want not to run as
+root, you can extend this. Create a new folder and at it to your
+`.bashrc` e.g. with:
+```
+ mkdir ~/.module
+ echo 'export MODULEPATH=$HOME/.module:$MODULEPATH'>> ~/.bashrc
+ source ~/.bashrc
+```
+
+name can by any name by which you will be later able to recognice
+this specific bout installation, e.g. `3d-metrics`. You can later
+load the module with `module load bout/name` e.g. `module load
+bout/3d-metrics`.
+
+bout-top is the root directory of the BOUT++ installation. Note that
+it only works with in-source-builds.""")
+
+def ask_modpath():
+    try:
+        modpath=os.environ['MODULEPATH']
+    except KeyError:
+        error("MODULEPATH not set. Please ensure environment modules are loaded.")
+
+    writeable=[]
+    for mp in modpath.split(':'):
+        if os.access(mp, os.W_OK):
+            writeable.append(mp)
+
+    if writeable == []:
+        error("Did not find a writeable directory in $MODULEPATH.\nPlease append a writable path.")
+
+    if len(writeable) == 1:
+        return mp[0]
+
+    print("Several paths found to store module files. Please select one:")
+    for i in range(len(writeable)):
+        print("%3d: %s"%(i, writeable[i]))
+    select = input("Your choice: ")
+    return writeable[int(select)]
+
+def ask_name():
+    print("The module will be prefixed with `bout/`")
+    print("Please enter the name of the module.")
+    name = input("bout/")
+    return name
+
+def ask_bouttop():
+    this = os.path.realpath(__file__)
+    top = "/".join(this.split("/")[:-2])
+    print("Do you want to use `%s` as BOUT_TOP?"%top)
+    while True:
+        inp = input("[Y/n/<alt-path>]")
+        if inp in ["", "y", "Y"]:
+            return top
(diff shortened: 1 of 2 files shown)


bin/add-mod-path.py
```diff
@@ -0,0 +1,128 @@
+#!/usr/bin/env python3
+
+import os
+import sys
+
+def error(*args):
+    print(*args)
+    sys.exit(1)
```
ZedThree, line 8: Why not just use `raise RuntimeError`?

bin/add-mod-path.py
```diff
@@ -0,0 +1,128 @@
...
+
+    if writeable == []:
+        error("Did not find a writeable directory in $MODULEPATH.\nPlease append a writable path.")
+
+    if len(writeable) == 1:
+        return mp[0]
+
+    print("Several paths found to store module files. Please select one:")
+    for i in range(len(writeable)):
+        print("%3d: %s"%(i, writeable[i]))
```
ZedThree, line 62:
```suggestion
    for index, directory in enumerate(writeable):
        print(f"{index:3d}: {directory}") 
```

bin/add-mod-path.py
```diff
@@ -0,0 +1,128 @@
...
+            return top
+        if os.path.isdir(inp):
+            return os.path.realpath(inp)
+        if inp not in "nN":
+            print("`%s` is not a valid directory"%inp)
+
+def create_mod(mp, name, top):
+    if not os.path.isdir(mp+"/bout"):
+        os.mkdir(mp+"/bout")
+    fn = mp+"/bout/"+name
```
ZedThree, line 88: `fn` -> `filename`

bin/add-mod-path.py
```diff
@@ -0,0 +1,128 @@
...
+def error(*args):
+    print(*args)
+    sys.exit(1)
+
+def info():
+    print(f"""This program, {sys.argv[0]} provides
+a convenient way to create modules for the environment module for
+BOUT++ that sets PYTHONPATH, PATH and BOUT_TOP, so that switching
+between different BOUT++ installations is easy.
+{sys.argv[0]} stores the loaded modules as prerequesit, so that you
```
ZedThree, line 15: `prerequesit` -> `prerequisites`

(It literally always takes me two attempts to type this word!)

bin/add-mod-path.py
```diff
@@ -0,0 +1,128 @@
...
+
+def ask_name():
+    print("The module will be prefixed with `bout/`")
+    print("Please enter the name of the module.")
+    name = input("bout/")
+    return name
+
+def ask_bouttop():
+    this = os.path.realpath(__file__)
+    top = "/".join(this.split("/")[:-2])
```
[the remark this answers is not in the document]
dschwoerer: I tried basename first, but that doesn't work, at least the one from os.path: https://docs.python.org/3.5/library/os.path.html#os.path.basename
If there is a different one, with different behaviour, I wouldn't call that intuitive.

I only found parent in purepath, but that does anyway the same I do, but by having an additional layer, I would assume it is better then that, so I think it is easier to just do it myself, but if you have prefer the parent one, I don't care that match ...
https://docs.python.org/3.5/library/pathlib.html?highlight=pathlib#pathlib.PurePath.parent

bin/add-mod-path.py
dschwoerer: I wanted to provide the current state as input arguments, but then got lazy ...

bin/add-mod-path.py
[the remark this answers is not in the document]
ZedThree: Well it's more that using a named function makes the code more explicit and easier to understand, even if it's doing exactly the same thing under the hood. This isn't something where we need to squeeze out every last drop of performance, so prefer to make code readable.

(btw `pathlib.Path` inherits from `PurePath`, but can do actual IO operations, `PurePath` doesn't need an actual filesystem)

bin/add-mod-path.py
```diff
@@ -0,0 +1,159 @@
...
+                "writeable_dir:{0} is not a writeable dir".format(prospective_dir))
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description=_info,
+                                     formatter_class=argparse.RawDescriptionHelpFormatter)
+    parser.add_argument('--modpath', '-m', action=writeable_dir, nargs=1,
+                        help='The modulepath to install the bout++ module')
+    parser.add_argument('--name', '-n', nargs=1,
+                        help='''can by any name by which you will be later able to recognice
```
ZedThree, line 140: Typos:

```suggestion
                        help='''Can be any name by which you will be later able to recognise
```

bin/add-mod-path.py
```diff
@@ -0,0 +1,159 @@
...
+
+def writeable_dir(prospective_dir):
+    if not os.path.isdir(prospective_dir):
+        raise Exception(
+            "readable_dir:{0} is not a valid path".format(prospective_dir))
+    if os.access(prospective_dir, os.W_OK):
+        return prospective_dir
+    else:
+        raise Exception(
+            "readable_dir:{0} is not a writeable dir".format(prospective_dir))
```
ZedThree, line 118: Still needed?

dschwoerer: Thanks, removed.

I also renamed it, it shouldn't matter for a user that it is a python script :-)

PR #2020, opened 2020-04-14
Experimental expression template pipelines

bendudson:
Warning: Experimental feature, potential time sink

Aim: To reduce the number of small loops, improving parallelisation, by grouping expressions using templates.

Inspired by ndarray-v2 (https://github.com/jzrake/ndarray-v2), this implements a kind of expression templates by overloading the `|` operator. This approach uses lambda functions, making this quite readable compared to pre-C++11 template expressions. If the nested lambda functions get inlined by the compiler, this should result in only one loop around the whole expression.

Advantages of this approach:
- Only requires overloading one operator `|` 
- Implementation is a completely separate header, which doesn't require any modifications to the fields e.g. assignment operator

Disadvantages include:
- Quite different syntax / pattern to existing code
- Implementing operators is a little fiddly: Functions which return functions of functions. It may be possible to simplify some of this with macros

```
#include "bout/pipelines.hxx"
using namespace bout::experimental;

Field3D field {1.0};
Field3D result = field | add(2.0) | collect();  // Chain of operators collect(add(2.0, field))
// result is filled with value 3.0
```
See unit tests.


ZedThree:
Very cool! Perhaps `toField` instead of `collect`?

Any idea how well this would work with C++20 ranges?

bendudson:
Good idea yes, `toField` sounds good. The syntax is basically the same as C++20 ranges, and I think this code could be adapted to use ranges in future without changing the API much. Operations like `add` are range views, which lazily evaluate; `toField` / `collect` is an action which eagerly evaluates the range.
One difference is that the domain of the range (begin/end) is determined here by `toField`, where it loops over the domain e.g. it could use:
```
Field3D result = field | add(2.0) | collect(field.region("RGN_ALL"));
```
i.e. define a pipeline and pull values through it, whereas a range-like API would use something like:
```
Field3D result = field[field.region("RGN_ALL")] | add(2.0) | collect(field.region("RGN_ALL"));
```
i.e. determine the range of values at the start, then push them through the pipe to the end.
I don't know how complicated it would be though to use something like `range-v3`, but probably worth thinking about whether we want something compatible with ranges.

bendudson:
Now with example for comparison against the single index operator approach. 
Single index operator:
```
BOUT_FOR(i, n.getRegion("RGN_NOBNDRY")) {
      ddt(n)[i] = bracket(n, phi, i)
                  - alpha * Div_par_Grad_par(phi_minus_n, i)
                  - kappa * DDZ(phi, i)
                  + Dn * Delp2(n, i);
}
```
Pipelines:
```
ddt(n) = bp::bracket(n, phi)
         | subtract(bp::Div_par_Grad_par(phi_minus_n) | multiply(alpha))
         | subtract(bp::DDZ(phi) | multiply(kappa))
         | add(bp::Delp2(n) | multiply(Dn))
         | toField(n.getRegion("RGN_NOBNDRY"));
```

johnomotani:
Random thought: The pipelines versions of operators aren't (to me at least) super easy to read, but they look like they might lend themselves well to some sort of code-translation script? E.g. get Python to search for everything on the RHS of an assignment operator (`=`, `+=`, `-=`, etc), and evaluate, somehow overloading e.g.
```
def __add__(self, other):
    return f" ( {eval(self)} | add({eval(other)}) )"
```
I don't know if there's a good way to define a class that would allow this, or if the recursion necessary can be handled sensibly, so this might just be nonsense... Also don't know if there's any actual advantage to having the pipelines involved in this, maybe it's actually a proposal for an alternative: use Python pre-processor to replace every assignment statement with a `BOUT_FOR`...

ZedThree:
This is essentially C++ code-gen for the single index operators. If we go full expression templates, then we can a layer of types where the arithmetic operators translate into code like this.

bendudson:
This was a fun experiment, but it looks like the single index operators are the more intuitive way, and probably more performant. Closing as not going to be merged.

PR #2022, opened 2020-04-15
Outer loop single index operators

bendudson:
Adds `include/bout/single_index_ops.hxx` which defines `bracket`, `DDX`, `DDY`, `DDZ` and `Delp2` operators which take a `SpecificInd` argument.
These operators can be used to put a loop around a section of the RHS function, reducing the number of loops and hopefully improving parallelisation with OpenMP or RAJA.

Includes unit tests which compare against original (field) operators.

The `blob2d-outerloop` example is a version of `blob2d` using these operators. Currently this gives a slightly different result, possibly small enough to be accumulated rounding error, but it's possible that there's still a bug.

Another example, `hasegawa-wakatani-3d` also uses single index operators, but is a 3D system which can be used to develop and test parallelisation methods.


johnomotani:
Could differences between `blob2d` and `blob2d-outerloop` be due to `blob2d` using FFT z-derivatives and `Delp2`? Looks like `blob2d-outerloop` is using central-difference.

bendudson:
Yes! Thanks @johnomotani !
`examples/blob2d`:
```
5.000e+01         54       1.01e+00    46.4   26.7    0.1    1.7   25.1
1.000e+02         19       3.64e-01    45.0   25.9    0.1    3.7   25.4
1.500e+02         21       4.00e-01    45.4   26.3    0.1    3.2   24.9
2.000e+02         34       6.42e-01    47.1   27.1    0.1    2.0   23.7
```
`examples/blob2d-outerloop`
```
5.000e+01         54       2.54e+00    78.8   10.7    0.0    0.6    9.8
1.000e+02         19       9.03e-01    77.8   10.5    0.0    1.5   10.2
1.500e+02         21       1.01e+00    77.9   10.7    0.0    1.3   10.0
2.000e+02         34       1.62e+00    79.6   10.7    0.0    0.8    8.9
```
The outer loop version is significantly slower than the original, not sure why. Perhaps looking up coordinates repeatedly?

johnomotani:
It doesn't look like there should be very much work inside each iteration of the main `BOUT_FOR` - is having the conditionals inside the tight loop actually slowing it down? Would it be better to make things like that compile-time options for `outerloop` style physics models?

d7919:
Hopefully something like scalasca profiling could help here.

bendudson:
Adding a public member to Field:
```
mutable Coordinates* coords {nullptr};
```
and then using that in the index operators reduces the time to:
```
5.000e+01         54       1.07e+00    49.3   25.8    0.1    1.5   23.3
1.000e+02         19       3.85e-01    47.7   24.9    0.1    3.5   23.8
1.500e+02         21       4.29e-01    48.1   24.3    0.1    3.0   24.5
2.000e+02         34       6.54e-01    50.3   25.9    0.1    1.9   21.8
```
i.e. pretty close to the original.

bendudson:
Changing the Field classes to have fast access to the Coordinates seems tricky. Some possibilities:
1. Add a public Coordinates* pointer to Field2D & Field3D. 
    a) If it's set on Field construction then mesh may not exist yet
    b) If it's set on assignment or when .allocate() is called, then this is a problem for the fields inside Coordinates: An infinite loop starts when we create coordinates. 
     c) Perhaps it could be set in communicate(), since a field should be communicated before being differentiated
     d) The user could manually set it, or call a function to set it
2. Add another set of FIeld types, for use in Coordinates. These would be stripped down versions of Field2D/3D without pointer to Coordinates (and maybe other things). The current Field2D/3D classes would inherit from the simpler classes.
3. The single index operators could take the Coordinates pointer as input
    a) The user passes it around, and if using staggered grids keeps track of several Coordinates
    b) The index could be bundled into a more complicated type with all the coordinates, so the user only passes one thing around

So far I'm leaning towards 1c), but suggestions welcome

Staggering also makes this difficult to do efficiently, if every operation on every index needs to check the cell location of all inputs. I think the easiest way for now is to have different operators that the user has to choose e.g. `Div_par<CtoYlow>(p, i)`. A possible optimisation would be to have (yet) another set of field types where the cell location was part of the type (known at compile time).

ZedThree:
Option 2 is quite appealing because it avoids a circular dependency, but would definitely be a bunch of work to implement.

Getting the staggering in via templated fields so it's known at compile time is also appealing, because we could then completely remove a bunch of runtime checks and just do it at compile time.

1c almost certainly the fastest way to get it done, in terms of development time, but I think getting more robust types would be a better long-term solution

johnomotani:
If staggered fields were templated types, then the operators could be templated just on the output location (so template parameter is a `CELL_LOC`), e.g. `Div_par<CELL_YLOW>(p, i)`, which I think is a bit nicer than `Div_par<CtoYlow>(p, i)`.

Another thing with staggering - at the moment we lose all location checking in the outerloop version. If we are templating the fields to different staggered locations, could we get compile-time location checking? If we make some typedefs `BoutReal_centre`, `BoutReal_ylow`, etc. (or is it possible to do `BoutReal<CELL_CENTRE>`, `BoutReal<CELL_YLOW>`...??) then a `Field3D<CELL_YLOW>` would have `BoutReal_ylow` data, and `Div_par<CELL_YLOW>(p, i)` could return `BoutReal_ylow`. I guess we can prevent the `BoutReal_centre`/`BoutReal_ylow` from being implicitly convertible, so if you made a mistake with the operators, you'd get a compiler error? And if `BoutReal` was implicitly convertible to any of the location-specific types, you could still do e.g. `Field3D p = 4.0;`? This does sound like quite a lot of work though!

Entirely separately from the changes needed for outerloop operators, having the location checking be compile-time rather than run-time would be very nice!

bendudson:
If the operators are returning types that the user doesn't often have to deal with directly, or can hide with `auto`, then something like `BoutReal<CELL_CENTRE>` should be possible, as a templated struct wrapping a `double`. There are libraries (e.g. xtensor) which have flexible types with more run-time checking, and specialised types with more compile-time checking. 

I think a lot depends on how useful and performant the single loop approach is: On a single thread it seems not much different, though that was on a far larger mesh than it would be on a large number of cores. Hopefully it parallelises better, but we'll need to test if the benefits are worth the effort.

bendudson:
I've tried 1c, but unfortunately Coordinates members are communicated so there would need to be some way of hacking it to avoid getting into an infinite loop. I think a similar infinite loop situation happened with yup/down fields, but can't quite remember how that one went. 

For now I think a obvious hack is better than one which makes too many changes behind the scenes and risks becoming permanent. I've added `fast_coord` to `Field`, which has to be set manually. This will enable us to test things out, and I think is gross enough to motivate replacing it asap.

dschwoerer:
The approach I took in the coord3d branch, is to use a different
communication, that avoids requiring having a valid coordinates.

Maybe it is worth to think more about the special type for the
coordinates type. For example it doesn't really make sense for the
coordinates fields to be transformed after creation, because they would
depend on them self ...

bendudson:
A suggestion for a way to use accessor types (a bit like SYCL) to get performance while wrapping some of the gross features used in this branch at the moment:
- The `Field` types include checking, smart pointers etc. which seem to produce significant penalties when used in inner loops. Examples are `Coordinates`, `yup` and `ydown` pointers.
- Currently the single index operators take `Field3D` types, but we have to set raw pointers outside the loop. 

Instead, we could have something like a `FieldAccessor` type, which has raw pointers and is unsafe but enables high performance. 
- Outside the loop, `FieldAccessor` types would be created e.g. 
```
auto n_acc = FieldAccessor(n); // Create accessor
```
- In the loop, the single index operators would take `FieldAccessor` types. To prevent `Field` types being passed and converted each time, the `FieldAccessor` constructor could be made explicit.
- We could define different `FieldAccessor` types, or a templated type, to allow specialisation on cell location. The compatibility of the cell location could then be checked once when the accessor is created:
```
auto v_acc = FieldAccessor<CELL_YLOW>(v); // Check at runtime that v.location is CELL_YLOW
```
I think this `FieldAccessor` set of types could be implemented without modifying the existing types, and provides a way to separate "safe" `Field` types  with checking, and unsafe types for high performance. 

Thoughts @ZedThree @johnomotani @dschwoerer ?

ZedThree:
That seems like a nice compromise to opt-in to the higher-performance but more complex system.

The checking might need a bit of work, as we often need to check two fields are compatible: are they at the same staggered location, in the same y-space, etc. I'm not sure how that would look like with `FieldAccessor`. I think we have something like `checkCompatible(a, b)` already? The user might just have to call that themselves.

```cpp
checkCompatible(n, T);
auto n_acc = FieldAccessor(n);
auto T_acc = FieldAccessor(T);
```

`checkCompatible` might need to be variadic to be most useful.

ZedThree:
Ah I see, yes that makes sense!

johnomotani:
I agree, this seems like a good compromise.

It might be nice to have the `Field*` classes themselves templated with properties like `CELL_LOC` so checks can be compile-time instead of run-time. That's probably a lot of work though, and I'm not sure how it would work or if it would need some redesigning. I have an impression that polymorphic inheritance and templates don't play well together...

ZedThree:
I very much like the idea of being able to move more checks to compile time, but we might need to be careful with runtime options.

As long as the base class is non-templated, having templated derived classes should be fine.

bendudson:
Replaced by PR #2397

PR #2023, opened 2020-04-15
Actually use extra flags

dschwoerer:
Unless I missed something, we check several flags, but then don't end up using them.

### DIFF
--- a/configure
+++ b/configure
@@ -6049,6 +6049,8 @@ fi
 done
 
 
+CXXFLAGS="$CXXFLAGS $extra_compiler_flags_test"
+
 # Disable checks if optimization > 2 is used
 if test -z $enable_checks && test "x$DISABLE_CHECK" = "xprobably"; then :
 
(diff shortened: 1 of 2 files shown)


ZedThree:
`extra_compiler_flags_test` is some extra flags to use just during the test to see if the compiler flags are supported, and for clang is just:

https://github.com/boutproject/BOUT-dev/blob/38eae3c7c80fcb65ce567221e1fb002a20180503/configure.ac#L218-L227

We do a similar thing for Intel, and for GCC with one other flag. We don't actually need (or necessarily _want_) these flags when we're actually compiling, it's just during the test for those flags.

`AX_APPEND_COMPILE_FLAGS` checks a list of compiler flags and appends them to `CXXFLAGS` if they work: https://www.gnu.org/software/autoconf-archive/ax_append_compile_flags.html

bendudson:
Thanks @ZedThree . I was going to suggest documenting this somewhere, but it's probably better to put that effort into moving to cmake.

dschwoerer:
Sorry, I was using extra_compile_flags_extra later, assuming that they would be used.

PR #2027, opened 2020-05-05
Move compile-time feature macros to header

ZedThree:
- Move most BOUT++ feature macros into a header `include/bout/build_defines.hxx`
    - This is generated at configure time, so no longer need to pass feature macros on command line
    - These macros are now always defined so `#ifdef` needs to be replaced with `#if`
- Rename feature macros consistently
    - Third party libraries are `BOUT_HAS_`, for example `BOUT_HAS_PETSC`
    - BOUT++ features are `BOUT_USE_`, for example `BOUT_USE_SIGFPE`
- Update `bin/bout-v5-macro-updater.py` to fix the above two changes automatically in user code
- Add `include/bout/build_config.hxx` which provides corresponding `constexpr` variables in the `bout::build::` namespace
    - `bout::build::has_petsc` is a `constexpr bool` so can be used in expressions
    - This is split from `build_defines.hxx` because that file is generated differently by CMake and configure, and I didn't want to complicate matters more than necessary
    - `build_config.hxx` is included in `bout.hxx`, so it should get into user code automatically
- Add `--config` command line option to print list of enabled/disabled features
    - Currently just human-readable

A command line option to print a machine-readable version would be nice. JSON is sort of an obvious choice, except that trailing commas are verboten, so maybe just Python dict?

```bash
$ ./hermes --config
{
  "has_petsc": True,
  "has_fftw": False,
  "use_sigfpe": True,
...
}
```
or BOUT++ input file format:
```bash
$ ./hermes --config
has_petsc = True
has_fftw = True
use_sigfpe = False
```

Another possible feature would be to add a `std::map`, but this couldn't be `constexpr`, so I've not done that. Also the values would probably need to be either `variant` or `any`.

I've not moved `CHECK`, `BOUT_LOCALE_PATH` or `BOUT_FLAGS_STRING` into `build_config.hxx`:
- `CHECK` is used everywhere, and most likely to be used in user code, so I'm a little worried about renaming it. Thoughts?
- `BOUT_LOCALE_CHECK` needs its value changing when installed, and I didn't quite understand how that's currently done
- `BOUT_FLAGS_STRING` in CMake is not fully determined until build time, while the `build_defines.hxx` gets created at configure time


ZedThree:
Using this branch and [cppyy](https://cppyy.readthedocs.io/en/latest/index.html) I was able to get all these flags into python pretty easily:

```python
import cppyy
cppyy.add_include_path("build/include")
cppyy.include("include/bout/build_config.hxx")
print({k: bool(getattr(cppyy.gbl.bout.build, k)) 
    for k in dir(cppyy.gbl.bout.build) 
    if k.startswith("has") or k.startswith("use")})                                                                                                                                                                                                                                                                

{'has_fftw': True,
 'has_gettext': True,
 'has_hdf5': False,
 'has_lapack': True,
 'has_netcdf': True,
 'has_petsc': False,
 'has_pretty_function': True,
 'has_pvode': True,
 'has_scorep': False,
 'has_slepc': False,
 'has_sundials': False,
 'use_backtrace': True,
 'use_color': True,
 'use_openmp': True,
 'use_output_debug': False,
 'use_sigfpe': False,
 'use_signal': True,
 'use_track': True}
```

bendudson:
Wow cppyy looks pretty impressive!

dschwoerer:
I do not like the `--config` option.

1. It calls exit. This is really bad, as it does not deallocate variables with automatic scope, i.e. most variables. This makes it quite tricky to use the bout++ library, as if there is for some reason the `--config` option set (which isn't bout specific) bout exits, terminating everything [1,2]

2. It doesn't really solve the issue for python - rather then needing the correct PYTHONPATH, I need a binary. Note also, that this binary needs not only be findable - it also needs to be executable, which is a lot of fun on machines like marconi, where you cannot run AVX512 binaries on the login nodes. I would much more prefer finding BOUT_TOP to put the config, then to find a binary.

3. It is not backwards compatible. If you try this on an old version of the binary, assuming it doesn't SIGILL due to 2, it will basically run forever, as you execute a most likely a very expensive task on a single core.

4. To make things worse, this can also lead to dataloss, as BOUT.dmp.0.nc is overwritten. This can happen due to 3, but also due to a regression, or due to a typo, or if this feature is removed at some point in the future.

5. It is not forward compatible either, see 4. We basically need to keep this forever.


I think a much better approach would be, to always create a file while running, and dump the config info in this file. Then for post processing, we anyway need the dmp files, so chances are the config file is in the same path. Thus no additional searching for most cases is needed. It cannot fail horribly. If the file isn't there, we didn't cause additional damage. You can still use cppyy to parse it. We do not need to exit from a library.

[1] https://en.cppreference.com/w/cpp/utility/program/exit
[2] https://en.cppreference.com/w/cpp/language/storage_duration

ZedThree:
To be clear, I think any python config reader should read directly from `include/bout/build_config.hxx`, rather than via running a binary with `--config`. The `--config` option is intended useful as a quick, human readable way to see what an executable was compiled with.

The config variables can also be dumped into the output file, and `BOUT.settings` too. I'm not sure there's need for another file.

ZedThree:
- Changed `--config` to `--print-config` to make it clearer that it's just human-readable info only
- Added the compile-time features to the dump file

Using `xbout`, getting the compile-time features out of the dump files was very easy:

```python
>>> import xbout
>>> d = xbout.open_boutdataset("data/BOUT.dump.*.nc")
>>> print({k: bool(v) for k, v in d.metadata.items() if k.startswith("has") or k.startswith("use")})
{'has_netcdf': True,
 'has_lapack': True,
 'has_hdf5': False,
 'has_slepc': False,
 'use_color': True,
 'use_signal': True,
 'use_backtrace': True,
 'has_fftw': True,
 'has_gettext': True,
 'has_pvode': True,
 'use_track': True,
 'use_output_debug': False,
 'has_pretty_function': True,
 'has_petsc': True,
 'has_scorep': False,
 'use_sigfpe': False,
 'has_sundials': False,
 'use_openmp': False}
```

ZedThree:
/clang-format

Using one of my PRs as a guinea pig for the new clang-format bot. We'll see how well I manage to actually run it myself! :smile:

configure.ac
```diff
@@ -300,51 +299,63 @@ AS_IF([test -z $enable_checks && test "x$DISABLE_CHECK" = "xprobably"], [
...
 AS_IF([test "x$enable_sigfpe" = "xyes"], [
   AC_MSG_NOTICE([Signaling floating point exceptions enabled])
-  CXXFLAGS="$CXXFLAGS -DBOUT_FPE"
+  BOUT_USE_SIGFPE=yes
 ], [
   AC_MSG_NOTICE([Signaling floating point exceptions disabled])
 ])
 
+BOUT_USE_OUTPUT_DEBUG=no
+AS_IF([test "x$enable_output_debug" = "xyes"], [
```
dschwoerer, line 352: Might be nice to enable if we have `BOUT_CHECK_LEVEL=3`?

configure.ac
```diff
@@ -1318,19 +1331,27 @@ AC_MSG_NOTICE([-------------------------])
...
-AC_MSG_NOTICE([  Natural language support: $HAS_NLS (path: $localedir)])
+AC_MSG_NOTICE([  PETSc support           : $BOUT_HAS_PETSC (has SUNDIALS: $PETSC_HAS_SUNDIALS)])
+AC_MSG_NOTICE([  SLEPc support           : $BOUT_HAS_SLEPC])
+AC_MSG_NOTICE([  IDA support             : $BOUT_HAS_IDA])
+AC_MSG_NOTICE([  CVODE support           : $BOUT_HAS_CVODE])
+AC_MSG_NOTICE([  ARKODE support          : $BOUT_HAS_ARKODE])
+AC_MSG_NOTICE([  FFTW support            : $BOUT_HAS_FFTW])
+AC_MSG_NOTICE([  NetCDF support          : $BOUT_HAS_NETCDF])
+AC_MSG_NOTICE([  Parallel-NetCDF support : $BOUT_HAS_PNETCDF])
+AC_MSG_NOTICE([  HDF5 support            : $BOUT_HAS_HDF5 (parallel: $BOUT_HAS_PHDF5)])
```
dschwoerer, line 1342: Are the parallel interfaces working?
I never managed to get them working, so might be better to remove them?

johnomotani: They don't work, but I think they're about 90% implemented. If anyone ever does have the motivation and time to finish them, I think the existing code's not necessarily useless. If we haven't documented somewhere that they don't work, then we should do that...

ZedThree: One of the blockers for v5, as far as I'm concerned at least, is refactoring `Datafile`, and that will include dealing with them one way or another!

configure.ac
ZedThree: Yes, that seems wise

ZedThree: It's now forced on at CHECK=3, ditto for CMake. Still requires `-v -v` to enable at runtime though

ZedThree:
Big warning to CMake users, all the options are now prefixed with `BOUT_`! This means we'll play nicer with other libraries/projects, as well as making it clear what the available options are for BOUT++ specifically. Also has the nice added advantage that `cmake-gui` groups all the `BOUT_` options together.

ZedThree:
The CMake option `GIT_SUBMODULE` is now `BOUT_UPDATE_GIT_SUBMODULE` to make it a bit clearer what it does

src/bout++.cxx
```diff
@@ -256,26 +263,39 @@ auto parseCommandLineArgs(int argc, char** argv) -> CommandLineArgs {
...
+      output.write(_("  -c, --color\t\t\tColor output using bout-log-color\n"));
 #endif
       output.write(
-          _("  -h, --help\t\tThis message\n"
-            "  restart [append]\tRestart the simulation. If append is specified, "
+          _("  --print-config\t\tPrint the compile-time configuration\n"
+            "  --list-solvers\t\tList the available time solvers\n"
+            "  --list-laplacians\t\tList the available Laplacian inversion solvers\n"
+            "  --list-laplacexz\t\tList the available LaplaceXZ inversion solvers\n"
+            "  --list-invertparts\t\tList the available InvertPar solvers\n"
```
bendudson, line 280: `invertparts` -> `invertpars`
```suggestion
            "  --list-invertpars\t\tList the available InvertPar solvers\n"
```

PR #2032, opened 2020-05-21
Add clang-format bot command for pull requests

ZedThree:
Simply have "/clang-format" as the first line in a comment to have clang-format automatically run over your PR! Commits are by `github-actions[bot]`. It uses `git-clang-format` to just format the changes in the PR.

I've tested this fairly extensively in a private repo, so I'm hoping it just works here. It needs to be merged into `master` and then it should Just Work. And it only took me this afternoon!

### DIFF
--- a/.github/workflows/clang-format-command.yml
+++ b/.github/workflows/clang-format-command.yml
@@ -1,0 +1,50 @@
+name: clang-format-command
+on:
+  repository_dispatch:
+    types: [clang-format-command]
+jobs:
+  clang-format:
+    runs-on: ubuntu-latest
+    steps:
+      # Checkout the pull request branch
+      - uses: actions/checkout@v2
+        with:
+          token: ${{ secrets.REPO_ACCESS_TOKEN }}
+          repository: ${{ github.event.client_payload.pull_request.head.repo.full_name }}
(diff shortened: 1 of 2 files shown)


.github/workflows/clang-format-command.yml
```diff
@@ -0,0 +1,50 @@
...
+          fi
+
+      # Execute clang-format and commit the change to the PR branch
+      - name: Commit to the PR branch
+        if: steps.format.outputs.format == 'true'
+        run: |
+          git clang-format-9 ${{ github.event.client_payload.pull_request.base.sha }}
+          git config --global user.name 'github-actions[bot]'
+          git config --global user.email '41898282+github-actions[bot]@users.noreply.github.com'
+          git commit -am "[clang-format-command] fixes"
```
dschwoerer, line 41: I just learned that there is the option `--ignore-rev` [1]

It might be worth to add it here :+1: 

[1] https://www.moxio.com/blog/43/ignoring-bulk-change-commits-with-git-blame

ZedThree: As in, add the commit to a `.git-blame-ignore-revs` file? I guess that's possible, although it would require two commits. Another possibility would be to just periodically dump commits from `github-actions[bot]` to that file.

PR #2035, opened 2020-05-23
Minor optimizations for ZHermiteSpline

johnomotani:
I tried running simulations to compare `shiftedmetric` and `shiftedmetricinterp`. I haven't got as far as comparing the results yet, but was expecting `shiftedmetricinterp` to run a bit faster, because of not needing to do FFTs. Actually, both seem to take similar numbers of iterations, but `shiftedmetricinterp` is slower - the whole simulation is about 10% slower, I haven't separated out the `toFieldAligned`/`fromFieldAligned` calls yet...

This PR is an attempt to optimize `ZHermiteSpline.interpolate()`, but doesn't seem to have made a noticeable difference.

Any suggestions welcome... I'm guessing this loop might be failing to vectorize?
https://github.com/boutproject/BOUT-dev/blob/83422038675977e361c63d7e270ea5a936c94302/src/mesh/interpolation/hermite_spline_z.cxx#L156-L179

The one other idea I had is that if we required an axisymmetric grid, we could use a `Field2D` index-shift instead of allowing the `Field3D` `k_corner`, and then there might be more optimizations possible. I'm not currently planning to run big simulations with `shiftedmetricinterp` that might make this worthwhile, and it would remove some functionality in the interests of performance, so I don't think it's worth pursuing (at least for the moment).

### DIFF
--- a/src/mesh/interpolation/hermite_spline_z.cxx
+++ b/src/mesh/interpolation/hermite_spline_z.cxx
@@ -32,11 +32,14 @@ ZHermiteSpline::ZHermiteSpline(BoutMask mask, int y_offset, Mesh* mesh)
       h10(localmesh), h11(localmesh) {
 
   // Index arrays contain guard cells in order to get subscripts right
-  k_corner.reallocate(localmesh->LocalNx, localmesh->LocalNy, localmesh->LocalNz);
+  const int n_total = localmesh->LocalNx*localmesh->LocalNy*localmesh->LocalNz;
+  k_corner.reallocate(n_total);
 
   // Initialise in order to avoid 'uninitialized value' errors from Valgrind when using
   // guard-cell values
-  k_corner = -1;
+  for (int i = 0; i < n_total; i++) {
+    k_corner[i] = -1;
+  }
 
   // Allocate Field3D members
   h00.allocate();
@@ -46,6 +49,8 @@ ZHermiteSpline::ZHermiteSpline(BoutMask mask, int y_offset, Mesh* mesh)
 }
 
 void ZHermiteSpline::calcWeights(const Field3D& delta_z, const std::string& region) {
+
+  const int ncz = localmesh->LocalNz;
 
   BOUT_FOR(i, delta_z.getRegion(region)) {
     const int x = i.x();
@@ -57,23 +62,20 @@ void ZHermiteSpline::calcWeights(const Field3D& delta_z, const std::string& regi
 
     // The integer part of zt_prime are the indices of the cell
     // containing the field line end-point
-    k_corner(x, y, z) = static_cast<int>(floor(delta_z(x, y, z)));
+    k_corner[i.ind] = static_cast<int>(floor(delta_z(x, y, z)));
 
     // t_z is the normalised coordinate \in [0,1) within the cell
     // calculated by taking the remainder of the floating point index
-    const BoutReal t_z = delta_z(x, y, z) - static_cast<BoutReal>(k_corner(x, y, z));
+    const BoutReal t_z = delta_z(x, y, z) - static_cast<BoutReal>(k_corner[i.ind]);
 
     // make k_corner be in the range 0<=k_corner<nz
-    const int ncz = localmesh->LocalNz;
-    BOUT_FOR(i, k_corner) {
-      k_corner[i] = ((k_corner[i] % ncz) + ncz) % ncz;
-    }
+    k_corner[i.ind] = ((k_corner[i.ind] % ncz) + ncz) % ncz;
 
     // Check that t_z is in range
     if ((t_z < 0.0) || (t_z > 1.0)) {
       throw BoutException(
           "t_z={:e} out of range at ({:d},{:d},{:d}) (delta_z={:e}, k_corner={:d})", t_z,
-          x, y, z, delta_z(x, y, z), k_corner(x, y, z));
+          x, y, z, delta_z(x, y, z), k_corner[i.ind]);
     }
 
     h00(x, y, z) = (2. * t_z * t_z * t_z) - (3. * t_z * t_z) + 1.;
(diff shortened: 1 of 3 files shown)


ZedThree:
`skip_mask` was used before we had `Region`. If we were to write these interpolations from scratch today, we'd just use `Region` instead. That's probably the next place to go.

src/mesh/interpolation/hermite_spline_z.cxx
```diff
@@ -32,11 +32,14 @@ ZHermiteSpline::ZHermiteSpline(BoutMask mask, int y_offset, Mesh* mesh)
...
-  k_corner.reallocate(localmesh->LocalNx, localmesh->LocalNy, localmesh->LocalNz);
+  const int n_total = localmesh->LocalNx*localmesh->LocalNy*localmesh->LocalNz;
+  k_corner.reallocate(n_total);
 
   // Initialise in order to avoid 'uninitialized value' errors from Valgrind when using
   // guard-cell values
-  k_corner = -1;
+  for (int i = 0; i < n_total; i++) {
+    k_corner[i] = -1;
+  }
```
ZedThree, line 42: Ah, this is because `Array<T>` doesn't have `operator=(T)`?

```suggestion
  std::fill(begin(k_corner), end(k_corner), -1);
```

src/mesh/interpolation/hermite_spline_z.cxx
```diff
@@ -125,32 +142,37 @@ Field3D ZHermiteSpline::interpolate(const Field3D& f, const std::string& region)
...
+        continue;
+    }
 
     // Due to lack of guard cells in z-direction, we need to ensure z-index
     // wraps around
-    const int ncz = localmesh->LocalNz;
-    const int z_mod = ((k_corner(x, y, z) % ncz) + ncz) % ncz;
-    const int z_mod_p1 = (z_mod + 1) % ncz;
+    const int z_mod = k_corner[i.ind];
+    const int z_mod_p1 = (z_mod < ncz - 1) ? (z_mod + 1) : 0;
```
ZedThree, line 169: What if made `k_corner` an `Array<Ind3D>`?

```suggestion
    const auto z_mod = k_corner[i.ind];
    const auto z_mod_p1 = z_mod.zp();
```

Similarly for `y_next`. Then `f_interp` calculation could be done entirely with single-indices

I think this moves some complexity into the construction of `k_corner`, but that's set-up cost so one-off.

johnomotani:
@ZedThree for my use-case I actually don't need a `skip_mask`. Agreed a custom `Region` is the long-term solution. My short term fix was to create two branches by templating the body of the `interpolate` method
https://github.com/boutproject/BOUT-dev/blob/34e5fff406391d51276e80377cfaf66644e09506/src/mesh/interpolation/hermite_spline_z.cxx#L129-L134
I think this should optimize out the conditional for `skip_mask` from the inside of the loop when I've never set one (so `has_mask = false`).

src/mesh/interpolation/hermite_spline_z.cxx
```diff
@@ -62,20 +62,23 @@ void ZHermiteSpline::calcWeights(const Field3D& delta_z, const std::string& regi
...
-    const BoutReal t_z = delta_z(x, y, z) - static_cast<BoutReal>(k_corner[i.ind]);
+    const BoutReal t_z = delta_z(x, y, z) - static_cast<BoutReal>(corner_zind);
+
+    // make corner_zind be in the range 0<=corner_zind<nz
+    corner_zind = ((corner_zind % ncz) + ncz) % ncz;
 
-    // make k_corner be in the range 0<=k_corner<nz
-    k_corner[i.ind] = ((k_corner[i.ind] % ncz) + ncz) % ncz;
+    // Convert z-index to Ind3D
+    k_corner[i.ind] = Ind3D((i.x()*ncy + i.y())*ncz + corner_zind, ncy, ncz);
```
ZedThree, line 75: This is because we want the `(x, y)` indices from the current loop iteration but we _know_ the z index?

src/mesh/interpolation/hermite_spline_z.cxx
```diff
@@ -62,20 +62,23 @@ void ZHermiteSpline::calcWeights(const Field3D& delta_z, const std::string& regi
...
-    k_corner[i.ind] = static_cast<int>(floor(delta_z(x, y, z)));
+    int corner_zind = floor(delta_z(x, y, z));
 
     // t_z is the normalised coordinate \in [0,1) within the cell
     // calculated by taking the remainder of the floating point index
-    const BoutReal t_z = delta_z(x, y, z) - static_cast<BoutReal>(k_corner[i.ind]);
+    const BoutReal t_z = delta_z(x, y, z) - static_cast<BoutReal>(corner_zind);
+
+    // make corner_zind be in the range 0<=corner_zind<nz
+    corner_zind = ((corner_zind % ncz) + ncz) % ncz;
```
ZedThree, line 72: Move into the initialisation of `corner_zind` to make it `const`?

src/mesh/interpolation/hermite_spline_z.cxx
```diff
@@ -165,16 +170,14 @@ Field3D ZHermiteSpline::interpolate_internal(const Field3D& f, const std::string
 
     // Due to lack of guard cells in z-direction, we need to ensure z-index
     // wraps around
-    const int z_mod = k_corner[i.ind];
-    const int z_mod_p1 = (z_mod < ncz - 1) ? (z_mod + 1) : 0;
-    const int y_next = y + y_offset;
+    const auto corner = k_corner[i.ind].yp(y_offset);
```
ZedThree, line 173: Comment on L171 now out-dated :smile:

src/mesh/interpolation/hermite_spline_z.cxx
johnomotani: Yes - basically we want `k_corner(x, y, z) = {x, y, floor(delta_z(x, y, z))}

I notice now I should have used `x` and `y` instead of `i.x()` and `i.y()`, since they're there already.

src/mesh/interpolation/hermite_spline_z.cxx
johnomotani: No, because we need the 'uncorrected' version to calculate `t_z` in l.69. I'll add a comment.

ZedThree: Ahh didn't spot that, thanks!

PR #2036, opened 2020-05-23
Travis: add Fedora job

dschwoerer:
Add job for fedora :smiley: 

It also adds a script to detect fedora, because one test was to slow.

I needed to change the detect travis script, because there was no un-versioned python command.

### DIFF
--- a/.travis_fedora.sh
+++ b/.travis_fedora.sh
@@ -44,8 +44,6 @@ then
     cp -a /tmp/BOUT-dev /home/test/
     chown -R test /home/test
     chmod u+rwX /home/test -R
-    restorecon -R /home/test || :
-    ls -alZ /home/test/BOUT-dev
     sudo -u test ${0/\/tmp/\/home\/test} $mpi
 ## If we are called as normal user, run test
 else
@@ -56,7 +54,7 @@ else
     cd
     cd BOUT-dev
     echo "starting configure"
-    time bash ./configure --with-petsc --enable-shared || cat config.log
+    time ./configure --with-petsc --enable-shared || cat config.log
     sed -e "s|-L/usr/lib64 ||g" -i make.config
     for f in tests/requirements/*[^y] ; do
 	echo -n "$f: "
--- a/.travis_fedora.sh
+++ b/.travis_fedora.sh
@@ -45,7 +45,6 @@ then
     chown -R test /home/test
     chmod u+rwX /home/test -R
     restorecon -R /home/test || :
-    ls -alZ /home/test
     ls -alZ /home/test/BOUT-dev
     sudo -u test ${0/\/tmp/\/home\/test} $mpi
 ## If we are called as normal user, run test
@@ -53,9 +52,9 @@ else
     . /etc/profile.d/modules.sh
     module load mpi/${1}-x86_64
     export OMPI_MCA_rmaps_base_oversubscribe=yes
+    export TRAVIS=true
     cd
     cd BOUT-dev
-    pwd || :
     echo "starting configure"
     time bash ./configure --with-petsc --enable-shared || cat config.log
     sed -e "s|-L/usr/lib64 ||g" -i make.config
--- a/.travis_fedora.sh
+++ b/.travis_fedora.sh
@@ -1,0 +1,68 @@
+#!/bin/sh
+## Copied script from arpack-ng. Please submit fixes also to
+## arpack-ng (if applicable)
+## -e : Make sure all errors cause the script to fail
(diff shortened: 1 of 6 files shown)


.travis_fedora.sh
```diff
@@ -0,0 +1,67 @@
...
+    # Ignore weak depencies
+    echo "install_weak_deps=False" >> /etc/dnf/dnf.conf
+    time dnf -y upgrade
+    time dnf -y install dnf-plugins-core {petsc,hdf5}-${mpi}-devel /usr/lib/rpm/redhat/redhat-hardened-cc1
+    time dnf -y builddep bout++
+    useradd test
+    cp -a /tmp/BOUT-dev /home/test/
+    chown -R test /home/test
+    chmod u+rwX /home/test -R
+    restorecon -R /home/test || :
```
ZedThree, line 47: `/tmp/BOUT-dev/.travis_fedora.sh: line 47: restorecon: command not found` [Travis](https://travis-ci.org/github/boutproject/BOUT-dev/jobs/690147764#L2219)

Is this important? I'm not familiar with `restorecon`

tests/integrated/test-multigrid_laplace/runtest
```diff
@@ -4,6 +4,9 @@
 # Run the test, check the error
 #
 
+# To slow on travis w. fedora
```
ZedThree, line 7: Typo:

```suggestion
# Too slow on travis with fedora
```

.travis_fedora.sh
```diff
@@ -0,0 +1,67 @@
...
+## If we are called as normal user, run test
+else
+    . /etc/profile.d/modules.sh
+    module load mpi/${1}-x86_64
+    export OMPI_MCA_rmaps_base_oversubscribe=yes
+    export TRAVIS=true
+    cd
+    cd BOUT-dev
+    echo "starting configure"
+    time bash ./configure --with-petsc --enable-shared || cat config.log
```
ZedThree, line 59: Is `bash` really necessary here? `configure` is supposed to be very shell-agnostic!

Personally I'd prefer we use CMake here too!

.travis_fedora.sh
dschwoerer: No, it is something selinux related. I had initially issues where configure failed. After I noticed `ls /` failed, even as root, i switch to a newer ubuntu, and it worked ...
Have removed it :+1:

.travis_fedora.sh
dschwoerer: Put it there for debugging, removed it now.

First I need to get used to using CMake. Is there a way to get tab completion to work?

ZedThree: Sure, that's reasonable! :smile: 

These files should probably be installed on your system somewhere: https://gitlab.kitware.com/cmake/cmake/-/tree/master/Auxiliary/bash-completion

I've only just discovered it and it seems to work quite well, so thanks for suggesting it!

Only slight annoyance is that it doesn't work with `cmake . -B <build dir>`, looks like you need to run `cmake ..` from the build directory for it to work

dschwoerer:
I think the issue is that you need to have run `cmake ..` in the past in
this directory. That makes it a bit of a pain ...
Also that breaks setting defaults based on other vars, as after the
first invocation they are already set, so the changed defaults aren't used.
So a workflow should probably be:
```
cmake || :
rm .* * ; cmake -D(tab)
```

I noticed only 2 flags are `BOUT_` prefixed - should they all start with
`BOUT_`. With tab completion cmake is so much easier to use :tada:

ZedThree:
Do you know which settings don't get updated? Most should be fine I think.

Oh, plan was to actually namespace everything in #2027, evidently I've not done that yet. Will fix!

dschwoerer:
On 5/26/20 2:33 PM, Peter Hill wrote:
> Do you know which settings don't get updated? Most should be fine I think.

BOUT_USE_SYSTEM_FMT:BOOL=ON
BOUT_USE_SYSTEM_MPARK_VARIANT:BOOL=ON
both stay OFF if called with -DGIT_SUBMODULE=OFF even if the
externalpackages have been purged.

More generally, maybe we should first check whether fmt and
mpark.variant is available, and only if that fails use the bundled ones?
I suppose installing them if the are already present is calling for trouble.

> Oh, plan was to actually namespace everything in #2027
> <https://github.com/boutproject/BOUT-dev/pull/2027>, evidently I've not
> done that yet. Will fix!

Tjamls

ZedThree:
Ah ok, think I've got that fixed with `cmake_dependent_option`!

PR #2037, opened 2020-05-24
filledFrom using lambda function

bendudson:
Simplifies a common pattern
```
Field3D result{emptyFrom(field)};
BOUT_FOR(i, result) {
  result[i] = ...;
}
```
and allows it to be used as an expression:
```
auto result = filledFrom(field, [&](const auto& i) {
                                  return ...;
                                });
```

Examples of use: https://github.com/bendudson/hermes-3/blob/master/src/collisions.cxx#L106

### DIFF
--- a/include/field.hxx
+++ b/include/field.hxx
@@ -251,6 +251,27 @@ inline T filledFrom(const T& f, BoutReal fill_value) {
   return result;
 }
 
+/// Return a field of some type derived from Field, with metadata copied from
+/// another field and a data array allocated and filled using a callable e.g. lambda function
+///
+/// e.g.
+///   Field3D result = filledFrom(field, [&](const auto& index) {
+///                                          return ...;
+///                                      });
+/// 
+/// An optional third argument is the region string
+template <
+    typename T, typename Function,
+    typename = decltype(std::declval<Function&>()(std::declval<typename T::ind_type&>()))>
+inline T filledFrom(const T& f, Function func, std::string region_string = "RGN_ALL") {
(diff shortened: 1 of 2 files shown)


include/field.hxx
```diff
@@ -251,6 +251,27 @@ inline T filledFrom(const T& f, BoutReal fill_value) {
...
+///
+/// e.g.
+///   Field3D result = filledFrom(field, [&](const auto& index) {
+///                                          return ...;
+///                                      });
+/// 
+/// An optional third argument is the region string
+template <
+    typename T, typename Function,
+    typename = decltype(std::declval<Function&>()(std::declval<typename T::ind_type&>()))>
```
ZedThree, line 265: Ahh, C++! :smile: The only bit I'm not sure of is the reference in `declval<Function&>` -- what am I missing?

bendudson: I'm also a bit unclear on this magic incantation. It's adapted from https://stackoverflow.com/questions/50859115/enable-if-template-param-is-lambda-with-particular-signature

ZedThree: Oh, I see, it's to guard against strange types! If `F` had a call operator like `void operator(ind_type) &&`, which is _rvalue_ qualified, it can only be called on a temporary. But we definitely have an _lvalue_ here, that is, a named variable. `F&` is a mutable reference, so it must be an _lvalue_, so it ensures that this will definitely be the right `operator()` overload.

Phew!

This might be a FIXME for when we get to move to C++17 and it can be `std::is_invocable`

PR #2040, opened 2020-05-27
Minor improvements and fixes to LaplaceIPT

ZedThree:
A couple of important fixes, and some other minor changes because I can't help myself:

- Fix a bug in `is_diagonally_dominant` where non-`included` levels were taking part
- Add new solver to CMake
- Make sure to pass a few things by `const&`
- Make `Level::init` overloads constructors
- Add some helper functions for reused bits
- Various fixes suggested by `clang-tidy`

### DIFF
--- a/src/invert/laplace/impls/iterative_parallel_tri/iterative_parallel_tri.hxx
+++ b/src/invert/laplace/impls/iterative_parallel_tri/iterative_parallel_tri.hxx
@@ -222,13 +222,13 @@ private:
   /// First and last interior points xstart, xend
   int xs, xe;
 
-  bool isGlobalFlagSet(int flag) {
+  bool isGlobalFlagSet(int flag) const {
     return (global_flags & flag) != 0;
   }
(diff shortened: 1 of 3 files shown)


src/invert/laplace/impls/iterative_parallel_tri/iterative_parallel_tri.hxx
```diff
@@ -195,17 +214,23 @@ private:
...
 
+  bool isGlobalFlagSet(int flag) {
+    return (global_flags & flag) != 0;
+  }
+  bool isInnerBoundaryFlagSet(int flag) {
+    return (inner_boundary_flags & flag) != 0;
+  }
+  bool isOuterBoundaryFlagSet(int flag) {
+    return (outer_boundary_flags & flag) != 0;
+  }
```
ZedThree, line 233: These helper functions might actually be more generally useful, and could maybe be moved to `Laplace` base class? Could then make the flag members `private` rather than `protected`

And I've just noticed these methods should all be `const` too

PR #2048, opened 2020-05-31
Use C++ style casts

dschwoerer:
/clang-format

Part of #2025 

Doesn't change anything, just makes it more explicit.

### DIFF
--- a/src/sys/boutexception.cxx
+++ b/src/sys/boutexception.cxx
@@ -53,8 +53,8 @@ std::string BoutException::getBacktrace() const {
     void * ptr=trace[i];
     if (dladdr(trace[i],&info)){
       // Additionally, check whether this is the default offset for an executable
-      if (info.dli_fbase != (void*)0x400000)
-        ptr=(void*) ((size_t)trace[i]-(size_t)info.dli_fbase);
+      if (info.dli_fbase != reinterpret_cast<void*>(0x400000))
+        ptr=reinterpret_cast<void*>(reinterpret_cast<size_t>(trace[i])-reinterpret_cast<size_t>(info.dli_fbase)));
     }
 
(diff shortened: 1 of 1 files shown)


ZedThree:
Just FYI, the clang-format command did work, but there was nothing to do. Maybe I should add some feedback for that case?

src/sys/boutexception.cxx
```diff
@@ -53,8 +53,8 @@ std::string BoutException::getBacktrace() const {
     void * ptr=trace[i];
     if (dladdr(trace[i],&info)){
       // Additionally, check whether this is the default offset for an executable
-      if (info.dli_fbase != (void*)0x400000)
-        ptr=(void*) ((size_t)trace[i]-(size_t)info.dli_fbase);
+      if (info.dli_fbase != reinterpret_cast<void*>(0x400000))
+        ptr=reinterpret_cast<void*>(reinterpret_cast<size_t>(trace[i])-reinterpret_cast<size_t>(info.dli_fbase)));
```
ZedThree, line 57: One too many closing brackets here!

```suggestion
        ptr=reinterpret_cast<void*>(reinterpret_cast<size_t>(trace[i])-reinterpret_cast<size_t>(info.dli_fbase));
```

dschwoerer:
/clang-format

Did the first one work? On the second there are emojis, so that seems to have worked, but the first one not?

Also I think the review with `/clang-format` didn't work :sweat_smile: 
Still think it is a great way to ensure all can use the same c`lang-format` :+1:

ZedThree:
I don't think the first one triggered it. The second one did trigger it but I think the syntax error fooled it, so it errored. The :tada: emoji on the third one indicates it finished successfully, although it says it was from me, rather than the bot!

Glad you think it's useful! I'll see if there's any polish can be done

PR #2058, opened 2020-07-07
Enable staggered versions of SplitFluxDerivativeType (master)

johnomotani:
`C2L` and `L2C` variants need to be explicitly included in the `Set<>` passed to `produceCombinations`.


ZedThree:
Was there a reason we didn't do this initially?

johnomotani:
> Was there a reason we didn't do this initially?

I suspect the line was copy-pasted from the FFT derivatives just above, which do not have a staggered implementation (yet). I've run a test simulation using a staggered version of split, which runs with only small errors that I'd expect (the operators were slightly changed since the expected data was generated).

PR #2060, opened 2020-07-07
Enable staggered versions of SplitFluxDerivativeType

johnomotani:
Apply fix from #2058 to `next`.


johnomotani:
Oops, got the base wrong when I created the PR. Fixed now, but that was why the notification mail had so many commits - only 1 plus a merge commit now!

PR #2062, opened 2020-07-24
Fix default file format when using legacy netcdf

ZedThree:


ZedThree:
@bshanahan Please could you check this works for you?

dschwoerer:
I am not sure legacy netcdf is the best name.
In some cases it would be nice to have "netcdf" as "any netcdf" - for example in the `bout-config` file (but also #2010)

It also breaks on ubuntu 18.04, failure from `test-delp2`:
```
        Option diffusion:useFFT = true (Command line)
Setting boundary for variable n
        core region:    Option all:bndry_all = dirichlet (data/BOUT.inp)

        sol region:     Option all:bndry_all = dirichlet (data/BOUT.inp)

        Option input:transform_from_field_aligned = true (default)
        Option input:max_recursion_depth = 0 (default)
        Option n:function = gauss(x-0.5, 0.2) * gauss(z - pi, 0.2*pi) (data/BOUT.inp)
        Option n:scale = 0.1 (data/BOUT.inp)
        Option all:evolve_bndry = 0 (default)
        Option n:evolve_bndry = 0 (default)
        Option datadir = data ()
        Option dump_format = nc (default)
Error encountered
====== Exception path ======
[bt] #6 ./test_delp2(+0x2cf2a) [0x562f7da9bf2a]
_start at ??:?
[bt] #5 /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xe7) [0x7fc7cefd8b97]
__libc_start_main at /build/glibc-2ORdQG/glibc-2.27/csu/../csu/libc-start.c:344
[bt] #4 ./test_delp2(+0x28972) [0x562f7da97972]
main at /usr/include/c++/7/bits/unique_ptr.h:821
 (inlined by) main at /home/IPP-HGW/dave/soft/BOUT-dev/tests/integrated/test-delp2/test_delp2.cxx:30
[bt] #3 ./test_delp2(+0x16bb2d) [0x562f7dbdab2d]
Solver::setModel(PhysicsModel*) at /home/IPP-HGW/dave/soft/BOUT-dev/src/solver/../../include/bout/physicsmodel.hxx:90
 (inlined by) Solver::setModel(PhysicsModel*) at /home/IPP-HGW/dave/soft/BOUT-dev/src/solver/solver.cxx:92
[bt] #2 ./test_delp2(+0x164fd7) [0x562f7dbd3fd7]
PhysicsModel::postInit(bool) at /usr/include/c++/7/bits/basic_string.h:211
 (inlined by) ?? at /usr/include/c++/7/bits/basic_string.h:737
 (inlined by) ?? at /home/IPP-HGW/dave/soft/BOUT-dev/src/physics/../../include/options.hxx:581
 (inlined by) PhysicsModel::postInit(bool) at /home/IPP-HGW/dave/soft/BOUT-dev/src/physics/physicsmodel.cxx:113
[bt] #1 ./test_delp2(+0x8d077) [0x562f7dafc077]
std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > Options::withDefault<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >) at /usr/include/c++/7/bits/basic_string.h:211
 (inlined by) ?? at /usr/include/c++/7/bits/basic_string.h:220
 (inlined by) ?? at /usr/include/c++/7/bits/basic_string.h:647
 (inlined by) ?? at /home/IPP-HGW/dave/soft/BOUT-dev/src/mesh/../../include/bout/../boutexception.hxx:28
 (inlined by) std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > Options::withDefault<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >) at /home/IPP-HGW/dave/soft/BOUT-dev/src/mesh/../../include/options.hxx:430
====== Back trace ======
 -> PhysicsModel::postInit on line 94 of 'physicsmodel.cxx'

Inconsistent default values for 'dump_format': 'nc' then 'h5'
```

Can we rename so that `netcdf` is "any netcdf" and then maybe have a better name then `legacy` - otherwise if there is a new netcdf version it will be confusing.

dschwoerer:
I have theses `netcdf`s installed:
```
libnetcdf-c++4/bionic,now 4.2-8 amd64 [installed,automatic]
libnetcdf-cxx-legacy-dev/bionic,now 4.2-8 amd64 [installed]
libnetcdf-dev/bionic,now 1:4.6.0-2build1 amd64 [installed]
libnetcdf13/bionic,now 1:4.6.0-2build1 amd64 [installed,automatic]
netcdf-bin/bionic,now 1:4.6.0-2build1 amd64 [installed]
python3-netcdf4/bionic,now 1.3.1-1 amd64 [installed]
```

ZedThree:
Maybe something like `bout::build::has_netcdf` (== any netcdf) and `bout::build::use_legacy_netcdf`. I'd prefer to stick with "legacy", as it's the official name for that release: https://www.unidata.ucar.edu/software/netcdf/release-notes-cxx-4.2.html

I do want to eventually remove the legacy version completely -- it's almost ten years old now! Plus maintaining two separate versions is not great.

That error is because we've repeated setting the default file format. I'll fix that.

Also, just to clarify the different versions that are available, because the naming is a bit of a mess:

- `libnetcdf13`/`libnetcdf-dev` is the C API, v4.6.0 -- latest is v4.7.4
- `libnetcdf-c++4`/`libnetcdf-cxx-legacy-dev` is the legacy C++ API, v4.2, provided for backwards compatibility shortly before the release of v4.2 of the C API, when they moved the Fortran and C++ APIs out of the main library
- `libnetcdf-c++4-1`/`libnetcdf-c++4-dev` is the "new" C++ API, v4.3.0 in Bionic -- latest is v4.3.1

Other distros may call the C++ API something like `netcdf-cxx4` . In BOUT++, the legacy version is called `NcFormat`, and the new one is `Ncxx4`

bshanahan:
Still broken for me.

ZedThree:
Please could you post the error message and/or config.log? I'm afraid I don't recall what the original issue was

bshanahan:
Sorry, the original problem is that configure did not recognize netcdf, so switched to h5.  This has been fixed.  I got confused because today I mentioned on slack that my code breaks if I try and collect while a simulation is running.  This PR does not fix that problem.

ZedThree:
`bout::build::has_netcdf` is now true for either interface. I've kept `bout::build::has_legacy_netcdf` to distinguish between the two.

Also (hopefully) fixed the problem @dschwoerer reported with `test-delp2` by just removing the default values (& upgrading to the new `Options` API at the same time, as it's much cleaner)

src/physics/physicsmodel.cxx
```diff
@@ -97,23 +97,17 @@ int PhysicsModel::postInit(bool restarting) {
...
-  options->get("restart_format", restart_ext, dump_ext);
+  auto& options = Options::root();
+
+  const std::string restart_dir = options["restartdir"]
+                                      .doc("Directory for restart files")
+                                      .withDefault(options["datadir"]);
+
+  const std::string restart_ext = options["restart_format"]
+                                      .doc("Restart file extension")
+                                      .withDefault(options["dump_format"]);
```
bendudson, line 108: Is it worth noting that we're assuming that `options["dump_format"]` has been set?

bendudson (dismissed):

johnomotani:
I checked the setup in #2182 again, and it's skipping lots of tests with `netcdf => False`. I configured just with `./configure`, and in the summary got
```
configure:   NetCDF support          : no (legacy: yes)
```
I'd expect the `netcdf` condition to be true, even though it's legacy netcdf.

src/physics/physicsmodel.cxx
ZedThree: Yes, good idea. It's only an issue if `BoutInitialise` isn't called, which is pretty rare

ZedThree:
@johnomotani Please can you check now?

johnomotani:
Unfortunately not, I still get the same symptoms as my previous comment. Here's the [config.log](https://github.com/boutproject/BOUT-dev/files/5766242/config.log) in case that helps.

ZedThree:
I found a better way of testing it on my machine -- this should actually work now!

johnomotani:
Ah, now it's trying to compile Ncxx4 and OptionNetCDF because `BOUT_HAS_NETCDF` is true, and failing because it only has legacy netCDF...

ZedThree:
:facepalm: I guess this worked for me because the netcdf-cxx4 headers are in a default search path, so the compiler could find them and it worked. I uninstalled it completely and the latest commit should actually, really, work this time!

johnomotani:
There are some tests for OptionsNetcdf that need to be switched off too: `tests/unit/sys/test_options_netcdf.cxx` and `tests/integrated/test-options-netcdf`. The unit tests would be easy, just change like 047ffa7, but the integrated test I think needs us to be able to do `#requires: not legacy_netcdf`, which isn't currently available.

johnomotani:
With a couple more tiny changes, this now works for me. I had to skip `test-io`, which is expected to fail with legacy netcdf. Also needed to add brackets to the `#requires` in `test-options-netcdf`, then realised that we don't need to check for Travis any more, and the test does pass on Fedora, so removed those checks.

johnomotani (approved):

PR #2064, opened 2020-08-06
Allow descriptions of output variables; save some diagnostics for solvers

johnomotani:
Descriptions of output variables
------------------------------------------
Adds an optional argument to `Datafile::add()`, `Solver::add()` and `bout_solve()` which can be passed a description as a `std::string`. The description is then added to the output files as an attribute of the variable. This should be useful for making the output files self-documenting. This PR adds the mechanism, but variables written by BOUT++ still need a `description` adding in a future PR (apart from the new solver diagnostic variables described below).

Save some diagnostics for solvers
----------------------------------------------
When `solver:diagnose=true` is set, CVODE and ARKODE print out some useful information. This PR saves those numbers to the dump files to make them easier to access in post-processing.

The diagnostics are always saved, regardless of `solver:diagnose` setting. The setting is useful to avoid cluttering up stdout, but I don't see any benefit to not saving into dump files - the diagnostics are scalars so don't take up much space, and it's nice to be able to go back and look at them without having decided to set the setting before the run.

The diagnostic variables have descriptions copied from the CVODE and ARKODE manuals which are written as attributes using the new `description` argument described above. These help make the output files a little less cryptic, since SUNDIALS variable names are not very descriptive (e.g. `nliters`).

- [x] add documentation of the 'description' arguments of `Datafile::add()` and `Solver::add()` to the manual.


johnomotani:
This PR implements descriptions for output variables, which would be useful for #2065, because I wanted to add descriptions of the diagnostic variables here. The description-handling parts (09b5fa2 and 6fbf9cb) could be split into a separate PR easily if that helps.

BTW the way this is implemented means `PhysicsModel`s can also add descriptions, with the new, optional 'description' arguments to `Datafile::add()` and `Solver::add()`. Should probably add some description of that in the manual, but let's agree on the interface design first...

johnomotani:
I think I fixed the memory errors. Had accidentally passed `int` or `BoutReal` by value before storing a pointer to the address.

johnomotani:
What's the clang-tidy bot supposed to do if you 'fix' the things it suggests? It doesn't seem to have removed the comments like I'd hope. The changes I made weren't directly what it suggested though (I passed by reference instead of using a `std::move`), so that might be why?

ZedThree:
No, I'm afraid it's not as clever as that. For now you can just click
"resolve" on the comment. I can probably make it do that automatically if
there's an API for it though

johnomotani:
> No, I'm afraid it's not as clever as that. For now you can just click
"resolve" on the comment. I can probably make it do that automatically if
there's an API for it though

Thanks @ZedThree! The PEP8 bot somehow does this, but I don't know how. Would be nice, but clicking 'resolve' on addressed issues is fine with me - I was just wondering if there was a feature that wasn't working right.

ZedThree:
I will investigate the PEP8 bot and steal features from it 😄

bendudson (approved):

PR #2071, opened 2020-08-11
Correct Grad2_par2 implementation in InvertParCR

johnomotani:
The 'B' coefficient of `InvertPar` is supposed to be the coefficient of `Grad2_par2`, but previously only had the `D2DY2` contribution and not the `DDY` contribution - compare the `sg` terms in the implementation of `Grad2_par2`
https://github.com/boutproject/BOUT-dev/blob/0448a8d12d5cf54b9dc75bcc7557af2e698dc86a/src/mesh/coordinates.cxx#L1357-L1377

Does change results compared to the existing code, but I assume nobody is relying on incorrect results... The LLNL bunch are probably the only ones using this heavily for their Landau-fluid closure - anyone have a clue if the change would have any impact on that? @xxu @umansky


bendudson:
Thanks John! Yes, agreed this looks odd. Also looking at the d/dz terms something doesn't look quite right: The coefficients are divided by dz 
https://github.com/boutproject/BOUT-dev/blob/c4e0fb7ec879180ff9aac8ac1caa72f40ec5fd3c/src/invert/parderiv/impls/cyclic/cyclic.cxx#L148
but then also multiplied by kwave which includes a Z length
https://github.com/boutproject/BOUT-dev/blob/c4e0fb7ec879180ff9aac8ac1caa72f40ec5fd3c/src/invert/parderiv/impls/cyclic/cyclic.cxx#L154

johnomotani:
Yes, the `D2DZ2` implementation does divide by `SQ(dz)`, but in `indexDDZ()` uses
https://github.com/boutproject/BOUT-dev/blob/58f130aa57e288c8fdcd408280fdf0ea63bc8ca7/src/mesh/index_derivs.cxx#L2044
so `kwave` is defined by dividing by `nz` rather than `zlength()`.

@bendudson do you think the best fix is to delete the ` / dz`?

bendudson:
If fixed, this might also improve preconditioners which often rely on this operator.

bendudson:
We could remove the `/ dz` lines, and then update the comment on this line:
https://github.com/boutproject/BOUT-dev/blob/c4e0fb7ec879180ff9aac8ac1caa72f40ec5fd3c/src/invert/parderiv/impls/cyclic/cyclic.cxx#L134
since it's then not 1/rad but 1/length

bendudson:
Sorry it looks like this was a bug in my original version, and lack of proper testing didn't catch it. Mea culpa. Thanks for checking!

johnomotani:
I think we might be able to catch the `dz` bug just by adding z-dependence to the input function in `test-invpar`... will have a look now.

bendudson:
The test currently uses a binary grid file:
https://github.com/boutproject/BOUT-dev/blob/master/tests/integrated/test-invpar/data/BOUT.inp#L8
but could be modified to use a grid with varying `dy`.

It looks like the tests do include cases with Z derivatives (https://github.com/boutproject/BOUT-dev/blob/master/tests/integrated/test-invpar/runtest#L28) but in that case the function defaults to `sin(2*y)` so the Z derivatives are zero. It's also possible that `dz` is set to 1. Definitely more thorough tests needed.

johnomotani:
I've removed the grid file from `test-invpar` and made it test the z-derivative terms. The test now passes, so I'm happy that the fix for the bug with dividing by `dz` that Ben found is correct. So the z-derivative terms in `InvertPar` have always been broken! I hope nobody was using them...

It took me hours to figure out why the test wouldn't pass until I realised that it wasn't using the FFT z-derivatives unless I explicitly ask for them in the options - not sure when that default changed!

johnomotani:
Should backport this PR to master and v4.4.0. I'll make PRs once everyone's happy with this one.

bendudson:
Adding to the BOUT.inp file:
```
dz = 0.3
g_22 = 1.1 + 0.2*cos(y)
g22 = 1 / g_22
```
and everything now passes, but taking out John's fixes causes it to fail. Strangely however, adding a variation of dy causes failure:
```
dy = 1.2 + 0.2*cos(y)
```

bendudson:
D2DY2 inside Grad2_par2 also includes a correction for non-uniform dy if non_uniform is true (which is the default now). Setting option non_uniform=false with varying dy and the tests pass. Looks like another correction needed, but this one less important because dy is usually constant (e.g. in tokamak grids).

bendudson:
The test now includes all kinds of variations, and I think is now properly checking that the forward and inverse operations are consistent. Thanks @johnomotani !

johnomotani:
@bendudson sorry about that, didn't see you'd made the fix before I pushed. I've removed my changes, but dismissed the review by mistake.

johnomotani:
One of the test fails was due to floating point exceptions when --sigfpe was turned on. This was due to calculating in the x-guard/x-boundary cells, combined with the change in this PR to include non-constant g_22 corrections which introduced NaNs because `DDY(1/sqrt(g_22)` is not calculated in guard cells.

I've fixed this by making `SurfaceIter` (by default) only iterate over the non-guard points. As far as I can see the only place `SurfaceIter` is used in the library is in `InvertParCR` so it should be fine to change the default behaviour. If `SurfaceIter` is used somewhere in user code, they probably don't really want guard cells either. Including or excluding guard cells is set by a new argument to the `SurfaceIter` constructor, so it's possible to include them if needed. Does anyone have a preference for keeping the old behaviour by default? Arguably it'd be better to change the default in `next` (since the new one seems more sensible) and not in `master` (to be backward compatible), but I'm inclined to change the default in master anyway because it's not likely to affect anyone.

johnomotani:
Seem to have fixed the tests except the Fedora issue #2073 - Should be ready to merge once a fix for that goes into `next` and we merge here.

johnomotani:
Only the codecov test failing now, so I think this is ready for review/merge.

ZedThree:
Failing checks are the codecov checks which turned themselves on without asking

src/invert/parderiv/impls/cyclic/cyclic.cxx
```diff
@@ -120,20 +130,25 @@ const Field3D InvertParCR::solve(const Field3D &f) {
...
+            E(x, y + localmesh->ystart)
+            + sg(x, y + localmesh->ystart)*B(x, y + localmesh->ystart);          // ddy
+
+        if (coord->non_uniform) {
+          ecoef += bcoef * coord->d1_dy(x, y + localmesh->ystart);
+        }
 
         bcoef /= SQ(coord->dy(x, y + localmesh->ystart));
-        ccoef /= coord->dy(x, y + localmesh->ystart) * coord->dz;
-        dcoef /= SQ(coord->dz);
```
bshanahan, line 136: Where did this line with dcoef go?

bendudson: The `dz` factor is already included in the definition of `kwave` used here, which uses `zlength`

bshanahan: Okay, thanks.

PR #2074, opened 2020-08-12
CI: Turn off patch coverage check

ZedThree:
Fixes some PRs failing because Codecov has decided to enable itself as a check.

Leaves project coverage check enabled, but at a target we can actually hit. We should re-enable the patch coverage check when we get coverage high enough.

This might need to also go into master.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -5,3 +5,10 @@ ignore:
 
 codecov:
   branch: next
+
+coverage:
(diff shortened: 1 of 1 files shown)


ZedThree:
Failing test is the Fedora job which is fixed in next (as seen in the PR check). This PR is also purely infrastructure, so it won't affect any code

PR #2077, opened 2020-08-12
Correct Grad2_par2 implementation in InvertParCR (v4.4.0)

johnomotani:
Merge the fixes from #2071 into `v4.4.0-alpha`. I've used the same branch as for merging them into `master`, which is based on `master` so will bring a couple of extra things in to v4.4.0-alpha, but I guess that doesn't hurt and it avoids duplicate commits being added to the history.


ZedThree:
I think the slash-command thing was triggered by a merge commit with a slash in it, which is not useful, so I'm merging anyway

PR #2084, opened 2020-08-20
Print more useful message when monitor fails/signals to quit

bendudson:
Error message wasn't being printed when Solver monitor returned an error code or threw an exception, making debugging difficult. Now prints error message before propagating error upwards.


ZedThree (dismissed):

ZedThree:
Could be backported to v4.4?

ZedThree:
@bendudson Just checking this is supposed to go into the 3D metrics branch, rather than next?

PR #2087, opened 2020-08-25
Staggered grids in InvertPar

johnomotani:
Allows `InvertPar` to work on staggered grids.

`CELL_YLOW` version requires changes to the boundary conditions. The Neumann boundary conditions have to be put in with one-sided differences, which (because the matrix in `InvertParCR` is tri-diagonal) are only first-order accurate. As long as `InvertPar` is only being used for preconditioners, hopefully this should not be a problem.

Also addresses a comment above the `InvertPar::Create()` method that it should be called `create` and take `Options*` (and now `CELL_LOC`) arguments.

Adds tests both for staggered grids and on open field lines (which were not previously tested).

- [x] `InvertPar::create()` supported in `bin/bout-b5-factory-upgrader.py`

### DIFF
--- a/bin/bout-v5-factory-upgrader.py
+++ b/bin/bout-v5-factory-upgrader.py
@@ -18,6 +18,7 @@ factories = {
         "type_name": "InvertPar",
         "create_method": "create",
         "old_create_method": "Create",
+        "arguments_changed": True,
     },
     "Mesh": {"factory_name": "Mesh", "type_name": "Mesh", "create_method": "Create"},
     "Laplacian": {
@@ -178,11 +179,22 @@ def fix_create_method(factory, source):
 
     if "old_create_method" not in factory:
         return source
+    old_create_pattern = re.compile(
+        r"({factory_name})\s*::\s*{old_create_method}\b".format(**factory)
+    )
+    if not old_create_pattern.findall(source):
+        return source
+
+    if factory.get("arguments_changed", False):
+        print(
+            "**WARNING** Arguments of {factory_name}::{create_method} have changed, and your current arguments may not work."
+            " Please consult the documentation for the new arguments.".format(**factory)
+        )
+
     return re.sub(
         r"({factory_name})\s*::\s*{old_create_method}\b".format(**factory),
         r"\1::{create_method}".format(**factory),
         source,
-        flags=re.VERBOSE,
     )
 
 
--- a/bin/bout-v5-factory-upgrader.py
+++ b/bin/bout-v5-factory-upgrader.py
@@ -16,7 +16,8 @@ factories = {
     "InvertPar": {
         "factory_name": "InvertPar",
         "type_name": "InvertPar",
-        "create_method": "Create",
+        "create_method": "create",
+        "old_create_method": "Create",
     },
     "Mesh": {"factory_name": "Mesh", "type_name": "Mesh", "create_method": "Create"},
     "Laplacian": {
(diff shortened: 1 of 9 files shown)


johnomotani:
I keep forgetting - do we have a list of breaking changes that I should add `InvertPar::Create`->`InvertPar::create` to? It should probably go into the `bout_4to5` script too I guess. @ZedThree can you help with that?

johnomotani:
The fedora test is failing :-(
It's failing test-petsc-laplace3d, which shouldn't be affected by any of the changes here. Could be a fedora and/or PETSc-version issue? I think the failure is a result only marginally over the tolerance anyway.

I don't have a fedora system to test on. Could someone please check the latest `next` on fedora rawhide? If it's not this PR, then `next` should have the same issue. @dschwoerer @ZedThree?

ZedThree:
I just merged a couple of PRs, I guess we'll see if the problem is also in `next` quite soon! I will also try to check Fedora locally, if I can remember how to an interactive session going in docker/podman.

[CHANGELOG.md](https://github.com/boutproject/BOUT-dev/blob/master/CHANGELOG.md) has a `Breaking Changes` section. There's also `bin/bout-b5-factory-upgrader.py` that's probably the best place to fix `Create` -> `create`. I'll try to do that today.

ZedThree (approved):

bin/bout-v5-factory-upgrader.py
```diff
@@ -170,6 +171,21 @@ def fix_deletions(variables, source):
...
+def fix_create_method(factory, source):
+    """Fix change of name of factory `create` method
+
+    """
+
+    if "old_create_method" not in factory:
+        return source
+    return re.sub(
+        r"({factory_name})\s*::\s*{old_create_method}\b".format(**factory),
+        r"\1::{create_method}".format(**factory),
```
johnomotani, line 183: Can we check that there's no argument to the `Create` method? The old version could take a `Mesh*` argument, which I'd be surprised if anyone was using, but if they were it'd be an error to pass as the first argument to the new `create` (which is an `Options*`). Don't think it's worth bothering to fix if there was an argument, but raising an error or a warning might be nice?

johnomotani (commented):
Looks good, thanks @ZedThree!

bin/bout-v5-factory-upgrader.py
ZedThree: Done. It's a bit dumb, in that it always warns if a factory where the arguments have changed is used. Hopefully the documentation is sufficiently useful for users to know what to do, otherwise we'll have to rely on the compiler. For me, using gcc 10.2, I get the following error (with nice colours too):

```
../tests/integrated/test-invpar/test_invpar.cxx: In function ‘int main(int, char**)’:
../tests/integrated/test-invpar/test_invpar.cxx:33:32: error: cannot convert ‘Mesh*’ to ‘Options*’
   33 |   auto inv = InvertPar::create(mesh);
      |                                ^~~~
      |                                |
      |                                Mesh*
In file included from ../tests/integrated/test-invpar/test_invpar.cxx:8:
../include/invert_parderiv.hxx:104:53: note:   initializing argument 1 of ‘static std::unique_ptr<InvertPar> InvertPar::create(Options*, CELL_LOC, Mesh*)’
  104 |   static std::unique_ptr<InvertPar> create(Options *opt_in = nullptr,
      |                                            ~~~~~~~~~^~~~~~~~~~~~~~~~
```

ZedThree (approved):

bin/bout-v5-factory-upgrader.py
johnomotani: looks good to me, thanks @ZedThree!

PR #2093, opened 2020-09-09
Handle FieldPerps in Datafile::varAdded() and Datafile::varPtr()

johnomotani:
Qian Xia noticed a test failure in `master` on Freia. `test-io` failed when double-adding variables. Turns out there was a bug where `FieldPerp` variables were not handled in `varAdded()` and `varPtr()`, so they were not skipped when added the second time.

Not sure what changed to make the test fail now when it didn't before, and doesn't on Travis. The missing `FieldPerp` support has been there since #1699, and should have been a bug in a tested feature since #1876.

### DIFF
--- a/tests/integrated/test-io/runtest
+++ b/tests/integrated/test-io/runtest
@@ -140,6 +140,7 @@ for check_incorrect_add in [
     except RuntimeError as e:
         if not (
             "Variable with name '" + check_incorrect_add + "' already added to Datafile"
+            == str(e)
         ):
             print(
                 "incorrect error message for check_incorrect_add=" + check_incorrect_add
--- a/tests/integrated/test-io/runtest
+++ b/tests/integrated/test-io/runtest
@@ -119,6 +119,40 @@ print("Checking with variables added twice")
 if not check_output():
     success = False
 
+# Test incorrectly double-adding variables - should throw exception
+for check_incorrect_add in [
+    "ivar",
+    "rvar",
+    "bvar",
+    "f2d",
+    "f3d",
+    "fperp",
+    "ivar_evol",
+    "rvar_evol",
+    "bvar_evol",
+    "v2d_evol",
+    "v3d_evol"
+]:
+    try:
+        s, out = launch_safe(
+            "./test_io check_incorrect_add="+check_incorrect_add, nproc=nproc, pipe=True
+        )
+    except RuntimeError as e:
+        if not (
+            "Variable with name '" + check_incorrect_add + "' already added to Datafile"
+        ):
+            print(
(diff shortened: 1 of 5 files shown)


ZedThree (approved):

ZedThree:
Pretty worrying that we have a test that should fail and isn't!

johnomotani:
It's not totally obvious to me that the test should fail - if a variable is added twice, I think it might depend how the library deals with it. Sometimes it might be OK if the library manages to just write once and then over-write the second time. When the test failed, the second time seemed to be appending to the first one, so we ended up with 6 points in the t-direction when the test only has 3 time-steps.

ZedThree:
Ah I see the issue with the test now, we check that adding the _same_ variable twice is _ok_, when we should be checking that adding _different_ variables twice _fails_

johnomotani:
> Ah I see the issue with the test now, we check that adding the same variable twice is ok, when we should be checking that adding different variables twice fails

The double-add test is testing what I intended it to test: if the exact same variable is added twice for some reason, there is no error. That was what was broken for `FieldPerp`s.
[I think my motivation for the feature and the test was if we have some variable which for some options in added to the solver, and in others is just added with `SAVE_REPEAT`, it might have `dump.add()` called twice and the least surprising thing to do then is to add it once and print a warning.]

But you're right, if we tested that adding a different variable with the same name did fail, then we would've caught this bug. :+1: for adding that test to this PR, but I'm busy at the moment...

johnomotani:
Fixes #2096

johnomotani:
> we should be checking that adding different variables twice fails

I've added another case to the `test-io` integrated test. Seems like it should be a unit-test thing, but I don't think we have I/O tests in the unit tests yet?

tests/integrated/test-io/runtest
```diff
@@ -119,6 +119,40 @@ print("Checking with variables added twice")
...
+    "v3d_evol"
+]:
+    try:
+        s, out = launch_safe(
+            "./test_io check_incorrect_add="+check_incorrect_add, nproc=nproc, pipe=True
+        )
+    except RuntimeError as e:
+        if not (
+            "Variable with name '" + check_incorrect_add + "' already added to Datafile"
+        ):
```
ZedThree, line 143: This is missing what to check the error message in :)

ZedThree (changes requested):

tests/integrated/test-io/runtest
johnomotani: Good spot! Fixed now.

ZedThree (approved):

ZedThree:
Failing tests are just coverage, fixed in #2103

PR #2095, opened 2020-09-09
Handle FieldPerps in Datafile::varAdded() and Datafile::varPtr() (next)

johnomotani:
Fix from #2093 into `next`.


johnomotani:
Added a test of attributes in `test-io` and `test-io_hdf5`. See https://github.com/boutproject/BOUT-dev/pull/2102#issuecomment-697962199.

I don't think we actually test the legacy netcdf interface, but if we do the tests will fail. It might not be a bad thing to fail and warn any users who are using legacy netcdf that there might be odd issues if they try to read netCDF4 files? We'd need to check in the test if it is using legacy netcdf in that case...

ZedThree:
I merged next in locally, and `test-io` still fails -- but I think that's because we've now changed the behaviour. `test-io.grd.nc` is read in, and `fperp` doesn't have a `yindex_global`. Previously it got 0, and now it gets 2. The benchmark file still contains 0 though, so we just need to update that.

johnomotani (commented):
@ZedThree your update looks good to me :+1:

ZedThree (approved):

ZedThree (approved):

PR #2100, opened 2020-09-19
Make examples relocatable

dschwoerer:
By not hard-coding BOUT_TOP it is easier to compile the examples
with a different BOUT_TOP, and thus use them with a different, e.g.
precompiled BOUT++ version.

### DIFF
--- a/examples/2Dturbulence_multigrid/makefile
+++ b/examples/2Dturbulence_multigrid/makefile
@@ -1,5 +1,5 @@
 
-BOUT_TOP	= ../../
+BOUT_TOP	?= ../../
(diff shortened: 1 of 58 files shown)


ZedThree (approved):

dschwoerer:
Forgot to mention, this would allow to drop Step 2b from the [instructions](https://omadesala.physics.dcu.ie/~dave/bout++/instructions.html) for the [live image](https://omadesala.physics.dcu.ie/~dave/bout++/) I created - which would allow to run a BOUT++ simulation within a minute after booting.

PR #2101, opened 2020-09-19
Copy changes from boututils/boutdata

dschwoerer:
boutdata/data.py not merged, as there are too many conflicts


ZedThree (dismissed):

johnomotani:
I can add the `data.py` changes to this. I merged them in the `boutdata` PR.

PS nudge for replacing `tools/pylib/boutdata` and `tools/pylib/boututils` with submodules to avoid these cross-merges.

ZedThree:
Yes submodules is a good idea. I should have another attempt at using xBOUT, as I'd prefer we just used that if we can.

ZedThree (approved):

PR #2102, opened 2020-09-21
Fix some HDF5 related issues and add Mesh::getLocal{X,Y,Z}Index

### LINKED ISSUE #2003: test-restarting fails without netcdf

ZedThree:
Fixes #2003 

Also:

- Fix bug in `DataFormat::writeFieldAttributes` with `FieldPerp`: need to use the global index that doesn't include boundaries
- Fix `test-io_hdf5` needing global `mesh/dump`
- Build with HDF5 on Travis
- Require netCDF for some tests that use grid files
- Add Mesh::getLocal{X,Y,Z}Index{NoBoundaries} methods: these are inverses of the `getGlobal` versions, replacing `{X,Y}LOCAL`, and are necessary for consistent, unique global indices into the boundaries for double-null tokamak grids

There are some tests that can't use HDF5 because they use grid files, and I don't think we want to have duplicate binaries just in different formats. There are some other tests that could use HDF5 but currently can't because we'd need to be able to change the file extension in the `runtest` script. And some others which should be run with both HDF5 and netCDF, including some duplicated ones that could be merged, if we could query what formats are available in `runtest`. Not sure how much I actually want to do there, I'm not terribly certain we want to support HDF5 into the future.

Fedora test is failing because it needs `h5py`, but I didn't know where to specify that -- @dschwoerer does it need to be in your BOUT++ Fedora recipe? If so, is it possible to give me maintainer access to that?

Some of this should be backported.

### DIFF
--- a/include/bout/mesh.hxx
+++ b/include/bout/mesh.hxx
@@ -585,11 +585,11 @@ class Mesh {
 
   /// Returns a local X index given a global index.
   /// Global index includes boundary cells, local index includes boundary or guard cells.
-  virtual int getLocalXIndex(int xlocal) const = 0;
+  virtual int getLocalXIndex(int xglobal) const = 0;
 
   /// Returns a local X index given a global index.
   /// Global index excludes boundary cells, local index includes boundary or guard cells.
-  virtual int getLocalXIndexNoBoundaries(int xlocal) const = 0;
+  virtual int getLocalXIndexNoBoundaries(int xglobal) const = 0;
 
   /// Returns a global Y index given a local index.
   /// Global index includes boundary cells, local index includes boundary or guard cells.
@@ -601,11 +601,11 @@ class Mesh {
 
   /// Returns a local Y index given a global index.
   /// Global index includes boundary cells, local index includes boundary or guard cells.
-  virtual int getLocalYIndex(int ylocal) const = 0;
+  virtual int getLocalYIndex(int yglobal) const = 0;
 
   /// Returns a local Y index given a global index.
   /// Global index excludes boundary cells, local index includes boundary or guard cells.
-  virtual int getLocalYIndexNoBoundaries(int ylocal) const = 0;
+  virtual int getLocalYIndexNoBoundaries(int yglobal) const = 0;
 
   /// Returns a global Z index given a local index.
   /// Global index includes boundary cells, local index includes boundary or guard cells.
@@ -617,11 +617,11 @@ class Mesh {
 
   /// Returns a local Z index given a global index.
   /// Global index includes boundary cells, local index includes boundary or guard cells.
-  virtual int getLocalZIndex(int zlocal) const = 0;
+  virtual int getLocalZIndex(int zglobal) const = 0;
 
   /// Returns a local Z index given a global index.
   /// Global index excludes boundary cells, local index includes boundary or guard cells.
-  virtual int getLocalZIndexNoBoundaries(int zlocal) const = 0;
+  virtual int getLocalZIndexNoBoundaries(int zglobal) const = 0;
 
   /// Size of the mesh on this processor including guard/boundary cells
   int LocalNx, LocalNy, LocalNz;
--- a/include/bout/mesh.hxx
+++ b/include/bout/mesh.hxx
@@ -564,14 +564,16 @@ class Mesh {
   /// Returns the global Y index given a local index
   /// The local index must include the boundary, the global index does not.
   [[deprecated("Use getGlobalYIndex or getGlobalYIndexNoBoundaries instead")]]
-  virtual int YGLOBAL(int yloc) const { return getGlobalYIndexNoBoundaries(yloc); }
+  int YGLOBAL(int yloc) const { return getGlobalYIndexNoBoundaries(yloc); }
 
   /// Returns the local X index given a global index
   /// If the global index includes the boundary cells, then so does the local.
-  virtual int XLOCAL(int xglo) const = 0;
+  [[deprecated("Use getLocalXIndex or getLocalXIndexNoBoundaries instead")]]
+  int XLOCAL(int xglo) const { return getLocalXIndex(xglo); };
   /// Returns the local Y index given a global index
   /// If the global index includes the boundary cells, then so does the local.
-  virtual int YLOCAL(int yglo) const = 0;
+  [[deprecated("Use getLocalYIndex or getLocalYIndexNoBoundaries instead")]]
+  int YLOCAL(int yglo) const { return getLocalYIndexNoBoundaries(yglo); };
 
   /// Returns a global X index given a local index.
   /// Global index includes boundary cells, local index includes boundary or guard cells.
@@ -581,6 +583,14 @@ class Mesh {
   /// Global index excludes boundary cells, local index includes boundary or guard cells.
   virtual int getGlobalXIndexNoBoundaries(int xlocal) const = 0;
 
+  /// Returns a local X index given a global index.
+  /// Global index includes boundary cells, local index includes boundary or guard cells.
+  virtual int getLocalXIndex(int xlocal) const = 0;
+
+  /// Returns a local X index given a global index.
+  /// Global index excludes boundary cells, local index includes boundary or guard cells.
+  virtual int getLocalXIndexNoBoundaries(int xlocal) const = 0;
+
   /// Returns a global Y index given a local index.
   /// Global index includes boundary cells, local index includes boundary or guard cells.
   virtual int getGlobalYIndex(int ylocal) const = 0;
@@ -589,6 +599,14 @@ class Mesh {
   /// Global index excludes boundary cells, local index includes boundary or guard cells.
   virtual int getGlobalYIndexNoBoundaries(int ylocal) const = 0;
 
+  /// Returns a local Y index given a global index.
+  /// Global index includes boundary cells, local index includes boundary or guard cells.
+  virtual int getLocalYIndex(int ylocal) const = 0;
+
+  /// Returns a local Y index given a global index.
+  /// Global index excludes boundary cells, local index includes boundary or guard cells.
+  virtual int getLocalYIndexNoBoundaries(int ylocal) const = 0;
+
   /// Returns a global Z index given a local index.
   /// Global index includes boundary cells, local index includes boundary or guard cells.
   virtual int getGlobalZIndex(int zlocal) const = 0;
@@ -596,6 +614,14 @@ class Mesh {
   /// Returns a global Z index given a local index.
   /// Global index excludes boundary cells, local index includes boundary or guard cells.
   virtual int getGlobalZIndexNoBoundaries(int zlocal) const = 0;
+
+  /// Returns a local Z index given a global index.
+  /// Global index includes boundary cells, local index includes boundary or guard cells.
+  virtual int getLocalZIndex(int zlocal) const = 0;
+
+  /// Returns a local Z index given a global index.
+  /// Global index excludes boundary cells, local index includes boundary or guard cells.
+  virtual int getLocalZIndexNoBoundaries(int zlocal) const = 0;
 
   /// Size of the mesh on this processor including guard/boundary cells
   int LocalNx, LocalNy, LocalNz;
--- a/src/fileio/dataformat.cxx
+++ b/src/fileio/dataformat.cxx
@@ -48,7 +48,7 @@ void DataFormat::writeFieldAttributes(const std::string& name, const FieldPerp& 
   int yindex = f.getIndex();
   if (yindex >= 0 and yindex < fieldmesh.LocalNy) {
     // write global y-index as attribute
-    setAttribute(name, "yindex_global", fieldmesh.getGlobalYIndex(f.getIndex()));
+    setAttribute(name, "yindex_global", fieldmesh.getGlobalYIndexNoBoundaries(f.getIndex()));
   } else {
     // y-index is not valid, set global y-index to -1 to indicate 'not-valid'
     setAttribute(name, "yindex_global", -1);
--- a/src/fileio/dataformat.cxx
+++ b/src/fileio/dataformat.cxx
@@ -48,7 +48,7 @@ void DataFormat::writeFieldAttributes(const std::string& name, const FieldPerp& 
   int yindex = f.getIndex();
   if (yindex >= 0 and yindex < fieldmesh.LocalNy) {
     // write global y-index as attribute
-    setAttribute(name, "yindex_global", fieldmesh.getGlobalYIndexNoBoundaries(f.getIndex()));
+    setAttribute(name, "yindex_global", fieldmesh.getGlobalYIndex(f.getIndex()));
   } else {
     // y-index is not valid, set global y-index to -1 to indicate 'not-valid'
     setAttribute(name, "yindex_global", -1);
@@ -79,8 +79,8 @@ void DataFormat::readFieldAttributes(const std::string& name, FieldPerp& f) {
   // Note: don't use DataFormat::mesh variable, because it may be null if the DataFormat
   // is part of a GridFromFile, which is created before the Mesh.
   if (getAttribute(name, "yindex_global", yindex_global)) {
-    f.setIndex(f.getMesh()->YLOCAL(yindex_global));
+    f.setIndex(f.getMesh()->getLocalYIndex(yindex_global));
   } else {
-    f.setIndex(f.getMesh()->YLOCAL(0));
+    f.setIndex(f.getMesh()->getLocalYIndex(0));
   }
 }
--- a/src/fileio/dataformat.cxx
+++ b/src/fileio/dataformat.cxx
@@ -48,7 +48,7 @@ void DataFormat::writeFieldAttributes(const std::string& name, const FieldPerp& 
   int yindex = f.getIndex();
   if (yindex >= 0 and yindex < fieldmesh.LocalNy) {
     // write global y-index as attribute
-    setAttribute(name, "yindex_global", fieldmesh.getGlobalYIndex(f.getIndex()));
+    setAttribute(name, "yindex_global", fieldmesh.getGlobalYIndexNoBoundaries(yindex));
   } else {
     // y-index is not valid, set global y-index to -1 to indicate 'not-valid'
     setAttribute(name, "yindex_global", -1);
@@ -79,8 +79,8 @@ void DataFormat::readFieldAttributes(const std::string& name, FieldPerp& f) {
   // Note: don't use DataFormat::mesh variable, because it may be null if the DataFormat
   // is part of a GridFromFile, which is created before the Mesh.
   if (getAttribute(name, "yindex_global", yindex_global)) {
-    f.setIndex(f.getMesh()->getLocalYIndex(yindex_global));
+    f.setIndex(f.getMesh()->getLocalYIndexNoBoundaries(yindex_global));
   } else {
-    f.setIndex(f.getMesh()->getLocalYIndex(0));
+    f.setIndex(f.getMesh()->getLocalYIndexNoBoundaries(0));
   }
 }
--- a/src/fileio/dataformat.cxx
+++ b/src/fileio/dataformat.cxx
@@ -46,9 +46,9 @@ void DataFormat::writeFieldAttributes(const std::string& name, const FieldPerp& 
 
   auto& fieldmesh = *f.getMesh();
   int yindex = f.getIndex();
-  if (yindex >= 0 and yindex < fieldmesh.LocalNy) {
+  if (yindex >= fieldmesh.ystart and yindex <= fieldmesh.yend) {
     // write global y-index as attribute
-    setAttribute(name, "yindex_global", fieldmesh.getGlobalYIndexNoBoundaries(yindex));
+    setAttribute(name, "yindex_global", fieldmesh.getGlobalYIndex(yindex));
   } else {
     // y-index is not valid, set global y-index to -1 to indicate 'not-valid'
     setAttribute(name, "yindex_global", -1);
@@ -79,8 +79,9 @@ void DataFormat::readFieldAttributes(const std::string& name, FieldPerp& f) {
   // Note: don't use DataFormat::mesh variable, because it may be null if the DataFormat
   // is part of a GridFromFile, which is created before the Mesh.
   if (getAttribute(name, "yindex_global", yindex_global)) {
-    f.setIndex(f.getMesh()->getLocalYIndexNoBoundaries(yindex_global));
+    f.setIndex(f.getMesh()->getLocalYIndex(yindex_global));
   } else {
+    // No boundary form here, so default value is on a grid cell
     f.setIndex(f.getMesh()->getLocalYIndexNoBoundaries(0));
   }
 }
--- a/src/fileio/dataformat.cxx
+++ b/src/fileio/dataformat.cxx
@@ -45,14 +45,18 @@ void DataFormat::writeFieldAttributes(const std::string& name, const FieldPerp& 
   writeFieldAttributes(name, static_cast<const Field&>(f));
 
   auto& fieldmesh = *f.getMesh();
-  int yindex = f.getIndex();
-  if (yindex >= fieldmesh.ystart and yindex <= fieldmesh.yend) {
-    // write global y-index as attribute
-    setAttribute(name, "yindex_global", fieldmesh.getGlobalYIndex(yindex));
-  } else {
-    // y-index is not valid, set global y-index to -1 to indicate 'not-valid'
-    setAttribute(name, "yindex_global", -1);
-  }
+  const int yindex = f.getIndex();
+  const int start = fieldmesh.hasBndryLowerY() ? 0 : fieldmesh.ystart;
+  const int end = fieldmesh.hasBndryUpperY() ? fieldmesh.LocalNy : fieldmesh.yend + 1;
+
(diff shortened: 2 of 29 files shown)


src/fileio/dataformat.cxx
```diff
@@ -48,7 +48,7 @@ void DataFormat::writeFieldAttributes(const std::string& name, const FieldPerp&
   int yindex = f.getIndex();
   if (yindex >= 0 and yindex < fieldmesh.LocalNy) {
     // write global y-index as attribute
-    setAttribute(name, "yindex_global", fieldmesh.getGlobalYIndex(f.getIndex()));
+    setAttribute(name, "yindex_global", fieldmesh.getGlobalYIndexNoBoundaries(f.getIndex()));
```
johnomotani, line 51: There is a potential problem with this fix, that someone might save two `FieldPerp`s on different processors with the same `getGlobalYIndex()` - i.e. at the inner and outer upper targets in a double-null topology. This is actually what I am using the `FieldPerp` support in restarts for, and because the sheath boundary condition is saved at a `CELL_YLOW` position, it is exactly the same global index, if we use `getGlobalYIndexNoBoundaries()`. If the outputs are saved into a single file, I'm not sure what would happen but probably we would only get one of the `FieldPerp`s saved.

I think the answer is we need `getLocalYIndex(int yglobal)` and `getLocalYIndexNoBoundaries(int yglobal_no_boundaries)` methods (sorry for adding more stuff to `Mesh`, but we can deprecate `YLOCAL()` in favour of `getLocalYIndexNoBoundaries()` so it's only one more small method!). Then we use `getLocalYIndex(yindex_global)` below https://github.com/boutproject/BOUT-dev/blob/e926569e23a634fe6b363cff30a9b7a0e70c8753/src/fileio/dataformat.cxx#L82

johnomotani (changes requested):

dschwoerer:
> And some others which should be run with both HDF5 and netCDF, including some duplicated ones that could be merged, if we could query what formats are available in `runtest`. 

The boutconf branch might help with that, as that lets you query how bout was configured.

> Not sure how much I actually want to do there, I'm not terribly certain we want to support HDF5 into the future.

I'd be in favor to only support one backend - makes it easier to maintain and test. Or what is the advantage of HDF5? Slighly easier to install compared to netcdf?

> Fedora test is failing because it needs `h5py`, but I didn't know where to specify that -- @dschwoerer does it need to be in your BOUT++ Fedora recipe? If so, is it possible to give me maintainer access to that?

I added it to the install line. I wouldn't mind giving you access to the package in fedora, but that [requires you to be a packager](https://fedoraproject.org/wiki/Join_the_package_collection_maintainers), which involves demonstrating you know the packaging rules of fedora. I certainly learned a lot in the process and it is fun, but also requires a bit of time ...

johnomotani:
I don't really remember (maybe @bendudson does?), but I think there was a thought at one point that some project (maybe the American SciDAC?) would standardise on HDF5, and it would be useful to be able to integrate with that. I happened to be working on an HDF5 interface for another project a few years ago, so thought I might as well put in a BOUT++ one while I was looking up the function calls.

I'd support standardising to one backend. netCDF seems better in my opinion, the ecosystem of tools around `xarray`, etc. all seems to be built mostly on netCDF. There shouldn't be any performance difference because netCDF is just HDF5 with extra conventions for metadata.

I would like to get rid of the netcdf-cxx4 dependency though if we're refactoring the backends. It always seems to be a bit of a pain to get onto clusters, and in my experience isn't provided in the modules.

src/fileio/dataformat.cxx
ZedThree: I thought something like that might be necessary! Ok, I'll add those two methods

include/bout/mesh.hxx
```diff
@@ -581,6 +583,14 @@ class Mesh {
   /// Global index excludes boundary cells, local index includes boundary or guard cells.
   virtual int getGlobalXIndexNoBoundaries(int xlocal) const = 0;
 
+  /// Returns a local X index given a global index.
+  /// Global index includes boundary cells, local index includes boundary or guard cells.
+  virtual int getLocalXIndex(int xlocal) const = 0;
```
johnomotani, line 588: I think here and in the other `getLocal*` methods, the argument should be `xglobal`, etc. as the argument is a global-index value.
```suggestion
  virtual int getLocalXIndex(int xglobal) const = 0;
```

src/fileio/dataformat.cxx
```diff
@@ -48,7 +48,7 @@ void DataFormat::writeFieldAttributes(const std::string& name, const FieldPerp&
   int yindex = f.getIndex();
   if (yindex >= 0 and yindex < fieldmesh.LocalNy) {
     // write global y-index as attribute
-    setAttribute(name, "yindex_global", fieldmesh.getGlobalYIndex(f.getIndex()));
+    setAttribute(name, "yindex_global", fieldmesh.getGlobalYIndexNoBoundaries(yindex));
```
johnomotani, line 51: see my earlier comment https://github.com/boutproject/BOUT-dev/pull/2102#discussion_r492238557
```suggestion
    setAttribute(name, "yindex_global", fieldmesh.getGlobalYIndex(yindex));
```

src/fileio/dataformat.cxx
```diff
@@ -79,8 +79,8 @@ void DataFormat::readFieldAttributes(const std::string& name, FieldPerp& f) {
   // Note: don't use DataFormat::mesh variable, because it may be null if the DataFormat
   // is part of a GridFromFile, which is created before the Mesh.
   if (getAttribute(name, "yindex_global", yindex_global)) {
-    f.setIndex(f.getMesh()->YLOCAL(yindex_global));
+    f.setIndex(f.getMesh()->getLocalYIndexNoBoundaries(yindex_global));
```
johnomotani, line 82:
```suggestion
    f.setIndex(f.getMesh()->getLocalYIndex(yindex_global));
```

src/fileio/dataformat.cxx
```diff
@@ -79,8 +79,8 @@ void DataFormat::readFieldAttributes(const std::string& name, FieldPerp& f) {
...
   } else {
-    f.setIndex(f.getMesh()->YLOCAL(0));
+    f.setIndex(f.getMesh()->getLocalYIndexNoBoundaries(0));
```
johnomotani, line 84: I think it makes sense to keep `getLocalYIndexNoBoundaries(0)` here, as that sets a default value on a grid cell rather than (possibly) in a boundary cell.

src/mesh/impls/bout/boutmesh.cxx
```diff
@@ -1630,8 +1630,11 @@ int BoutMesh::getGlobalXIndexNoBoundaries(int xlocal) const {
   return xlocal + PE_XIND * MXSUB - MXG;
 }
 
-/// Returns a local X index given a global index
-int BoutMesh::XLOCAL(int xglo) const { return xglo - PE_XIND * MXSUB; }
+int BoutMesh::getLocalXIndex(int xlocal) const { return xlocal - PE_XIND * MXSUB; }
```
johnomotani, line 1633:
```suggestion
int BoutMesh::getLocalXIndex(int xglobal) const { return xglobal - PE_XIND * MXSUB; }
```

src/mesh/impls/bout/boutmesh.cxx
```diff
@@ -1630,8 +1630,11 @@ int BoutMesh::getGlobalXIndexNoBoundaries(int xlocal) const {
...
+
+int BoutMesh::getLocalXIndexNoBoundaries(int xlocal) const {
+  return xlocal - PE_XIND * MXSUB + MXG;
```
johnomotani, line 1636:
```suggestion
int BoutMesh::getLocalXIndexNoBoundaries(int xglobal) const {
  return xglobal - PE_XIND * MXSUB + MXG;
```

src/mesh/impls/bout/boutmesh.hxx
```diff
@@ -188,13 +188,16 @@ class BoutMesh : public Mesh {
 
   int getGlobalXIndex(int xlocal) const override;
   int getGlobalXIndexNoBoundaries(int xlocal) const override;
+  int getLocalXIndex(int xlocal) const override;
+  int getLocalXIndexNoBoundaries(int xlocal) const override;
```
johnomotani, line 192:
```suggestion
  int getLocalXIndex(int xglobal) const override;
  int getLocalXIndexNoBoundaries(int xglobal) const override;
```

src/mesh/impls/bout/boutmesh.hxx
```diff
@@ -188,13 +188,16 @@ class BoutMesh : public Mesh {
...
   int getGlobalYIndex(int ylocal) const override;
   int getGlobalYIndexNoBoundaries(int ylocal) const override;
+  int getLocalYIndex(int ylocal) const override;
+  int getLocalYIndexNoBoundaries(int ylocal) const override;
```
johnomotani, line 196:
```suggestion
  int getLocalYIndex(int yglobal) const override;
  int getLocalYIndexNoBoundaries(int yglobal) const override;
```

src/mesh/impls/bout/boutmesh.hxx
```diff
@@ -188,13 +188,16 @@ class BoutMesh : public Mesh {
...
   int getGlobalZIndex(int zlocal) const override;
   int getGlobalZIndexNoBoundaries(int zlocal) const override;
-
-  int XLOCAL(int xglo) const override;
-  int YLOCAL(int yglo) const override;
+  int getLocalZIndex(int zlocal) const override;
+  int getLocalZIndexNoBoundaries(int zlocal) const override;
```
johnomotani, line 200:
```suggestion
  int getLocalZIndex(int zglobal) const override;
  int getLocalZIndexNoBoundaries(int zglobal) const override;
```

include/bout/mesh.hxx
```diff
@@ -581,6 +583,14 @@ class Mesh {
...
+
+  /// Returns a local X index given a global index.
+  /// Global index excludes boundary cells, local index includes boundary or guard cells.
+  virtual int getLocalXIndexNoBoundaries(int xlocal) const = 0;
```
johnomotani, line 592:
```suggestion
  virtual int getLocalXIndexNoBoundaries(int xglobal) const = 0;
```

include/bout/mesh.hxx
```diff
@@ -589,6 +599,14 @@ class Mesh {
   /// Global index excludes boundary cells, local index includes boundary or guard cells.
   virtual int getGlobalYIndexNoBoundaries(int ylocal) const = 0;
 
+  /// Returns a local Y index given a global index.
+  /// Global index includes boundary cells, local index includes boundary or guard cells.
+  virtual int getLocalYIndex(int ylocal) const = 0;
```
johnomotani, line 604:
```suggestion
  virtual int getLocalYIndex(int yglobal) const = 0;
```

include/bout/mesh.hxx
```diff
@@ -589,6 +599,14 @@ class Mesh {
...
+
+  /// Returns a local Y index given a global index.
+  /// Global index excludes boundary cells, local index includes boundary or guard cells.
+  virtual int getLocalYIndexNoBoundaries(int ylocal) const = 0;
```
johnomotani, line 608:
```suggestion
  virtual int getLocalYIndexNoBoundaries(int yglobal) const = 0;
```

include/bout/mesh.hxx
```diff
@@ -597,6 +615,14 @@ class Mesh {
   /// Global index excludes boundary cells, local index includes boundary or guard cells.
   virtual int getGlobalZIndexNoBoundaries(int zlocal) const = 0;
 
+  /// Returns a local Z index given a global index.
+  /// Global index includes boundary cells, local index includes boundary or guard cells.
+  virtual int getLocalZIndex(int zlocal) const = 0;
```
johnomotani, line 620:
```suggestion
  virtual int getLocalZIndex(int zglobal) const = 0;
```

include/bout/mesh.hxx
```diff
@@ -597,6 +615,14 @@ class Mesh {
...
+
+  /// Returns a local Z index given a global index.
+  /// Global index excludes boundary cells, local index includes boundary or guard cells.
+  virtual int getLocalZIndexNoBoundaries(int zlocal) const = 0;
```
johnomotani, line 624:
```suggestion
  virtual int getLocalZIndexNoBoundaries(int zglobal) const = 0;
```

johnomotani (changes requested):

src/fileio/dataformat.cxx
ZedThree: I did try it this way, but it fails `test-io` because it ends up writing `fperp2` on processors 2 and 3, both with `yindex_global = 11`. 

I think maybe the conditional also needs changing to 

```cpp
if ((yindex >= fieldmesh.myg) and (yindex < fieldmesh.LocalNy - fieldmesh.myg)) {
```

src/fileio/dataformat.cxx
ZedThree: I'm not sure about this, might be better to just fail?

include/bout/mesh.hxx
ZedThree: Thanks! I did change it in some places, but evidently didn't follow it through

src/fileio/dataformat.cxx
johnomotani: I wanted to not fail so we can successfully read old restart files, which might (I think? haven't checked the actual history...) exist with FieldPerps but without `yindex_global` attributes.

johnomotani:
I think this needs #2095 for the tests to pass.

BTW, I tested with the netcdf-legacy and the tests pass, but `fperp2` is not read or written on the correct processors! The legacy netcdf interface just fails to read attributes from the benchmark file. Maybe because it's a NetCDF4-format and legacy netcdf doesn't support NetCDF4 properly? I changed the `runtest` scripts in #2095, so I'll add a check of the attributes in that PR.

johnomotani (dismissed):
Other than a merge of #2095, this looks good to me.

dschwoerer:
I guess we should drop legacy netcdf and hdf5 with the 5.0 release.
I think it is better to fail early and with a clear error, then to try to do the wrong thing, but produce slightly wrong results.

src/fileio/dataformat.cxx
ZedThree: That sounds reasonable, I'm just not sure how you'd deal with such a field? You'd need to set the index via some other means

ZedThree:
It looks like this should just go into #2095 ?

> I guess we should drop legacy netcdf and hdf5 with the 5.0 release.

After we get 4.3.2 and 4.4.0 out, I want to look at completely replacing datafile with `OptionsNetCDF`

johnomotani:
> It looks like this should just go into #2095 ?

That sounds sensible.

src/fileio/dataformat.cxx
```diff
@@ -46,9 +46,9 @@ void DataFormat::writeFieldAttributes(const std::string& name, const FieldPerp&
 
   auto& fieldmesh = *f.getMesh();
   int yindex = f.getIndex();
-  if (yindex >= 0 and yindex < fieldmesh.LocalNy) {
+  if (yindex >= fieldmesh.ystart and yindex <= fieldmesh.yend) {
```
ZedThree, line 49: What are valid values for `yindex` and `yindex_global`? `0 < yindex < local_ny` and `0 < yindex_global < ny - 1`? Or `ystart <= yindex <= yend` and `myg <= yindex_global < ny - myg`? (that is, can you have `FieldPerp`s in the boundaries?)

I feel like either the form I've changed it to, where only interior points get saved, or the previous form, where points on this process get saved must be correct, but then one of `test-io` or `test-restart-io` ends up failing.

The problem might be that `collect` assumes that a `FieldPerp` will have `yindex_global == -1` in all but one file. But using `getLocalYIndex`, we can easily end up saving it on multiple processors, due to the guard cells.

For example, in `test-io` we check that reading and then writing a restart file is correct for a `FieldPerp` with `yindex_global = 11`. The grid has `ny = 16` and `myg = 2`, and the failing test is on 4 cores with `NYPE = 4`, so the `LocalNy = 7` (with 3 interior points).

With this set up, we have the following values for `getLocalYIndex(yindex_global)`:

| Processor | Local index |
|------------|------------------|
| 0 | 11 |
| 1 | 8 |
| 2 | 5 |
| 3 | 2 | 

So processors 2 & 3 both save the same point using the old form of the conditional here, which triggers an `assert` in `collect`.

On the other hand, `test-restart-io` tries to check with two `FieldPerp`s, one with `yindex_global = 0`, the other with `16` (note I think this should be `ny - 1 = 15` anyway). These are both in the boundary, so fail with the new form of the conditional.

Do we ever want to read/write `FieldPerp`s in boundaries (and not guards)? If no, then I think this is easily solved by just changing the numbers in `test-restart-io`. Otherwise, I guess we'll need to check `firstY/lastY` and modify the conditional accordingly? If we want to be able to save in _guards_ as well, `collect` will need to be modified.

Now that I've written all this out, I suspect we probably do want them in boundaries!

Note, I have no idea how xBOUT handles this.

johnomotani: Sorry, I meant to think about this earlier but got distracted by something else...

We definitely want `FieldPerp`s in boundaries. For example, to save an upper-target boundary condition I'd want to save a `CELL_YLOW` `FieldPerp` at `yend+1`.

Short version of what I say below: I think we should use `if (yindex >= 0 and yindex < fieldmesh.LocalNy)` to test for validity, but only add `fperp2` to `dump` in the tests if its index is on a non-guard point. We allow users to save a `FieldPerp` in guard cells if they want to, but they are responsible for the errors from `collect()` if their `FieldPerp`s clash or are inconsistent.

#### long version

Not sure about guard cells. I'd be tempted to allow this to be done - it might be useful for debugging or something, but if a model saves `FieldPerp`s from different processors at the same global yindex (as defined by `getGlobalYIndex()`, then `collect` should give an error and the user can sort it out by hand.
Actually, I think that ideally a given `FieldPerp` (at least if it's going to be saved and not just used as a temporary, e.g. inside `Laplacian::solve()`) should only be allowed to exist with a single global y-index, and on a single set of processors with the same `PE_YIND`. In this view, we shouldn't actually call for example `dump.add(fperp2, "fperp2", false);` on all the processors, we should only call it on the ones where `fperp2.getIndex()` is not in a guard cell. Unfortunately I don't see any way to enforce this, but we could document it and raise a helpful error in `collect()`.

Sorry, the previous paragraph wasn't very well written... The idea I had in mind in saving with `yindex_global` was that if we're going to have a physical quantity represented by a `FieldPerp` then it should be at a well-defined and unique physical position (with a global y-index in either grid or boundary cells, but guard cells would not make sense). Only then would it be allowed to save it. Other processors where the physical field is not present should not really allocate it or add it to `dump`.

If we do want to check for boundaries explicitly, we should use
https://github.com/boutproject/BOUT-dev/blob/b810b18886c9f93d6399eb247e6e6cab681f8f06/include/bout/mesh.hxx#L457-L458
because `firstY()` and `lastY()` do not account for an upper target (double null configuration). Although their docstrings suggest that they do
https://github.com/boutproject/BOUT-dev/blob/b810b18886c9f93d6399eb247e6e6cab681f8f06/include/bout/mesh.hxx#L388-L391
I think the docstrings are wrong and should be changed (or we need to change `firstY()` and `lastY()` to do what they say).
Limiter configurations are still a bit dodgy because we could have processors where some x-indices have a y-boundary and others don't, but I don't think it's worth worrying about whether to and how to support this case until somebody actually wants it and can make the implementation support their use case.

xBOUT doesn't support `FieldPerp` at all at the moment. We should update that some time after this issue has defined the correct behaviour.
https://github.com/boutproject/xBOUT/blob/1f2090e0167b5147a0489889151a76c625b47fc9/xbout/load.py#L700-L707

ZedThree: One problem with allowing saving `FieldPerp`s in guard cells is that at high enough core counts, all of the interior points are also guard cells on neighbouring processors, so it's inevitable that a `FieldPerp` would get written by two processors. This is exactly the problem with `test-io`.

Options:

1. Allow saving in guard cells, and `collect` either just uses the first one it comes across (and assumes that guard & grid cell values agree, possibly checks), or works out which one is a grid cell and uses that
2. Only allow reading/writing to grid & boundary cells
3. Only allow `FieldPerp`s to exist on grid & boundary cells (involves changes to `Datafile::read_fperp` and `GridFile::readField`)

The last option feels the "cleanest" in some sense, but I don't know in what circumstances someone would want a `FieldPerp` in guard cells and therefore when this would break someone's code. `Laplacian` has a `include_yguards` option, but I think that's just a misnomer, and should be `include_y_boundaries`.

The first option feels the most pragmatic, but I don't think it's going to be very easy for `collect` work out if a cell is a boundary cell or a guard cell. We could of course write more information about the mesh/processor decomposition to the restart/dump files, but then we need to worry about how xBOUT or `restart.redistribute` handles that information.

So, the second option then? Actually, this still writes guard cells, it's just that we'd set the `yindex_global` attribute to -1. `collect` wouldn't read it, but it would be possible to read through the python `DataFile` or lower level APIs. The reverse is not true though, and it would not be possible to read guard cells.

johnomotani: Thanks @ZedThree. I agree your option 2 is the way to go. :+1:

For the record, I think what I wrote above still wasn't quite right. My actual use case is that I have some sheath boundary conditions that create `FieldPerp`s, and save those to output and restart files. So I have a field like `psi_sheath_lower` that is created and saved anywhere there is a 'lower' sheath boundary. In double-null there are two of those, so I do save `psi_sheath_lower` twice with different `yindex_global` values (which contradicts what I said should be allowed earlier...). Because of the way we handle topology, the two locations where I save `psi_sheath_lower` can never be on the same processor, so restarting works fine.

I'll make an issue on xBOUT about this now. I think I know what we want to implement...

johnomotani (approved):

PR #2104, opened 2020-09-23
Bugfix release: v4.3.2

ZedThree:
Time for a release!

4.3.2 is a bugfix release:
- Make downloading the submodules a bit nicer
- Make `dz` in the Python API a property
- Make `Div_par_K_Grad_par` check the staggered location of its inputs
- Enable split-flux derivatives on staggered fields
- Fix `Grad2_par2` implementation in `InvertParCR`
- Fix an issue writing `FieldPerp`s
- Make it easier to compile the examples with different versions of BOUT++

I think #2102 probably wants backporting to this, at the very least.


ZedThree:
v4.3.2 ready for release!

ZedThree:
`test-io` timed out again, but the output still wasn't flushed, even with a `sleep` command in `after_failure` :(

I'm trying always sleeping in `after_script` instead, as well as an explicit `flush=True` in `test_suite` when printing the failed tests. That's in a separate branch that I'll merge straight into here if it actually works.

bendudson (approved):

PR #2107, opened 2020-09-24
Add configure option to use system mpark.variant

dschwoerer:
I have [a patch to remove the bundled mpark.variant](https://src.fedoraproject.org/rpms/bout++/blob/master/f/remove-mpark.patch) for fedora.
That breaks however on every release, thus it would make my life easier if it would be part of BOUT++ - and I thought it might also be easier for others, if they already have mpark.variant installed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,7 +75,7 @@ jobs:
     env:
       - *default_env
       - CONFIGURE_OPTIONS="--enable-shared --with-petsc --with-slepc --with-sundials=$HOME/local"
-      - SCRIPT_FLAGS="-uim -t python -t shared"
+      - SCRIPT_FLAGS="-uim -t shared -t python"
   - name: "4to5 test"
     env:
       - *default_env
--- a/.travis_script.sh
+++ b/.travis_script.sh
@@ -38,7 +38,7 @@ do
 	    TESTS=1
 	    ;;
         t) ### Set target to build
-            MAIN_TARGET="$OPTARG"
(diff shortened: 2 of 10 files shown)


dschwoerer:
Anybody know what the `slashCommandDispatch` check is, and why it hasn't been run?

ZedThree:
That's the clang-format tool. I wonder if it was triggered from the URL in your PR comment? Pretty sure it's only supposed to trigger if the command is the first thing on the first line

ZedThree:
This should make things easier on initial checkout for other people too, but as @johnomotani pointed out in #2110, we're still left with a potential problem when changing branches etc.

At the risk of over-engineering things, how about we move the submodule recipe into a new file `makefile.submodules`, and then `configure` could call `make -f makefile.submodules` while `makefile` can `include makefile.submodules`. This way we can have a `make submodules` target that we can tell people to run instead of `git submodule update --init --recursive` without getting in the way of using a system version.

ZedThree:
@dschwoerer Please could you check 845397d is still suitable for your use-case?

dschwoerer:
Strangely that seems to break boutcore for me, as the BOUT++ flags aren't any more injected.

What is also worrying is that boutcore isn't build as part of travis - I thought it would be part "Shared library + Python" - but it is not build.

dschwoerer:
Looks good to me, [a test build](https://copr.fedorainfracloud.org/coprs/davidsch/f33/build/1698093/) failed, but as far as I can tell, that was due to #2108 also being required.

ZedThree (approved):

Issue #2026, opened 2020-05-04
PETSc installation advice in docs is outdated


ZedThree:
https://bout-dev.readthedocs.io/en/latest/user_docs/advanced_install.html#petsc

- While we sort of support PETSc 3.4+, we definitely recommend 3.7+, and we should recommend using the latest available version
- The advice to tell PETSc to use a pre-compiled version of SUNDIALS is not great; we should recommend either using a standalone SUNDIALS or compiling PETSc with `--download-sundials` but not both

johnomotani:
I think our advice should be that even if you configure with SUNDIALS, PETSc does not need to be linked to SUNDIALS. I believe the only reason to link PETSc to SUNDIALS would be to call SUNDIALS time-solvers via the PETSc interface: I don't think the PETSc time-solver interface in BOUT++ is widely used, and even if it is used I doubt there is a benefit to calling the SUNDIALS solvers through PETSc rather than directly. If a user really wants PETSc linked to SUNDIALS for some reason, I think it's fair to assume that they're sufficiently expert to decide how to link it themselves.

PR #2109, opened 2020-09-25
Update documentation on processor splitting and advanced installation

### LINKED ISSUE #2106: documentation on processor splitting
The documentation on what values of `NXPE` and `NYPE` are allowed is a bit terse - from l.297 onward here https://github.com/boutproject/BOUT-dev/blame/next/manual/sphinx/user_docs/bout_options.rst.

> In BOUT++, grids can be split between processors in both X and Y directions. By default BOUT++ automatically divides the grid in both X and Y, finding the decomposition with domains closest to square, whilst satisfying
constraints. These constraints are:
>
>- Every processor must have the same size and shape domain
>
>- Branch cuts, mostly at X-points, must be on processor boundaries. This is because the connection between grid points is modified in BOUT++  by changing which processors communicate.

I think all the following information is useful, but it's too long-winded, and I'm having a brain-freeze at the moment about how to write something nicer

* the option `nx` is the total number of points in the x-direction including boundary points, so `nx` is usually "(number of grid points in x-direction) + 4". [This will probably change for the 5.0 release to be nx is the number of grid points, but I don't think it has been changed yet]
* the option `ny` is the number of points in the y-direction *not* including boundary cells. [i.e. this is inconsistent at the moment with the way `nx` is defined]
* NXPE must divide the number of grid points in the x-direction (i.e. nx-4)
* for simple slab grids NYPE must divide the number of grid points in the y-direction (i.e. ny)
* for general topologies, the number of points per processor `mysub=ny/nype` (which must be an integer) must divide the number of points in each region (2 divertor legs and core for single-null; 4 divertor legs, inner core and outer core for double-null; here when I write 'core' I mean 'the region including the core but also part of the SOL' - sorry, I don't know a good word for it so 'core' is sometimes used in BOUT++ when talking about the grid topology as a shorthand even though it's not totally accurate).
* When BOUT++ automatically chooses NXPE and NYPE it finds all valid pairs which give (total number of procs)=NXPE*NYPE and satisfy the constraints above, then chooses the one which makes the grid on each processor as close to square as possible.

ZedThree:
Fixes #2106, #2026

### DIFF
--- a/manual/sphinx/user_docs/bout_options.rst
+++ b/manual/sphinx/user_docs/bout_options.rst
@@ -239,12 +239,23 @@ of the input file (see :ref:`sec-gridgen` for more information):
 .. code-block:: cfg
 
     [mesh]
-    nx = 16  # Number of points in X
+    nx = 20  # Number of points in X
     ny = 16  # Number of points in Y
     nz = 32  # Number of points in Z
 
+Due to historical reasons, ``nx`` is defined differently to ``ny`` and ``nz``:
+
+- ``nx`` is the number of points in X **including** the boundaries
+- ``ny`` and ``nz`` are the number of points in Y and Z **not including** the
+  boundaries
+
+The default number of boundary points in X is 2, so taking into account the
+boundary at each end of the domain, ``nx`` usually means "the number of interior
+grid points in X plus four". In the example above, both X and Y have 16 interior
+grid points.
+
 It is recommended, but not necessary, that this be :math:`\texttt{nz}
-= 2^n`, i.e.  :math:`1,2,4,8,\ldots`. This is because FFTs are usually
+= 2^n`, that is :math:`1,2,4,8,\ldots`. This is because FFTs are usually
 slightly faster with power-of-two length arrays, and FFTs are used
 quite frequently in many models.
 
@@ -302,6 +313,29 @@ given, ``NXPE`` takes precedence and ``NYPE`` is ignored):
 .. code-block:: cfg
 
     NYPE = 1  # Set number of Y processors
+
+When choosing ``NXPE`` or ``NYPE``, they must also obey some constraints:
+
+- ``NXPE`` must be a factor of the number of grid points in the x-direction
+  - That is, ``(nx - 4) / NXPE`` must be an integer, assuming the usual two
+    boundary points
+- For simple slab grids, ``NYPE`` must be a factor of the number of grid points
+  in the y-direction
+  - That is, ``ny / NYPE`` must be an integer
+- For more general topologies, the number of points per processor ``ny / NYPE``
+  must also be a factor of the number of points in each region. For example, in
+  the usual tokamak topologies:
+  - in single-null there are two divertor leg and one core regions
+  - in double-null there are four divertor leg, one inner core and one outer
+    core regions
+
+Please note that here "core" means "core and adjacent SOL". See
+:ref:`sec-bout-topology` for a more detailed explanation of these regions.
+
+When BOUT++ automatically chooses ``NXPE`` and ``NYPE`` it finds all valid pairs
+which give ``total number of processors = NXPE * NYPE`` and also satisfy the
+constraints above. It then chooses the pair that makes the grid on each
+processor as close to square as possible.
 
 If you need to specify complex input values, e.g. numerical values
 from experiment, you may want to use a grid file. The grid file to use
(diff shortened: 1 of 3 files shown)


ZedThree:
Skipped travis builds as this is just documentation, so only the readthedocs check is useful

manual/sphinx/user_docs/bout_options.rst
```diff
@@ -303,6 +314,34 @@ given, ``NXPE`` takes precedence and ``NYPE`` is ignored):
...
     NYPE = 1  # Set number of Y processors
 
+When choosing ``NXPE`` or ``NYPE``, they must also obey some constraints:
+
+- ``NXPE`` must be a factor of the number of grid points in the x-direction
+
+  - That is, ``(nx - 4) / NXPE`` must be an integer, assuming the usual two
+    boundary points
+
+- For simple slab grids, ``NYPE`` must be a factor of the number of grid points
```
dschwoerer, line 324: I think it would be easier to read if it wouldn't say "For simple slab grids" - because it is true for all cases.

manual/sphinx/user_docs/bout_options.rst
```diff
@@ -303,6 +314,34 @@ given, ``NXPE`` takes precedence and ``NYPE`` is ignored):
...
+  - in double-null there are four divertor leg, one inner core and one outer
+    core regions
+
+Please note that here "core" means "core and adjacent SOL". See
+:ref:`sec-bout-topology` for a more detailed explanation of these regions.
+
+When BOUT++ automatically chooses ``NXPE`` and ``NYPE`` it finds all valid pairs
+which give ``total number of processors = NXPE * NYPE`` and also satisfy the
+constraints above. It then chooses the pair that makes the grid on each
+processor as close to square as possible.
```
dschwoerer, line 343: constraints above. It then chooses the pair that minimizes the boundaries, i.e. results ideally on a square domain on each process.

"as close to square" is not well defined (imho) - other option would be an aspect ratio close to one.

manual/sphinx/user_docs/bout_options.rst
johnomotani: I agree

manual/sphinx/user_docs/bout_options.rst
johnomotani: Agreed 'as close to square' isn't great. It sounds a bit like geometry rather than number of grid points. 'aspect ratio' could too. Maybe just 'It then chooses the pair with number of points in the x- and y-directions as close to equal as possible.'?

ZedThree: "Minimise the difference between NXPE and NYPE"?

johnomotani: > "Minimise the difference between NXPE and NYPE"?

It's really 'minimise the difference between MXSUB and MYSUB' which isn't quite the same. I think it's equivalent to that. The actual code is
https://github.com/boutproject/BOUT-dev/blob/5461a40d963e400c2543ea6e1e1bffa018364169/src/mesh/impls/bout/boutmesh.cxx#L322
https://github.com/boutproject/BOUT-dev/blob/5461a40d963e400c2543ea6e1e1bffa018364169/src/mesh/impls/bout/boutmesh.cxx#L398-L400

ZedThree: Oh yes, sorry, good point!

How about: "It then chooses the pair that makes the grid on each processor as close to square as possible (technically it chooses the pair that minimises `((nx - 4) * NPES / ny) - NXPE`, where `NPES` is total number of processors)"

johnomotani: sounds good to me

johnomotani:
BTW thanks to Qian Xia for orignially pointing out this area for improvement in the docs.

johnomotani (approved):

ZedThree:
I think the slash command dispatch actually triggers on every pull request (and possibly comment?), it's just it usually doesn't end up doing anything. I'm not sure why it hasn't reported its status or how to make it do so.

manual/sphinx/user_docs/bout_options.rst
dschwoerer: `((nx - 4) * NPES / ny) - NXPE` is minimal for maximal NXPE
You need to add `abs`

dschwoerer: I am even more confused - why is this linear in NPES?

The code uses a square root:

https://github.com/boutproject/BOUT-dev/blob/5461a40d963e400c2543ea6e1e1bffa018364169/src/mesh/impls/bout/boutmesh.cxx#L322

Maybe:
find NXPE as close as possible to `sqrt((nx-4) / ny * NPES)` - I like putting nx and ny together to make it clear the ratio is important. Maybe splitting  it in 2 `sqrt` makes it even more clear - part of the contribution is the number of cores, the other the ratio ...

ZedThree: Oops, sorry, only just saw this. Yes, a typo there. That's what you get for trying to do two things at once!

I'll fix this directly in the v4.3.2-rc PR next week.

PR #2111, opened 2020-09-30
Suggestions for gnu build on Marconi

### LINKED ISSUE #1898: netCDF configuration
On Marconi, I have to explicitly pass the `lib` (in `EXTRA_LIBS`) and `include` (in `EXTRA_INCS`) search paths for `netcd-c` to `configure`*. I suspect the problem may be due to `configure` only using `ncxx4-config` provided by `netcdf-cxx4` to find the paths to add. It would be nice if `configure` would **also** get paths from `nc-config` (provided by `netcdf-c`). I guess this might need an extra config flag: maybe we could have both `with-netcdf` and `with-netcdf-cxx` or something like that, for systems which need both?

\* possibly part of the problem is that sometimes there is a very old `netcdf` provided by the system, which gets found instead of one provided by modules, or installed by the user, etc.

johnomotani:
Fixes #1898.

### DIFF
--- a/manual/sphinx/user_docs/advanced_install.rst
+++ b/manual/sphinx/user_docs/advanced_install.rst
@@ -178,6 +178,84 @@ to enable AVX512 vectorization.
           ``boutcore.pyx`` needs to be manually triggered, if
           ``boutcore.pyx`` has already been created.
 
+Marconi with gnu compilers
+**************************
+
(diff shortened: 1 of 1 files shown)


johnomotani:
Cancelled the Travis build since this PR only touches the manual.

ZedThree (approved):

ZedThree:
Awesome, thanks @johnomotani !

PR #2043, opened 2020-05-29
Fixes for tokamak-2fluid example

johnomotani:
There were several staggered-grid issues with `examples/tokamak-2fluid` that I've started trying to fix here.

Should fix #2041

PR #2113, opened 2020-10-01
Fix Div_par when using more than one y guard cell

ZedThree:
I think this is the final fix needed for #2043 

With this (and #2043), `tokamak-2fluid` runs for me (once I drop the timestep low enough to get any output on my machine). No idea if the output is actually _correct_ of course


johnomotani (approved):

ZedThree:
Failing test is just RTD which was fixed in #2114

PR #2114, opened 2020-10-01
Update breathe for readthedocs

ZedThree:
Was breaking with newer version of Sphinx (which RTD seems to have upgraded to without any warning?)


ZedThree:
This is infrastructure so I'm using admin-fiat to merge it. It can be checked in #2104 if needed :)

PR #2116, opened 2020-10-05
Fix reading/writing std::string attributes when using HDF5

ZedThree:
`H5Awrite` and `H5Aread` actually need `(const) char**`

Partial backport of #2102 -- `test-io_hdf5` passes (on my machine) with just this, so hopefully this is all that's needed.


johnomotani (approved):
There might be some bugs/oddities in the `FieldPerp` reading/writing in `master` (which were being addressed by other parts of #2102), but it was working already for the use case I needed, so I think we can just regard `FieldPerp` I/O as a bit experimental until v5.

Issue #2117, opened 2020-10-05
`test-io` on Travis randomly times out


ZedThree:
It seems to be on `test-io`, but that could just be the last thing printed and might be something else later that's hanging, although it is pretty consistently `test-io`.
Happens across configurations, including debug and optimised builds.

Address sanitizer didn't reveal anything obvious, and it seems to always work for me. At this point, I'm not sure if it's BOUT++ or Travis that's having issues.

bshanahan:
`test-io` is constantly failing for me. I can't look up exactly why at the moment, but wanted to clarify that this might not just be Travis.

ZedThree:
Thanks @bshanahan . What branch is that with? And how are you configuring BOUT++?

ZedThree:
Three separate jobs in the PR build for #2116  failed. Given that the CMake job runs tests in a different order also stopped on `test-io`, pretty sure that is to blame.

ZedThree:
I've got a branch that just runs `test-io` multiple times in each configuration, none of them hang. They do all pause on `Checking variable f2d ... ` for 4 processors with `NYPE=1` for some time, but I think that might just be a reflection on how Travis buffers and displays the output, because it then shows the other half of the test output all at once.

So, still none the wiser.

bshanahan:
It turns out I had a stale next branch, and it works for me on master now.

ZedThree:
It is definitely `test-io` that is hanging: https://travis-ci.org/github/boutproject/BOUT-dev/jobs/733405537#L4351

Annoyingly, there's an issue with Travis quitting before flushing stdout, but of course now I've got something in to work around that, the random hang isn't happening any more!

ZedThree:
I've looked at several recent Travis runs, and test-io usually takes around 25 seconds, except that occasionally it takes much longer. I've seen ~70, ~80, ~120 seconds, and of course > 10 minutes.

I've now tested it on a local 16.04 machine, running `test-io` 20 times in a row, and it was consistently ~25 seconds. Now running 100 times in a row on my machine.

`valgrind` also turned up nothing interesting.

My working theory is that there is some strange interaction with Travis's filesystem somehow, but why it would only be that one test, which isn't even doing anything particularly weird, I don't know.

ZedThree:
This seems to have stopped being a problem now we've moved to Github Actions -- I'm blaming some weird filesystem issue on Travis

PR #2118, opened 2020-10-06
Add timeout to tests

ZedThree:
Adds `BOUT_TEST_TIMEOUT` environment variable to timeout tests if they take too long. Default is 10 minutes, but set this to 5 minutes on Travis, so we don't get caught by their timeout.

CMake's `ctest` has this built in with `--timeout`.

This is really to help investigate #2117 , but might actually be a useful feature too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -163,6 +163,15 @@ script:
   # Configure, compile and run test_suite
   - "./.travis_script.sh ${SCRIPT_FLAGS}"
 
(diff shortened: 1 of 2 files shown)


dschwoerer (dismissed):
I think that is a really useful feature. I had once a test stuck for hours because there was no timeout. For running the full test-suite the timeout most likely needs to be ajusted ...

dschwoerer (approved):

Issue #1940, opened 2020-03-03
cmake: boutcore


dschwoerer:
This requieres a fix for #1934 

And then all the fun can begin ...

ZedThree:
[scikit-build](https://scikit-build.readthedocs.io/en/latest/cmake-modules.html) has some CMake modules that might be helpful here

Issue #1934, opened 2020-03-02
cmake: generate files


dschwoerer:
I cannot find code that seems to generate the generated fieldops?

Also adding `#error` to `src/field/gen_fieldops.jinja` does not break the build with cmake.

dschwoerer:
Should be doable with:
https://cmake.org/cmake/help/latest/command/add_custom_command.html

I'll have a look ...

ZedThree:
Fixed in #2119

PR #2119, opened 2020-10-08
CMake: Generate Field arithmetic operators

dschwoerer:
Add support for generating files

Should solve #1934  and thus enable fixing #1940

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -306,6 +306,16 @@ set(BOUT_SOURCES
   ${CMAKE_CURRENT_BINARY_DIR}/include/bout/version.hxx
   )
 
+
+ADD_CUSTOM_COMMAND( OUTPUT ${CMAKE_CURRENT_SOURCE_DIR}/src/field/generated_fieldops.cxx
+  COMMAND python gen_fieldops.py --filename generated_fieldops.cxx.tmp
+  COMMAND clang-format generated_fieldops.cxx.tmp -i
+  COMMAND mv generated_fieldops.cxx.tmp generated_fieldops.cxx
+  DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/src/field/gen_fieldops.jinja  ${CMAKE_CURRENT_SOURCE_DIR}/src/field/gen_fieldops.py
+  WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/src/field/
+  COMMENT "Generating source code" )
+
+
 add_library(bout++
   ${BOUT_SOURCES}
   )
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -306,6 +306,16 @@ set(BOUT_SOURCES
   ${CMAKE_CURRENT_BINARY_DIR}/include/bout/version.hxx
   )
 
+
+ADD_CUSTOM_COMMAND( OUTPUT ${CMAKE_CURRENT_SOURCE_DIR}/src/field/generated_fieldops.cxx
+  COMMAND python gen_fieldops.py --filename generated_fieldops.cxx.tmp
```
ZedThree, line 311: If we add [`find_package(Python)`](https://cmake.org/cmake/help/latest/module/FindPython.html) we can using `Python_EXECUTABLE` to make this nice and portable. I think that also prefers to find python 3 first too, which is nice.

The same package can also find the numpy include headers, which might be useful for building boutcore?

CMakeLists.txt
```diff
@@ -306,6 +306,16 @@ set(BOUT_SOURCES
...
+  COMMAND clang-format generated_fieldops.cxx.tmp -i
```
ZedThree, line 312: We can also add `FindClangFormat` from [here](https://github.com/ttroy50/cmake-examples/blob/64bd54a1a11dcc83c44280d3c255218ffc1b2d66/04-static-analysis/clang-format/cmake/modules/FindClangFormat.cmake) and do a similar thing for `clang-format`.

CMakeLists.txt
```diff
@@ -306,6 +306,16 @@ set(BOUT_SOURCES
...
+  COMMAND mv generated_fieldops.cxx.tmp generated_fieldops.cxx
```
ZedThree, line 313: We can also make this lovely and portable too:

```suggestion
  COMMAND ${CMAKE_COMMAND} -E rename generated_fieldops.cxx.tmp generated_fieldops.cxx
```

https://cmake.org/cmake/help/latest/manual/cmake.1.html#run-a-command-line-tool

ZedThree (commented):
Thanks @dschwoerer ! I've made some suggestions to make it more portable.

ZedThree (approved):

ZedThree:
I've added an option to skip generating the field ops if, for example, you can't/don't want to install Jinja.

CMakeLists.txt
dschwoerer: Sorry for not replying earlier.

`mv` is defined by POSIX, thus only non-POSIX complient OS would be affected by that.
I think `mv` is lovely :smiley: - calling cmake to move a file however feels like overkill.
As this would only cause issues on an OS that is intentionally non-POSIX compliant thus I disagree with aiming for portability to such a degree. I think such an OS shouldn't be catered for, and I would prefer we wouldn't support that. But the LGPL gives you every right to ignore my preference :+1:

PR #2123, opened 2020-10-22
Fix links to open-mpi.org and lam-mpi.org

tobyjamez:
Links weren't specified using https so were linking to nonexistent pages on readthedocs.io.


ZedThree (approved):

ZedThree:
Failing tests are just coverage tests that can't be run on forks. Might be worth us seeing if it's possible to disable them in that case

PR #2131, opened 2020-11-02
Allow setting FFTW_EXHAUSTIVE

johnomotani:
May sometimes make FFTs faster than the previously available FFTW_MEASURE setting.

Also uses a `BOUT_ENUM_CLASS()` for the option, which should help ensure that that doesn't get broken in future :-)

Unfortunately the macros don't work inside the `bout::fft` namespace, I think due to `FFT_FLAG` needing to be passed as a template parameter in various places. I don't understand why that didn't work though.

Performance - FFTs seemed to be faster with the exhaustive flag on Marconi, but I haven't checked with BOUT++ yet. Having the option for the exhaustive flag can't hurt anyway.

Todo:
* [x] document new option in manual

### DIFF
--- a/bin/bout-v5-input-file-upgrader.py
+++ b/bin/bout-v5-input-file-upgrader.py
@@ -12,14 +12,32 @@ from boututils.boutwarnings import AlwaysWarning
 # This should be a list of dicts, each containing "old", "new" and optionally "values".
 # The values of "old"/"new" keys should be the old/new names of input file values or
 # sections. The value of "values" is a dict containing replacements for values of the
-# option.
+# option. "type" optionally specifies the type of the old value of the option; for
+# example this is needed for special handling of boolean values.
 REPLACEMENTS = [
     {"old": "mesh:paralleltransform", "new": "mesh:paralleltransform:type"},
     {"old": "fci", "new": "mesh:paralleltransform"},
     {"old": "interpolation", "new": "mesh:paralleltransform:xzinterpolation"},
     {"old": "fft:fft_measure", "new": "fft:fft_measurement_flag",
-        "values": {"false": "estimate", "true": "measure"}}
+        "type": bool, "values": {False: "estimate", True: "measure"}}
 ]
+
+
+def parse_bool(bool_expression):
+    try:
+        bool_expression_lower = bool_expression.lower()
+    except AttributeError:
+        # bool_expression was not a str: no need to lower
+        bool_expression_lower = bool_expression
+
+    if bool_expression_lower in ["true", "y", "t", 1, True]:
+        return True
+    elif bool_expression_lower in ["false", "n", "f", 0, False]:
+        return False
+    else:
+        raise RuntimeError(
+            f"Expected boolean option. Could not parse {bool_expression}"
+        )
 
 
 def fix_replacements(replacements, options_file):
@@ -38,11 +56,35 @@ def fix_replacements(replacements, options_file):
                 "\n\t{1}".format(options_file.filename, e.args[0], **replacement)
             ) from e
         else:
-            if "values" in replacement:
-                old_value = options_file[replacement["new"]]
-                options_file[replacement["new"]] = replacement["values"].get(
-                    old_value, old_value
-                )
+            if "type" in replacement:
+                # Special handling for certain types, replicating what BOUT++ does
+                if replacement["type"] is bool:
+                    # The original value must be something that BOUT++ recognises as a
+                    # bool.
+                    # replacement["values"] must contain both True and False keys.
+                    old_value = parse_bool(options_file[replacement["new"]])
+                    options_file[replacement["new"]] = replacement["values"][old_value]
+                else:
+                    raise ValueError(
+                        f"Error in REPLACEMENTS: type {replacement['type']} is not handled"
+                    )
+            else:
+                # Option values are just a string
+                if "values" in replacement:
+                    old_value = options_file[replacement["new"]]
+                    try:
+                        old_value = old_value.lower()
+                    except AttributeError:
+                        # old_value was not a str, so no need to convert to lower case
+                        pass
+
+                    try:
+                        options_file[replacement["new"]] = replacement["values"][
+                            old_value
+                        ]
+                    except KeyError:
+                        # No replacement given for this value: keep the old one
+                        pass
 
 
 def apply_fixes(replacements, options_file):
--- a/bin/bout-v5-input-file-upgrader.py
+++ b/bin/bout-v5-input-file-upgrader.py
@@ -17,6 +17,8 @@ REPLACEMENTS = [
     {"old": "mesh:paralleltransform", "new": "mesh:paralleltransform:type"},
     {"old": "fci", "new": "mesh:paralleltransform"},
     {"old": "interpolation", "new": "mesh:paralleltransform:xzinterpolation"},
+    {"old": "fft:fft_measure", "new": "fft:fft_measurement_flag",
+        "values": {"false": "estimate", "true": "measure"}}
 ]
 
 
--- a/bin/bout-v5-input-file-upgrader.py
+++ b/bin/bout-v5-input-file-upgrader.py
@@ -9,9 +9,10 @@ import warnings
 from boutdata.data import BoutOptionsFile
 from boututils.boutwarnings import AlwaysWarning
 
-# This should be a list of dicts, each containing exactly two keys:
-# "old" and "new". The values of these keys should be the old/new
-# names of input file values or sections
+# This should be a list of dicts, each containing "old", "new" and optionally "values".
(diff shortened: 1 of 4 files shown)


src/invert/fft_fftw.cxx
```diff
@@ -61,13 +61,47 @@ void fft_init(Options* options) {
...
     options = Options::getRoot()->getSection("fft");
   }
-  fft_init((*options)["fft_measure"]
-               .doc("Perform speed measurements to optimise settings?")
-               .withDefault(false));
+  bool fft_measure = (*options)["fft_measure"]
+                    .doc("Perform speed measurements to optimise settings?")
+                    .withDefault(false);
+  fft_flag = (*options)["fft_flag"]
+                    .doc("Level speed measurements to optimise FFT settings: [estimate], measure, exhaustive")
```
d7919, line 68: Also might be worth noting that estimate is the only option certain to be exactly repeatable (as the other two make measurements and can choose different algorithms with each run).

src/invert/fft_fftw.cxx
```diff
@@ -61,13 +61,47 @@ void fft_init(Options* options) {
...
+    if ((*options)["fft_flag"].isSet()) {
+      throw BoutException("Cannot set both fft_measure and fft_flag");
+    }
+    fft_init(fft_measure);
+  } else {
+    fft_init(fft_flag);
+  }
+}
+
+unsigned int get_flags(FFT_FLAG fft_flag) {
```
d7919, line 81: Currently this only deals with measurement flags but I guess we might want other options in the future as well like use wisdom or not so this may benefit from a more specific name (although may this is just something that could be done if we ever get to that situation)?

johnomotani: Good point, maybe something like `fft_measurement_flag` would be better.

johnomotani:
The new option needs documenting in the manual. I'll do that before we merge.

ZedThree:
Do we want to deprecate the `fft_measure` input? There's a v5 upgrader for renaming input parameters, it could probably also change the values at the same time.

The issue with putting a `BOUT_ENUM_CLASS` declaration in a namespace is that the specialisation of `Options::as` needs to be in the same namespace as `Options`. I managed to get it to work like this:

```cpp
namespace bout {
enum class FFT_FLAG;
FFT_FLAG BOUT_MAKE_FROMSTRING_NAME(FFT_FLAG)(const std::string& s);
}

BOUT_ENUM_CLASS(bout::FFT_FLAG, estimate, measure);
```

but I can't see how to make it work in the macro without knowing the namespace name, i.e. `BOUT_ENUM_CLASS(bout, FFT_FLAG, ...)` can work.

src/invert/fft_fftw.cxx
ZedThree: +1 for `fft_measurement_flag`

src/invert/fft_fftw.cxx
ZedThree: Good thing to note in the docs

src/invert/fft_fftw.cxx
```diff
@@ -108,10 +142,7 @@ void rfft(MAYBE_UNUSED(const BoutReal *in), MAYBE_UNUSED(int length), MAYBE_UNUS
      */
     fout = static_cast<fftw_complex*>(fftw_malloc(sizeof(fftw_complex) * (length/2 + 1)));
 
-    unsigned int flags = FFTW_ESTIMATE;
-    if (fft_measure) {
-      flags = FFTW_MEASURE;
-    }
+    auto flags = get_flags(fft_flag);
```
ZedThree, line 145: This is much nicer than what we had before, even without the extra flag. Thanks!

ZedThree (approved):

ZedThree:
BTW, the reason I worked out why the namespace wasn't working was I tried compiling with clang, which gave me this error:

```
/home/peter/Codes/BOUT-dev/working/test_enum.cxx:8:1: error: function template specialization of 'as' not in class 'Options' or an enclosing namespace
BOUT_ENUM_CLASS(FFT_FLAG, estimate, measure);
^
/home/peter/Codes/BOUT-dev/include/bout/bout_enum_class.hxx:96:36: note: expanded from macro 'BOUT_ENUM_CLASS'
template <> inline auto ::Options::as<enumname>(const enumname&) const -> enumname { \
                                   ^
/home/peter/Codes/BOUT-dev/include/options.hxx:355:5: note: explicitly specialized declaration is here
  T as(const T& UNUSED(similar_to) = {}) const {
    ^
1 error generated.
```

whereas gcc gave me

```
In file included from /home/peter/Codes/BOUT-dev/working/test_enum.cxx:1:
/home/peter/Codes/BOUT-dev/working/test_enum.cxx:8:17: error: ‘FFT_FLAG’ does not name a type
    8 | BOUT_ENUM_CLASS(FFT_FLAG, estimate, measure);
      |                 ^~~~~~~~
/home/peter/Codes/BOUT-dev/working/test_enum.cxx:8:17: error: ‘FFT_FLAG’ does not name a type
    8 | BOUT_ENUM_CLASS(FFT_FLAG, estimate, measure);
      |                 ^~~~~~~~
```

Not quite as helpful as clang!

src/invert/fft_fftw.cxx
johnomotani: > Good thing to note in the docs

It already was - 39a00aeecaa1555be41e37de6aeb25b9319e64f1 by @ZedThree :smiley:

bin/bout-v5-input-file-upgrader.py
```diff
@@ -34,6 +37,12 @@ def fix_replacements(replacements, options_file):
                 "Could not apply transformation: '{old}' -> '{new}' to file '{0}', due to error:"
                 "\n\t{1}".format(options_file.filename, e.args[0], **replacement)
             ) from e
+        else:
+            if "values" in replacement:
+                old_value = options_file[replacement["new"]]
+                options_file[replacement["new"]] = replacement["values"].get(
+                    old_value, old_value
```
johnomotani, line 44: @ZedThree should this be
```suggestion
                    old_value.lower(), old_value
```
?

ZedThree: Good question! `options` is case insensitive for keys, I don't think it is for values, although BOUT++ _is_. I think `.lower()` is correct.

I'm a little worried about failure modes here. We might not want to use `dict.get` here, and instead `try... except` and print a nice error message instead?

bin/bout-v5-input-file-upgrader.py
```diff
@@ -9,13 +9,16 @@
...
+# This should be a list of dicts, each containing "old", "new" and optionally "values".
+# The values of "old"/"new" keys should be the old/new names of input file values or
+# sections. The value of "values" is a dict containing replacements for values of the
+# option.
 REPLACEMENTS = [
     {"old": "mesh:paralleltransform", "new": "mesh:paralleltransform:type"},
     {"old": "fci", "new": "mesh:paralleltransform"},
     {"old": "interpolation", "new": "mesh:paralleltransform:xzinterpolation"},
+    {"old": "fft:fft_measure", "new": "fft:fft_measurement_flag",
+        "values": {"false": "estimate", "true": "measure"}}
```
ZedThree, line 21: BOUT++ actually converts `y/Y/t/T/1` to `true` and `n/N/f/F/0` to `false`. I'm not sure what the best way to handle that here is, without this suddenly becoming massively complicated. My first thought was to add another key, `type`, and if that's `bool`, apply a conversion function first. Then use `True`/`False` literals in the `value` dict.

bin/bout-v5-input-file-upgrader.py
johnomotani: I'd intended to allow replacing just some of the possible values of an option, so if the old value isn't in replacments["values"], then we just keep the old value, so no need for an error message. This 'feature' isn't actually used now because in fixing the `bool` issue I've required both `True` and `False` to be present for bool-type options; if it turns out it's better to actually have an error, then we can change it.

bin/bout-v5-input-file-upgrader.py
johnomotani: Agreed. I've made an update implementing a `parse_bool()` function. I think it's not too horrible.

I've tested `bout-v5-input-file-upgrader.py` by 'upgrading' `fft_measure` from all of `true/y/Y/t/T/1` and `false/n/N/f/F/0`.

ZedThree (approved):

bin/bout-v5-input-file-upgrader.py
ZedThree: Lovely, thanks @johnomotani ! We can now deprecate both input keys and values with a nice upgrade path :tada:

ZedThree (approved):

PR #2138, opened 2020-11-06
CMake: Add option to download and build netCDF C++ API

### LINKED ISSUE #2105: Providing netcdf-cxx4
I think we discussed a while back what interface to use to netcdf if/when we refactor `DataFile`, etc. and the consensus was that we'd want to keep using [netcdf-cxx4](https://github.com/Unidata/netcdf-cxx4/tree/master/cxx4) because otherwise we'd end up basically wanting a C++ wrapper around the netcdf C-interface anyway. Issue is that `netcdf-cxx4` is often not provided on clusters, or only provided for a specific compiler (e.g. for icc and I want to use gcc). Could something be done to make life easier? Maybe include `netcdf-cxx4` as a submodule, or provide a `--download-netcdf-cxx4` option (or cmake-equivalent - I guess we're going to cmake-only for 5.0?) like PETSc does for external packages?

ZedThree:
Fixes #2105

Currently uses my fork -- I've opened a PR, but it might be a while before it gets merged.

@johnomotani Any chance you could check this works on MARCONI?


johnomotani:
Is it possible to automate this? ie get `cmake` to pass `$(nc-config --prefix)` to the `netcdf-cxx4` build by default? As I understand `nc-config` is a standard utility?

ZedThree:
Yes, I think I just need to split the current `FindNetCDF.cmake` into two parts, one to find the C library, and one for the C++ API. That way, I can find the C library before downloading the C++ API, which should then let it skip its checks. I might then contribute the C version back too.

johnomotani (dismissed):
This works for me, configuring with just `cmake -DBOUT_DOWNLOAD_NETCDF_CXX4=ON ..` and tests pass. I wondered briefly whether it'd be good to add the `-DBOUT_DOWNLOAD_NETCDF_CXX4=ON` flag in a test, but we only have one run using cmake at the moment so I don't think it's worth it.

ZedThree:
My changes had broken things for the older version of netCDF-C on Travis, should be fixed now

ZedThree:
Also, I should note that the CMake module names have changed a little bit to be more in line with the upstream netCDF names, which are `netCDF` and `netCDFCxx`. This affects things like finding netcdf via `netCDF_ROOT`/`netCDFCxx_ROOT` CMake variables

johnomotani (approved):
auto-downloading netcdf-cxx4 still working on Marconi :+1:

PR #2145, opened 2020-11-18
Check DataFile grid sizes match those in existing mesh

### LINKED ISSUE #2141: Restart files can be read incorrectly
I have some tests that load restart files, run and then check the output against expected output. I changed `MXG` in the input file and ran the tests. Unsurprisingly they failed, but they failed by producing incorrect output rather than throwing an error. I think what happened is that since `LocalNx` and `LocalNy` when I tried to run with `MXG=1` are smaller than when the restart files were created with `MXG=2`, the arrays in the restart files are big enough that they never go out of bounds when being read, and BOUT++ reads in values from the file offset by 1 from where it should be reading.

The simplest fix would be to just check that relevant things are the same in the restart file as in the run: [`MXG`, `MYG`, `MZG`] and/or the array size compared to [`LocalNx`, `LocalNy`, `LocalNz`].

Since we generally don't need the guard or boundary cells from restart files, in principle it should be possible to load from them even if `MXG`, `MYG` or `MZG` change. We'd need to get the values of those variables that the restart file was created with, and use those to calculate the appropriate offsets, etc. This would be a lot more work though, and I'd be happy with just throwing an error.

ZedThree:
Fixes #2141

### DIFF
--- a/src/fileio/datafile.cxx
+++ b/src/fileio/datafile.cxx
@@ -902,6 +902,33 @@ void Datafile::add(Vector3D &f, const char *name, bool save_repeat) {
   }
 }
 
+namespace {
+// Read a value from file and check it matches reference_value, throw if not
+void checkGridValue(DataFormat* file, const std::string& name,
+                    const std::string& filename, const int reference_value) {
+  int file_value;
+  if (!file->read(&file_value, name)) {
+    throw BoutException("Could not read {} from file '{}'", name, filename);
+  }
+
+  if (file_value != reference_value) {
+    throw BoutException("{} ({}) in file '{}' does not match value in mesh ({})", name,
+                        file_value, filename, reference_value);
+  }
+}
+
+// Check that the array sizes in \p file match those in existing \p mesh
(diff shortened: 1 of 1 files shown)


src/fileio/datafile.cxx
```diff
@@ -902,6 +902,33 @@ void Datafile::add(Vector3D &f, const char *name, bool save_repeat) {
   }
 }
 
+namespace {
+// Read a value from file and check it matches reference_value, throw if not
+void checkGridValue(DataFormat* file, const std::string& name,
+                    const std::string& filename, const int reference_value) {
```
bendudson, line 908: unneeded `const` on `int`?

bendudson (approved):
LGTM

johnomotani (approved):
Looks good to me. Would have told me exactly what the problem was that led me to open #2141

src/fileio/datafile.cxx
ZedThree: Makes no different to the signature, but the local variable is `const` :)

ZedThree:
Failing test is just Fedora build that we fixed today

PR #2146, opened 2020-11-19
Relax security controls on Fedora test image

### LINKED ISSUE #2143: Fedora rawhide test failing due to out of date shell
The Fedora test is failing with the following error:

```
./configure: This script requires a shell more modern than all
./configure: the shells that I found on your system.
./configure: Please tell bug-autoconf@gnu.org and bd512@york.ac.uk
./configure: about your system, including any error possibly output
./configure: before this message. Then install a modern shell, or
./configure: manually run the script under such a shell if you do
./configure: have one.
```

I guess Fedora rawhide recently changed /bin/sh, but I can't find what that change was. @dschwoerer any ideas what's caused this?

Not sure what we need to do to fix this -- possibly run `autoreconf -fvi` as part of the fedora test? Or maybe just shift it over to use CMake instead?

ZedThree:
For some reason, this addition is necessary so that `test -x /` works in `configure`

Fixes #2143

### DIFF
--- a/.travis_fedora.sh
+++ b/.travis_fedora.sh
@@ -21,8 +21,9 @@ then
     test . != ".$2" && mpi="$2" || mpi=openmpi
     test . != ".$3" && version="$3" || version=latest
     time $cmd pull registry.fedoraproject.org/fedora:$version
-    time $cmd create --name mobydick registry.fedoraproject.org/fedora:$version \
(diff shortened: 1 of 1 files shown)


dschwoerer (approved):
Let's merge the workaround.

I'll try to keep digging :+1:

PR #2147, opened 2020-11-19
Add a clamp function to FieldFactory

bendudson:
Name taken from C++17 std::clamp, but works the same as NumPy clip.

Includes docs and tests


ZedThree (approved):

ZedThree:
Failing tests are the Fedora build that we fixed today, and that one test that keeps timing out on Travis

PR #2149, opened 2020-11-20
Generate random run ID, track restarts

bendudson:
Improve data provenance tracking, so that sets of runs can be linked to each other, and the run used to generate data identified.

Each time the solver is run, generate a random number which is broadcast to all processors. This is stored as "run_id" in:
- log files
- BOUT.settings, though at the moment only if the run finishes
- restart files
- dump files

When a simulation is restarted, the ID of the run it started from is also recorded as "run_restart_from".

If restart files are used which don't have a run_id, then this will either cause the run to fail, or set run_id to 0. The
`run_restart_from` ID therefore has two special values:
- 1 means no restart, the run was started from scratch
- 0 means restart from unknown run_id (missing -> run_id set to 0)

There may well be a better way to handle missing run_id values in restart files.


src/solver/solver.cxx
```diff
@@ -468,8 +468,28 @@ int Solver::solve(int NOUT, BoutReal TIMESTEP) {
     throw BoutException(_("Failed to initialise solver-> Aborting\n"));
   }
 
+  // Set the run ID to a random number
+  run_restart_from = run_id; // Restarting from the previous run ID
+
+  if (MYPE == 0) {
+    srand(static_cast<unsigned int>(time(nullptr)));
```
johnomotani, line 475: If I'm reading right, `time()` returns output in seconds. If a script started lots of runs in a scan in parallel, you might get a clash. Maybe we could use a high resolution clock http://www.cplusplus.com/reference/chrono/high_resolution_clock/now/? To be really paranoid we could also combine with a PID, multiplying by some long-ish amount of time so different PIDs aren't the same as small changes in time, something like
```
srand(static_cast<unsigned int>(std::chrono::high_resolution_clock::now()) + 1000000000000u * unsigned_int(get_pid()));
```
maybe?

src/solver/solver.cxx
```diff
@@ -468,8 +468,28 @@ int Solver::solve(int NOUT, BoutReal TIMESTEP) {
     throw BoutException(_("Failed to initialise solver-> Aborting\n"));
   }
 
+  // Set the run ID to a random number
+  run_restart_from = run_id; // Restarting from the previous run ID
```
johnomotani, line 472: It would be nice to have `run_restart_from` be a `std::vector<unsigned_int>`, so if you restart from a restart (etc.) you know the IDs of all the previous runs. I think this needs an extra method or two adding to `DataFile` to handle `std::vector` (or `Array`?), but shouldn't be too complicated because `DataFormat` already supports C-arrays.

johnomotani (commented):
:heart: This would be a really nice feature to have!

src/solver/solver.cxx
ZedThree: We should probably at least use the C++ `<random>`:

```cpp
    std::random_device rd;  // cross platform way to get a random seed
    std::mt19937 gen(rd());
    std::uniform_int_distribution<> dis(2, 255); // ensures it's never 0 or 1, probably want std::numeric_limits<init>::max()
    run_id = dis(gen);
```

We might like to go one step further and generate something like a UUID, there are a few header-only libraries out there to do this.

src/solver/solver.cxx
bendudson: This would be nice! These might need to be vectors in dump files, as at the moment if a simulation restarts and appends to the .dmp files then the run_id and run_restart_from will be overwritten. In that case possibly `run_id` should be a vector of the runs whose outputs are contained in the dump file, and `run_restart_from` could be a vector of previous runs?

johnomotani: I'm having a go at adding `std::vector<int>` to `Datafile`s. Hopefully PR coming soon...

johnomotani: See #2150, which allows a `std::vector<int>` to be added to a `Datafile`.

Not sure what to do about simulations that append. In #2150 I've tried to treat a `std::vector<int>` generically - I named the dimension "vec<n>" so we could add any number of vectors of any size, but that doesn't make sense if we want to append to them.

It wouldn't be crazy to have a dimension that we append one entry per run to - I'd actually quite like to have that for saving the git hashes of BOUT++ and physics models to - but that would be a bigger change to the `Datafile` interface. I hesitate to put more work into `Datafile` as we're going to (?) replace it with `OptionsNetCDF` in v5. One possible temporary work-around would be to just make `run_id` a time-dependent `int` variable. It's only a slight waste of disk space and would cure the appending problem. I should think we can put a special case into `xbout`'s `create_restarts()` method to get rid of duplicate entries and turn the time-dependent `run_id` into a `run_restart_from` vector. Then I guess we skip writing `run_restart_from` to dump files if we're appending, as the info would already be there in the original `run_restart_from` (if there was one) plus time-dependent `run_id`. Finally, open a feature request for whatever I/O system we end up with in v5 to tidy this up.

ZedThree: I've started work on replacing `Datafile` with `OptionsNetCDF`, although not very far in yet.

Having a "number of restarts-length" unlimited dimension is a good idea. In the `OptionsNetCDF` approach, we also add everything in `Options` to the output file. This includes all the input parameters. Given that these could change in a restarted simulation, we would have to make basically _everything_ have this dimension.

An alternative approach might be to create a group structure:

```
/
|-- time evolving variables
| |-- n
| \-- T
|-- build info
| |-- run001
| | |-- git hash
| | |-- has_petsc
| | ...
| \-- run002
|   |-- git hash
|   |-- has_petsc
|   ...
\-- input parameters
  |-- run001
  | |-- run_id
  | |-- nx
  | ...
  \-- run002
    |-- run_id
    |-- nx
    ...
```

or perhaps

```
/
|-- time evolving variables
| |-- n
| \-- T
|-- run001
| |-- run_id
| |-- build info
| | |-- git hash
| | |-- has_petsc
| | ...
| \-- input parameters
|   |-- nx
|   ...
|-- run001
| |-- run_id
| |-- build info
| | |-- git hash
| | |-- has_petsc
| | ...
| \-- input parameters
|   |-- nx
|   ...
...
```

I know that xarray `Dataset` doesn't understand groups, but if we know what groups we're expecting `xBOUT`  can handle it -- flattening the separate run groups for instance.

bendudson: I quite like this group structure approach: If the evolving variables are in the root directory as now, and the metadata is in groups, then it should be backwards compatible. Then we're not having to add lots of dimensions to things that probably don't need them most of the time.

johnomotani: This sounds like a pretty nice idea. Would need an xbout update, but seems like it should be straightforward and would make things nicer. At the moment xbout puts all scalars into a 'metadata' attribute, because most of them are settings (like `nx`, etc.), but there could also be scalar variables which it would be better to keep separate. xarray can load groups from a netCDF file (using the `group` kwarg to `open_dataset`), so we could instead create the metadata with something like
```
ds.attrs["metadata"] = open_dataset(\<first_file_paths\>, group="run001/input parameters")
```
or something a bit fancier to deal with multiple runs that might have the same input parameters (so we can use a single `metadata`) or might have changed values (concatenate the Datasets along a 'run' dimension maybe?). Anyway seems like a nice thing to have.

While we're discussing - it might be nice to have the non-time-dependent variables (`SAVE_ONCE()` ones) in the `run001`, `run002`, ..., groups in case they change. I'm not sure if we overwrite or keep the original value when appending at the moment...

johnomotani: All that said, I would love a backportable version of this PR (i.e. works with existing `Datafile`), even if (for example) it doesn't support using `append` nicely.

ZedThree:
Fedora tests are currently failing, but it looks like @dschwoerer  is already on it :rocket: : https://bugzilla.redhat.com/show_bug.cgi?id=1924231

ZedThree:
The other test that's failing is `test-restart-io`, with `Missing data for run_id in input. Set init_missing=true to set to zero.`. I think that's expected, and the test just needs updating

johnomotani (approved):
Thanks @bendudson @ZedThree, I think this looks good now! I'll try and have a look at a backport to v4.4.0-alpha sometime...

ZedThree:
The only thing nagging me about this PR at the mo is that all this lives in `Solver`, and it doesn't feel quite like it belongs there. I think it's there because `Solver::solve` actually does a bunch of the simulation setup/teardown, and maybe that would all be better off living somewhere else. I have been thinking about a `BoutSim<PhysicsModel>` class or something, that basically has `BoutInitialise` as its constructor and `BoutFinalise` as its destructor, and we move lots of the non-solver bits out of `Solver` into that.

That is for another PR though!

I do want to make two more changes: make all the run ID bits `private` in `Solver`, rather than `protected` (I don't think anything else actually needs these members? We can always add a getter if so), and install `uuid-dev` for clang-tidy-review. One of the Fedora builds passed, so hopefully that's fixed now too.

johnomotani:
Actually a getter would be good to have for anything that creates a separate set of output files (e.g. https://github.com/johnomotani/BoutFastOutput that writes some stuff with higher frequency) so the run-id can be added there. Eventually if/when we can have multiple time dimensions in the dump file the need might go away, but for the moment would be useful.

dschwoerer:
The Fedora issue is fixed, and the fix was released Yesterday, so as soon as the mirrors catch up, it should work :+1:

include/bout/solver.hxx
```diff
@@ -350,6 +350,13 @@ public:
     pargv = &v;
   }
 
+  /// Generate a random UUID (version 4) and broadcast it to all processors
+  std::string createRunID() const;
```
johnomotani, line 354: Maybe we want to keep `createRunID()` private, to ensure it only gets called once?

I wondered if we could just put a check in `createRunID()` to see if it had already been called, and if it has then just return `run_id`, but `getRunRestartFrom()` always needs to be called after the restart files have been loaded (and if `run_id` is set before restart files are loaded then it would be over-written, which we don't want). Anywhere that wants to save run_id should ideally also save run_restart_from, so it makes sense to restrict their use to after restart files are loaded, which is pretty much the same as after `createRunID()` is called. So I'd suggest making `createRunID()` private, and making it an error to call `getRunID()` or `getRunRestartFrom()` before `createRunID()`. We could add an extra flag to record when `createRunID()` is called, but maybe it'd be good enough to just check whether `run_id` still has its default value - that check would be wrong between the restarts being loaded and `createRunID()` being called, but probably nothing could happen in that time? Or maybe it's better to have a specific flag to be safe and future proof?

ZedThree: Good points. Maybe just have `getRunID` and `getRunRestartFrom` throw if `run_id` is still its default value?

The guts of `createRunID` may be more generally useful if we want UUIDs for other things (not that I can think of any off the top of my head), but yes, should be private for now.

johnomotani: > Maybe just have getRunID and getRunRestartFrom throw if run_id is still its default value?

sounds good to me.

src/solver/solver.cxx
```diff
@@ -472,8 +473,21 @@ int Solver::solve(int NOUT, BoutReal TIMESTEP) {
...
+
+  // Put the run ID into the options tree
+  // Forcing in case the value has been previously set
+  Options::root()["run"]["run_id"].force(run_id, "Solver");
+  Options::root()["run"]["run_restart_from"].force(run_restart_from, "Solver");
+
   /// Run the solver
   output_info.write(_("Running simulation\n\n"));
+  output_info.write("Run ID: {:s}\n", run_id);
+  if (run_restart_from != "zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz") {
```
johnomotani, line 488:
```suggestion
  if (run_restart_from != default_run_id) {
```

johnomotani (approved):
Looks good to me. One tiny suggestion - defining `default_run_id` was a good idea!

src/solver/solver.cxx
ZedThree: Good catch, thanks!

Thinking about this some more, is there a reason why we couldn't just make the default the empty string, then check for `run_{id,restart_from}.empty()`? Does it interact funny when writing to disk?

johnomotani: I think having 36-character strings is sometimes necessary. The Datafile string I/O assumes that the length of the string doesn't change between adding the variable to the Datafile object, and reading/writing. I think that means `run_id` and `run_restart_from` need to be 36-character strings by default. Possibly we could get away with `run_restart_from` being an empty string, but I think it's less confusing and potentially safer to have both be 36-character strings.

ZedThree: Ok, we may be able to change this when we do away with Datafile

johnomotani (approved):
:+1:

ZedThree:
I'm not sure if we quite have the full tracking of previous restarts working yet, but getting the group structure of runs is a task for the datafile refactoring

johnomotani:
Yeah, I think full tracking of previous restarts is a job for another PR, after we move I/O to OptionsNetCDF. What we've got here is definitely useful, and back-portable (I'm working on a back-port to 4.4 now). I think backporting full tracking would be a pain.

PR #2150, opened 2020-11-21
I/O for std::vector<int> and std::vector<char>

johnomotani:
Allows a `std::vector<int>` to be added to a `Datafile` instance. Since the length of the vector is not necessarily related to nx, ny or nz, need to create a new dimension (for netCDF files) - the dimension is named `"vec" + std::to_string(size)`, where `size` is the length of the vector.

Includes a `std::vector<int>` in `test-io` and `test-io_hdf5`.

Intended to support #2149.


johnomotani:
One current issue with this feature is that `collect()` doesn't handle the vector<int> variables - it would need a special case. For the tests I've just used `boututils.datafile.DataFile` directly. I would like `boutdata.squashoutput()` to work though, so I'll fix `collect()` in the `boutdata` repo once we finalise this.
I think `xbout` might handle them happily already - if not it should be a fairly simple work-around.

johnomotani:
I've added support for `std::vector<char>` because I want to be able to save a git hash as a time-dependent variable (in case code is recompiled between restarts). It's possible `std::string` would be better (?), I didn't think about it because the `std::vector<char>` version that I've added was a fairly simple copy-paste of the `std::vector<int>` version.

I also rebased to squash the commits changing `test-io` and `test-io_hdf5` into one, to minimise the number of binary files in the history. I can revert if anyone has objections to the `std::vector<char>` I/O.

src/fileio/datafile.cxx
```diff
@@ -204,7 +208,21 @@ bool Datafile::openw(const std::string& filename_) {
       throw BoutException("Failed to add int variable {:s} to Datafile", var.name);
     }
   }
+
+  // Add vectors of integers
+  for(const auto& var : int_vec_arr) {
+    if (!file->addVarIntVec(var.name, var.save_repeat, var.ptr->size())) {
+      throw BoutException("Failed to add int vector variable {:s} to Datafile", var.name);
+    }
```
johnomotani, line 216: I've added some checks that the size of the `std::vector`s haven't been changed when adding/reading/writing. Possibly we could make this a warning and use the smaller size, but that risks losing information, so I think an error forcing the user to fix the input is safer.

src/fileio/impls/netcdf4/ncxx4.cxx
```diff
@@ -352,6 +352,97 @@ bool Ncxx4::addVarInt(const string &name, bool repeat) {
...
+        throw BoutException("Found existing variable '{:s}' with size {}. Trying to add "
+                            "with size {}.", name, var.getDim(1).getSize(), size);
+      }
+    } else {
+      ASSERT0(var.getDimCount() == 1);
+      if (var.getDim(0).getSize() != size) {
+        throw BoutException("Found existing variable '{:s}' with size {}. Trying to add "
+                            "with size {}.", name, var.getDim(0).getSize(), size);
+      }
+    }
```
johnomotani, line 395: I didn't bother adding these checks to the hdf5 implementation, because there wasn't a simple way to add them there. Plus I don't think hdf5 is used much (at all?) and is probably being removed for v5.

johnomotani (commented):
Not sure why `test-io_hdf5` is/was failing. It passes on my laptop. Am trying to debug on debug-hdf5-io-fail.

johnomotani:
I think I found the cause of the test failure. When I added `cvar_vec` and `cvar_vec_evol` to `benchmark.out.0.hdf5` I'd created variable-length strings rather than fixed-length strings. I guess newer versions of `h5py` decode some (I think variable-length) strings to `str` and others (fixed-length?) to `bytes`, which made `runtest` on Travis evaluate the output and benchmark as not equal. Should be fixed now, I created the variables with, e.g.
```
import h5py
import numpy as np
f = h5py.File("benchmark.out.0.hdf5", "a")
f.createVariable("cvar_vec", data=np.array([b"a", b"b"]))
```
instead of
```
f.createVariable("cvar_vec", data=["a", "b"])
```
and according to `h5dump` they are identical in the benchmark and my BOUT++ output now.

src/fileio/datafile.cxx
ZedThree: Yes, explicit error is good

src/fileio/impls/hdf5/h5_format.cxx
```diff
@@ -348,21 +349,21 @@ bool H5Format::addVar(const std::string &name, bool repeat, hid_t write_hdf5_typ
...
       hsize_t init_size[3];
       if (parallel) {
-        init_size[0] = mesh->GlobalNx - 2 * mesh->xstart;
+        init_size[0] = lx == 0 ? mesh->GlobalNx - 2 * mesh->xstart : lx;
         if (datatype == "FieldPerp") {
-          init_size[1] = mesh->GlobalNy - mesh->numberOfYBoundaries() * 2 * mesh->ystart;
+          init_size[1] = lz == 0 ? mesh->GlobalNz : lz;
         } else {
-          init_size[1] = mesh->GlobalNz;
+          init_size[1] = ly == 0 ? mesh->GlobalNy - mesh->numberOfYBoundaries() * 2 * mesh->ystart : ly;
```
ZedThree, line 356: I guess you just caught this bug! Do we now have a test that covers it?

ZedThree:
LGTM, but I do think `std::string` will end up being both easier to use and more useful than `std::vector<char>`.  Although it's pretty easy to convert between them, so you might just want to add an overload to do that?

```cpp
std::vector<char> char_vec(string.begin(), string.end());
char_vec.push_back("\0"); // to get null-termination
```

Also, I think this PR just shows how much copy-pasting is required with the current API :slightly_frowning_face:

src/fileio/impls/hdf5/h5_format.cxx
johnomotani: No test - it's in the `parallel=true` branch which doesn't work... So it's useless, but thought I might as well fix while I spotted it!

johnomotani:
Now works with `std::string` too.

I realised that so far there are no tests for reading the new types with `Datafile`. As far as I know the only place we test reads is in `test-restart-io`, but that needs the Python `DataFile` and `collect()` to support the new things too...

ZedThree (approved):

src/fileio/impls/hdf5/h5_format.cxx
ZedThree: Fair enough!

johnomotani:
Having got this far, I'm thinking it would be better to get rid of the `std::vector<char>` in favour of just having a `std::string` interface. Then we could do without the null-termination bit which I think would simplify things a bit (e.g. Python doesn't need a null-terrminator, so we'd probably need to strip it out again). I'll have a think about re-jigging this.

ZedThree:
Replaced by #2153

PR #2152, opened 2020-11-24
Add ability to use FieldPerp in Options

ZedThree:
I've started work on trying to completely replace use of `DataFile` with `OptionsNetCDF`. In order to prevent that PR from being way too big, I'm trying to split it into smaller chunks. This is one necessary part that is useful even if we decide not to use `OptionsNetCDF` and could probably be backported?

This PR actually has a couple of other features itself:

- Add `FieldPerp::getGlobalIndex`, `FieldPerp::setIndexFromGlobal` -- moved code from `DataFormat` methods to `FieldPerp` methods
  - Names maybe not the best?
- Add `Options::hasAttribute` to check existence of attribute
  - This leads to slightly more readable, but less efficient, code

```cpp
const auto attribute = options.attributes.find("key");
if (attribute != options.attributes.end()) {
  use(attribute->second);
} else {
  use(something_else);
}
```
vs
```cpp
if (options.hasAttribute("key")) {
  use(options.attributes.at("key"));
} else {
  use(something_else);
}
```

### DIFF
--- a/src/field/fieldperp.cxx
+++ b/src/field/fieldperp.cxx
@@ -114,6 +114,31 @@ const Region<IndPerp> &FieldPerp::getRegion(const std::string &region_name) cons
   return fieldmesh->getRegionPerp(region_name);
 }
 
+int FieldPerp::getGlobalIndex() const {
+  auto& fieldmesh = *getMesh();
+  const int start = fieldmesh.hasBndryLowerY() ? 0 : fieldmesh.ystart;
+  const int end = fieldmesh.hasBndryUpperY() ? fieldmesh.LocalNy : fieldmesh.yend + 1;
+
+  // Only use the global y index if it's either an interior (grid)
+  // point, or a boundary point. Otherwise, use -1 to indicate a guard
+  // cell or an invalid value. The actual FieldPerp value is still
+  // written to file
+  return (yindex >= start and yindex < end) ? fieldmesh.getGlobalYIndex(yindex) : -1;
+}
+
+FieldPerp& FieldPerp::setIndexFromGlobal(int y_global) {
+  auto& fieldmesh = *getMesh();
+  const int start = fieldmesh.hasBndryLowerY() ? 0 : fieldmesh.ystart;
+  const int end = fieldmesh.hasBndryUpperY() ? fieldmesh.LocalNy : fieldmesh.yend + 1;
+
+  // Only use the global y index if it's either an interior (grid)
+  // point, or a boundary point. Otherwise, use -1 to indicate a
+  // guard cell or an invalid value
+  const int yindex_local = fieldmesh.getLocalYIndex(y_global);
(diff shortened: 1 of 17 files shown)


src/field/fieldperp.cxx
```diff
@@ -48,6 +48,16 @@ FieldPerp::FieldPerp(BoutReal val, Mesh *localmesh) : FieldPerp(localmesh) {
...
 }
 
+FieldPerp::FieldPerp(Array<BoutReal> data_in, Mesh* localmesh, CELL_LOC location_in,
+                     int yindex_in, DirectionTypes directions)
+    : Field(localmesh, location_in, directions), yindex(yindex_in),
+      nx(fieldmesh->LocalNx), nz(fieldmesh->LocalNz), data(std::move(data_in)) {
+
+  ASSERT1(data.size() == nx * nz);
+
+  setLocation(location_in);
```
johnomotani, line 58: Is this call necessary? If `setLocation()` is doing anything that's needed, shouldn't it be in the `Field()` constructor rather than here?

src/sys/options.cxx
```diff
@@ -463,6 +470,88 @@ template <> Field2D Options::as<Field2D>(const Field2D& similar_to) const {
...
+                                  : similar_to.getLocation();
+
+    // Convert from a string using FieldFactory
+    if (bout::utils::holds_alternative<std::string>(value)) {
+      return FieldFactory::get()->createPerp(bout::utils::get<std::string>(value), this,
+                                             similar_to.getMesh(), location);
+    } else if (bout::utils::holds_alternative<Matrix<BoutReal>>(value)) {
+      auto localmesh = similar_to.getMesh();
+      if (!localmesh) {
+        throw BoutException("mesh must be supplied when converting Tensor to Field3D");
```
johnomotani, line 510:
```suggestion
        throw BoutException("mesh must be supplied when converting Matrix to FieldPerp");
```

johnomotani (commented):
Looks good to me, although I'm not at all familiar with OptionsNetCDF.

src/field/fieldperp.cxx
ZedThree: Good point! I copied this from the equivalent `Field3D` constructor. `setLocation` is non-trivial, and normalises/checks `location` (`CELL_DEFAULT` -> `CELL_CENTRE`, throws if `CELL_VSHIFT`, etc). You're right that this is better done in the `Field` constructor though

ZedThree:
@bendudson Would you mind have a quick glance over the `OptionsNetCDF` changes?

bendudson (approved):
Thanks @ZedThree ! Looks good to me.

ZedThree:
We've moved from travis-ci.org to .com, so the failed PR test was restarted, but on .com which doesn't clear the .org version... Basically all tests have passed, we've just got an old notification sticking around. This is likely to the be the case with a few open PRs

PR #2139, opened 2020-11-09
Merge v4.3.2 into next

ZedThree:
There were a few changes to master that didn't make it into next

PR #2153, opened 2020-11-24
I/O for std::vector<int> and std::string

johnomotani:
Refactor of #2150 removing support for `std::vector<char>`.

Closes #2150.


johnomotani:
BTW looks like there was a bad merge of `test-io_hdf5/runtest` in #2139 (duplicated the `check_output()` function). I've fixed in 7f6eb24 - could make a separate PR for just that if it's useful.

ZedThree (approved):

PR #2154, opened 2020-11-25
Fix test-restart-io and FieldPerp writes

johnomotani:
`test-restart-io` was failing but not failing the CI because it always returned 0.

The test bug was hiding a bug in setting the y-index when reading `FieldPerp`s. This should now be mostly consistent with `next`. The exception is that we don't have the upgraded globalYIndex() methods in v4, so the global indices of `FieldPerp`s will be wrong for y-indices higher than the location of the upper target (this is a known, won't-fix bug as the fix changes existing behaviour of the global index methods).

Note these bugs were not present on `next`.


ZedThree (dismissed):

ZedThree:
Thanks @johnomotani !

ZedThree:
This is failing `test-io`:

```
   4 processor.... NXPE=1
      Checking NXPE ... Pass
      Checking variable ivar ... Pass
      Checking variable rvar ... Pass
      Checking variable bvar ... Pass
      Checking variable f2d ... Pass
      Checking variable f3d ... Pass
      Checking variable fperp ... Pass
      Checking variable fperp2 ... Traceback (most recent call last):
  File "/home/travis/build/boutproject/BOUT-dev/build/tests/integrated/test-io/runtest", line 113, in <module>
    if not check_output():
  File "/home/travis/build/boutproject/BOUT-dev/build/tests/integrated/test-io/runtest", line 44, in check_output
    result = collect(v, path="data", info=False)
  File "/home/travis/build/boutproject/BOUT-dev/tools/pylib/boutdata/collect.py", line 596, in collect
    assert fieldperp_yproc is None or fieldperp_yproc == pe_yind
AssertionError
```

johnomotani:
The problem is `fperp2` and `fperp2_evol` get written by two different processes when NYPE=4. I changed the `yindex_global` from excluding y-boundaries to including the yboundary at the beginning of the global grid. Simplest fix is to say 'writing from two processors should never be done' and just change the yindex_global attribute in the benchmark data from 11 to 13. I'd be in favour of that as the simplest solution for v4, the whole issue is better in `next` after #2152... The possibly better, but more work, alternative would be to update the index checks so that `FieldPerp`s in guard cells are not read in, which we should then really test properly.

ZedThree:
How about changing `yindex_global` in the test to make it pass, and noting something like your comment

> The exception is that we don't have the upgraded globalYIndex() methods in v4, so the global indices of FieldPerps will be wrong for y-indices higher than the location of the upper target (this is a known, won't-fix bug as the fix changes existing behaviour of the global index methods).

in `CHANGELOG.md` under a "Known Bugs" section?

ZedThree (approved):

ZedThree:
The Travis job has stopped because we've already run out of Travis credits :tada: 

Looks like we'll have to switch over to github actions in master already.

PR #2156, opened 2020-11-27
Migrate to GitHub Actions

ZedThree:
I've disabled Travis for the time being, rather than delete the `.travis.yml`, as I want to check this runs the coverage tests on the PR correctly without Travis running at the same time.

This is going into `next` -- I'm not sure if the action becomes available in all branches.

I've also tweaked the jobs a little bit, so they're not exactly the same as on Travis. Most are now running on ubuntu-18.04 (Travis is all 16.04), except the CMake one is on 20.04.

It doesn't seem to be any faster than Travis, which is a little bit of a shame.

EDIT: Looks like the coverage jobs aren't running, so that'll be a fun job for Monday.


johnomotani:
> This is going into next -- I'm not sure if the action becomes available in all branches.

I think the actions run on a PR are the ones in the `.github/workflows` directory in that branch. If I remember right I've been debugging them on other repos with pushes to a branch - which is the other way around but I suspect means actions are only available on branches that have merged the `.github/workflows` changes.

ZedThree:
Coverage job still not running on the PR build... think I'll need to investigate in a fork so as to not spam this repo!

I've deleted the `.travis.yml` file and re-enabled Travis so it will run on other branches for now.

ZedThree:
:facepalm: Ignore all that, meant to push those debugging commits to my fork

ZedThree:
Coverage job now working, I've also added a `clang-tidy` job which comments directly on the PR if it finds any issues.

Assuming that all the tests now actually run and pass, this is ready to go in. As it gets merged into other branches, they'll switch over from Travis to GitHub Actions. I've set these tests to be required, but I think I'll have to undo that for the time being, otherwise it's going to be a little confusing as they won't run on other branches

bendudson (approved):

PR #2163, opened 2020-12-01
Stringify BOUT_REVISION when passed as a define on the command line

### LINKED ISSUE #2159: Revision not printed correctly
Running the `elm-pb` example on `next` (actually `fix-shiftoutput`, but the small changes there should be unrelated), the start of the terminal output was:
```
BOUT++ version 5.0.0-alpha
Revision: BOUT_REVISION
MD5 checksum: 22e894bfbc60fd70a63f64d02b558b04
Code compiled on Nov 29 2020 at 12:53:23
```
I think `BOUT_REVISION` should be the git hash. Presumably related to #1920?

BOUT was configured with `./configure --enable-optimize=3`.

Full [BOUT.log.0](https://github.com/boutproject/BOUT-dev/files/5612193/BOUT.log.0.txt). One interesting thing, apparently the hash was passed to the compiler
```
	Compiled with flags : "-DMD5SUM=22e894bfbc60fd70a63f64d02b558b04 -DBOUT_REVISION=df7c980c3bec3adee17067a3dd9190c2ff79f0db -g -Wall -Wextra -Wnull-dereference -Wno-cast-function-type -O3 -march=native -funroll-loops -DBOUT_LOCALE_PATH=/usr/local/share/locale"
```

ZedThree:
Fixes #2159

### DIFF
--- a/include/bout/revision.hxx.in
+++ b/include/bout/revision.hxx.in
@@ -15,7 +15,12 @@ namespace version {
 #ifndef BOUT_REVISION
 constexpr auto revision = "@BOUT_REVISION@";
 #else
-constexpr auto revision = "BOUT_REVISION";
+// Stringify value passed at compile time
+#define BUILDFLAG1_(x) #x
+#define BUILDFLAG(x) BUILDFLAG1_(x)
+constexpr auto revision = BUILDFLAG(BOUT_REVISION);
+#undef BUILDFLAG1
+#undef BUILDFLAG
 #endif
 } // namespace version
 } // namespace bout


ZedThree:
Alternative fix: remove the macro definition from here: https://github.com/boutproject/BOUT-dev/blob/708c718ab0bb34090f3da60583bc1a3bdb4c695e/src/makefile.in#L8

However, this would leave the revision fixed at configure-time.

johnomotani (approved):
This works for me - produces correct terminal output (with the git hash) from `elm-pb`.

Also successfully changes the hash when recompiling on a new commit without reconfiguring (using `./configure`). If I understand right, this should always work as long as `bout++.cxx` is recompiled, which it always is with `./configure` - is it with cmake too?

johnomotani:
Cool, so this feature should work fine with cmake then. The date and time could be a bit confusing since they may be 'wrong', but I think that's a separate issue.

ZedThree:
I actually implemented this header for CMake, because passing the commit hash as a macro definition on the compile line made CMake recompile everything each commit. CMake uses this bit: https://github.com/boutproject/BOUT-dev/blob/2a83b7c6349b9c030077774718c818e984879476/include/bout/revision.hxx.in#L15-L17

PR #2165, opened 2020-12-05
Tidy up coordinate systems with explicit directions

bendudson:
Define the directions of the poloidal and toroidal angles more carefully, so that at each step the direction of each coordinate is clearer.

The new calculations clarify what happens when the poloidal field is negative. For the standard coordinate system (where J can be >0 or <0), the metric tensor doesn't change from what it was before. The calculation of the field line pitch and integrated shear however has been clarified (in terms of which toroidal direction is used).

A new coordinate system is defined, with the Y coordinate reversed for negative poloidal field. This always has a positive Jacobian, and some metric elements reverse sign, but has the same properties (being Clebsch).

### DIFF
--- a/manual/sphinx/user_docs/coordinates.rst
+++ b/manual/sphinx/user_docs/coordinates.rst
@@ -20,36 +20,39 @@ BOUT++ tokamak models, and useful derivations and expressions.
 Orthogonal toroidal coordinates
 ===============================
 
-Starting with an orthogonal toroidal coordinate system
-`\left(\psi, \theta, \zeta\right)`, where `\psi` is the
-poloidal flux, `\theta` the poloidal angle (from `0` to
-`2\pi`), and `\zeta` the toroidal angle (also `0` to
-`2\pi`). We have that the magnetic field
-`{\boldsymbol{B}}` can be expressed as
+Starting with an orthogonal, right-handed toroidal coordinate system
+`\left(r, \theta, \phi\right)`. `\theta` is the poloidal angle (from
+`0` to `2\pi`) in the clockwise direction in the right R-Z
+plane. `\phi` is the toroidal angle (also `0` to `2\pi`) going
+anti-clockwise from the top of the tokamak.
+
+We define the poloidal magnetic field `B_{pol}` as the component of
+the magnetic field in the `\theta` direction, and the toroidal field `B_{tor}`
+as the component of the magnetic field in the `\phi` direction.
+
+We now introduce the poloidal flux `\psi` as the new radial
+coordinate.  If the poloidal magnetic field `B_{pol}` is positive
+then `\psi` increases with radius; if `B_{pol}` is negative then
+`\psi` decreases with radius. To keep the coordinate system
+right-handed, we define a new toroidal coordinate `\zeta` which is
+defined as `\zeta = \sigma_{B\theta}\phi`, where the sign of the
+poloidal magnetic field is `\sigma_{B\theta} \equiv {B_{\text{pol}}}/
+\left|{B_{\text{pol}}}\right|`. If `B_{pol} > 0` then `\zeta` is
+anti-clockwise looking down from above the tokamak, and if `B_{pol} <
+0` then `\zeta` is clockwise. This coordinate system `\left(\psi,
+\theta, \zeta\right)` is orthogonal and right-handed.
+
+The magnitudes of the basis vectors are
 
 .. math::
 
    \begin{aligned}
-    {\boldsymbol{B}}=& B_\theta \nabla \theta + B_\zeta \nabla \zeta\\ =& B_\theta
-       {\boldsymbol{e}}_\theta + B_\zeta {\boldsymbol{e}}_\zeta\\ =& {B_{\text{pol}}}h_\theta {\boldsymbol{e}}_\theta + {B_{\text{tor}}}
-       R {\boldsymbol{e}}_\zeta\\ =& {B_{\text{pol}}}\hat{{\boldsymbol{e}}}_\theta + {B_{\text{tor}}}\hat{{\boldsymbol{e}}}_\zeta\end{aligned}
+   \left|{\boldsymbol{e}}_\psi\right| = \frac{1}{R\left|{B_{\text{pol}}}\right|} \qquad \left|\boldsymbol{e}_\theta\right| = {h_\theta}
+   \qquad \left|\boldsymbol{e}_\zeta\right| = R
+   \end{aligned}
 
-The magnitudes of the unit vectors are
-
-.. math::
-
-   \begin{aligned}
-   \left|\hat{{\boldsymbol{e}}}_\psi\right| = \frac{1}{R\left|{B_{\text{pol}}}\right|} \qquad \left|\hat{{\boldsymbol{e}}}_\theta\right| = {h_\theta}
-   \qquad \left|\hat{{\boldsymbol{e}}}_\zeta\right| = R\end{aligned}
-
-where `{h_\theta}` is the poloidal arc length per radian. The
-coordinate system is right handed, so
-`\hat{{\boldsymbol{e}}}_\psi\times\hat{{\boldsymbol{e}}}_\theta
-= \hat{{\boldsymbol{e}}}_\zeta`,
-`\hat{{\boldsymbol{e}}}_\psi\times\hat{{\boldsymbol{e}}}_\zeta =
--\hat{{\boldsymbol{e}}}_\theta` and
-`\hat{{\boldsymbol{e}}}_\theta\times\hat{{\boldsymbol{e}}}_\zeta
-= \hat{{\boldsymbol{e}}}_\psi`.  The covariant metric coefficients are
+where `{h_\theta}` is the poloidal arc length per radian.
+The non-zero covariant metric coefficients are
 
 .. math::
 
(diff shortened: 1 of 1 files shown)


ZedThree:
Note, built docs are here: https://bout-dev--2165.org.readthedocs.build/en/2165/user_docs/coordinates.html

manual/sphinx/user_docs/coordinates.rst
```diff
@@ -216,8 +252,113 @@ and the covariant metric tensor:
...
+Right-handed field-aligned coordinates
+======================================
+
+If the poloidal magnetic field is negative, i.e. anti-clockwise in the right-hand R-Z plane, then the above
+coordinate system is left-handed and the Jacobian `J` is negative.
+To obtain a consistently right-handed coordinate system, we have to reverse the direction of the `y` coordinate
+when `B_{pol} < 0`:
+
+This `\left(x,y,z\right)` coordinate system is defined by:
+.. math::
```
ZedThree, line 278: I think this needs a blank line before the block:

```suggestion

.. math::
```

manual/sphinx/user_docs/coordinates.rst
```diff
@@ -65,76 +68,109 @@ and the magnitudes of the reciprocal vectors are therefore
...
 .. math::
-   :label: eq:coordtransform
 
    \begin{aligned}
-   x = {\sigma_{B\theta}}\left(\psi - \psi_0\right) \qquad y = \theta \qquad z = \sigma_{B\theta}
-   \left(\zeta - \int_{\theta_0}^{\theta}\nu\left(\psi,\theta\right)d\theta\right)
-   \end{aligned}
+   \nu\left(\psi, \theta\right) = \frac{{\boldsymbol{B}}\cdot\nabla\phi}{{\boldsymbol{B}}\cdot\nabla\theta} =
+   \frac{{B_{\text{tor}}}{h_\theta}}{{B_{\text{pol}}}R}
+
```
ZedThree, line 116: Looks like the environment close statement got clobbered:

```suggestion
   \end{aligned}
```

manual/sphinx/user_docs/coordinates.rst
```diff
@@ -65,76 +68,109 @@ and the magnitudes of the reciprocal vectors are therefore
...
 
    \begin{aligned}
-   \nabla x = {\sigma_{B\theta}}\nabla \psi \qquad \nabla y = \nabla \theta \qquad \nabla z =
-   {\sigma_{B\theta}}\left(\nabla\zeta - \left[\int_{\theta_0}^\theta{\frac{\partial \nu\left(\psi,
-   \theta\right)}{\partial \psi}} d\theta\right] \nabla\psi - \nu\left(\psi, \theta\right)\nabla\theta\right)\end{aligned}
-
+   \nabla x = {\sigma_{B\theta}}\nabla \psi \qquad
+   \nabla y = \nabla \theta \qquad
+   \nabla z = \nabla\zeta - \sigma_{B\theta}\left[\int_{\theta_0}^\theta{\frac{\partial \nu\left(\psi,\theta\right)}{\partial \psi}} d\theta\right] \nabla\psi
+   - \sigma_{B\theta}\nu\left(\psi, \theta\right)\nabla\theta\right)
```
ZedThree, line 147: I think this closing bracket is superfluous now? Either that, or there's a missing `\left(` somewhere

```suggestion
   - \sigma_{B\theta}\nu\left(\psi, \theta\right)\nabla\theta
```

manual/sphinx/user_docs/coordinates.rst
```diff
@@ -65,76 +68,108 @@ and the magnitudes of the reciprocal vectors are therefore
...
 
+  
+The basis vectors are:
+
+.. math::
+   
+   \begin{aligned}
+   \boldsymbol{e}_x =& J\left(\nabla y \times \nabla z\right) = {\sigma_{B\theta}} {\boldsymbol{e}}_\psi + I{\boldsymbol{e}}_\zeta \\
+   \boldsymbol{e}_y =& J\left(\nabla z \times \nabla x\right) = {\boldsymbol{e}}_\theta + \nu{\boldsymbol{e}}_\phi \\
+   \boldsymbol{e}_z =& J\left(\nabla x \times \nabla y\right) = {\boldsymbol{e}}_\zeta
```
johnomotani, line 164: Following up to this point, I get a `\sigma_{B\theta}` everywhere in these basis vectors. Sorry! @bendudson did you maybe change a definition somewhere above?
```suggestion
    \boldsymbol{e}_x =& J\left(\nabla y \times \nabla z\right) = {\boldsymbol{e}}_\psi + {\sigma_{B\theta}} I{\boldsymbol{e}}_\zeta \\
   \boldsymbol{e}_y =& J\left(\nabla z \times \nabla x\right) = {\sigma_{B\theta}} \left({\boldsymbol{e}}_\theta + \nu{\boldsymbol{e}}_\phi\right) \\
   \boldsymbol{e}_z =& J\left(\nabla x \times \nabla y\right) = {\sigma_{B\theta}} {\boldsymbol{e}}_\zeta
```

bendudson: Thanks @johnomotani but I don't think that can be right: Where `I = 0`, surely `e_x = sigma_B e_psi` because e_x is outwards, while e_psi would be inwards if sigma_B < 0, and if the toroidal field is zero then `e_y = e_theta`

bendudson: For `e_x` I have:
`e_x = J (e^y x e^z) = h_theta / B_theta * ( e^theta x e^zeta - sigma_B * I * e^theta x e^psi)`

then `e^theta x e^zeta = |B_theta| / h_theta e_psi`  (note modulus, because (psi, theta, zeta) is right-handed) and `e^theta x e^psi = - |B_theta| / h_theta e_zeta`.

bendudson: There is definitely potential for confusion, because the `J` in this section is for the (x,y,z) coordinates, which is different (Bpol vs |Bpol|) from the J for the (psi, theta, zeta) system in the section above

johnomotani: Ah, I think that would do it. So the `J` in the 'Field-aligned coordinates' section is for {x,y,z}, and is negative when the coords are left-handed, which they are when Bp<0. But the `J` in 'Orthogonal toroidal coordinates' is for {psi,theta,zeta} which are always right-handed so there J is +ve always. I think different symbols and a comment that their sign might be different would be helpful :smiley:

johnomotani: Maybe `J_{\psi\theta\zeta}` and `J_{xyz}`?

bendudson: Ah yes that's a good idea. Thanks @johnomotani !

johnomotani: Maybe I shouldn't be doing this so late, but is the `J` in the 'Magnetic field' section also the {x,y,z} one that's `J = \sigma_{B\theta} B_\mathrm{pol} / h_\theta`? Which would make `\boldsymbol{B} = \sigma_{B\theta} \frac{B_\mathrm{pol}}{h_\theta} \boldsymbol{e}_y, which was what we wanted to avoid? That seems to add up with `y=\theta` meaning y is always clockwise, so \nabla_y is in the opposite direction to Bp if Bp is negative (anticlockwise)... Maybe we should have `y = \sigma_{B\theta} \theta`? If we can/should do that it would have the advantage of x/y/z always being right-handed.

bendudson: Thanks John! Yes, in the next section I've written out the results for `y = \sigma_{B\theta} \theta`. Following the earlier confusion, I've re-labelled `y` to make `\eta = \sigma_{B\theta} \theta` for this right-handed coordinate system.

johnomotani (commented):
A couple of typos and a clarification in the review comments.

Another question: should the Laplacian and perpendicular Laplacian expressions be written in terms of metric components so that if we're in the xyz coords, they're implicitly using d/dy and g_{yz}, etc. with \sigma_{B\theta} included?

I think we're going to end up using the 'x\etaz' coordinates for simulations with negative Bp? In that case, would it be better to swap the names y<->\eta so that 'xyz' is the (always right-handed) coordinate system that we'll actually use?

manual/sphinx/user_docs/coordinates.rst
```diff
@@ -65,76 +68,112 @@ and the magnitudes of the reciprocal vectors are therefore
...
-   :label: eq:coordtransform
 
    \begin{aligned}
-   x = {\sigma_{B\theta}}\left(\psi - \psi_0\right) \qquad y = \theta \qquad z = \sigma_{B\theta}
-   \left(\zeta - \int_{\theta_0}^{\theta}\nu\left(\psi,\theta\right)d\theta\right)
+   \nu\left(\psi, \theta\right) = \frac{{\boldsymbol{B}}\cdot\nabla\phi}{{\boldsymbol{B}}\cdot\nabla\theta} =
+   \frac{{B_{\text{tor}}}{h_\theta}}{{B_{\text{pol}}}R}
    \end{aligned}
+The sign of the poloidal field `{B_{\text{pol}}}` and toroidal field 
+`{B_{\text{tor}}}` can be either + or -.
```
ZedThree, line 117: Just to clarify, are the signs independent of each other? Or do they both need to be the same sign?

manual/sphinx/user_docs/coordinates.rst
```diff
@@ -216,8 +255,170 @@ and the covariant metric tensor:
...
+
+   \begin{aligned}
+   \texttt{zShift}\left(x, y\right) &= \int_{y = 0}^{y}\frac{{\boldsymbol{B}}\cdot\nabla z}{{\boldsymbol{B}}\cdot\nabla y} dy \\
+   &= \int_{\theta = 0}^{\theta}\frac{{\sigma_{B\theta}}{B_{\text{tor}}}{h_\theta}}{{B_{\text{pol}}}R} d\theta \\
+   &= {\sigma_{B\theta}} \int_{\theta = 0}^{\theta} \nu d\theta
+   \end{aligned}
+
+The `\texttt{ShiftAngle}` is then defined as the change in
+`\texttt{zShift}` between `y=0` and `y=2\pi`: It is the change in the
+`Z` coordinate after one poloidal circuit in `Y`.
```
ZedThree, line 289: Maybe lowercase `z`, `y` for consistency?

```suggestion
`z` coordinate after one poloidal circuit in `y`.
```

ZedThree (approved):
LGTM, assuming @johnomotani is happy with the maths! :)

manual/sphinx/user_docs/coordinates.rst
bendudson: Ah yes, they are independent, but that could be clearer in the text

ZedThree (dismissed):

johnomotani (commented):
A few more minor comments...

manual/sphinx/user_docs/coordinates.rst
```diff
@@ -216,8 +255,170 @@ and the covariant metric tensor:
...
+field. It is the `z` angle of a point on a field line relative to a
+reference location:
+
+.. math::
+
+   \begin{aligned}
+   \texttt{zShift}\left(x, y\right) &= \int_{y = 0}^{y}\frac{{\boldsymbol{B}}\cdot\nabla z}{{\boldsymbol{B}}\cdot\nabla y} dy \\
+   &= \int_{\theta = 0}^{\theta}\frac{{\sigma_{B\theta}}{B_{\text{tor}}}{h_\theta}}{{B_{\text{pol}}}R} d\theta \\
+   &= {\sigma_{B\theta}} \int_{\theta = 0}^{\theta} \nu d\theta
+   \end{aligned}
```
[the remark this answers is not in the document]
bendudson: B dot e_z = 0, but B dot nabla z is the toroidal field (with a +/-)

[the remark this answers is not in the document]
johnomotani: I'm not sure about that:
* nabla z != nabla zeta, and B dot nabla zeta is the toroidal field (with a +/-)
* we can write \vec{B} = Bp / htheta * \vec{e}_y, and e_i dot \nabla x^j = delta_i^j, so e_y dot \nabla z = 0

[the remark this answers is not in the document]
johnomotani: Isn't it B dot e_z that's !=0?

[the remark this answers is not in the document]
bendudson: Ah yes, quite right sorry: `B = nabla x cross nabla z` so `B dot nabla z` must be zero

johnomotani:
Do we still need the 'Shifted radial derivatives' subsection? I guess it's for 'BOUT06-style' `IncIntShear` derivatives that we don't use any more (and should remove for v5?)? I've skipped checking that one...

johnomotani (approved):
@bendudson I think we can merge this if you're happy with it.

PR #2166, opened 2020-12-07
Boutcore improvements

dschwoerer:
I think the documentation should go to master.

The fix is also probably fine in master, but that might have unintended side effects (the language level part)

1d51901 might as well go some where else, but as it is an addition, should also be fine in master.


bendudson (approved):

dschwoerer:
Will master be merged into next, or should I rebase onto next?

ZedThree:
Probably easiest/quickest to rebase off next right now

Issue #2128, opened 2020-10-27
Clean up Coordinates() constructor


johnomotani:
At the moment the `Coordinates::Coordinates()` constructor (plus some utility methods/functions) does a lot of interpolation, extrapolation and checking. This was useful to support new features, but has become more complicated and hard to maintain. It is also difficult for users to find and fix problems with grids if they arise. Extrapolation (especially) is still useful, because geqdsk files may not define the poloidal flux function far enough past the walls for a grid generator to create a grid with boundary points, or the grid may run into coils/singularities, etc.

A proposal for discussion:
1. Change `Coordinates::Coordinates()` so that it does no extrapolation or interpolation. Grid files are required to include boundary points, and if staggered `Coordinates` are created, the grid file must include metrics, etc. at the staggered grid locations. It should check for positivity, etc. as appropriate, and also require sane values in corner guard/boundary points.
2. Write a grid file validation and conversion tool. This would take a grid file that might be missing staggered quantities and/or boundary points and apply extrapolation/interpolation to generate them. It should also allow users to easily check grid files and fix problems.

Desired features of the conversion tool:
* A GUI similar to `hypnotoad-gui` would be ideal, that like `hypnotoad` can also create an input file for reproducible command-line conversions.
* Can take input grid files with/without staggered quantities and with/without boundary cells.
* To start with have at least the same extrapolation/interpolation algorithms currently implemented in `coordinates.cxx`, but would be easier to add new ones (e.g. to ensure positivity in problematic cases).
* Check for errors (same checks as `Coordinates::Coordinates()`, e.g. negative `B`, etc. Ideally provide some options to fix the problems.
* Would be nice to have functionality for further validation of grid files, e.g. a standard set of plots so users can check visually that there are no excessively large or small values, and that quantities are smooth. Probably a combination of 2d color-plots for all quantities and 1d line plots for selected quantities (e.g. `dx`, `hthe`/`hy`).
  * **Edit** would be nice to have a check of the 'no slip condition' from Dudson et al. (2009): Delta psi * Delta theta * d nu / d psi <= Delta zeta
* Possibly this could be implemented as an extra tab/mode in `hypnotoad`? The stuff above would certainly always be useful for validation of `hypnotoad` outputs.

See also #1472 - I think addressing that would be much nicer in a Python conversion tool than as more complication in `coordinates.cxx`.

johnomotani:
***Help wanted!***

[...]

PR #2167, opened 2020-12-07
Testing of Coordinates 

bendudson:
PR #2165 opened a bit of a can of worms... I'd like to be able to test the Coordinates constructor, to handle negative Jacobian cases and other things. Looking through the Coordinates code my head started to hurt and so I started a refactoring... I've now stepped back from the brink, and decided to put tests in place first before changing everything.

I'd like to remove the `Mesh->get` routines, and all the `GridDataSource` stuff; I think it would be better if the values needed were just taken from an `Options` input. That could then be read from NetCDF, BOUT.inp or other source. 

For now, `FakeGridDataSource` can be given some values in the tests, allowing testing of parts of the code (e.g. Coordinates) which use `mesh->get`. 

Basic tests of `Coordinates` included. Doesn't test lots of the functionality, in particular
- Extrapolation into boundary cells
- Staggered coordinates, and interpolation. Interpolation needs >= 2 boundary cells (for 4th-order?), which `FakeMesh` currently doesn't have.


johnomotani:
See also #2128?

ZedThree:
+1 for replacing `Mesh::get` with `Options` -- it's got a much nicer interface. I think it should be reasonably easy to mechanically convert between the two as well.

I'm not sure `readability-named-parameter` is a very useful warning for us from clang-tidy. You can use `MAYBE_UNUSED` to silence it, but it might be better if we just turn it off.

ZedThree:
A whole bunch of unhelpful comments here from googletest internals :( Even with the latest googletest, there's still a ton of warnings.

I turned the magic-numbers warnings in another PR, as that is spectacularly unhelpful.

Might just need to turn off clang-tidy on the unit tests, this is a bit too spammy.

ZedThree (approved):

PR #2170, opened 2020-12-08
Convert Travis to Github Actions (master)

ZedThree:
Currently failing on `test-restart-io_hdf5`: https://github.com/boutproject/BOUT-dev/runs/1516674244?check_suite_focus=true#step:9:1113

I think the updates to the netcdf files in https://github.com/boutproject/BOUT-dev/pull/2154/files are also required for the HDF5 files


johnomotani (approved):
The problem with `test-io_hdf5` is fixed.

Looks good to me. Just to check: the coverage check has run on this PR - is that intended? I wasn't sure if the `is_master_or_next` thing was intended to restrict coverage runs to any PR into `master` or `next` (in which case it should have run here) or is only meant to be run on commits _to_ `master` or `next` (in which case I guess it shouldn't have run)?

ZedThree:
Sorry, it's not the best name -- it's commits _or_ PRs to either `master` or `next`

Issue #2169, opened 2020-12-07
Problem running hasegawa-wakatani-3d example in next-hypre-outerloop branch with petsc3damg type for Laplacian


ADimits:
Opening this issue to communicate/document the problem - remind me how best to pass information, e.g., environment, run logs, etc.

Bout.inp:

[mesh]

nx = 68  # Note 4 guard cells in X
ny = 16 
nz = 64  # Periodic, so no guard cells in Z

dx = 0.2
dy = 1.
dz = 0.2

NXPE = 1

# Set periodic boundary condition in y
ixseps1 = nx
ixseps2 = nx 

[laplace]

flags = 0   # Flags for Laplacian inversion
type = petsc3damg

.
.

$ mpirun -n 1 hw
.
.

Run started at  : Mon 07 Dec 2020 10:23:48 AM PST
	Option restart = false (default)
	Option append = false (default)
	Option dump_on_restart = 1 (default)
	Option wall_limit = -1 (default)
	Option stopCheck = 0 (default)
Sim Time  |  RHS evals  | Wall Time |  Calc    Inv   Comm    I/O   SOLVER

0.000e+00          1       6.82e-01    98.8    0.0    0.1    0.8    0.3
1.000e+00        435       2.54e+01    95.5    0.0    0.3    0.1    4.1
.
.

$ mpirun -n 2 hw

.
.
Finding value for NXPE (ideal = 2.828427)
	Candidate value: 1
	 -> Good value
	Candidate value: 2
	 -> Good value
	Domain split (NXPE=2, NYPE=1) into domains (localNx=32, localNy=16)
.
.

Run started at  : Mon 07 Dec 2020 10:23:16 AM PST
	Option restart = false (default)
	Option append = false (default)
	Option dump_on_restart = 1 (default)
[0]PETSC ERROR: ------------------------------------------------------------------------
[0]PETSC ERROR: Caught signal number 11 SEGV: Segmentation Violation, probably memory access out of range
.
.

ADimits:
Runs OK for several of the other solver options (e.g. cyclic)

Results of key unit and integrated tests:

Unit tests - all serial?

$ make check-unit-tests

[==========] 1918 tests from 127 test suites ran. (4851 ms total)
[  PASSED  ] 1918 tests.


With the following in
tests/integrated/test-laplace-petsc3d/runtests
.
.
test_directories = [('data_slab_core', 4),
                    ('data_slab_sol', 4),
[...]

PR #2171, opened 2020-12-09
Update docs on supported PETSc versions

johnomotani:
The manual on `next` had been updated to note we support up to (at least) PETSc-3.13, but `master` had not. The version from `master` appears (I think) on bout-dev.readthedocs.io/ and was out of date, which has caused some confusion (https://github.com/boutproject/BOUT-dev/issues/2169#issuecomment-741462821). This PR updates the `master` docs to match `next` (I have compiled some version of 4.3 with PETSc-3.13 so this should be fine).

Also adds a note that newer versions of PETSc may work, and copies across extra installation hints for Marconi from the `next` docs.

By the way, the very latest PETSc is 3.14 - has anyone tried compling/testing that? If so we could update here before merging.

### DIFF
--- a/manual/sphinx/user_docs/advanced_install.rst
+++ b/manual/sphinx/user_docs/advanced_install.rst
@@ -178,6 +178,84 @@ to enable AVX512 vectorization.
           ``boutcore.pyx`` needs to be manually triggered, if
           ``boutcore.pyx`` has already been created.
 
+Marconi with gnu compilers
+**************************
+
+It is also possible to configure on Marconi using gnu compilers, which may give better performance. A set of modules which work as of 30/9/2020 is
+
+.. code-block:: bash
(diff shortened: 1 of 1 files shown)


dschwoerer:
Yes, fedora rawhide ships with bout++ compiled against sundials 5.5.0 
and petsc 3.14.1

johnomotani:
Thanks @dschwoerer!

I haven't updated the install instructions for PETSc from 3.13 to 3.14. If anyone has time to check the installation and update them, then feel free! Otherwise, I think this is ready to merge once the docs CI passes.

ZedThree (approved):

ZedThree:
Failing test is coverage on master has somehow slipped 0.06% below the target 30%

PR #2173, opened 2020-12-09
Allow registering "unavailable" Factory options

ZedThree:
This is so we can give a more helpful error message when trying to use, for example, the PETSc solver but BOUT++ was not configured with PETSc

I've gone for "unavailable", rather than "unsupported" -- we do support these things, they're just not able to be used in this configuration.

Used like:

```cpp
#if not BOUT_HAS_PETSC

namespace {
RegisterUnavailableLaplaceXZ
    registerlaplacexzpetsc("petsc", "BOUT++ was not configured with PETSc");
}

#else

// actual implementation
```

Output looks like:

```
$ ./my_model --list-solvers
adams-bashforth
euler
karniadakis
power
pvode
rk3ssp
rk4
rkgeneric
splitrk

The following Solvers are currently unavailable:
arkode (BOUT++ was not configured with ARKODE/SUNDIALS)
cvode (BOUT++ was not configured with CVODE/SUNDIALS)
ida (BOUT++ was not configured with IDA/SUNDIALS)
imdexbdef2 (BOUT++ was not configured with PETSc)
petsc (BOUT++ was not configured with PETSc)
slepc (BOUT++ was not configured with SLEPc)
snes (BOUT++ was not configured with PETSc)
```

and on trying to use an unavailable thing:

```
Error when trying to create a Solver: 'petsc' is not available because BOUT++ was not configured with PETSc
Available Solvers are:
adams-bashforth
euler
karniadakis
power
pvode
rk3ssp
rk4
rkgeneric
splitrk
```

I just saw John also suggested using a second constructor to `RegisterInFactory` that takes the reason. It does cut down on a little bit of code, but I think the `RegisterUnavailableInFactory` is only slightly more code and makes it a bit more explicit what we're doing.

### DIFF
--- a/include/bout/generic_factory.hxx
+++ b/include/bout/generic_factory.hxx
@@ -64,6 +64,12 @@
 template <class BaseType, class DerivedFactory,
           class TypeCreator = std::function<std::unique_ptr<BaseType>(Options*)>>
 class Factory {
+  /// Storage of the creation functions
+  std::map<std::string, TypeCreator> type_map;
+
+  /// Known implementations that are unavailable, along with the reason
+  std::map<std::string, std::string> unavailable_options;
+
 protected:
   // Type returned from the creation function
   using ReturnType = typename TypeCreator::result_type;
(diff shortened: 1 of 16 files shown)


johnomotani (dismissed):
Looks good to me :+1:

johnomotani (approved):

PR #2175, opened 2020-12-10
Add requirements to boutcore/print test

ZedThree:
Missing requirements was causing test to be built during `build-check` despite not needing, and run even if `boutcore` was not built

### DIFF
--- a/tests/integrated/test-boutcore/print/runtest
+++ b/tests/integrated/test-boutcore/print/runtest
@@ -1,4 +1,8 @@
 #!/bin/sh
+
+# requires boutcore
+# requires not make
+
 set -ex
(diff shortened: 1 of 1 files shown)


dschwoerer (approved):
:+1:

dschwoerer:
I see tests didn't run on the other PR, that is why I didn't notice. Thanks for fixing it :+1:

ZedThree:
I think the codecov is failing now because when switching to github actions, I only added a single coverage build -- previously we were running a coverage build with everything enabled and one with everything disabled. I guess that is actually enough to drop the coverage by 3%!

Added to my to-do list for when I get back in the new year.

PR #2177, opened 2020-12-10
Completely remove LaplaceShoot

ZedThree:


bendudson (approved):
Thanks! It seemed like a good idea at the time, but really wasn't.

PR #2179, opened 2020-12-15
Simplify Coordinates in 3d metrics branch

johnomotani:
Removes call to `Coordinates::geometry()` from the `Coordinates` constructor. This results in a 2-step initialisation of `Coordinates` and means that at the places where we have to calculate derivatives, the `Field`s have a valid `Coordinates` pointer. We can then remove some work-arounds for not-existing Coordinates.


johnomotani:
This PR now includes a merge of recent updates in `next`, because I needed #2183, which was based on `next`. I've also fixed a couple of tests, and skipped a couple of others that seemed like they ought to fail - now the test suite passes with 3d metrics enabled, on my laptop and on CI.

The main commits to review are just 11ba037, 91174cc, 5a77878 and 6a9e00b.

9e58e8a and cd0500f add some checking to `ShiftedMetric`.

8d30f3c, a84789f, 865d274, c0fda06, 98cbe71 and ff6b801 were fixing or skipping tests (not all directly related to this PR).

src/mesh/mesh.cxx
```diff
@@ -433,11 +438,13 @@ std::shared_ptr<Coordinates> Mesh::createDefaultCoordinates(const CELL_LOC locat
...
     // Initialize coordinates from input
-    return std::make_shared<Coordinates>(this, options);
+    auto result = std::make_shared<Coordinates>(this, options);
+    return result;
   } else {
     // Interpolate coordinates from CELL_CENTRE version
-    return std::make_shared<Coordinates>(this, options, location,
+    auto result = std::make_shared<Coordinates>(this, options, location,
         getCoordinates(CELL_CENTRE), force_interpolate_from_centre);
+    return result;
```
ZedThree, line 447: Was this change because you were going to do `result.geometry()` before returning it? Was there an issue with that?

ZedThree (approved):

src/mesh/mesh.cxx
johnomotani: Yes and yes. 6a9e00b1553ed9414402677937aef53c5e4fe45b moved the `geometry()` call to `getCoordinatesSmart()` to make sure that `field.getCoordinates()` never returns `nullptr`.

johnomotani, line 447:
```suggestion
    return std::make_shared<Coordinates>(this, options, location,
        getCoordinates(CELL_CENTRE), force_interpolate_from_centre);
```

src/mesh/mesh.cxx
```diff
@@ -433,11 +438,13 @@ std::shared_ptr<Coordinates> Mesh::createDefaultCoordinates(const CELL_LOC locat
 
   if (location == CELL_CENTRE || location == CELL_DEFAULT) {
     // Initialize coordinates from input
-    return std::make_shared<Coordinates>(this, options);
+    auto result = std::make_shared<Coordinates>(this, options);
+    return result;
```
johnomotani, line 442:
```suggestion
    return result = std::make_shared<Coordinates>(this, options);
```

johnomotani:
The test failures are just timeouts (which should be fixed if we merged next again now #2197's gone in), so I think this is ready to go. I'll leave merging to the 3d metrics team :smiley:

PR #2180, opened 2020-12-15
Check fieldCoordinates is not nullptr

johnomotani:
While debugging #2179, I ran into segfaults due to `Field.getCoordinates()` returning `nullptr`. In a few places a check for `field.getCoordinates() == nullptr` was used, but by modifying those, we can add a check to `Field.getCoordinates()` that the result to be returned is not `nullptr` to avoid those segfaults (at least when `CHECK > 0`).

Could @dschwoerer, @bshanahan or @ZedThree check the logic around the `ParallelTransform` pointers in the `Coordinates` constructor please? I had to change the conditionals to be things like `field.getLocation() == location` instead of `field.getCoordinates() == nullptr`. I think the result is the same, but it's quite complicated to follow. Hopefully we can simplify later in resolving #2128...

Making this as a separate PR into the #2179 branch to make it easier to review. If #2179 is merged into `coords3d_merged2` we could merge this directly into `coords3d_merged2` instead.

### DIFF
--- a/include/bout/mesh.hxx
+++ b/include/bout/mesh.hxx
@@ -658,6 +658,12 @@ class Mesh {
 
     auto found = coords_map.find(location);
     if (found != coords_map.end()) {
+#if CHECK > 0
+      if (found->second == nullptr) {
+        throw BoutException("Coordinates pointer at {} is nullptr. Probably means "
+                            "Coordinates are being constructed.", toString(location));
+      }
+#endif
       // True branch most common, returns immediately
       return found->second;
     }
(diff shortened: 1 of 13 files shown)


johnomotani:
This PR may be a bad idea... I fixed most of the tests, but `test-coordinates-initialisation` failed. From the discussion in #1369, it seems that calling `getCoordinates()` in `setLocation()` is necessary to avoid weird bugs. But we can't do that while forbidding `getCoordinates()` from returning `nullptr`, because staggered `Coordinates` have to (?) call `setLocation()` on their members. Possibly one option would be to move all the initialisation into a `Coordinates::load()` method that could be called after `Coordinates` is constructed so the `fieldCoordinates` pointers could always be valid. That seems like a choice between two bad things, so (a) we might as well stick with the one we've got and (b) I definitely don't want to hold up #2025 with this: I'll close this for now. Maybe we can revisit if we've simplified the `Coordinates` constructor (#2128).

PR #2184, opened 2020-12-17
Removing hypnotoad grid generator

bendudson:
Both IDL and python versions are now replaced with the python Hypnotoad in separate repository https://github.com/boutproject/hypnotoad

- Don't really want to maintain the IDL version, but it could be made a separate repository
- The python version in BOUT-dev was never completed, and only works for particular cases (only in the core iirc).


johnomotani:
Should also update the manual here https://github.com/boutproject/BOUT-dev/blame/next-remove-gridgen/manual/sphinx/user_docs/input_grids.rst#L328-L335

johnomotani (approved):

Issue #2185, opened 2020-12-29
test-invpar timing out


johnomotani:
`test-invpar` keeps timing out in the debug CI build. Options:
* make a shorter version of the test, with fewer combinations of options, and mark the current one as `all_tests`
* increase the timeout for `test-invpar`
* increase the timeout for the debug CI build
* other suggestions?

ZedThree:
`test-drift-instability-staggered` is also timing out.

Actually, looks like everything has slowed down across the board for debug builds:

|                                 | #2164     | 7d4d427      |
|---------------------------------|-------------|--------------|
| test-code-style                 | ✓   0.071 s | ✓   0.225 s |
| test-command-args               | ✓  12.123 s | ✓  16.719 s |
| test-communications             | ✓  72.113 s | ✓  85.884 s |
| test-coordinates-initialization | ✓   0.768 s | ✓   1.183 s |
| test-cyclic                     | ✓  10.978 s | ✓  17.765 s |
| test-delp2                      | ✓  32.179 s | ✓  60.645 s |
| test-drift-instability-staggered| ✓ 148.480 s | 💤 300.006 s |
| test-fieldfactory               | ✓   3.581 s | ✓   5.856 s |
| test-griddata                   | ✓   0.877 s | ✓   1.601 s |
| test-griddata-yboundary-guards  | ✓  24.318 s | ✓  28.254 s |
| test-gyro                       | ✓   4.227 s | ✓   6.282 s |
| test-initial                    | ✓  18.776 s | ✓  28.354 s |
| test-interchange-instability    | ✓ 130.239 s | ✓ 188.292 s |
| test-interpolate                | ✓  12.111 s | ✓  21.008 s |
| test-interpolate-z              | ✓   3.555 s | ✓   4.084 s |
| test-invertable-operator        | ✓   1.815 s | ✓   2.328 s |
| test-invpar                     | ✓ 195.222 s | 💤 300.016 s |
| test-io                         | ✓  54.116 s | ✓  87.950 s |
| test-io_hdf5                    | ✓  34.380 s | ✓  52.547 s |
| test-laplace                    | ✓   5.453 s | ✓   8.991 s |
| test-laplace-petsc3d            | ✓  24.500 s | ✓  47.673 s |
| test-laplacexy-fv               | ✓  46.032 s | ✓  60.678 s |
| test-laplacexy-short            | ✓  84.505 s | ✓ 115.010 s |
[...]

PR #2188, opened 2021-01-02
Fix reading of char* in Ncxx4

johnomotani:
Fix for feature introduced in #2153. I guess it needed a better test that uses the DataFormat::read() methods. I don't have time to work on one now, if anyone does, feel free to push!


johnomotani:
Test failures seem unrelated to this bug-fix. The 'debug' tests are timing out (#2185), but the cmake tests are failing in test-squash, which is a bit worrying...

bendudson (approved):

PR #2196, opened 2021-01-03
Replace boutdata and boututils directories with submodules

johnomotani:
Get boutdata and boututils from the separate git repos (github.com/boutproject/boutdata and github.com/boutproject/boututils), including them as submodules.

I did check that the current versions of `boututils` and `boutdata` in `next` are identical to the versions in the separate git repos.


johnomotani:
I added symlinks `tools/pylib/boutdata -> externalpackages/boutdata/boutdata` and `tools/pylib/boututils -> externalpackages/boututils/boututils` so the changes should be totally backward-compatible, as long as the submodules have been checked out.

bendudson (dismissed):
Thanks @johnomotani removing this duplication is good.

johnomotani:
I think there's a problem with the automatic version numbering, that was recently added to boututils and boutdata, possibly related to the way I used symlinks. Investigating now...

johnomotani:
When https://github.com/boutproject/boututils/pull/16 and https://github.com/boutproject/boutdata/pull/26 are merged, we can update the submodules and then the tests will pass.

johnomotani:
The Fedora builds are failing. I need to add the Python `setuptools_scm` package, but I'm not sure how to do that for Fedora. It looks like there is a Fedora package https://src.fedoraproject.org/rpms/python-setuptools_scm. @dschwoerer @ZedThree where/how can I add it? Thanks!

johnomotani:
> Is setuptools_scm now a hard dependency to use the boutdata/boututils submodules in this repo?

Yes. It's not required when boutdata/boututils are installed, but since we use them straight from the git repos in the the submodules it is required.

johnomotani:
I added `python-setuptools_scm` to the Fedora build as @ZedThree suggested, and `setuptools_scm` is now there. Now it complains that git isn't there, and therefore setuptools_scm can't get the version number for boututils. I'm a bit lost now. @dschwoerer what should we be doing with submodules in Fedora?

ZedThree:
Just adding `git` to the same line should be sufficient I think? The Fedora build creates a docker image and copies the checked-out source into it, so the image doesn't actually use git.

johnomotani:
> The Fedora build creates a docker image and copies the checked-out source into it, so the image doesn't actually use git.

If it does the same for the submodules then I guess git wouldn't actually help, but if we go with the suggestion in https://github.com/boutproject/BOUT-dev/pull/2198#issuecomment-754695834 to avoid a hard requirement on `setuptools_scm` then we'd get around this issue anyway.

ZedThree (approved):

PR #2197, opened 2021-01-04
Fix test-squash

ZedThree:
Ignore some variables when comparing squashed outputs in test-squash:

The CVODE/ARKODE solver internal diagnostics may depend on processor
count, etc., so should not be compared across runs with different
processor counts


ZedThree:
Increasing the timeout to 6m seems to have been sufficient. The pull request seems to be the slow one this time round, overall about 5 minutes slower than the push build.

johnomotani (approved):
Looks good to me.

PR #2198, opened 2021-01-04
Replace boutdata and boututils directories with submodules (v4.4)

johnomotani:
Backport of #2196.

Get `boutdata` and `boututils` from the separate git repos (github.com/boutproject/boutdata and github.com/boutproject/boututils), including them as submodules.

There were some differences between the existing versions of `boututils` and `boutdata` in v4.4.0-alpha and the versions in the git repos - I think these were just fixes and updates that had not been backported to the v4.4.0-alpha branch.


ZedThree:
Is there a way to work around needing `setuptools_scm`? Otherwise this is a bit of breaking change, so I'm not so sure about backporting it to 4.4

johnomotani:
> Is there a way to work around needing setuptools_scm? Otherwise this is a bit of breaking change, so I'm not so sure about backporting it to 4.4

We could just set `__version__ = "dev"` or something like that as a fall-back, and print a warning saying to get a version number you need `setuptools_scm`?

ZedThree:
Yes, that sounds reasonable! :+1:

ZedThree (approved):

PR #2201, opened 2021-01-06
Minor FCI improvements

ZedThree:
- Remove some dead code
  - The `i/k_corner` and `t_x/z` stuff is no longer used so can just go
- Convert nested-loop to `BOUT_FOR`
  - Actually `BOUT_FOR_SERIAL` as it's got a `vector.push_back` in the loop and I suspect that's not thread safe 
- Pass in `dy` to avoid null-dereference
  - It turns out we don't actually have a test for any parallel boundaries, but if we did they wouldn't work! We construct the FCI transform inside the `Coordinates` constructor, which means we can't call `Mesh::getCoordinates`. Instead, pass in `dy`

(It's a bit easier to view the diff if you select "hide whitespace changes")


bshanahan (approved):
Looks good to me.

PR #2204, opened 2021-01-08
Correct Laplacexz petsc implementation of g13 terms

bendudson:
Off-diagonal terms for non-orthogonal meshes. Now includes ASCII art to explain what each term is doing.


ZedThree (approved):

ZedThree:
Thanks @bendudson !

PR #2206, opened 2021-01-10
LaplaceHypre3d

johnomotani:
Here is an attempt at a 3d Laplace solver using the Hypre interface. @YiningQin @ZedThree @bendudson @acfisher @jonesholger @xxu

It does not pass the integrated test. The first thing I'd like to check is whether this is just due to the solver tolerances - we had to use pretty tight tolerances (rtol=1e-10, atol=1e-13) for the PETSc 3d Laplace solver to pass. I could not find a method to set tolerances in the Hypre interface - @acfisher did I miss something?

I had to change the definition of the `HypreMalloc` to compile with the CPU version, and also change the calls from, e.g. `HypreMalloc(&I, vsize*sizeof(HYPRE_BigInt))` to `HypreMalloc(I, vsize*sizeof(HYPRE_BigInt))` - not sure if the `&` had some significance for the GPU version.

There seems to be a bug in `configure.ac`. I couldn't configure using `--with-hypre` and `--with-petsc` at the same time. When I did, the check for `libHYPRE` failed; it seemed to be because it was trying to link with `-L $(PETSC_LIB)` instead of the path to the Hypre library, but I don't understand why when the check worked without PETSc...

### DIFF
--- a/include/bout/hypre_interface.hxx
+++ b/include/bout/hypre_interface.hxx
@@ -835,6 +835,15 @@ public:
 
     solve_err = checkHypreError(HYPRE_ParCSRGMRESSolve(solver, A->getParallel(), b->getParallel(), x->getParallel()));
 
+    if (solve_err) {
+      // Aaron Fisher warns that it is possible that the error code is non-zero even
+      // though the solve was successful. If this occurs, we might want to make this a
+      // warning, or otherwise investigate further.
+      char* error_cstr;
+      HYPRE_DescribeError(solve_err, error_cstr);
+      throw BoutException("Hypre solve failed with error code {:d}: {:s}",
+                          solve_err, error_cstr);
+    }
     return solve_err;
   }
 
--- a/include/bout/hypre_interface.hxx
+++ b/include/bout/hypre_interface.hxx
@@ -707,13 +707,12 @@ private:
   MPI_Comm comm;
   HYPRE_Solver solver;
   HYPRE_Solver precon;
-  //bool pcg_setup;
   bool gmres_setup;
 public:
   HypreSystem(Mesh& mesh)
   {
     HYPRE_Init();
-# if 1
+#if 0
     hypre_HandleDefaultExecPolicy(hypre_handle()) = HYPRE_EXEC_DEVICE;
 #endif
 
@@ -741,7 +740,7 @@ public:
     HYPRE_BoomerAMGSetMaxLevels(precon, 20);
     HYPRE_BoomerAMGSetKeepTranspose(precon, 1);
     HYPRE_BoomerAMGSetTol(precon, 0.0);
-
+    HYPRE_BoomerAMGSetPrintLevel(solver, 3);
 
     gmres_setup = false;
   }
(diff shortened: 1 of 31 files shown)


YiningQin:
Thanks John, that's really helpful. Let's discuss it at Monday's meeting.

Best,

Yining

ZedThree:
There's a unit test on the PETSc 3D AMG that's probably worth copying too

johnomotani:
> There's a unit test on the PETSc 3D AMG that's probably worth copying too

Good point, I've (partially) copied now. Many of the PETSc unit tests use a matrix-vector product, which I couldn't see implemented for Hypre. It would be useful to have though, because the solve-test is failing, saying the calculated solution is zero everywhere. I don't have any idea why, and don't have time to debug this solver any more...

johnomotani:
BTW I had to hack a couple of things on this branch for a non-RAJA build:
on l.24 of `include/bout/hypre_interface.hxx`
```
-#if 1   //HYPRE with Cuda enabled : todo use appropriate flag
+#if 0   //HYPRE with Cuda enabled : todo use appropriate flag
```
on l.16 of `single_index_ops.hxx`
```
-using EXEC_POL = RAJA::loop_exec;
+//using EXEC_POL = RAJA::loop_exec;
```
and somewhere in the unit tests (with these hacks, don't know if it's related) there's a segfault. I worked around that by using `--gtest_filter` to only run the `LaplaceHypre3d` tests.

include/bout/hypre_interface.hxx
```diff
@@ -130,8 +130,8 @@ public:
 
     location = f.getLocation();
     initialised = true;
-    HypreMalloc(&I, vsize*sizeof(HYPRE_BigInt)); 
-    HypreMalloc(&V, vsize*sizeof(HYPRE_Complex));
+    HypreMalloc(I, vsize*sizeof(HYPRE_BigInt));
+    HypreMalloc(V, vsize*sizeof(HYPRE_Complex));
```
jonesholger, line 134: @johnomotani In cudaMallocManaged mode of the define the call is expecting a void** (i.e. Handle), so some finesse still needed:  

__host__ ​cudaError_t cudaMallocManaged ( void** devPtr, size_t size, unsigned int  flags = cudaMemAttachGlobal )

johnomotani: Ah, OK. Thanks @jonesholger. Trouble with me not being able to test the GPU version...  Without these changes the CPU version refused to compile. Please feel free to push fixes to this branch!

johnomotani:
This should possibly be working and passing tests now (but see #2225, which also applies to the tests here).

@ZedThree - this is so similar to LaplacePetsc3dAmg that it would be really nice to remove the code duplication (otherwise maintenance will be a headache!). One way to approach that might be to add a `PetscSystem` similar to Aaron's `HypreSystem` so we can move the PETSc calls `KSPSolve()`, etc.) under there. Then I hope it would be relatively simple to template a 3d Laplace solver on the PETSc and Hypre interfaces, or make the interfaces implementations of an abstract `LinearSolver` so `Laplace3d` doesn't need to know which one it's using..
It would be nice to merge the integrated test too, so there's only one `test-laplace3d` to maintain, which runs both implementations (if they're available).
I'm less sure about merging the unit tests (`test_laplace_hypre3d.cxx` and `test_laplace_petsc3d.cxx`). At the moment they are different because the interface to matrix-vector multiplication is different, but probably that could be massaged away. I don't know how gtest would play with templated tests? Maybe it's easier if the interfaces are implementations of an abstract `LinearSolver`, and the tests can just be parameterised on the type?

johnomotani:
Hacks still needed from the `next-hypre-outerloop-cuda-debug` branch, which could do with a better fix somewhere (but aren't part of the LaplaceHypre3d solver):
* Commenting out `HypreFinalize()` in the `HypreSystem` class to stop unit tests crashing. Probably the correct solution is to create a HypreLib reference counter like the PetscLib one.
* Comment out `using EXEC_POL = RAJA::loop_exec;` in `single_index_ops.hxx` to compile without RAJA

johnomotani:
Closing since @jonesholger merged the changes manually.

PR #2208, opened 2021-01-13
Remove HDF5 support

johnomotani:
HDF5 support is not planned for v5. Remove now to get rid of the effort required to maintain support?


ZedThree:
I'm on-board with doing this now. Do we also want to remove legacy/parallel netCDF at the same? I say that realising we just merged a PR fixing the legacy netCDF.

We could also remove `dump_format` from the code and the .inp files

bendudson:
Yes I think this is a good idea: I don't think hdf5 has really been used much. Legacy NetCDF can probably be kept for now, but parallel NetCDF I don't think works, and certainly isn't used.

johnomotani:
I'd support keeping legacy NetCDF around for the moment. I've been using it to avoid linking issues recently...

ZedThree (approved):

ZedThree:
I'm going to merge this manually in order to add a line to the breaking changes in CHANGELOG without triggering needless CI runs

PR #2214, opened 2021-01-18
Replace zoidberg code with submodule

johnomotani:
zoidberg now exists in the repo https://github.com/boutproject/zoidberg . Import this as a submodule instead of having zoidberg code in BOUT-dev.

I assume the code in the repo is the same as or more up to date than the current zoidberg code in next - it's hard to check quickly because the repo seems to have had `black` run on it, so there were lots of whitespace/newline changes.


johnomotani:
PS not sure if the symlink in `tools/pylib` is necessary for zoidberg. It was useful for boutdata/boututils to avoid changing the config and people's existing environments. I've included it here in case it is, but if you'd rather import directly from `externalpackages/zoidberg/zoidberg` then we could remove it?

ZedThree:
Ah, I think it needs to wait till 3D metrics goes in -- there's a PR over there with at least some of the changes from that branch.

johnomotani:
Whenever it's time to go, just need to go into `externalpackages/zoidberg/`, do a `git pull` to update the master branch and commit the change.

dschwoerer:
https://pypi.org/project/zoidberg/ is available (again) :tada:

ZedThree:
There's some conflicts in https://github.com/boutproject/zoidberg/pull/2 that need fixing before this can be updated.

I also struggled to merge `next` into this branch manually, hopefully Github will have better luck

dschwoerer:
Replaced by #2505

PR #2230, opened 2021-02-14
Save provenance tracking info from grid file

johnomotani:
Save (if present) the `grid_id`, `hypnotoad_version`, `hypnotoad_git_hash`, `hypnotoad_git_diff` and `hypnotoad_geqdsk_filename`. I've checked that this works on the v4.4 backport using a gridfile generated with a recent version of hypnotoad.

Also update the docs to note useful attributes/variables for reproducibility and provenance tracking.

### DIFF
--- a/src/mesh/impls/bout/boutmesh.hxx
+++ b/src/mesh/impls/bout/boutmesh.hxx
@@ -263,6 +263,13 @@ private:
 
   int MXG, MYG, MZG; // Boundary sizes
 
+  // Grid file provenance tracking info
+  std::string grid_id = "";
+  std::string hypnotoad_version = "";
+  std::string hypnotoad_git_hash = "";
+  std::string hypnotoad_git_diff = "";
+  std::string hypnotoad_geqdsk_filename = "";
+
   void default_connections();
   void set_connection(int ypos1, int ypos2, int xge, int xlt, bool ts = false);
   void add_target(int ypos, int xge, int xlt);
--- a/manual/sphinx/user_docs/output_and_post.rst
+++ b/manual/sphinx/user_docs/output_and_post.rst
@@ -621,3 +621,61 @@ NetCDF files.
 
     u = bread(f, "U")  # Finally read the variable
 
+.. _sec-reproducibility:
+
+Reproducibility and provenance tracking
+=======================================
+
+To help with reproducibility of simulations and provenance tracking of
+data, BOUT++ saves some metadata into output files.
+
+.. table:: Provenance tracking metadata attributes
+
+   +---------------------------------------------------------------------------+
+   | File attributes                                                           |
+   +=============================+=============================================+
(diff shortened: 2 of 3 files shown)


johnomotani:
Would be nice to save the filename of the grid file too, but that looks like a faff at the moment. Hopefully it might be easier after the I/O refactor...

Also would be nice to save the `hypnotoad_inputs` and `hypnotoad_input_geqdsk_file_contents` but they're saved as string variables so not currently accessible through `Mesh::get()`. The reason for not saving them as attributes is that they are very long strings, so they clutter up the output of `ncdump -h gridfile.nc`. Any opinions on whether it's better to just make them attributes anyway, or to wait or the I/O refactor, when hopefully it will be possible to read and save them more easily?

ZedThree:
Probably wait for the I/O refactor? Hopefully won't be too long. Although I guess that doesn't help for the backport.

It also looks like it should be possible to make `Mesh::get` read string variables and fallback to getting from an attribute if not? Something like this maybe?

```cpp
bool GridFile::get(Mesh *UNUSED(m), std::string &sval, const std::string &name, const std::string& def) {
  Timer timer("io");
  TRACE("GridFile::get(std::string)");
  
  if (!file->is_valid()) {
    throw BoutException("File cannot be read");
  }
  
  // Try to read from a variable first, maximum of 2048 characters
  constexpr int max_string_size = 2048;
  std::string temp_var;
  temp_var.reserve(max_string_size);
  const bool success_var = file->read(&(*temp_var.begin()), name, max_string_size);
  if (success_var) {
    sval = std::move(temp_var);
    return success_var;
  }

  const bool success = file->getAttribute("", name, sval);
  if (!success) {
    sval = def;
  }

  print_read_option(sval, name, filename, success);

  return success;
}
```

Not quite I suppose because `max_string_size` should be the exact size of the string to read. There's `DataFormat::getSize`, but I'm not sure that does exactly what we want here.

src/mesh/impls/bout/boutmesh.hxx
```diff
@@ -263,6 +263,13 @@ private:
 
   int MXG, MYG, MZG; // Boundary sizes
 
+  // Grid file provenance tracking info
+  std::string grid_id = "";
+  std::string hypnotoad_version = "";
+  std::string hypnotoad_git_hash = "";
+  std::string hypnotoad_git_diff = "";
+  std::string hypnotoad_geqdsk_filename = "";
```
ZedThree, line 271: Just to note that when we get the new I/O stuff in, we won't need to have these as class members

ZedThree (approved):

johnomotani:
> Probably wait for the I/O refactor? 

I think that's best. What's here is good enough for now, and I don't think it's worth putting more work into modifying `Datafile` `GridFile`, etc.

PR #2234, opened 2021-02-16
ny_inner label misleading in topology schematic

johnomotani:
Qian Xia pointed out that the `n_inner` labels on the topology schematic are misleading. The label at upper outer divertor definitely should not be `n_inner+1`. This PR has a 'corrected' version of the figure, and also adds the .svg file (updated version of the one I got from @lukeeasy) in case someone wants to edit it in future (can be edited with inkscape).

I've changed `n_inner` to `ny_inner` because I think it's clearer.

I think the most consistent label for the outer boundaries is the number of the last grid point, i.e. `nx-1`, `ny_inner-1`, `ny-1` instead of `nx`, `ny_inner`, `ny`, so I've updated that too.


johnomotani:
I've cancelled the CI (apart from readthedocs build) since this PR is only documentation (and I forgot to add a `[skip ci]` tag to the commit message).

ZedThree:
I think the filename should be `topology_cross_section.*` -- `schematic` is the following figure (see [current next](https://bout-dev.readthedocs.io/en/latest/user_docs/input_grids.html#id3) and [this PR](https://bout-dev--2234.org.readthedocs.build/en/2234/user_docs/input_grids.html#id3))

Is it possible to crop the SVG a bit too? Otherwise, looks good!

johnomotani:
Thanks @ZedThree, I've made those changes. I've force-pushed the branch to avoid possibly adding more binary files to the history.

ZedThree (approved):

PR #2239, opened 2021-02-21
Add Backward Euler solver (rebased)

bendudson:
Replaces #2238, based on an updated `next` branch to avoid conflicts.


ZedThree:
Fiddling about this to try and get it to pass `test-solver`, it seems the biggest effect comes from setting the output timestep. Adding the following to `test_solver.cxx` is "sufficient" to get it to pass, but then it does take ~35 seconds:

```cpp
  constexpr int beuler_nout = 1e5;
  root["beuler"]["nout"] = beuler_nout;
  root["beuler"]["output_step"] = end / beuler_nout;
```

Making the initial timestep very small, or tightening the tolerances makes very little difference.

tests/integrated/test-beuler/CMakeLists.txt
```diff
@@ -0,0 +1 @@
+bout_add_integrated_test(test_beuler SOURCES test_beuler.cxx)
```
ZedThree, line 1:
```suggestion
bout_add_integrated_test(test_beuler SOURCES test_beuler.cxx
  REQUIRES BOUT_HAS_PETSC)
```

johnomotani: Does this need fixing in the v4.4.0-alpha port?

ZedThree (approved):

PR #2237, opened 2021-02-16
Fix LaplacePetsc3dAmg

johnomotani:
The bug in `test-laplace-petsc3d that @ZedThree found (#2225) was hiding a couple of other bugs in `ZInterpolation` and `ZHermiteSpline` that made the matrix for 3d Laplace_perp inversion incorrect with non-trivial parallel derivatives. These bugs should be fixed by this PR.

I've also added some extra checks (when `CHECK > 2`) for finiteness of matrix elements and interpolation weights, and bounds of indices passed to `getWeightsForYApproximation()`, which helped to find the bugs.

Fixes #2225.

PR #2242, opened 2021-02-23
Laplace3d bugfix backport

johnomotani:
Backport of #2237, and extends the bugfix to `LaplaceHypre3d`.

Also includes a fix for `operator+=()` in the Hypre interface, which was previously over-writing values instead of adding to them.


ZedThree (approved):

jonesholger:
elmpb is still unstable while running against laplacexy-hypre. While I do have a few stability fixes in the works, we absolutely need to work on initialization and assembly checks. But, this PR looks good with the couple of inline edits I added earlier.

PR #2243, opened 2021-02-23
Remove <optional> include from uuid

bendudson:
Causes trouble on older compilers without C++17 support


johnomotani (approved):

ZedThree:
clang-tidy-review failing because it doesn't handle PRs with only deleted lines in very well -- fix incoming

PR #2244, opened 2021-02-24
Bump clang-tidy-review

ZedThree:
Fixes failures when the PR only removes lines, or if a comment comes from a file not in the PR (e.g. a header)

Also don't check for a couple of clang-tidy warnings that basically only come from PETSc (array-decay-to-pointer, this is essentially only from macros that make `const char[]` constants that we pass to `const char*` -- this is fine; and calling C-style variadic argument functions, these are now all third-party library functions, so out of our control)


ZedThree:
Cancelled CI as this is CI infrastructure changes, so won't affect tests!

dschwoerer (approved):

PR #2245, opened 2021-03-01
Add unit tests for BoutMesh::load

ZedThree:
Adds various unit tests for `BoutMesh::load` by splitting it into smaller helper methods and testing those, plus unit tests for some other `BoutMesh` methods. This enables parameterised tests, for instance, checking a double-null configuration can be split across various processor combinations. Each of these tests takes less than 3ms, and it's very easy to add a new combination to test, so it would be interesting to get some more variations in.

There's still some more bits left to pull out into helper functions and test, so this is WIP till then.

It should also be possible to test the communication routines to some extent as well, as we've got `MpiWrapper` that allows us to use GoogleMock to test routines that use MPI.


ZedThree:
Think I'm done with this now. Adds 137 unit tests for most of the non-communication `BoutMesh` methods for various possible tokamak configurations. Coverage of `boutmesh.cxx` should increase from about ~46% to ~62%.

Also expand a bit on the topology documentation. @bendudson please could you double check this?

This is supposedly 8k+ lines, but 5.5k lines are from converting the topology schematic to SVG, so it's nice and sharp in the docs.

ZedThree:
Now also fixes a whole host of clang-tidy warnings

ZedThree:
clang-tidy is worried about a few MPI calls, mostly when `async_send` is true. I've not managed to follow the logic through myself yet, but could be something we need to look at.

ZedThree:
Not really related to this PR, but I think we can now squash `BoutMesh::load` into the constructor and remove `Mesh::load` entirely. It looks like we need to remove the call to `Mesh::getCoordinates` at the end of `load` to initialise the `Coordinates`, and rely on fields creating them as needed. All the tests pass, but maybe there's still some cases that this would break.

johnomotani:
> maybe there's still some cases that this would break.

I can't think of any immediately. Anything that uses `Coordinates` members will need to get a pointer to some `Coordinates` using `Mesh::getCoordinates()` (either directly or through `Field*::getCoordinates()`), so I think `Coordinates` should always be created when they are needed.

ZedThree:
I guess I was thinking of if we create a staggered field before anything else -- but that issue is sorted now, right?

I'll make another PR for that change, this one is already way, way bigger than I was expecting/hoping!

johnomotani:
> I guess I was thinking of if we create a staggered field before anything else -- but that issue is sorted now, right?

Good point, I hadn't thought about that! But I think it's OK, when we create staggered `Coordinates`, they'll just get/create the `CELL_CENTRE` ones first - I think we fixed the problem with initialising things in the `coords_map`.
https://github.com/boutproject/BOUT-dev/blob/e7920b760ee896adf3c476b3238b073212133bc2/src/mesh/mesh.cxx#L433-L444

tests/unit/mesh/test_boutmesh.cxx
```diff
@@ -35,3 +270,1889 @@ TEST(BoutMeshTest, SingleCoreDecomposition) {
...
+
+  EXPECT_EQ(mesh.getRegion("RGN_UPPER_INNER_Y").size(), rgn_upper_y[0]);
+  EXPECT_EQ(mesh.getRegion("RGN_UPPER_OUTER_Y").size(), rgn_upper_y[1]);
+  EXPECT_EQ(mesh.getRegion("RGN_UPPER_Y").size(), rgn_upper_y[2]);
+
+  EXPECT_EQ(mesh.getRegion("RGN_INNER_X").size(), rgn_x[0]);
+  EXPECT_EQ(mesh.getRegion("RGN_OUTER_X").size(), rgn_x[1]);
+}
+
+// These next few tests chec both default_connections and the Region
```
johnomotani, line 1666:
```suggestion
// These next few tests check both default_connections and the Region
```

tests/unit/mesh/test_mesh.cxx
```diff
@@ -55,6 +55,27 @@ TEST_F(MeshTest, GetRegionPerpFromMesh) {
...
+}
+
+TEST_F(MeshTest, HasRegion2D) {
+  localmesh.createDefaultRegions();
+  EXPECT_TRUE(localmesh.hasRegion2D("RGN_ALL"));
+  EXPECT_TRUE(localmesh.hasRegion2D("RGN_NOBNDRY"));
+  EXPECT_FALSE(localmesh.hasRegion2D("SOME_MADE_UP_REGION_NAME"));
+}
+
+TEST_F(MeshTest, HasRegionPMesh) {
```
johnomotani, line 72: typo?
```suggestion
TEST_F(MeshTest, HasRegionPerp) {
```

johnomotani (approved):
Looks good to me, although I only skimmed over the unit test files very quickly.

Having test coverage of BoutMesh is awesome, thanks @ZedThree! :tada:

src/mesh/impls/bout/boutmesh.cxx
```diff
@@ -73,113 +77,42 @@ BoutMesh::~BoutMesh() {
...
-  GlobalNyNoBoundaries = ny;
-  GlobalNzNoBoundaries = nz;
+void BoutMesh::setYDecompositionIndices(int jyseps1_1_, int jyseps2_1_, int jyseps1_2_,
+                                        int jyseps2_2_, int ny_inner_) {
+  // Set member variables
+  this->jyseps1_1 = jyseps1_1_;
+  this->jyseps2_1 = jyseps2_1_;
+  this->jyseps1_2 = jyseps1_2_;
+  this->jyseps2_2 = jyseps2_2_;
+  this->ny_inner = ny_inner_;
```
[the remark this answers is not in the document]
ZedThree: I think I changed my mind about how best to write this and didn't clean it up properly. `this->jyseps*` was to make it a bit clearer that we're setting member variables from the function arguments, but I think the trailing underscore is probably sufficient. I'll drop `this->`

Also, to explain these functions a little bit, `setXDecompositionIndices` is really only useful for the tests, so it takes a helper struct, while `setYDecompositionIndices` is used in "real" code too. The latter has an overload with a helper struct too, which is useful for the tests. 

These helper structs are mostly to separate tons of literal doubles in the tests, as they were getting really quite unreadable

ZedThree:
Thanks @johnomotani , sorted your comments

PR #2247, opened 2021-03-03
Fixes and test for LaplaceXY2Hypre; runtime options for Hypre interface

johnomotani:
* Integrated test for `LaplaceXY2Hypre`
* Fixes for bugs found with integrated test
* Add an `Options &options` argument to the `HypreSystem` constructor, use it to allow options to be set at run-time
* Simplify options handling in `LaplaceHypre3d`, removing things now handled by `HypreSystem`
* Setting to control command-line output from `LaplaceXY2Hypre` (off by default)

### DIFF
--- a/include/bout/hypre_interface.hxx
+++ b/include/bout/hypre_interface.hxx
@@ -821,10 +821,16 @@ public:
 
     comm = std::is_same<T, FieldPerp>::value ? mesh.getXcomm() : BoutComm::get();
 
+    auto print_level = options["hypre_print_level"]
+                         .doc("Verbosity for Hypre solver. Integer from 0 (silent) to 4 "
+                              "(most verbose).")
+                         .withDefault(0);
+
     if (solver_type == HYPRE_SOLVER_TYPE::gmres) {
       HYPRE_ParCSRGMRESCreate(comm, &solver);
       /* set the GMRES parameters */
       //HYPRE_GMRESSetKDim(solver, 5); // commented out to let Hypre pick default
+      HYPRE_ParCSRGMRESSetPrintLevel(solver, print_level);
     } else {
       throw BoutException("Unsupported hypre_solver_type {}", solver_type);
     }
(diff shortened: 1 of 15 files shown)


johnomotani:
By the way, I think `LaplaceXY2` is broken (_not_ `LaplaceXY`, which should be fine).

I don't think the intention was ever for `LaplaceXY2` to be a long-term option, so not sure what the best thing to do is. I guess either copy the test in this PR for `LaplaceXY`, and fix any bugs (fixes can probably be copied from `LaplaceXY2Hypre`), or remove it. If we remove it, I'd be nice to convert `LaplaceXY` to use the PETSc interface, but that's probably a job for later (maybe after merging the PETSc and Hypre interfaces into a common structure).

include/bout/hypre_interface.hxx
```diff
@@ -810,11 +821,26 @@ public:
...
+                         .withDefault(0);
+
+    if (solver_type == HYPRE_SOLVER_TYPE::gmres) {
+      HYPRE_ParCSRGMRESCreate(comm, &solver);
+      /* set the GMRES parameters */
+      //HYPRE_GMRESSetKDim(solver, 5); // commented out to let Hypre pick default
+      HYPRE_ParCSRGMRESSetPrintLevel(solver, print_level);
+    } else {
+      throw BoutException("Unsupported hypre_solver_type {}", solver_type);
+    }
```
ZedThree, line 836: I usually prefer flipping the conditionals for preconditions to exit early, which leaves the rest of the code less-indented:

```suggestion
    if (solver_type != HYPRE_SOLVER_TYPE::gmres) {
      throw BoutException("Unsupported hypre_solver_type {}", solver_type);
    }
    HYPRE_ParCSRGMRESCreate(comm, &solver);
    /* set the GMRES parameters */
    //HYPRE_GMRESSetKDim(solver, 5); // commented out to let Hypre pick default
    HYPRE_ParCSRGMRESSetPrintLevel(solver, print_level);
```

I think you could also get away with removing the checks in the member functions, as they would already need an instance that was created with an acceptable `solver_type`

ZedThree:
I really like the `BOUT_ENUM_CLASS` and its ability to set an `enum` from `Options` -- such a nice feature! Thanks @johnomotani!

include/bout/hypre_interface.hxx
johnomotani: The intention is to have some other solver types too (CG and BiCGStab maybe?) that need different HYPRE functions calling, so this should become an `if...else if...else if...else`. Here I was just setting up the option so Holger/Aaron can add the extra cases.

I tend to put the `else {throw ...}` everywhere just in case one of the places is forgotten when implementing a new `solver_type`. It's true they don't catch user errors, so don't matter as long as the developers are careful, so I'm happy to remove the checks in member functions if you prefer @ZedThree.

ZedThree: ah yes, actually looking at the Hypre function names it's more obvious now! In that case, maybe consider a `switch` instead? Missing off the `default` case and not throwing is a possibility -- the compiler should warn about any missing enum values, and `Options` should have thrown already if it's not an acceptable value.

This isn't a blocker, just a suggestion :)

PR #2248, opened 2021-03-03
Use zero for initial guess in test-laplace-petsc3d

johnomotani:
Previously was using the expected solution for the initial guess, which is probably not the best test.


ZedThree:
Good catch! Is there a reason to keep the code setting the boundaries on `guess`? Rather than just have `solve(rhs, 0.0)`?

bendudson:
I think it needs to be given the correct boundary conditions, or it'll never converge to the correct solution

ZedThree (approved):

PR #2256, opened 2021-03-09
Make shifted paralleltransform message more helpful

bendudson:
If using `type = shifted`, and `getWeightsForYApproximation` is used, suggest `type = shiftedinterp` in the error message.

### DIFF
--- a/include/bout/paralleltransform.hxx
+++ b/include/bout/paralleltransform.hxx
@@ -224,7 +224,7 @@ public:
   const FieldPerp fromFieldAligned(const FieldPerp& f,
                                    const std::string& region = "RGN_ALL") override;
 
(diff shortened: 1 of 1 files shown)


johnomotani (approved):
Thanks @bendudson!

include/bout/paralleltransform.hxx
```diff
@@ -224,6 +224,13 @@ public:
   const FieldPerp fromFieldAligned(const FieldPerp& f,
                                    const std::string& region = "RGN_ALL") override;
 
+  virtual std::vector<PositionsAndWeights>
```
ZedThree, line 227:
```suggestion
  std::vector<PositionsAndWeights>
```

ZedThree (approved):

PR #2262, opened 2021-03-11
Fix use of uninitialised value in Delp2

johnomotani:
Restricting the y-loop in `Delp2()` to only grid cells (`ystart`->`yend`) avoids use of potentially uninitialised values in corner cells (I found this chasing a warning from valgrind), and is also slightly more optimal. We generally do not provide output from derivative operators in guard cells anyway, so I don't think this change should break anything.

### DIFF
--- a/src/mesh/coordinates.cxx
+++ b/src/mesh/coordinates.cxx
@@ -1391,8 +1391,8 @@ Field3D Coordinates::Delp2(const Field3D& f, CELL_LOC outloc, bool useFFT) {
     auto ft = Matrix<dcomplex>(localmesh->LocalNx, ncz / 2 + 1);
     auto delft = Matrix<dcomplex>(localmesh->LocalNx, ncz / 2 + 1);
 
-    // Loop over all y indices
-    for (int jy = 0; jy < localmesh->LocalNy; jy++) {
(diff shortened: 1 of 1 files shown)


johnomotani:
I have a vague memory that this came up before at some point, and there was some reason why we didn't make this change, but I can't find it on a quick search now, or remember what the reason might have been...

ZedThree:
Not ringing any bells for me. I think we do generally compute derivatives in the guards in the "other" direction, but as this is a 2D operator, then it shouldn't do any guards. I note we do still take the forward FFTs over the whole X domain, although that skips the corners.

ZedThree (approved):

ZedThree:
I'll just hold off on merging these till next week in case anybody else wants to chime in.

bendudson:
I suspect this came from not wanting to have to impose boundary conditions on both a quantity and Delp2 of it, but I can't remember what the quantity was. Since the corner cells are not set, the calculation is wrong, and so this is really a bug fix. 
Perhaps a comment would be useful, to explain why the calculation can't be done in the y guard cells? Otherwise this may repeat...

ZedThree (approved):

PR #2265, opened 2021-03-19
Backport of Backward Euler solver to v4.4

bendudson:
Useful for finding steady-state solutions (e.g. transport problems in 1D, 2D); terrible for time-evolving.

### DIFF
--- a/src/solver/impls/snes/snes.cxx
+++ b/src/solver/impls/snes/snes.cxx
@@ -21,153 +21,262 @@
  * This function assumes the context void pointer is a pointer
  * to an SNESSolver object.
  */
-static PetscErrorCode FormFunction(SNES UNUSED(snes), Vec x, Vec f, void *ctx) {
+static PetscErrorCode FormFunction(SNES UNUSED(snes), Vec x, Vec f, void* ctx) {
   return static_cast<SNESSolver*>(ctx)->snes_function(x, f);
 }
 
 int SNESSolver::init(int nout, BoutReal tstep) {
 
   TRACE("Initialising SNES solver");
-  
+
   /// Call the generic initialisation first
-  if (Solver::init(nout, tstep))
+  if (Solver::init(nout, tstep) != 0) {
     return 1;
-  
+  }
+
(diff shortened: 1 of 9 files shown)


src/solver/impls/snes/snes.cxx
```diff
@@ -21,153 +21,262 @@
...
+        } else if (its >= upper_its) {
+          // Reduce timestep slightly
+          timestep *= 0.9;
+        }
+      }
+    } while (looping);
+
+    // Put the result into variables
+    {
+      const BoutReal* xdata = nullptr;
```
johnomotani, line 224: Why make this `const`? Does PETSc need it? Naively it looks like you just strip it off with `const_cast`.

src/solver/impls/snes/snes.cxx
```diff
@@ -21,153 +21,262 @@
...
 
 // f = rhs
 PetscErrorCode SNESSolver::snes_function(Vec x, Vec f) {
-  const BoutReal *xdata;
-  BoutReal *fdata;
-  int ierr;
-
   // Get data from PETSc into BOUT++ fields
-  ierr = VecGetArrayRead(x,&xdata);CHKERRQ(ierr);
+  const BoutReal* xdata = nullptr;
```
johnomotani, line 246: again `const`?

johnomotani (approved):
:+1: Would be nice to have a test, but I'm happy to leave that for a future PR since I guess this is a somewhat experimental feature at the moment.

src/solver/impls/snes/snes.cxx
bendudson: yes, PETSc needs it to be const, but `load_vars` needs non-const. Probably should fix `load_vars`.

src/solver/impls/snes/snes.cxx
bendudson: just for `VecGetArrayRead`. Since we're reading not writing I think it's best to make it const, though `load_vars` needs the const casting away

johnomotani (approved):
Nice test! Thanks @bendudson.

PR #2269, opened 2021-03-23
Return non-const should be preferred

dschwoerer:
Returning non-const objects has no disadvantage for const correctness
and is easier to read.

(as suggested by GHA bot)


ZedThree (approved):

ZedThree:
Nice, thanks @dschwoerer !

PR #2272, opened 2021-04-07
Fix segfault when creating LaplaceXZ without passing Options

ZedThree:
Ensures `options` is not `nullptr`


bendudson (approved):
Thanks @ZedThree !

ZedThree:
Failing test is readthedocs -- this was an issue their end, should be fixed now.

PR #2277, opened 2021-04-09
Require `Options` to be exactly one of either a section or a value

ZedThree:
This is a breaking change, but I don't think it will affect very many people, and turns out to have basically no effect on the logic anywhere, but does automatically fix a small bug: when creating a new section, it would _not_ be marked as a section:

```cpp
auto section1 = options.getSection("new section"); // old-style API
auto section2 = options["new section"]; // new-style API
```

Previously, both `section1` and `section2` would have `is_section == false` and `is_value == false`. This PR fixes both cases: an `Options` is a section until it has a value assigned to it, when it becomes a value.

This does prevent us having an input file like the following:

```ini
restart = true
[restart]
guards = false
```

but I think that's probably a good thing. I think `restart` is the only case of this in the library, but of course users might have other instances.

Renames the `restart` section to `restart_files` -- this doesn't appear in any test or example input files, so I suspect it's likely not used in the wild either, but the input file upgrader can fix it if it does.


johnomotani:
`[restart]` is sometimes used if we need to set `restart:init_missing`, but that's not too common I guess. I'm in favour of this change - the new system will throw an error for any uses of `[restart]` anyway, so we won't surprise anyone by failing silently.

ZedThree:
Hmm, I've just double-checked that and it doesn't throw an error -- ah, you mean in the case-sensitive `Options` PR? Yes, that is true. With this PR, it looks like a `[restart]` section will silently clobber `restart = true` in the input file, but `-restart`/`restart=true` on the command line wins over the `[restart]` section.

I just quickly tested what happens when this gets merged into #2210, and it looks like it needs a bit more work to play nice with this change.

It might actually need to throw an error in this PR if something that was a value is used as a section. This PR just silently converts between values and sections, which on reflection is not a good idea!

johnomotani:
Sorry, yes I meant with #2210 I thought it would be (should be!) an error.

ZedThree:
Ok, this now catches `[restart]` and flags it as an error. It also catches the other case of an option being both a value and a section, which is in the MMS test that failed.

johnomotani (dismissed):
:+1:

ZedThree:
Had to fix the inevitable conflicts with #2210!

johnomotani (approved):

PR #2280, opened 2021-04-12
Equation numbers in manual

### LINKED ISSUE #2279: Equation numbers in manual
It would be nice to have equation numbers shown in the manual on readthedocs.io, in particular for the 'Field-aligned coordinates' section. I'm not a reStructuredText expert does anyone know if there's an easy way to do this? A few equation numbers seem to be shown, I think it might be when there's a reference to them somewhere else, but it would be nice for all of them to be numbered for when we want to refer to them in an external document, email, etc.

johnomotani:
Add labels to equations in `coordinates.rst`, so that they get numbered.

I haven't checked if there are other equations in the manual that could do with being numbered. If anyone feels like looking for and labelling them, feel free to push to this PR.

Fixes #2279.


ZedThree:
Think I've found a fix for the equation numbering placement: https://stackoverflow.com/a/52509369/2043465
Seems to work locally at least

I'm also piggy-backing on this to fix a bunch of other RST issues!

johnomotani:
Thanks @ZedThree, that's awesome! Other updates look good to me too.

ZedThree:
I'm half-tempted to try and automatically add labels to all the other equations in the manual too, but they would not be descriptive names.

It's also a pity that Sphinx doesn't understand multiple equations in one `:math:` block. I've not found a good solution to that.

ZedThree (approved):

PR #2282, opened 2021-04-14
Make conversion of Option to bool stricter

johnomotani:
Previously, `Options::as<bool>()` only tested (case-insensitively) if first character was 'n', 'f', '0', 'y', 't', or '1'. Now only allow (still case-insensitively but checking full strings) 'n', 'no', 'f', 'false', '0', 'y', 'yes', 't', 'true', or '1'. See discussion on https://github.com/boutproject/boutdata/issues/32.

I found a nice intro to parameterised tests here https://www.sandordargo.com/blog/2019/04/24/parameterized-testing-with-gtest, which I used for the unit tests.

### DIFF
--- a/src/sys/options.cxx
+++ b/src/sys/options.cxx
@@ -328,11 +328,17 @@ template <> bool Options::as<bool>(const bool& UNUSED(similar_to)) const {
   
   } else if(bout::utils::holds_alternative<std::string>(value)) {
     auto strvalue = bout::utils::get<std::string>(value);
+
+    // case-insensitve check, so convert string to lower case
+    for (auto& c : strvalue) {
+      c = std::tolower(c);
+    }
   
-    auto c = static_cast<char>(toupper((strvalue)[0]));
-    if ((c == 'Y') || (c == 'T') || (c == '1')) {
+    if ((strvalue == "y") or (strvalue == "yes") or (strvalue == "t")
+        or (strvalue == "true") or (strvalue == "1")) {
       result = true;
-    } else if ((c == 'N') || (c == 'F') || (c == '0')) {
+    } else if ((strvalue == "n") or (strvalue == "no") or (strvalue == "f")
+        or (strvalue == "false") or (strvalue == "0")) {
       result = false;
     } else {
       throw BoutException(_("\tOption '{:s}': Boolean expected. Got '{:s}'\n"), full_name,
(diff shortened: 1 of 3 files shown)


ZedThree:
Tests are failing because we check the old behaviour:

https://github.com/boutproject/BOUT-dev/blob/b5e2c325b795af172210a42aeb771a78d3bd807b/tests/unit/sys/test_options.cxx#L293-L296

Deleting those lines should be sufficient

src/sys/options.cxx
```diff
@@ -328,11 +328,17 @@ template <> bool Options::as<bool>(const bool& UNUSED(similar_to)) const {
   
   } else if(bout::utils::holds_alternative<std::string>(value)) {
     auto strvalue = bout::utils::get<std::string>(value);
+
+    // case-insensitve check, so convert string to lower case
+    for (auto& c : strvalue) {
+      c = std::tolower(c);
+    }
```
ZedThree, line 335: We've got a `lowercase` function in `utils.hxx` :)

```suggestion
    // case-insensitve check, so convert string to lower case
    const auto strvalue = lowercase(bout::utils::get<std::string>(value));
```

src/sys/options.cxx
```diff
@@ -328,11 +328,17 @@ template <> bool Options::as<bool>(const bool& UNUSED(similar_to)) const {
...
   
-    auto c = static_cast<char>(toupper((strvalue)[0]));
-    if ((c == 'Y') || (c == 'T') || (c == '1')) {
+    if ((strvalue == "y") or (strvalue == "yes") or (strvalue == "t")
```
ZedThree, line 337: Just for my own curiosity, I had a stab at rewriting this to avoid the long-ish chain of comparisons, but I think this is probably worse:

```cpp
auto one_of = [&strvalue](const std::vector<std::string>& values) -> bool {
  return std::any_of(
      values.begin(), values.end(),
      [&strvalue](const auto& value) { return value == strvalue; });
};
if (one_of({"y", "yes", "t", "true", "1"})) {
  result = true;
} else if (one_of({"n", "no", "f", "false", "0"})) {
  result = false;
} else {
```

Could be worth if it `one_of` were a function we could reuse somewhere else.

I think it just betrays my love of lambdas and C++'s love of verbose grammar. Here it is in Python:

```python
if strvalue.lower() in ["y", "yes", "t", "true", "1"]:
  result = True
elif if strvalue.lower() in ["n", "no", "f", "false", "0"]:
   result = False
else:
   raise ValueError()
```

ZedThree:
Looks good, thanks @johnomotani !

ZedThree (approved):

PR #2283, opened 2021-04-15
Fix some Solvers not always using user preconditioner/Jacobian

ZedThree:
CVODE, ARKODE and PETSc `Solver`s only use the preconditioner or Jacobian if they were set using `Solver::setPrecon/setJacobian`, and not if they were set through the `PhysicsModel` versions.

This required adding `Solver::hasUserJacobian/runJacobian` in order to access the model or solver Jacobian seamlessly.

Discovered because I've started removing the old `Solver` API and realised the PETSc solver stored its own pointers to the preconditioner/Jacobian functions! I'm not sure if anyone actually uses these functions, or this combination of solver and model functions, so possibly this doesn't affect anyone.

Should the name be `Solver::hasJacobian` instead of `hasUserJacobian`?

### DIFF
--- a/include/bout/solver.hxx
+++ b/include/bout/solver.hxx
@@ -193,7 +193,7 @@ public:
   /// Specify a preconditioner (optional)
   void setPrecon(PhysicsPrecon f) { prefunc = f; }
   /// Specify a Jacobian (optional)
-  virtual void setJacobian(Jacobian UNUSED(j)) {}
+  virtual void setJacobian(Jacobian jacobian) { user_jacobian = jacobian; }
   /// Split operator solves
   virtual void setSplitOperator(rhsfunc fC, rhsfunc fD);
 
@@ -422,6 +422,11 @@ protected:
   bool have_user_precon();
   int run_precon(BoutReal t, BoutReal gamma, BoutReal delta);
 
+  /// Do we have a user Jacobian?
+  bool hasUserJacobian();
+  /// Run the user Jacobian
+  int runJacobian(BoutReal time);
+
   // Loading data from BOUT++ to/from solver
   void load_vars(BoutReal* udata);
   void load_derivs(BoutReal* udata);
@@ -456,6 +461,8 @@ private:
   rhsfunc phys_run{nullptr};
   /// The user's preconditioner function
   PhysicsPrecon prefunc{nullptr};
+  /// The user's Jacobian function
+  Jacobian user_jacobian{nullptr};
   /// Is the physics model using separate convective (explicit) and
   /// diffusive (implicit) RHS functions?
   bool split_operator{false};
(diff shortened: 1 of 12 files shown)


johnomotani:
Just a comment: at least for CVODE the preconditioner-setting was working (I've used it in 4.4 and it seemed to do something), and it looks like you didn't need to change the preconditioner stuff here (apart from deleting the overrides in PetscSolver, which I've never used, and hiding away the use of function pointers).

I'd vote for `hasJacobian()` rather than `hasUserJacobian()` and rename `has_user_precon()` to `hasPrecon()` while you're here. I can't imagine we'll ever be able to automatically build preconditioners (even something like algbraic multigrid with Hypre you'd have to build it a matrix somehow...) so there's no non-user preconditioner or Jacobian to distinguish.

ZedThree:
The issue is that there is a `PhysicsModel::setPrecon` method that takes a physics model member function which is stored in the model, while the `Solver`s took a pointer to a free function that's stored in the `Solver`. The `Solver`s fixed here had overrides for `Solver::setPrecon` and stored the function pointer themselves, and only checked/used that.

So this was broken:

```cpp
class MyModel : public PhysicsModel {
public:
  int my_preconditioner(...) {...}

  int init(bool) {
    ...
    setPrecon((preconfunc) &MyModel::my_preconditioner); // Set in PhysicsModel but not called in CVODE etc
  }
};
```

but this worked:

```cpp
int my_preconditioner(...) {...}
class MyModel : public PhysicsModel {
public:
  int init(bool) {
    ...
    solver->setPrecon(my_preconditioner); // Set in CVODE
  }
};
```

johnomotani:
Agree fixes here are needed, and PetscSolver was broken. I think CVODE managed to work (at least precon, I haven't used Jacobian or thought about it now). CVODE would call
https://github.com/boutproject/BOUT-dev/blob/1d1307a1ce5ab1deec1eb74e7701ad2223b1d510/src/solver/solver.cxx#L1427-L1435
and get into `model->runPrecon()`
https://github.com/boutproject/BOUT-dev/blob/1d1307a1ce5ab1deec1eb74e7701ad2223b1d510/src/physics/physicsmodel.cxx#L62-L66
which does use the `userprecon` function set by `PhysicsModel::setPrecon()`
https://github.com/boutproject/BOUT-dev/blob/1d1307a1ce5ab1deec1eb74e7701ad2223b1d510/include/bout/physicsmodel.hxx#L225

Sorry, I'm going on about this to check I'm not missing something - I think for people who were using preconditioning with CVODE, it was doing what they expected, and it was only possibly setting the Jacobian or preconditioning with other solvers that was previously broken.

ZedThree:
Apologies, you're entirely correct. For CVODE, it was just the jacobian that was not called correctly. The PETSc time solver was the only that didn't call the preconditioner correctly.

bendudson (approved):

PR #2288, opened 2021-04-18
Make blob2d example self contained

dschwoerer:
Makes testing the python interface easier.

### DIFF
--- a/examples/boutcore/blob2d.py
+++ b/examples/boutcore/blob2d.py
@@ -9,28 +9,14 @@
 #
 ###################################################################
 
-# from boutcore import *
-# init("-d mini -q -q -q")
-
-# class MyModel(PhysicsModel):
-#          def init(self,restart):
-#                      self.n=create3D("dens:function")
-#                              self.solve_for(dens=self.n)
-#                                  def rhs(self,time):
-#                                              self.n.ddt(DDX(self.n))
-
-#                                              model=MyModel()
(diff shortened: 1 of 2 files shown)


examples/boutcore/blob2d.py
```diff
@@ -154,7 +145,145 @@ def rhs(self, time):
...
-# Start the simulation
-blob2d.solve()
+# Ensure the blob folder exists
+def ensure_blob():
+    if not os.path.isdir("blob"):
+        print("Setting up folder blob for simulation ...")
+        os.mkdir("blob")
+    if not os.path.exists("blob/BOUT.inp"):
+        with open("blob/BOUT.inp", "w") as f:
+            f.write(
```
ZedThree, line 155: I guess you're writing this as a string, rather than from a `BoutOptions` is to preserve the comments?

ZedThree (approved):

examples/boutcore/blob2d.py
dschwoerer: Mostly because a string is easy to store in a source file. If you know how to create the code for generating a BoutOptions, I wouldn't mind that either. But having a string is very easy to copy-paste between the two files and keep them in sync.

ZedThree: It's mostly like a dict, though I think constructing one from a dict is less than ideal at the mo.

Issue #1979, opened 2020-03-20
next builds much slower then master


dschwoerer:
~next~ master (EDIT: PH) takes 12 to 22 minutes to build e.g. [1](https://travis-ci.org/github/boutproject/BOUT-dev/builds/664434451) [2](https://travis-ci.org/github/boutproject/BOUT-dev/builds/664332537) while next takes 21 to 33 minutes [3](https://travis-ci.org/github/boutproject/BOUT-dev/builds/664144208) [4](https://travis-ci.org/github/boutproject/BOUT-dev/builds/661610524)

This coincides with a significantly increased size of `libbout++.a` from ~400 MB to ~800 MB.
This results in a full build taking roughly 14 GB of disk space, with test-compile-examples that roughly doubles again.

Git bisect blames fe3eb2be0e9c0b7eef4615d273e889ce702dbaaf which increased the size from 537821428 to 805149572 bytes.

Is this an issue?

800 MB seems huge, that is a size similar to a Linux distro live installer, and the size seems to be still increasing ...

ZedThree:
I'd noticed the build times, but not the library increase. That's pretty big! Even 400MB seems like a lot.

I suspect I messed up forwarding the parameter pack. Suspect it should look like:

```cpp
  template <class S, class... Args>
  int push(S&& format, Args&&... args) {
    return push(fmt::format(std::foward<S>(format), std::forward<Args>(args)...));
  }
```

Thanks for the bisect, that's very helpful!

ZedThree:
Also I include `fmt/format.h` everywhere, when `fmt/core.h` should be sufficient almost everywhere. I've pushed a branch to see if this helps on Travis.

But neither changing the header nor forwarding the arguments makes any measurable difference on my machine :/

ZedThree:
Doesn't look like that made it any faster :disappointed:  https://travis-ci.org/github/boutproject/BOUT-dev/builds/664940917

ZedThree:
Interesting article on compile times: https://cor3ntin.github.io/posts/compiletime/

I thought I had made an attempt at using ccache on Travis, but I can't find it now. It might be a branch on my work desktop that I currently don't have access to.

I'll also try getting gold working again, and maybe switch to ninja for the CMake build.

dschwoerer:
> Doesn't look like that made it any faster disappointed https://travis-ci.org/github/boutproject/BOUT-dev/builds/664940917

You missed on reference ;-) Now it helped a lot.

[...]

PR #2292, opened 2021-04-21
Split index_derivs into two files

ZedThree:
There was some issue with the IBM XL compiler choking on `index_derivs.cxx` -- I can't find the issue, and I don't recall all the details.

This PR splits it into two files, with the extra file having the upwind, flux and FFT derivatives. 

Some experimentation, and use of `ClangBuildAnalyzer` (see #1979), this seems to be the optimal split. As there some headers that need to be in each file, and these cost some time to include and parse, splitting up into further files actually increases compilation time.

As there hasn't been much noise about the original issue, and it doesn't seem to be of more general benefit, I'd be happy to not merge this.


ZedThree:
Oh, this has a fun issue, when building a static library, nothing in `index_derivs_flux.cxx` counts as being used so the linker chucks out the whole file.

We had a similar problem with the generic factory, so we have `DerivedFactory::ensureRegistered();` which gets implemented in the .cxx files for each factory, and called in the header. So we could do something similar here too.

ZedThree:
I'm closing this for now, it didn't lead to any major wins, there are issues with building a static library that I don't have time to solve, and no one seems to be shouting about it.

Issue #1933, opened 2020-03-01
cmake: bout++.cxx is not rebuild


dschwoerer:
With configure, whenever any file is rebuild, `bout++.cxx` is rebuild, to get the correct time stamp.

ZedThree:
I've not worked out how to do this that doesn't result in `bout++.cxx` getting marked as always out of date -- that is, every time you build, it rebuilds `bout++.cxx` even if nothing has changed.

Might be possible to use an intermediate object library with all the files except `bout++.cxx`, and add a dependency between that file and the object library. Will investigate at some point, but it's not a high priority for me right now.

ZedThree:
Fixed in #2301

PR #2294, opened 2021-04-28
Faster recompile

dschwoerer:
Especially nice if working with shared libraries, and only a single, small file needs to be recompiled.

Recompiling `bout++.cxx` can take a long time, and this issue issue is avoided this way.

Due to #1933 not useful for cmake ...

I accidentally included 73e7ad8 - but that might be useful anyway to avoid a regression?


src/bout++.cxx
```diff
@@ -454,7 +455,7 @@ void printStartupHeader(int MYPE, int NPES) {
 #ifdef MD5SUM
   output_progress.write("MD5 checksum: {:s}\n", BUILDFLAG(MD5SUM));
 #endif
-  output_progress.write(_("Code compiled on {:s} at {:s}\n\n"), __DATE__, __TIME__);
+  output_progress.write(_("Code compiled on {:s} at {:s}\n\n"), boutcompiletime, boutcompiledate);
```
ZedThree, line 458: Typo:

```suggestion
  output_progress.write(_("Code compiled on {:s} at {:s}\n\n"), boutcompiledate, boutcompiletime);
```

ZedThree (changes requested):
Nice, thanks @dschwoerer ! One small change please :)

ZedThree:
The black test is failing because it's trying to run on everything and two files are not compatible with python 3.

dschwoerer:
> The black test is failing because it's trying to run on everything and two files are not compatible with python 3.

I am confused why it is trying to do that now. Shouldn't black already be running on next?
I further thought that this has already succeded, where the files reintroduced accidentially?

ZedThree:
black is failing on this file: https://github.com/boutproject/BOUT-dev/blob/d2750d5cd3fa7b17b75ffbc9001c9163839a2967/tools/tokamak_grids/pyGridGen/aeqdsk.py

which no longer exists in `next`. I think this is just because this branch was off an old version of `next`. I suggest merging in latest `next`

dschwoerer:
It seems I forgot to fetch before merging next. Thanks for the catch, it now works :+1:

ZedThree (approved):

PR #2296, opened 2021-04-30
cyclic reduction solver documentation

bendudson:
Confusingly, the `CyclicReduce` tridiagonal solver, used in the `cyclic` potential and parallel inversion solvers, doesn't implement cyclic reduction. Instead it implements a two-level partitioning algorithm which is described in this LANL report from 2004: https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.104.9454&rep=rep1&type=pdf

Should the solver and file be renamed?

### DIFF
--- a/include/cyclic_reduction.hxx
+++ b/include/cyclic_reduction.hxx
@@ -1,6 +1,10 @@
 /************************************************************************
- * Cyclic reduction for direct solution of a complex tridiagonal system
+ * Direct solution of a complex tridiagonal system in parallel
  *
+ * Algorithm from:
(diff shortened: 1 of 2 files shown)


johnomotani:
:+1: for renaming (in next).

d7919:
Agreed with renaming in next (and adding to the documentation). Is there any value in a actual cyclic reduction implementation?

ZedThree:
:+1:  for renaming in next. We've now also got a way of "registering" disabled implementations, so we can use that to point users to the new name

PR #2297, opened 2021-04-30
Enable formatting of SpecificInd objects

bendudson:
Basic formatting of `SpecificInd` objects
```
Ind3D ind(11, 2, 3)
output.write("{}",  ind); // => "(1, 1, 2)" 
```
which can be formatted to output components (`"{}"` or `"{:c}"`) or the index (`"{:i}"`).
Includes tests, some docs, but could probably do with more documentation.


ZedThree (approved):

ZedThree:
Nice! Thanks @bendudson !

include/output_bout_types.hxx
```diff
@@ -0,0 +1,47 @@
...
+      throw format_error("invalid format");
+
+    // Return an iterator past the end of the parsed range:
+    return it;
+  }
+
+  // Formats the point p using the parsed format specification (presentation)
+  // stored in this formatter.
+  template <typename FormatContext>
+  auto format(const Ind3D& ind, FormatContext& ctx) {
```
d7919, line 34: Worth adding `Ind2D` as well?

tests/unit/sys/test_output.cxx
```diff
@@ -294,3 +295,21 @@ TEST_F(OutputTest, DummyJustPrint) {
...
+  local_output.write("{}", ind);
+
+  EXPECT_EQ(buffer.str(), "(1, 1, 2)");
+}
+
+TEST_F(OutputTest, FormatInd3Di) {
+  Ind3D ind(11, 2, 3);
+
+  Output local_output;
+  local_output.write("{:i}", ind);
```
d7919, line 312: Worth adding `:c` version as well?

How about testing an invalid format behaves as expected as well?

include/output_bout_types.hxx
ZedThree: Instead of overloading, we cpi;d take a `SpecificInd<N>` here, and then check `N == 3` to find out whether we use `ind.z()` or not.

ZedThree: This was pretty easy to do with a switch-case to work for `Ind2D` and `IndPerp` as well.

@bendudson Are you ok if I push straight to this branch? I might also move the header to `include/bout` if that's ok.

bendudson: Great, thanks @ZedThree . Yes, happy for you to push straight to this. Agree probably best to put in `include/bout`.

tests/unit/sys/test_output.cxx
ZedThree: Added tests for `{:c}` and the invalid `{:b}`

ZedThree (approved):

PR #2298, opened 2021-05-01
CMake: Build Python API

### LINKED ISSUE #1940: cmake: boutcore

### LINKED ISSUE #2289: cmake summary wrong
after running `cmake -B build`

I get:
```*** Now run `cmake --build .` to compile BOUT++ ***```

which fails for obvious reasons. It would be better to give no (wrong) advice if correct advise isn't possible.

### PR DESCRIPTION
**Pull request by dschwoerer**
Add cmake instructions for boutcore

Resolves #1940 
Resolves #2289 

I need to remove the commented out code, but maybe @ZedThree also has some further ideas to cleanup.

I am not sure about auto-enabling. The issue is that in order to build a .so, `-fPIC` is needed, which seems to be only inserted with `BUILD_SHARED_LIBS=ON`.

I also just noticed I need to move this a bit further down, so it is only enabled when we are building the interface.
https://github.com/boutproject/BOUT-dev/blob/1dd83c6e4fd5bd628c052e66198510d4bdf33839/CMakeLists.txt#L428

### CODE REVIEW

**Comment by ZedThree** [comment]
Awesome, thanks @dschwoerer !

Overall it looks good. To make it a bit cleaner, I'd probably move the numpy stuff into `cmake/FindNumpy.cmake` (similarly for Cython), and then maybe the bulk of the boutcore-specific stuff to `tools/pylib/boutcore_/CMakeLists.txt` (and then `add_subdirectory(...)`)

I'm not sure about forcing shared libs on. It might be nicer to instead have an overall default of shared libs and then a check and fatal error if they're not enabled.

File: tools/pylib/_boutcore_build/CMakeLists.txt [hunk 1]
```diff
@@ -0,0 +1,135 @@
...
+
+  # No errors? We can build the interface!
+  if ("${BOUT_PYTHON}" STREQUAL "MAYBE")
+    set(BOUT_PYTHON ON)
+  endif()
+
+
+  if (NOT "${BOUT_PYTHON}" STREQUAL "ON")
+    message(WARNING "Python interface will not be build, see warnings above")
+  else()
```
ZedThree, line 75: Could also return here, and then less indentation below?

```suggestion
    message(WARNING "Python interface will not be build, see warnings above")
    return()
  endif()
```

ZedThree (commented):
Looks good, a few pointers on how to polish the CMake a bit

ZedThree (approved):

ZedThree:
@dschwoerer I've made some changes, are you ok with them?

dschwoerer:
Yes, looks good. Sorry I didn't get around to it myself. Much appreciated :+1:

ZedThree (approved):

PR #2300, opened 2021-05-04
CMake: Build documentation

### LINKED ISSUE #1937: cmake: documentation is missing
`make doc` does not exist.

ZedThree:
Resolves #1937 

Pretty bare bones. There's a more complete method described in https://devblogs.microsoft.com/cppblog/clear-functional-c-documentation-with-sphinx-breathe-doxygen-cmake/

Currently we always run doxygen (via `subprocess.call` in `conf.py`), a nicer method would be to only run it when needed, which the above link describes how to do.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -791,7 +791,7 @@ endif()
 
 option(BOUT_BUILD_DOCS "Build the documentation" ON)
 if (BOUT_BUILD_DOCS)
-  add_subdirectory(manual)
+  add_subdirectory(manual EXCLUDE_FROM_ALL)
 endif()
 
 ##################################################
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -789,7 +789,7 @@ endif()
 ##################################################
 # Documentation
 
-option(BOUT_BUILD_DOCS "Build the documentation" ON)
+option(BOUT_BUILD_DOCS "Build the documentation" OFF)
 if (BOUT_BUILD_DOCS)
   add_subdirectory(manual EXCLUDE_FROM_ALL)
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -787,6 +787,14 @@ if(BOUT_BUILD_EXAMPLES)
 endif()
 
 ##################################################
+# Documentation
+
+option(BOUT_BUILD_DOCS "Build the documentation" ON)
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -786,6 +786,14 @@ if(BOUT_BUILD_EXAMPLES)
   add_subdirectory(examples)
 endif()
 
+##################################################
+# Documentation
+
+option(BOUT_BUILD_DOCS "Build the documentation" ON)
```
dschwoerer, line 792: Would it make sense to only default to ON if it can be build?
Most users will not need it ...

CMakeLists.txt
```diff
@@ -786,6 +786,14 @@ if(BOUT_BUILD_EXAMPLES)
...
+if (BOUT_BUILD_DOCS)
+  add_subdirectory(manual)
```
dschwoerer, line 794: should probably add `EXCLUDE_FROM_ALL`

manual/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+# BUOT++ Documentation
```
dschwoerer, line 1: typo
```suggestion
# BOUT++ Documentation
```

CMakeLists.txt
ZedThree: Fair point that most users won't need it, as the docs are available online. Might be best to default `OFF` then.

CMakeLists.txt
ZedThree: Because the docs are built with `add_custom_target`, they need explicitly adding to `all`, which I've not done. Happy to add `EXCLUDE_FROM_ALL` here too though

manual/CMakeLists.txt
ZedThree: Good catch, thanks!

CMakeLists.txt
dschwoerer: I hadn't tested it, from the documentation it sounded like without it, that would be added to `all`

dschwoerer (approved):

CMakeLists.txt
ZedThree: It does, just for "normal" targets that are created with `add_library` or `add_executable`. `add_custom_target/command` have an `ALL` option to control this separately. I'm not sure why they felt the need to have different behaviour!

PR #2301, opened 2021-05-05
CMake: Ensure compilation date/time are updated with file changes

### LINKED ISSUE #1933: cmake: bout++.cxx is not rebuild

ZedThree:
Resolves #1933

### DIFF
--- a/cmake/GenerateDateTimeFile.cmake
+++ b/cmake/GenerateDateTimeFile.cmake
@@ -1,10 +1,7 @@
 # Creates "bout++-time.cxx" in the build directory with the
 # compilation date and time as variables
 
-string(TIMESTAMP bout_date "%b %d %Y")
-string(TIMESTAMP bout_time "%H:%M:%S")
-
 set(bout_date_time_file
-  "const char* boutcompiledate{\"${bout_date}\"}; const char* boutcompiletime{\"${bout_time}\"};")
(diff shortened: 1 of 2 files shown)


cmake/GenerateDateTimeFile.cmake
```diff
@@ -0,0 +1,10 @@
+# Creates "bout++-time.cxx" in the build directory with the
+# compilation date and time as variables
+
+string(TIMESTAMP bout_date "%b %d %Y")
+string(TIMESTAMP bout_time "%H:%M:%S")
+
+set(bout_date_time_file
+  "const char* boutcompiledate{\"${bout_date}\"}; const char* boutcompiletime{\"${bout_time}\"};")
```
dschwoerer, line 8: Why not use the compiler macros `__DATE__` and `__TIME__`?
If it is purely so that the file changes, adding a comment with "generated at" should do the job?

ZedThree: Good shout, using the macros makes it simpler. This file is always re-generated if any other file changes, so the macros are sufficient

ZedThree:
readthedocs is failing because this branch doesn't have the fixes in yet

dschwoerer (approved):

PR #2303, opened 2021-05-05
CVODE constraints and max_noinlinear_iterations options

johnomotani:
Interface to two more options provided by CVODE solver:
* `solver:apply_positivity_constraints` - if set to true, read an option `positivity_constraint` in each variable subsection, which can be set to `none` (the default), `positive`, `non_negative`, `negative`, or `non_positive`. If any of these constraints is violated at the end of an internal timestep, CVODE will go back to the previous step and continue with a shorter timestep.
* `solver:max_nonlinear_iterations` - allow changing the maximum number of nonlinear iterations allowed for each timestep. CVODE default is 3. Might be useful to increase if the number of linear iterations per Newton iteration is small, but a very brief test did not observe any benefit.

### DIFF
--- a/src/solver/impls/cvode/cvode.cxx
+++ b/src/solver/impls/cvode/cvode.cxx
@@ -314,53 +314,8 @@ int CvodeSolver::init(int nout, BoutReal tstep) {
   }
 #else
   if (apply_positivity_constraints) {
-    std::vector<BoutReal> f2d_constraints;
-    f2d_constraints.reserve(f2d.size());
-    std::transform(begin(f2d), end(f2d), std::back_inserter(f2d_constraints),
-                   [](const VarStr<Field2D>& f2) {
-                     auto f2_options = Options::root()[f2.name];
-                     const auto value = f2_options["positivity_constraint"]
-                                        .doc(fmt::format(
-                                             "Constraint to apply to {} if "
-                                             "solver:apply_positivity_constraint=true. "
-                                             "Possible values are: none (default), "
-                                             "positive, non_negative, negative, or "
-                                             "non_positive.", f2.name))
-                                        .withDefault(positivity_constraint::none);
-                     switch (value) {
-                       case positivity_constraint::none: return 0.0;
-                       case positivity_constraint::positive: return 2.0;
-                       case positivity_constraint::non_negative: return 1.0;
-                       case positivity_constraint::negative: return -2.0;
-                       case positivity_constraint::non_positive: return -1.0;
-                       default: throw BoutException("Incorrect value for "
-                                                    "positivity_constraint");
-                     }
-                   });
-
-    std::vector<BoutReal> f3d_constraints;
-    f3d_constraints.reserve(f3d.size());
-    std::transform(begin(f3d), end(f3d), std::back_inserter(f3d_constraints),
-                   [](const VarStr<Field3D>& f3) {
-                     auto f3_options = Options::root()[f3.name];
-                     const auto value = f3_options["positivity_constraint"]
-                                        .doc(fmt::format(
-                                             "Constraint to apply to {} if "
-                                             "solver:apply_positivity_constraint=true. "
-                                             "Possible values are: none (default), "
-                                             "positive, non_negative, negative, or "
-                                             "non_positive.", f3.name))
-                                        .withDefault(positivity_constraint::none);
-                     switch (value) {
-                       case positivity_constraint::none: return 0.0;
-                       case positivity_constraint::positive: return 2.0;
-                       case positivity_constraint::non_negative: return 1.0;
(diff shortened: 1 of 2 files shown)


src/solver/impls/cvode/cvode.cxx
```diff
@@ -650,33 +739,34 @@ static int cvode_jac(N_Vector v, N_Vector Jv, realtype t, N_Vector y, N_Vector U
...
+                                               int& p, std::vector<BoutReal>& f2dtols,
+                                               std::vector<BoutReal>& f3dtols, bool bndry)
+{
   // Loop over 2D variables
   for (std::vector<BoutReal>::size_type i = 0; i < f2dtols.size(); i++) {
     if (bndry && !f2d[i].evolve_bndry) {
       continue;
     }
-    abstolvec_data[p] = f2dtols[i];
+    option_data[p] = f2dtols[i];
```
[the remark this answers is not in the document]
johnomotani: Don't understand this error - I don't think there's any pointer arithmetic here, just accessing a C-array with an `int` index...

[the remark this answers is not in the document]
ZedThree: The C-array indexing is basically sugar for pointer arithmetic: the compiler turns it into `*(option_data + p) = f2dtols[i]` (and this is why `array[0] == 0[array]` works even though it looks like bad magic). This is a bit more obvious looking at the function signature, which takes a `BoutReal*` rather than an array type.

There isn't a great deal we can do to do this properly, given that we don't own `N_Vector` (and also that it's pretty much an opaque pointer). In C++20, we could take a `std::span` here instead of `BoutReal*`.

I think we _could_ use `N_VMake_Parallel` to create `abstolvec` with our own data (e.g. `Array`), and then pass that directly to `set_abstol_values`.

(Incidentally, I'm pretty sure we could do something like that to avoid copying the derivatives in and out of the solver's data structures if we could somehow allocate a big block of consecutive memory for the derivatives in the first place)

src/solver/impls/cvode/cvode.cxx
```diff
@@ -288,6 +294,89 @@ int CvodeSolver::init(int nout, BoutReal tstep) {
...
+                     switch (value) {
+                       case positivity_constraint::none: return 0.0;
+                       case positivity_constraint::positive: return 2.0;
+                       case positivity_constraint::non_negative: return 1.0;
+                       case positivity_constraint::negative: return -2.0;
+                       case positivity_constraint::non_positive: return -1.0;
+                       default: throw BoutException("Incorrect value for "
+                                                    "positivity_constraint");
+                     }
+                   });
```
ZedThree, line 339: Might be worth pulling this out into a helper function? `VarStr<T>` is private, so you might need to template it like this:

```cpp
template <class FieldType>
std::vector<BoutReal> create_constraints(const std::vector<FieldType>& fields) {
    std::vector<BoutReal> field_constraints;
    field_constraints.reserve(fields.size());
    std::transform(begin(fields), end(fields), std::back_inserter(field_constraints),
                   [](const FieldType& field) {
...
```

or else make it a lambda and use `auto` for the lambda arguments.

src/solver/impls/cvode/cvode.cxx
```diff
@@ -288,6 +294,89 @@ int CvodeSolver::init(int nout, BoutReal tstep) {
...
+                                             "non_positive.", f2.name))
+                                        .withDefault(positivity_constraint::none);
+                     switch (value) {
+                       case positivity_constraint::none: return 0.0;
+                       case positivity_constraint::positive: return 2.0;
+                       case positivity_constraint::non_negative: return 1.0;
+                       case positivity_constraint::negative: return -2.0;
+                       case positivity_constraint::non_positive: return -1.0;
+                       default: throw BoutException("Incorrect value for "
+                                                    "positivity_constraint");
```
ZedThree, line 337: Do we need this? Does `Options` already throw if the value is not correct? Leaving it off should let the compiler warn us if the switch doesn't cover all the enum values.

ZedThree (dismissed):
LGTM, maybe a couple of things to polish

src/solver/impls/cvode/cvode.cxx
johnomotani: Without it I got a compiler warning about "missing return from a non-void function".

src/solver/impls/cvode/cvode.cxx
johnomotani: Good plan, done.

src/solver/impls/cvode/cvode.cxx
ZedThree: Ah fair enough! I guess the compiler isn't clever enough to reason about the two things together.

ZedThree (approved):

PR #2305, opened 2021-05-06
Add instructions for MPCDF clusters

dschwoerer:
The instructions are based on the marconi ones.

Would it make sense to put them rather in a bash file (where?) and reference that instead?


manual/sphinx/user_docs/advanced_install.rst
```diff
@@ -263,6 +263,144 @@ Ubgl
...
+            fi
+        else
+            ln -s lib $PREFIX/lib64
+        fi      
+        mkdir -p $BUILD
+    }
+    
+    hdf5() {
+        cd $BUILD
+        wget --delete -c https://support.hdfgroup.org/ftp/HDF5/releases/hdf5-1.12/hdf5-${HDF5VER}/src/hdf5-${HDF5VER}.tar.bz2
```
ZedThree, line 300: `--delete` is not an option in the version of `wget` on my machine (`GNU Wget 1.21.1 built on linux-gnu.`). Maybe `--unlink`? Or is this a different version of `wget` on this machine?

manual/sphinx/user_docs/advanced_install.rst
```diff
@@ -263,6 +263,144 @@ Ubgl
...
+    # Only allow one bout-dep module to be loaded at a time
+    conflict bout-dep
+    # Require all modules that where loaded at generation time
+    prereq $(echo $LOADEDMODULES | tr : \ )
+    
+    
+    setenv        BOUT_DEP         $PREFIX
+    prepend-path  PATH             $PREFIX/bin
+    prepend-path  LD_LIBRARY_PATH  $PREFIX/lib
+    """
```
ZedThree, line 386: I think the triple-quotes here might be python muscle memory kicking in :)

ZedThree (approved):
Could put the bash script in something like `bin/bout-build-on-draco.sh`?

We might be able to incorporate a fair bit of this into the CMake build. We already have an option for downloading and building the netCDF C++ API for instance.

dschwoerer:
@johnomotani I updated the marconi GNU instructions - I am happy to revert if you prefer your version ...

johnomotani:
> I updated the marconi GNU instructions - I am happy to revert if you prefer your version ...

Looks good to me, but it'd be nice to add a bit of documentation of `bout-build-deps.sh` in `advanced_install.rst` - e.g. the environment flags that should be set, `PREFIX`, etc. It'd also be nice to add a suggestion to run `make check` or `make test` for the libraries - we had some issues with a few tests failing for netcdf depending on versions...

ZedThree:
readthedocs is failing because we need #2308 first

ZedThree:
A short docstring in `bout-build-dep.sh` with a `--help` flag would be lovely too.

ZedThree:
An alternative to this is to use [spack](https://spack.io). There's some spack files already in the GPU branches, e.g. https://github.com/boutproject/BOUT-dev/blob/next-hypre-outerloop-cuda/scripts-config/lassen-examples/spack_env/bout/spack.yaml

It looks like it might also not be very much effort to create a spack package for BOUT++ itself, which might make it even easier to install all the dependencies.

Using spack would require users to learn another tool, but might mean less maintenance for BOUT++ developers.

bshanahan (dismissed):
Works well for me.

bin/bout-build-deps.sh
```diff
@@ -0,0 +1,295 @@
...
+    test $of0 && diff $of0 $of -q && rm $of && of=$dup && ext=$ext0
+    test $reload && msg="
+Re-login in or source ~/.bashrc
+" || msg=
+    cat <<EOF
+####################################################################
+###                  Summary and Info follows!                   ###
+####################################################################
+$msg
+Activate the dependency module by runing
```
bshanahan, line 259: typo runing->running

dschwoerer:
I haven't tested the openmp flags well, they seem to configure without issues, but I have not checked that it actually uses treads ...

dschwoerer:
For intel based tools:
`module load intel/21.3.0 impi/2021.3 vtune/2021.5 advisor/2021.3 itac/2021.3 hdf5-serial/1.12.1 fftw-serial/3.3.9 mkl/2021.3 netcdf-serial/4.7.0 petsc-real-double/3.15`

ZedThree:
> For the future, it might be easier to integrate downloading + compiling of dependencies within cmake ...?

Yes, there's a few options for this: Conan, vcpkg, CPM. Or we can have a "superproject" that includes BOUT++ and downloads and compiles all the dependencies.

This looks ok to me. We could move the script into https://github.com/boutproject/BOUT-configs

bendudson (approved):
This looks good to me. Would it be possible to use bout-build-deps.sh in one of the github tests, so we can make sure that it is updated when needed?

PR #2307, opened 2021-05-07
Add Mesh method to get all possible boundary regions

ZedThree:
Add a method to return a set of all possible boundary region names in a given configuration.

Merging into #2245 to reduce diff

Needed for #2210

### DIFF
--- a/src/mesh/impls/bout/boutmesh.cxx
+++ b/src/mesh/impls/bout/boutmesh.cxx
@@ -47,6 +47,10 @@
 #include <output.hxx>
 #include <utils.hxx>
 
+#include <algorithm>
+#include <iterator>
+#include <set>
+
 /// MPI type of BoutReal for communications
 #define PVEC_REAL_MPI_TYPE MPI_DOUBLE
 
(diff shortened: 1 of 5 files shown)


src/mesh/impls/bout/boutmesh.cxx
```diff
@@ -975,6 +989,53 @@ void BoutMesh::createYBoundaries() {
...
+
+  // This is sufficient to get the SOL boundary, if it exists
+  get_boundaries_on_different_rank(NXPE - 1, 0);
+
+  // Now we need to work out which Y rank at each of the possible
+  // targets and branch cuts. This makes sure we get a processor on
+  // each possible boundary. We only need to do this at PE_XIND==0, as
+  // the other X boundary is covered by the case above
+  for (const auto& y_index :
+       {0, jyseps1_1, jyseps1_2, ny_inner - 1, ny_inner, jyseps2_1, jyseps2_2, ny - 1}) {
```
johnomotani, line 1032: Is it necessary to do `jyseps*` here? I think the targets (where we actually apply y-boundary conditions) are only at `0`, `ny-1`, and (if `jyseps2_1==jyseps1_2`) `ny_inner-1` and `ny_inner`.

johnomotani (approved):

PR #2308, opened 2021-05-07
Fix a couple of documentation issues

ZedThree:
- Typo in the CMake messages for building the docs
- Latest version of `breathe` added a new option for the auto-doc tool that we call in a slightly weird way

### DIFF
--- a/manual/CMakeLists.txt
+++ b/manual/CMakeLists.txt
@@ -9,11 +9,11 @@ set(BOUT_SPHINX_BUILD ${CMAKE_CURRENT_BINARY_DIR}/docs)
 add_custom_target(sphinx-html
   COMMAND ${SPHINX_EXECUTABLE} -b html ${BOUT_SPHINX_SOURCE} ${BOUT_SPHINX_BUILD}
   WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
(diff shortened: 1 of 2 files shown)


ZedThree:
Using my admin privileges to merge this so that the readthedocs checks pass, otherwise other PRs have check failures that aren't relevant. Shout at me later if this turns out to cause problems :slightly_smiling_face:

ZedThree:
> I wanted to ask whether the targets should be renamed from `sphinx-` to `manual-` to match the autotools version and make it easier to find them? Or would `make doc(s)-` be even easier to guess?

Yeah that's a fair-point. I was translating `manual/Makefile` which uses the `sphinx-` names, but the top-level Makefile calls them `manual-`. So it would be better to use that instead. `docs` is also a good name.

> More general, do we need to discover sphinx at configure time? I think it is nice to only check for sphinx as the user tries to build the docs?

Currently, CMake tries to find them if you configure with `BOUT_BUILD_DOCS`, which is default off, so turning it indicates the user probably wants to build the docs. More practically, it's just much easier with CMake to find things at configure time.

ZedThree:
Now that I've merged this, RTD is pulling in Sphinx 4.0 which removes `app.add_stylesheet`, so there's a fix for that incoming shortly. I'll also add a target `docs` that builds `sphinx-html`

PR #2309, opened 2021-05-07
CMake: Add separate configure-time option for MsgStack

dschwoerer:
Formatting took over 50% of the run time, so I added an option to disable it while keeping the full checks enabled.

This is really useful if you want to debug on a small mesh, as you get the stack trace anyway with e.g. gdb.


ZedThree:
Might make sense to make it just on/off, with a default value of `CHECK > 1`? If so, the name should be `BOUT_USE_MSGSTACK`

Please could you also add it to `include/bout/build_config.hxx`, and write it out in `src/bout++.cxx`, in `printCompileTimeOptions`, `setupDumpFile`?

The other option would be to just bump up the `CHECK` level that it gets enabled at.

dschwoerer:
> Might make sense to make it just on/off, with a default value of `CHECK > 1`? If so, the name should be `BOUT_USE_MSGSTACK`

I don't mind, will change :+1: 

> Please could you also add it to `include/bout/build_config.hxx`, and write it out in `src/bout++.cxx`, in `printCompileTimeOptions`, `setupDumpFile`?

Sure :+1: 

> The other option would be to just bump up the `CHECK` level that it gets enabled at.

Even at level 4 I wouldn't want them, as they add no value if I already have compiled with `-g`.
Without `-g` and/or without the native stacktrace the msgstack would add value, but I prefer `-g` as that gives much more info at smaller cost.

dschwoerer:
I cannot find anything in the docs on this, is it possible to evaluate a statement without if?

I want to  assign the result of the comparison `CHECK > 1` - is there another way to:
```
if (CHECK GREATER 1)
  set(VALUE ON)
else ()
  set(VALUE OFF)
endif()
```

ZedThree:
Yeah, I think that's the cleanest way. Everything is stringly-typed in CMake, so boolean expressions basically only work inside `if`

CMakeLists.txt
```diff
@@ -434,6 +434,17 @@ message(STATUS "Runtime checking level: CHECK=${CHECK}")
...
+
+if (CHECK GREATER 1)
+  set(default ON)
+else()
+  set(default OFF)
+endif()
+set(BOUT_USE_MSGSTACK ${default} CACHE STRING "Set the Mesagestack level" )
+set_property(CACHE BOUT_USE_MSGSTACK PROPERTY TYPE BOOL)
+message(STATUS "Message stack: BOUT_USE_MSGSTACK=${BOUT_USE_MSGSTACK}")
+target_compile_definitions(bout++ PUBLIC "BOUT_USE_MSGSTACK=${BOUT_USE_MSGSTACK}")
```
ZedThree, line 446: Now that I come to look at this properly, I see the [`cmake_dependent_option`](https://cmake.org/cmake/help/latest/module/CMakeDependentOption.html) immediately below should do exactly what we want :facepalm: 

```suggestion
cmake_dependent_option(BOUT_ENABLE_MSGSTACK "Enable debug message stack" ON
  "CHECK GREATER 1" OFF)
message(STATUS "Message stack: BOUT_USE_MSGSTACK=${BOUT_ENABLE_MSGSTACK}")
set(BOUT_USE_MSGSTACK ${BOUT_ENABLE_MSGSTACK})
```

What this does is: if `CHECK > 1`, then present the option to turn `MsgStack` on/off, but defaults to on. Otherwise, turn it off.

What this doesn't do is allow the user to turn `MsgStack` on for `CHECK <= 1` -- it's always off in that case.

If we do actually want complete control, then the current method like this works fine, although I would make the following changes:

```suggestion
if (CHECK GREATER 1)
  set(bout_use_msgstack_default ON)
else()
  set(bout_use_msgstack_default OFF)
endif()
set(BOUT_ENABLE_MSGSTACK ${bout_use_msgstack_default} CACHE BOOL "Enable debug message stack")
message(STATUS "Message stack: BOUT_USE_MSGSTACK=${BOUT_ENABLE_MSGSTACK}")
set(BOUT_USE_MSGSTACK ${BOUT_ENABLE_MSGSTACK})
```

The naming convention is `BOUT_ENABLE_*` for the cmake option, and then `BOUT_USE_*` for the `#define`. Third-party dependencies are `BOUT_USE` for the option, `BOUT_HAS` for the define. I can't really remember why I settled on this particular naming convention, historical reasons I suspect. 

`target_compile_definitions` isn't needed, it's all handled through the `build_defines.hxx` header.

ZedThree (changes requested):
LGTM, pick one of the suggestions depending on how you want to use this :)

CMakeLists.txt
dschwoerer: I think it might be useful to keep that options - with optimisations the formatting should be much faster - and for large meshes probably negligible.

PR #2313, opened 2021-05-12
Refactor `FieldData` to be base class of `Field`

ZedThree:
The reason for doing all of this was originally to be able to remove uses of `bout::global::mesh` from `FieldData`, so that I could add the fix for the case-sensitive `Options` for ignoring the boundary region inputs without adding more uses of the global mesh.  There were a few other plus points to doing this, namely removing the double-inheritance from `Field2/3D`, and tidying up the code a bit. It has turned out to be a bit of a bigger effort than I had anticipated/hoped for!

---

Make `FieldData` the base class of `Field`, and move some members for `Field` into `FieldData`. This does a few things:

- remove the double-inheritance from `Field2D` and `Field3D`
- `FieldPerp` now also ultimately inherits from `FieldData`, where it previously was only `Field`
- move the `Mesh`, `Coordinates`, and `CELL_LOC` members from `Field` to `FieldData`
    - this means `Vector2/3D` now have `getMesh()` and `getCoordinates()` methods, which means we don't need to do the hacky `v.x.getMesh()` etc.
- replace `Field::copyFieldMembers` with more idiomatic `Field::operator=(rhs)` calls in the derived classes, and makes sure the copy-constructor methods delegate to the `Field` constructor
   - this makes sure all the copy operations are done consistently down the whole class hierarchy
- make most of the `FieldData` members `private` rather than `protected`
   - I understand why `protected` members are bad now! They make refactoring that much more awkward
   - use getters instead
   - `fieldmesh` is still `protected` because it's used in a _lot_ of places. I should probably also bite the bullet and make this `private` too
- remove several unused methods from `FieldData`, and the unused experimental `FieldVisitor`
- remove use of `bout::globals::mesh` from `FieldData`

Also includes #2306 so that I didn't have to deal with the conflicts separately.

---

If we were to go further here, I think we could do some or all of the following:
- make `FieldData::applyBoundary(bool)` have a default implementation -- it appears to be basically the same between all the implementations
- rename `Field` to `ScalarField` and add a `VectorField` base class to remove duplicated code from the vector classes
- move the time-derivative member (`deriv`) into one of the base-classes and make it a `unique_ptr` -- or even just a value?


bendudson (dismissed):
Nice! Thanks @ZedThree

ZedThree:
@bendudson I just fixed the few clang-tidy warnings, please could you re-approve?

bendudson (approved):

PR #2318, opened 2021-05-17
Fix for FindPETSc not being reentrant if pkg-config used (next)

ZedThree:


dschwoerer:
Looks good. I think there is still one issue:
```
-- Not building test test-options-netcdf, requirement not met: CONFLICTS
```
`test-io` also requires `not legacy_netcdf` - but has no equivalent in the cmake version. `test-stopCheck-file` conflicts with travis - again no equivalent in cmake. Are conflicts not supported?

ZedThree:
Hmm, looks like `CONFLICTS` hasn't actually made it into `next` yet, I think it might just be in the 3D metrics branch. Should be simple to port across though. I'll do that in a separate PR.

There's no equivalent for `not travis` yet, but it's possible the tests with that condition were just working round limitations of Travis that don't apply to GHA. `test-stopCheck-file` seems to work fine on GHA at least, and only takes ~7 seconds, so I don't think that needs fixing. It looks like the only other test with that requirement is multigrid-laplace which is slow on Fedora+Travis. We need to see if that's still the case.

dschwoerer (approved):

PR #2322, opened 2021-05-18
Fix make -q

dschwoerer:
By using real, empty targets `make -q` always returned false. Using order
only dependencies is the proper solution and fixes this issue


dschwoerer:
Looks like there is something wrong with GHA :disappointed:

ZedThree:
Yeah, GHA having problems: https://www.githubstatus.com

ZedThree (approved):

ZedThree:
Lovely, good stuff @dschwoerer !

PR #2323, opened 2021-05-18
Do not include inner corner cells in FCI

dschwoerer:
The inner corner values (i.e. values in the parallel slices, that are in the inner x-boundary) are normally not set, and should thus not be used.

I think this can also go to next, and then be merged into the 3D branch.

### DIFF
--- a/src/mesh/parallel/fci.cxx
+++ b/src/mesh/parallel/fci.cxx
@@ -132,14 +132,13 @@ FCIMap::FCIMap(Mesh& mesh, const Field2D& dy, Options& options, int offset_,
 
       xt_prime_corner[i] = -1.0;
       zt_prime_corner[i] = -1.0;
-      continue;
+    } else {
+      xt_prime_corner[i] =
+          0.25 * (xt_prime[i] + xt_prime[i_xplus] + xt_prime[i_zplus] + xt_prime[i_xzplus]);
+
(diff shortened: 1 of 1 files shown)


src/mesh/parallel/fci.cxx
```diff
@@ -132,14 +132,13 @@ FCIMap::FCIMap(Mesh& mesh, const Field2D& dy, Options& options, int offset_,
...
       xt_prime_corner[i] = -1.0;
       zt_prime_corner[i] = -1.0;
-      continue;
-    }
-
-    xt_prime_corner[i] =
-        0.25 * (xt_prime[i] + xt_prime[i_xplus] + xt_prime[i_zplus] + xt_prime[i_xzplus]);
+    } else {
+      xt_prime_corner[i] =
+          0.25 * (xt_prime[i] + xt_prime[i_xplus] + xt_prime[i_zplus] + xt_prime[i_xzplus]);
```
ZedThree, line 137: I'm not sure I see any difference, this change is functionally identical? Am I missing something?

dschwoerer: Sorry, yes, that is just a cleanup - I thought the continue was confusing, if the else branch is so short.

Probably should have gone into another commit ...

bshanahan (approved):

PR #2326, opened 2021-05-18
[3D metric] Ensure boundaries are valid

dschwoerer:
After communication we might have invalid points, this is fixed by
applying boundaries.


bshanahan (approved):

src/mesh/coordinates.cxx
```diff
@@ -1626,6 +1626,7 @@ Field3D Coordinates::DDY(const Field3D& f, CELL_LOC outloc, const std::string& m
   if (! f.hasParallelSlices() and ! transform->canToFromFieldAligned()) {
     Field3D f_parallel = f;
     transform->calcParallelSlices(f_parallel);
+    f_parallel.applyParallelBoundary("parallel_neumann");
```
ZedThree, line 1629: Should this not be the same parallel boundary as `f`?

ZedThree:
The CMake tests are failing because this needs e6e02e8 from #2328

dschwoerer:
The python interface doesn't currently work with 3D + out of source builds.

I am tempted to fix this based on #2351 ... so disable that in CI for now.

PR #2327, opened 2021-05-18
Autotools: Link against PETSc first to avoid unresolved symbols from parallel HDF5

dschwoerer:
This avoids the error of unresolved symbol in libpetsc as it uses
parallel hdf5 features, but BOUT++ is otherwise linking with the
serial versions.


ZedThree:
Does this imply we're not getting the full list of PETSc libraries we need to link against? Is this only a problem for static builds, or does it affect shared builds too?

ZedThree (approved):

dschwoerer:
The petsc library is already linked to hdf5 - but we are setting `-L/usr/lib64` which takes precedence over the already defined linkage for the petsc, and thus breaks petsc.

For petsc only `-lpetsc` is required, and I think it would be nice if we would switch to the same for libbout++, but that is a different story, and probably something we should try to do with cmake, rather then autoconf.

ZedThree:
PETSc only needs `-lpetsc` when it's built as a shared library, for a static build it needs all its dependencies. We include all our dependencies on the link line because historically we only had static builds.

CMake handles this all automatically, including using rpath so that the user doesn't need to set `LD_LIBRARY_PATH`.

PR #2330, opened 2021-05-19
Parallel cyclic reduction

JosephThomasParker:
As mentioned in #2296, cyclic reduction wasn't actually performing cyclic reduction.  Also, as noted in #2314, its scaling with processor count on Archer2 is much worse than we've seen on other machines.

This PR adds a parallel cyclic reduction solver, incorporating code from [this repo](https://github.com/jihoonakang/parallel_tdma_cpp), published under MIT.  It can be selected using `type=pcr`.

This solver has a much better scaling performance than "old cyclic" while still being a parameter-free direct solver. The graph below shows run times on Archer2 (with 128 cores per node) for blob2d with (nx,nz)=(8192,1024) and rk4 time advance. Old cyclic (in magenta) doesn't scale nearly so well as parallel cyclic reduction (in red). The multigrid-Thomas solver (blue) is available in BOUT++ as `type=ipt` (`impls/iterative_parallel_tri`), while multigrid (green) is _not_ the multigrid in `next` - it hasn't been pushed yet.  However, PCR is a good drop-in replacement for old cyclic, without the need to fiddle with parameters (and possible non-convergence) that comes with multigrid.  

run_time_8196


ZedThree:
Should we name this one "cyclic" and rename the direct partitioned one to something else?

We might also want to put the citation for this solver into `CITATION.cff`

bendudson:
I think on balance renaming the old one (to `partitioned`?) and calling the new one is good. This will change the algorithm used for the same input file, but the result should be the same so shouldn't affect reproducibility.

JosephThomasParker:
There are some limitations:

- [ ] **requires `NYPE=1`** - probably not necessary, but I need to play around with the implementation to fix this
- [ ] **cannot run in serial** - again probably not a necessary constraint
- [ ] **requires `nx=2^n** and `NXPE=2^m with `n >= 2*m`** - this is from the library README, so it's unlikely to find a simple fix for this.

Other possible enhancements
- [ ] The wrapper around the actual tridiagonal solve is identical to that in `cyclic_laplace`. This should probably be shared between them somehow.
- [ ] The repo (and code here) can also support two other solvers: (non-parallel) cyclic reduction and hybrid Thomas-PCR, both of which could probably implemented quite easily, given code here.

JosephThomasParker:
@bendudson Yes, both are direct solvers and their solutions agree to roundoff.

bendudson:
Ah the restriction to running in serial is tricky... we probably can't make it the default then. Renaming current cyclic, but keeping it as the default might be good for now. If someone has set `type = cyclic` then they would from now on get actual cyclic.

src/invert/laplace/impls/cyclic/cyclic_laplace.cxx
```diff
@@ -476,3 +477,74 @@ Field3D LaplaceCyclic::solve(const Field3D& rhs, const Field3D& x0) {
...
 }
+
+void LaplaceCyclic :: verify_solution(const Matrix<dcomplex> &a_ver, const Matrix<dcomplex> &b_ver, const Matrix<dcomplex> &c_ver, const Matrix<dcomplex> &r_ver, const Matrix<dcomplex> &x_sol, const int nsys)
+{
+    output.write("Verify solution\n");
+    const int nx = xe - xs + 1;  // Number of X points on this processor,
+                                 // including boundaries but not guard cells
+    const int myrank = localmesh->getXProcIndex();
+    const int nprocs = localmesh->getNXPE();
+    int i;
```
ZedThree, line 488: Is this needed outside of the loop at the end?

src/invert/laplace/impls/cyclic/cyclic_laplace.cxx
```diff
@@ -476,3 +477,74 @@ Field3D LaplaceCyclic::solve(const Field3D& rhs, const Field3D& x0) {
 
   return x;
 }
+
+void LaplaceCyclic :: verify_solution(const Matrix<dcomplex> &a_ver, const Matrix<dcomplex> &b_ver, const Matrix<dcomplex> &c_ver, const Matrix<dcomplex> &r_ver, const Matrix<dcomplex> &x_sol, const int nsys)
+{
+    output.write("Verify solution\n");
```
ZedThree, line 483: Maybe

```suggestion
    output_progress.write("Verify solution\n");
```

or `output_info`/`output_verbose`?

src/invert/laplace/impls/cyclic/cyclic_laplace.cxx
```diff
@@ -476,3 +477,74 @@ Field3D LaplaceCyclic::solve(const Field3D& rhs, const Field3D& x0) {
...
+    }
+    
+    BoutReal max_error = 0.0;
+    for(int kz=0;kz<nsys;kz++){
+      for(i=0;i<nx;i++) {
+        y_ver(kz,i) = a_ver(kz,i)*x_ver(kz,i)+b_ver(kz,i)*x_ver(kz,i+1)+c_ver(kz,i)*x_ver(kz,i+2);
+        error(kz,i) = y_ver(kz,i) - r_ver(kz,i);
+	if(abs(error(kz,i)) > max_error){
+	  max_error = abs(error(kz,i));
+	}
```
ZedThree, line 545:
```suggestion
        max_error = std::max(max_error, std::abs(error(kz, i)));
```

src/invert/laplace/impls/pcr/pcr.cxx
```diff
@@ -0,0 +1,1137 @@
...
+  if (localmesh->hasBndryLowerY()) {
+    if (include_yguards)
+      ys = 0; // Mesh contains a lower boundary and we are solving in the guard cells
+
+    ys += extra_yguards_lower;
+  }
+  if (localmesh->hasBndryUpperY()) {
+    if (include_yguards)
+      ye = localmesh->LocalNy -
+           1; // Contains upper boundary and we are solving in the guard cells
```
ZedThree, line 343: Might want to move the comment to the previous line so clang-format does do this :)

src/invert/laplace/impls/pcr/pcr.cxx
```diff
@@ -0,0 +1,1137 @@
...
+}
+
+
+/** 
+ * @brief   Initialize local private variables from global input parameters.
+ * @param   n Size of global array
+ * @param   np_world Number of MPI process
+ * @param   rank_world rank ID in MPI_COMM_WORLD
+*/
+void LaplacePCR :: setup(int n, int np_world, int rank_world)
```
ZedThree, line 537: Can this be inlined into the constructor? It doesn't look like it's called anywhere else

src/invert/laplace/impls/pcr/pcr.cxx
```diff
@@ -0,0 +1,1137 @@
...
+}
+
+/** 
+ * @brief   Forward elimination of CR until a single row per MPI process remains.
+ * @details After a single row per MPI process remains, PCR or CR between a single row is performed.
+*/
+void LaplacePCR :: cr_forward_multiple_row(Matrix<dcomplex> &a,Matrix<dcomplex> &b,Matrix<dcomplex> &c,Matrix<dcomplex> &r)
+{
+    MPI_Comm comm = BoutComm::get();
+    int i, l;
```
ZedThree, line 675: Clang-tidy will probably complain about these, but I think these declarations can be moved into the loop init statements

JosephThomasParker:
Turns out that "cyclic reduction" is not the serial version of "parallel cyclic reduction" :monocle_face: 
but in any case, a small tweak to the parallel case makes it work on one core (guarding some communication).

src/invert/laplace/impls/cyclic/cyclic_laplace.cxx
JosephThomasParker: Nope, was a dumb copy. Fixed.

src/invert/laplace/impls/cyclic/cyclic_laplace.cxx
```diff
@@ -482,7 +482,7 @@ void LaplaceCyclic ::verify_solution(const Matrix<dcomplex>& a_ver,
                                      const Matrix<dcomplex>& b_ver,
                                      const Matrix<dcomplex>& c_ver,
                                      const Matrix<dcomplex>& r_ver,
-                                     const Matrix<dcomplex>& x_sol, const int nsys) {
+                                     const Matrix<dcomplex>& x_sol, int nsys) {
```
dschwoerer, line 485: Keeping the value const in the cxx file is perfectly fine, and probably even nicer.

The `const` doesn't have to be in the interface file (but it is no error to have it there), as it does not change the interface.

Removing the const from the implementation changes however the implementation, as `nsys` can now be (accidentally) changed.

src/invert/laplace/impls/pcr/pcr.cxx
```diff
@@ -0,0 +1,1086 @@
...
+  a.reallocate(nmode, n);
+  b.reallocate(nmode, n);
+  c.reallocate(nmode, n);
+  xcmplx.reallocate(nmode, n);
+  bcmplx.reallocate(nmode, n);
+
+  xproc = localmesh->getXProcIndex(); // Local rank in x proc space
+  const int yproc = localmesh->getYProcIndex();
+  nprocs = localmesh->getNXPE();                    // Number of processors in x
+  myrank = yproc * nprocs + xproc;                  // Global rank for communication
```
ZedThree, line 130: Could just be `BoutComm::rank()`?

src/invert/laplace/impls/pcr/pcr.cxx
```diff
@@ -0,0 +1,1086 @@
...
+  if ((global_flags & INVERT_BOTH_BNDRY_ONE) || (localmesh->xstart < 2)) {
+    inbndry = outbndry = 1;
+  }
+  if (inner_boundary_flags & INVERT_BNDRY_ONE)
+    inbndry = 1;
+  if (outer_boundary_flags & INVERT_BNDRY_ONE)
+    outbndry = 1;
+
+  if (dst) {
+    BOUT_OMP(parallel) {
```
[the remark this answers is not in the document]
ZedThree: This is wrong, I think it's just confused by the openmp loop turning into a function call.

ZedThree:
What do we want to do about the names? I'm happy for this to go in now, and we can sort out the names in another PR

JosephThomasParker:
PCR is a parameter-free direct solver, so it is a like-for-like replacement for `cyclic` that scales better. The only concern I have is the restriction that we require powers of two for both processors and points in x. If we change the default name, then there's a good chance that previously-working submission scripts will error (particularly on Marconi, 48 cores per node). We could change the name, and also make the old cyclic the fallback solver when the procs/points conditions aren't met, but then there would be an unexplained performance drop-off.

The way around this to make the solver node-aware: solving within a node using (something like) the Thomas algorithm or the old cyclic (that doesn't care about point or core counts), and between nodes with PCR. This would only require using 2^n nodes, which is much less restrictive. The "within a node" bit will require work beyond this PR though.  The Kang library is set up to do Thomas + PCR but it only has MPI so I doubt it would be node-aware out-of-the-box. Doing old cyclic on-node and PCR off-node might be nice though (and wouldn't require OpenMP (or SYCL :stuck_out_tongue: ))

JosephThomasParker:
After a bit of code-digging:
* Thomas+PCR would get rid of the `nx=2^n` restriction without needing to be node-aware, but we'd still need `NXPE=2^m`. This could be implemented reasonably easily from the Kang code.
* old cyclic + PCR would get rid of all restrictions except `nnodes=2^m`.  This would require thinking about the algorithm though - the PCR bit requires coefficients as inputs (in the form that they happen to come out from the Thomas algorithm). On the other hand, solvers tend to give "the solution" as their output. It takes some work to translate between "solution to a node's subsystem" and "what this means in terms of coefficients".

ZedThree (approved):

src/invert/laplace/impls/pcr/pcr.cxx
```diff
@@ -114,285 +157,732 @@ void transpose(Matrix<dcomplex>& m_t, const Matrix<dcomplex>& m) {
...
+      // including boundary conditions
+      BOUT_OMP(for nowait)
+      for (int ind = 0; ind < nsys; ind++) {
+        // ind = (iy - ys) * nmode + kz
+        int iy = ys + ind / nmode;
+        int kz = ind % nmode;
+
+        BoutReal zlen = coords->dz * (localmesh->LocalNz - 3);
+        BoutReal kwave =
+            kz * 2.0 * PI / (2. * zlen); // wave number is 1/[rad]; DST has extra 2.
```
dschwoerer, line 407: It sounds like "DST has extra 2" means that there are two extra cells, but why is then `nz - 3` used?

PR #2331, opened 2021-05-21
CMake: Add option to download SUNDIALS at configure time

ZedThree:
Similarly to the netCDF C++ API, `-DBOUT_DOWNLOAD_SUNDIALS=ON` will download and build SUNDIALS as part of the BOUT++ build. This should make it much easier to use SUNDIALS on systems that don't have it installed.

Currently needs my (Peter's) fork because the latest release is
missing an `export()` call that allows it to work with
`FetchContent`.

Also a small change to `FindSundials` to make the
download/non-download options work in the same way


bendudson (dismissed):
LGTM, works on YPI server (Avalanche). I've made a PR into this branch with a bit of documentation.

bendudson (approved):

PR #2333, opened 2021-05-26
Add wrapper for `LaplacePCR` that falls back to `LaplaceCyclic`

ZedThree:
The new PCR solver pretty much always beats cyclic, but has a couple of preconditions that mean it's not suitable as the default Laplacian type -- it needs both `NXPE` and `nx - 2*mxg` to be powers of two.

This PR adds a wrapper `LaplacePCRorCyclic` (better suggestions for names welcome!) that uses an internal `LaplacePCR` if its preconditions are met, otherwise it uses `LaplaceCyclic`. The default `Laplacian` is now `pcr_or_cyclic` (also better suggestions welcome!)

One downside to this approach is the repetition of the PCR preconditions in this new solver. That could be changed to doing this instead:

```cpp
try {
  pcr_or_cyclic = std::make_unique<LaplacePCR>(opt, loc, mesh_in);
} catch (const BoutException&) {
  pcr_or_cyclic = std::make_unique<LaplaceCyclic>(opt, loc, mesh_in);
}
```

but there may be other ways `LaplacePCR` could fail during construction?

---

After I wrote this, I realised an alternative to this whole PR would be to move this logic into `LaplaceFactory::create` -- it turns out we already have a custom implementation for this, which could be something like:

```cpp
  ReturnType create(Options* options = nullptr, CELL_LOC loc = CELL_CENTRE,
                    Mesh* mesh = nullptr) {
    options = optionsOrDefaultSection(options);
    if (not options->isSet("type")) {
      try {
        return Factory::create("pcr", options, loc, mesh);
      } catch (const BoutException&) {
        return Factory::create("cyclic", options, loc, mesh);
      }
    }
    return Factory::create(getType(options), options, loc, mesh);
  }
```

This would probably handle the majority of cases -- it looks like most examples use `Laplacian::create` rather than `LaplaceFactory::create`. The former just calls this particular overload.

---

The benefit of the wrapper class is that it does mean you can explicitly request it with `laplace:type=pcr_or_cyclic`. I'm not sure if that's useful or not.

### DIFF
--- a/tests/integrated/test-laplace/runtest
+++ b/tests/integrated/test-laplace/runtest
@@ -52,7 +52,7 @@ for v in vars:
 print("Running Laplacian inversion test")
 success = True
 
-for solver in ["cyclic", "pcr"]:
+for solver in ["pcr", "cyclic", "pcr_or_cyclic"]:
     for nproc in [1, 2, 4]:
         nxpe = 1
         if nproc > 2:
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -186,6 +186,8 @@ set(BOUT_SOURCES
   ./src/invert/laplace/impls/naulin/naulin_laplace.hxx
   ./src/invert/laplace/impls/pcr/pcr.cxx
   ./src/invert/laplace/impls/pcr/pcr.hxx
+  ./src/invert/laplace/impls/pcr_or_cyclic/pcr_or_cyclic.cxx
+  ./src/invert/laplace/impls/pcr_or_cyclic/pcr_or_cyclic.hxx
   ./src/invert/laplace/impls/pdd/pdd.cxx
   ./src/invert/laplace/impls/pdd/pdd.hxx
   ./src/invert/laplace/impls/petsc/petsc_laplace.cxx
--- a/include/bout/mesh.hxx
+++ b/include/bout/mesh.hxx
@@ -375,10 +375,10 @@ class Mesh {
   [[deprecated("This experimental functionality will be removed in 5.0")]]
   virtual comm_handle receiveFromProc(int xproc, int yproc, BoutReal *buffer, int size, int tag) = 0;
   
-  virtual int getNXPE() = 0; ///< The number of processors in the X direction
-  virtual int getNYPE() = 0; ///< The number of processors in the Y direction
-  virtual int getXProcIndex() = 0; ///< This processor's index in X direction
-  virtual int getYProcIndex() = 0; ///< This processor's index in Y direction
+  virtual int getNXPE() const = 0; ///< The number of processors in the X direction
+  virtual int getNYPE() const = 0; ///< The number of processors in the Y direction
+  virtual int getXProcIndex() const = 0; ///< This processor's index in X direction
+  virtual int getYProcIndex() const = 0; ///< This processor's index in Y direction
   
   // X communications
   virtual bool firstX() const = 0;  ///< Is this processor first in X? i.e. is there a boundary to the left in X?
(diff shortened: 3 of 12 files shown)


tests/integrated/test-laplace/runtest
```diff
@@ -52,7 +52,7 @@ for v in vars:
 print("Running Laplacian inversion test")
 success = True
 
-for solver in ["cyclic", "pcr"]:
+for solver in ["pcr", "cyclic", "pcr_or_cyclic"]:
```
ZedThree, line 55: Not really sure adding this to this test is useful -- it's always PCR. The grid file would need changing such that PCR was sometimes not acceptable, but I think that basically necessitates an entirely new test, or _only_ using `pcr_or_cyclic`

d7919:
From a user point of view I guess the most interesting feature is that these are direct solvers rather than iterative ones. Could the user request this as just "direct" with the code then sorting out which is the best option (as is done here)? We could also have "iterative" to pick our recommended iterative method. This would be a way of the user saying I don't care what method we use as long as it's from a particular class. We're then free to change how we pick what methods those actually correspond to. This would mean that what an input file gets you could change between versions, but by not specifying an exact type you've already said that you don't care.

bendudson:
The `direct` and `iterative` classes of solvers sounds like a good idea

dschwoerer:
I am not sure catching an exception is the best way to go.

- It is slow
- It is not specific - we throw BoutException for any reason
- It makes `catch throw` in `gdb` debugging confusing

It seems the test are fairly simple - would it be better to have an extra function `is_suitable()` with the same arguments as the constructor, that defaults to `True`.

While that does do lead to code duplication, it seems the criteria isn't that complicated ...

ZedThree:
I've been thinking for quite some time about adding something like a metadata struct to the Laplacians, maybe:

```cpp
struct LaplacianMetadata {
  /// Works on one processor
  bool serial;
  /// Works on multiple processors
  bool parallel;
  /// Is a direct solver
  bool direct;
  /// Is an iterative solver
  bool iterative
  /// Other information?
  ...
};
```

This could be queried in various ways, e.g. maybe a `--help-laplacian=pcr` command line flag to print this information out?

One way to implement the "give me the default direct solver` would be through overriding the factory `getType` method:

```cpp
  std::string getType(Options* options = nullptr) const {
    options = optionsOrDefaultSection(options);
    const std::string type = (*options)[DerivedFactory::option_name].withDefault(default_type);
    if (type == "direct") {
      return default_direct_solver;
    }
    if (type == "iterative") {
      return default_iterative_solver;
    }
  }
```

This won't be able to handle the issue in this PR of "PCR or Cyclic", so that would still need to be dealt with elsewhere.

--- 

@dschwoerer I'm not sure about your points on the exception-based method:

- it should usually only hit the exception at most once during initialisation, unless you're creating lots of solvers, so usually shouldn't be too much impact (and then only if the solver type isn't specified and the preconditions not met -- if it's all fine, there won't be any slowdown)
- we could subclass `BoutException` with `PreconditionNotMetException` -- could be more generally useful too?
- it's true debugging becomes a little more painful, but not very much, and the STL is _much_ more painful for debugging

But yes, having a `is_suitable()` method is probably a nicer solution, as it can also handle the serial/parallel solvers. Even better would be to return 

```cpp
struct IsSuitable {
  bool suitable;
  std::string reason; // is suitable is false, the reason why
};
```

No need to duplicate anything, we can just move all the preconditions into that function and call it in the constructor.

We could go even further, and if a solver can't be used, go through the list of available solvers and tell the user which ones are suitable. That probably needs a try-catch in the factory `create` method in order to do that -- and then having a `PreconditionNotMetException` would be useful to be specific.

ZedThree:
I'm not sure a method like `is_suitable` is possible to implement. How do we call it? It either needs to be:

1. virtual -- in which case, we need to create the particular solver first in order to call the function, but by that point we've either successfully created it, or we've thrown an exception.
2. static -- in which case, we need to know the exact type in order to call it, and there's no way of getting that information dynamically, we'd need to hardcode the list of types.

ZedThree:
Closing this in favour of #2336

PR #2334, opened 2021-05-27
CMake: Add integration test for bout-config

dschwoerer:


ZedThree (approved):

ZedThree:
Thanks @dschwoerer !

Might be an idea to have `set -ex` in the script?

dschwoerer:
Ah, I had set that in the debug commits, that I have removed during rebase :sweat_smile: 

Otherwise it will only fail if the last command fails, and ignores the test results completely ...

PR #2336, opened 2021-06-01
Command-line options help

ZedThree:
This is an initial proof-of-concept for a system for printing help for specific solver, laplacian, etc. implementations. This current version lets users do this:

```
$ ./2fluid --help-laplacian ipt
atol = 1e-20
max_cycle = 1
max_level = 0
maxits = 100
predict_exit = false
rtol = 1e-07
```

Still lots to be done, but I'm interested in any feedback people might have.

The way it works is that when we register a `Thing` (`Solver`, `Laplacian`, etc), we also register a type like:

```cpp
namespace bout {
template <>
struct ArgumentHelper<Thing> : public ArgumentHelperBase {
  ArgumentHelper(Options& options) :
    arg1(options["arg1"].doc("docstring").withDefault(value)),
    arg2(options["arg2"].doc("docstring").withDefault(value)) {}
  int arg1;
  bool arg2;
};
}
```

There's a default implementation of `ArgumentHelper<T>`, which means we can always register a type like this, even if it hasn't been specialised for `Thing`. It needs to inherit from a common base type so we can stick it in a `map<string, unique_ptr<ArgumentHelperBase>>`

Then when parsing the command line, we can do something like:
```cpp
    if (current_arg == "--help-laplacian") {
      // Needs check for argument
      std::cout << LaplaceFactory::getInstance().help(argv[++i]) << std::endl;
      std::exit(EXIT_SUCCESS);
    }
```

The `help` function here constructs a new `Options` and passes it to the correct `ArgumentHelper<T>`. We can then see exactly which options it uses, and print them to screen.

We can avoid duplicating code by pulling all of the use of `Options` out of the `Thing` constructor into its `ArgumentHelper` class, and then do:
```cpp
class Thing : public ThingBase {
public:
  Thing(const ArgumentHelper<Thing>& args, <other args>) :
    arg1(args.arg1), arg2(args.arg2) { <rest of current ctor> }
  Thing(Options& options, <other args>) : 
    Thing(ArgumentHelper<Thing>(options, <other args>)) {}
```

All of this is "opt-in", meaning we get the extra functionality if we specialise the `ArgumentHelper<T>` for some particular class, but it still works if we don't.

We can go even further, and add a `checkPrecondition` method to the `ArgumentHelper<T>` that takes the usual constructor arguments. This would allow us to check a type's preconditions before we construct one. We might not be able to do this during the command-line argument parsing because e.g. we might not have a `Mesh` yet, but we would be able to do this in any place we currently construct a `Thing`, e.g. in `ThingFactory::create`. This would replace the functionality in #2333, and allow us to provide fallbacks when we can't construct the default type, or for having "vague" options like specifying just a `type = direct` Laplacian.

---

Some polish that would be required:

- adding a `Formatter` to `Options` so we can print out the docstring (plus default section names?)
- fixing the template for `RegisterInFactory` so we don't need to redefine it if a type takes different arguments than plain `Options`
- making e.g. `ArgumentHelper<LaplacianIPT>` inherit from `ArgumentHelper<Laplacian>` so that we also pick up the options from the base class
- work out what's needed to enable this for user models as well

On that last point, I think it is possible, but might need a fair bit of restructuring so that we can get the model type name during the command line parsing.


ZedThree:
With #2341, we can now get some docstrings and types with our input help:

```
$ ./2fluid --help-laplacian ipt
laplace:all_terms = true                # type: bool, doc: Include first derivative terms?
laplace:async = true            # type: bool, doc: Use asyncronous MPI send?
laplace:atol = 1e-20            # type: BoutReal, doc: Absolute tolerance
laplace:extra_yguards_lower = 0         # type: int, doc: Exclude some number of points at the lower boundary, useful for staggered grids or when boundary conditions make inversion redundant
laplace:extra_yguards_upper = 0         # type: int, doc: Exclude some number of points at the upper boundary, useful for staggered grids or when boundary conditions make inversion redundant
laplace:filter = 0              # type: BoutReal, doc: Fraction of Z modes to filter out. Between 0 and 1
laplace:global_flags = 0                # type: int, doc: Default flags
laplace:include_yguards = false         # type: bool, doc: Solve Laplacian in Y guard cells?
laplace:inner_boundary_flags = 0                # type: int, doc: Flags to set inner boundary condition
laplace:low_mem = false         # type: bool, doc: If true, reduce the amount of memory used
laplace:max_cycle = 1           # type: int, doc: Maximum number of iterations per coarse grid
laplace:max_level = 0           # type: int, doc: Maximum number of coarse grids
laplace:maxits = 100            # type: int, doc: Maximum number of iterations
laplace:maxmode = 999           # type: BoutReal, doc: The maximum Z mode to solve for. Default is based on filter and mesh size
laplace:nonuniform = true               # type: bool, doc: Use non-uniform grid corrections? Default is the coordinates setting (top-level option 'non_uniform').
laplace:outer_boundary_flags = 0                # type: int, doc: Flags to set outer boundary condition
laplace:predict_exit = false            # type: bool, doc: Predict when convergence will be reached, and skip expensive convergence checks at earlier iterations
laplace:rtol = 1e-07            # type: BoutReal, doc: Relative tolerance
```

Might be nice to do a first pass through a section to line up all the docstrings

ZedThree:
Nice experiment, but ultimately replaced by the much simpler #2483

PR #2340, opened 2021-06-03
Add PCR Thomas Laplacian inversion solver

JosephThomasParker:
This adds another direct parallel solver, `pcr_thomas`, based on [Laszlo, Gilles and Appleyard, Manycore Algorithms for Batch Scalar and Block Tridiagonal Solvers, ACM TOMS, 42, 31 (2016)](https://ora.ox.ac.uk/objects/uuid:6985082c-8c58-4549-bfb8-6051b72b1fdc/download_file?file_format=pdf&safe_filename=paper.pdf&type_of_work=Journal+article) and adapted from the [parallel_tdma_cpp](https://github.com/jihoonakang/parallel_tdma_cpp) library.

It is very similar to the parallel cyclic reduction solver (#2330), except instead of performing PCR directly on the distributed tridiagonal matrix, it uses the Thomas algorithm on each core to get a reduced tridiagonal system with 2 equations per processor. It then calls PCR on this reduced system. This reduces the number of communications from `O(log2(nx))` to `O(log2(nxpe))`, without changing the `O(nx)` complexity.  

This solver passes the Laplace integrated tests, but I haven't done any performance analysis on it.  This is really intended as a stepping stone to getting a node-aware solver, which would further reduce the communication count to `O(log2(nnodes))` inter-node communications. To do this, we need to replace the Thomas solver with a parallel direct solver and work out how to extract the coefficients to feed into a node-level PCR. This PR gives the code structure to do this, but since it also adds a complete solver, I thought I'd add it as an option.


src/invert/laplace/impls/pcr_thomas/pcr_thomas.cxx
```diff
@@ -0,0 +1,1219 @@
...
+        }
+      }
+
+      // Get elements of the tridiagonal matrix
+      // including boundary conditions
+      BOUT_OMP(for nowait)
+      for (int kz = 0; kz < nmode; kz++) {
+        BoutReal zlen = coords->dz * (localmesh->LocalNz - 3);
+        BoutReal kwave =
+            kz * 2.0 * PI / (2. * zlen); // wave number is 1/[rad]; DST has extra 2.
```
dschwoerer, line 194: Similar to what I asked for the `pcr` - why is there a 3 but the comment says 2?

Besides that, would that also work if `dz` is a `Field2D`?

I assume variation of `dz` along `z` wouldn't work due to FFTs require being uniform ...

ZedThree:
This would work best when transformed into a node-aware solver. The best way to do that might be running one MPI rank per node, and then filling up the node with OpenMP threads. So, we might want to wait for the outer-loop stuff in order to get the best performance.

ZedThree (approved):

PR #2345, opened 2021-06-07
Fix for conditionally used boundary condition inputs in 'all'

ZedThree:
Boundary conditions like `all:bndry_all` also need to be marked as
conditionally used in order to avoid unused input errors in parallel
simulations


bshanahan (approved):
Fixes my issue.

PR #2347, opened 2021-06-09
CMake rename PACKAGE_TESTS to BOUT_TESTS

bendudson:
If including BOUT++ as a dependency, this namespacing makes it easier to select which tests should be run: In some cases (e.g. CI, Github actions) only the tests for the physics model should be run, and not the BOUT++ tests.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -748,9 +748,9 @@ target_compile_definitions(bout++
 ##################################################
 # Tests
 
-option(PACKAGE_TESTS "Build the tests" ON)
+option(BOUT_TESTS "Build the tests" ON)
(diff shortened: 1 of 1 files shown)


ZedThree:
It might also be useful to set the default value like this:

```cmake
# Are we building BOUT++ directly, or as part of another project
string(COMPARE EQUAL 
  "${PROJECT_NAME}" "${CMAKE_PROJECT_NAME}"
  PROJECT_IS_TOP_LEVEL
)
option(BOUT_TESTS "Build the tests" ${PROJECT_IS_TOP_LEVEL})
```

ZedThree (approved):

PR #2348, opened 2021-06-10
Convert ersatz MMS tests to use library MMS features

ZedThree:
Two MMS tests did lots of the source/solution/boundary condition stuff manually. This PR converts them to use the library features:

- move the source/solution functions into the input file
- add `solver:mms = true`
- delete unneeded header

Useful for #2209 so I don't need to faff about writing the error fields from the custom monitors, instead this will let the library take care of them automatically.


bendudson (approved):
Excellent! That simplifies things quite a bit. I think these were some early tests, which were developed before or while the library MMS support was implemented.

PR #2349, opened 2021-06-10
Delete integrated test of FieldFactory: replaced by unit tests

ZedThree:
It's plausible that `test-fieldfactory` is useful, as it runs on multiple processors, but it only compares against a benchmark and only tests a few options, while the unit tests are much more comprehensive.

`test-fieldfactory2` doesn't contain a real test, so should definitely go.


bendudson (approved):
Thanks @ZedThree .  I agree that the unit tests cover this much better than these integrated tests

PR #2357, opened 2021-07-01
Fix test_coordinates

dschwoerer:
The test doesn't actually test the default behaviour, as the default is `dz` of 2pi/nz.

For some reason if dz is a Field, it gets the 2pi/nz value in the unit test, but not if it is a BoutReal.

### DIFF
--- a/tests/unit/mesh/test_coordinates.cxx
+++ b/tests/unit/mesh/test_coordinates.cxx
@@ -17,6 +17,8 @@ using bout::globals::mesh;
 
 using CoordinatesTest = FakeMeshFixture;
 
+const BoutReal default_dz{2 * 3.141592653589793 / 7};
+
 TEST_F(CoordinatesTest, ZLength) {
   Coordinates coords{
       mesh,         Field2D{1.0}, Field2D{1.0}, BoutReal{1.0}, Field2D{1.0}, Field2D{1.0},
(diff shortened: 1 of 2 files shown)


tests/unit/mesh/test_coordinates.cxx
```diff
@@ -17,6 +17,8 @@ using bout::globals::mesh;
 
 using CoordinatesTest = FakeMeshFixture;
 
+const BoutReal default_dz{2 * 3.141592653589793 / 7};
```
ZedThree, line 20:
```suggestion
constexpr BoutReal default_dz{TWOPI / CoodinatesTest::nz};
```

Might need to `#include "constants.hxx"`

tests/unit/test_extras.hxx
```diff
@@ -371,58 +371,66 @@ public:
...
       fval = values[name].as(FieldPerp(0.0, mesh));
       return true;
     }
+    fval = def;
     return false;
   }
 
   bool get(Mesh*, std::vector<int>&, const std::string&, int, int = 0,
            Direction = GridDataSource::X) override {
+    throw BoutException("NI");
```
ZedThree, line 428:
```suggestion
    throw BoutException("Not implemented");
```

Assuming you're not making a Monty Python reference :)

tests/unit/test_extras.hxx
```diff
@@ -371,58 +371,66 @@ public:
...
     return false;
   }
   bool get(Mesh*, std::vector<BoutReal>&, const std::string&, int, int = 0,
            Direction UNUSED(dir) = GridDataSource::X) override {
+    throw BoutException("NI");
```
ZedThree, line 433:
```suggestion
    throw BoutException("Not implemented");
```

ZedThree (changes requested):

ZedThree:
This is also quite worrying behaviour

ZedThree (approved):

PR #2358, opened 2021-07-01
CMake: Add SOVERSION; fix FindSUNDIALS

dschwoerer:
* Add a soname (maybe should be the same as the autotool one)?

* Fix sundials issue


cmake/FindSUNDIALS.cmake
```diff
@@ -68,9 +68,10 @@ if (SUNDIALS_DEBUG)
     " SUNDIALS_nvecparallel_LIBRARY = ${SUNDIALS_nvecparallel_LIBRARY}")
 endif()
 
-if (SUNDIALS_nvecparallel_LIBRARY)
-  list(APPEND SUNDIALS_LIBRARIES "${SUNDIALS_nvecparallel_LIBRARY}")
+if (${SUNDIALS_nvecparallel_LIBRARY} STREQUAL "SUNDIALS_nvecparallel_LIBRARY-NOTFOUND")
```
ZedThree, line 71: I think the following should be sufficient:

```suggestion
if (NOT SUNDIALS_nvecparallel_LIBRARY)
```

Variables that end in `-NOTFOUND` are considered false.

cmake/FindSUNDIALS.cmake
```diff
@@ -89,9 +90,10 @@ foreach (LIB ${SUNDIALS_COMPONENTS})
       " SUNDIALS_${LIB}_LIBRARY = ${SUNDIALS_${LIB}_LIBRARY}")
   endif()
 
-  if (SUNDIALS_${LIB}_LIBRARY)
-    list(APPEND SUNDIALS_LIBRARIES "${SUNDIALS_${LIB}_LIBRARY}")
+  if (${SUNDIALS_${LIB}_LIBRARY} STREQUAL "SUNDIALS_${LIB}_LIBRARY-NOTFOUND")
```
ZedThree, line 93:
```suggestion
  if (NOT SUNDIALS_${LIB}_LIBRARY)
```

ZedThree (changes requested):

ZedThree (approved):

PR #2359, opened 2021-07-01
LaplaceXZ is buggy

dschwoerer:


ZedThree:
See #2207 for some discussion

ZedThree:
Possible directions for investigation:

- convergence checks for the round-trip test here: if the issue is one of mismatched discretisations, the error should decrease as we increase the resolution
- analytic tests: this could be done in both uniform and spatially-dependent metric tensors by solving the forward operator with something like sympy or mathematica -- this would give us analytic tests for both forward and inverse operators

PR #2362, opened 2021-07-02
Added tcsh compatible build scripts for cori

ejm930:
Just some scripts to allow tcsh users to build on cori.

### DIFF
--- a/scripts-config/cori-examples/config-bout-cgpu.csh
+++ b/scripts-config/cori-examples/config-bout-cgpu.csh
@@ -1,0 +1,99 @@
+#!/usr/bin/env tcsh
+# this script is designed to be run in the BOUT source dir
+set arch = `uname -m`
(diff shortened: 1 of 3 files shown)


bendudson:
Thanks @ejm930 ! It would be nice to document how these scripts should be used, also so that users know they exist. There is a section of the manual on machine-specific installation: https://bout-dev.readthedocs.io/en/latest/user_docs/advanced_install.html#machine-specific-installation
The document source is in ReStructuredText here: https://github.com/boutproject/BOUT-dev/blob/master/manual/sphinx/user_docs/advanced_install.rst#machine-specific-installation

ejm930:
Thanks for letting me know about that. I'll look into adding the documentation if applicable. I'm not sure yet if these scripts can be used by others who don't have access to the spack path, so let me check on that!

@jonesholger might have some insight.

PR #2363, opened 2021-07-02
[3D Metric] Update to next

dschwoerer:
The unit test failed, thus I need to explicitly call `.geometry()` in ff1e152


dschwoerer:
I didn't notice @ZedThree did the same already :+1:

PR #2368, opened 2021-07-05
Build BOUT++ on RTD

dschwoerer:


dschwoerer:
cmake 3.10 (from ubuntu 18.04, used on rtd) is to old for findPython3 ... maybe worth to implement some fallback (naive backport failed)

.readthedocs.yaml
```diff
@@ -0,0 +1,34 @@
...
+   - requirements: manual/sphinx/requirements.txt
+
+
+build:
+  apt_packages:
+    - mpich
+    - libmpich-dev
+    - libfftw3-dev
+    - libnetcdf-dev
+    - libnetcdf-cxx-legacy-dev
```
ZedThree, line 29: Maybe prefer the cxx4 version?
```suggestion
    - libnetcdf-c++4-dev
```

ZedThree:
CMake is available through pip, so could either be installed manually or added to a `requirements.txt`?

.readthedocs.yaml
dschwoerer: I updated it in the manual as well :+1:

dschwoerer:
Increases the build time from 300..400 to over 600 secs ( 900 is the limit)

I rebased to get rid of all the debug commits.

I also changed the requirements.txt - which isn't strictly necessary - if they shouldn't be changed I can revert.

Thanks @ZedThree for the help :+1:

manual/sphinx/conf.py
```diff
@@ -36,21 +37,34 @@
...
-    ]
-    sys.modules.update((mod_name, Mock()) for mod_name in MOCK_MODULES)
+    print(os.environ)
+    print(sys.argv)
+    python = sys.argv[0]
+    pydir = "/".join(python.split("/")[:-2])
+    os.system("which clang-format")
+    os.system("which clang-format-6.0")
+    os.system(
+        "git clone https://github.com/mpark/variant.git ../../externalpackages/mpark.variant"
```
ZedThree, line 47: I'm guessing this means RTD doesn't clone the top-level repo so we can't use the submodules?

ZedThree (approved):

manual/sphinx/conf.py
dschwoerer: Yes, I forgot the exact failure but git complained it wouldnt know the url for cloning.

ZedThree:
Does this also create any documentation for boutcore? I thought it would appear here, but I don't see it: https://bout-dev--2368.org.readthedocs.build/en/2368/api_reference.html

Also, building on RTD does worry me a little bit, partly in terms of the resources used, partly I'm worried it might be fragile, but I'm not sure how else to handle this.

Another way might be to have Github Actions generate the docs and commit them as part of one of the builds? I'm not sure what they actually need to generate, and that still has fun engineering challenges.

dschwoerer:
The documentation is at:
https://bout-dev--2368.org.readthedocs.build/en/2368/user_docs/python_boutcore.html#module-boutcore

You are right, it should be moved to the API section. I noticed it broke the zoidberg docs:
https://bout-dev--2368.org.readthedocs.build/en/2368/_apidoc/zoidberg.html
https://bout-dev.readthedocs.io/en/latest/_apidoc/zoidberg.html
```
WARNING: autodoc: failed to import module 'zoidberg'; the following exception was raised:
No module named 'scipy'
```
so I need to add scipy again to the mock models. Further boututils and boutdata should be installed or removed from the docs.

That is generated from the python module, which needs to be loaded at run time. I thought about building the python module in GHA and sharing that with RTD, but that would e.g. require the jobs to be run sequantial, and we would need to have the correct libraries installed on RTD, which I thought would be more stable.

Alternatively I thought about building all the docs on GHA, and only publish them using RTD, but that might not be easier.

Concerning resources, it doesn't even quite double the execution time. I was surprised how fast that was, and we are still below the limit for the free account ...

ZedThree:
Now that I think about it, compiling all the tests probably more than doubles the compilation time, so I guess it's not too bad.

Let's go with this solution rather than trying to create something more complicated with more moving parts.

Really we need some overarching, joined up solution for the documentation for all the BOUT++ projects, but for now maybe just restore the mocking how it was.

ZedThree (approved):

ZedThree:
I'm going to merge this now, as RTD is currently failing due to the Sphinx/Breathe issues, but the boutcore docs could do with a little polish, e.g. https://bout-dev--2368.org.readthedocs.build/en/2368/_apidoc/boutcore.html#boutcore.PhysicsModel

The boutcore API docs page is also really big without much structure -- I'm not sure what can be done, but that can happen in a future PR

PR #2369, opened 2021-07-06
Fix documentation for examples/wave-slab

ZedThree:
Copy README from `next`, and add note about needing to run `generate.py` to create the grid file first.

Also apply black.


dschwoerer (approved):
Might be nice to add `./generate.py` to the `makefile`, to make the instructions easier.

PR #2371, opened 2021-07-06
Fix for compiling without FFTW

### LINKED ISSUE #2370: compling without FFTW is broken
Either with `./configure --without-fftw` or `cmake -DBOUT_USE_FFTW=OFF`
```
fft_fftw.cxx: In function ‘unsigned int bout::fft::get_measurement_flag(FFT_MEASUREMENT_FLAG)’:
fft_fftw.cxx:74:14: error: ‘FFTW_ESTIMATE’ was not declared in this scope
   74 |       return FFTW_ESTIMATE;
      |              ^~~~~~~~~~~~~
fft_fftw.cxx:76:14: error: ‘FFTW_MEASURE’ was not declared in this scope
   76 |       return FFTW_MEASURE;
      |              ^~~~~~~~~~~~
fft_fftw.cxx:78:14: error: ‘FFTW_EXHAUSTIVE’ was not declared in this scope
   78 |       return FFTW_EXHAUSTIVE;
      |              ^~~~~~~~~~~~~~~
```

ZedThree:
Guard use of FFTW constants

Fixes #2370

### DIFF
--- a/src/invert/fft_fftw.cxx
+++ b/src/invert/fft_fftw.cxx
@@ -69,18 +69,20 @@ void fft_init(Options* options) {
 }
 
 #if BOUT_HAS_FFTW
+namespace {
 unsigned int get_measurement_flag(FFT_MEASUREMENT_FLAG fft_measurement_flag) {
   switch (fft_measurement_flag) {
-    case FFT_MEASUREMENT_FLAG::estimate:
-      return FFTW_ESTIMATE;
-    case FFT_MEASUREMENT_FLAG::measure:
-      return FFTW_MEASURE;
-    case FFT_MEASUREMENT_FLAG::exhaustive:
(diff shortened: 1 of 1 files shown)


src/invert/fft_fftw.cxx
```diff
@@ -68,6 +68,7 @@ void fft_init(Options* options) {
   fft_init(fft_measurement_flag);
 }
 
+#if BOUT_HAS_FFTW
```
dschwoerer, line 71: Should we put only the function body in the `#if` to have a stabile ABI?

ZedThree: This function is only used internally, so it should really be in an anonymous namespace. Let me fix that now too.

dschwoerer (approved):
LGTM

Issue #2350, opened 2021-06-23
Bug in LaplacePCR


ZedThree:
In #2336 I've added an alternative to #2333: https://github.com/boutproject/BOUT-dev/blob/93454ecaded94667725d48c3cb9b991c23f239bc/src/invert/laplace/invert_laplace.cxx#L106-L126

Basically check that we can use PCR as the default Laplacian, otherwise fall back to cyclic. This essentially makes PCR the new default in most cases.

Unfortunately (or possibly fortunately?) this throws up a few issues with PCR in the following tests:

- test-gyro, which is a regression test, comparing the inversion to a benchmark result:
```
Running Gyro-average inversion test
   1 processors (nxpe = 1)....
      Checking variable pade1 ... Fail, maximum difference = 0.021458830684423447
      Checking variable pade2 ... Fail, maximum difference = 0.0043205489637330174
   2 processors (nxpe = 1)....
      Checking variable pade1 ... Fail, maximum difference = 0.021458830684423447
      Checking variable pade2 ... Fail, maximum difference = 0.0043205489637330174
   4 processors (nxpe = 2)....
      Checking variable pade1 ... Fail, maximum difference = 0.021458830684423447
      Checking variable pade2 ... Fail, maximum difference = 0.0043205489637330174
```
- test-petsc-laplace-MAST-grid, which applies the forward operator and inverts the result, checking round-trip accuracy, tests 4 and 8 use the default solver (i.e. PCR), all others use PETSc (errors are the same across processor count and grid):
```
      Checking max_error1 ... Pass
      Checking max_error2 ... Pass
      Checking max_error3 ... Pass
      Checking max_error4 ... Fail, maximum error is = 0.9939465685503767
      Checking max_error5 ... Pass
      Checking max_error6 ... Pass
      Checking max_error7 ... Pass
      Checking max_error8 ... Fail, maximum error is = 1.0057517623023355
```
- test-petsc-laplace, which is similar to the previous test, but the PETSc solvers use the default Laplacian as a preconditioner:
```
KSPConvergedReason is -5
BoutException occured in invert->solve(b1): petsc_laplace: inversion failed to converge.

Test 1: PETSc 2nd order
Magnitude of maximum absolute error is -1
KSPConvergedReason is -5
BoutException occured in invert->solve(b1): petsc_laplace: inversion failed to converge.

Test 2: PETSc 4th order
Magnitude of maximum absolute error is -1
KSPConvergedReason is -5
[...]

PR #2375, opened 2021-07-08
Fix INVERT_BNDARY_ONE case for PCR solver

JosephThomasParker:
This fixes cases where we use the PCR solver with boundary conditions that use the `INVERT_BNDRY_ONE` flag.

As input, the PCR solver takes a tridiagonal matrix with number of rows/columns that is a power of 2. In BOUT++ test cases, the problem domain is typically a power of 2 plus some guard cells. In the original implementation, we did a row elimination on the first and last non-guard cells so that we could decouple the 2^n-sized problem domain and pass it to the solver.

In some tests cases however, `INVERT_BNDRY_ONE` is set while also specifying that there are two guard cells. In such cases, the problem domain actually as 2^n + 2 points, so the solver fails.

This PR introduces a workaround for those cases where we do forward elimination across all lower boundary cells and backwards elimination across all upper boundary cells to decouple the middle 2^n points to give to the PCR solver.

Schematically:
**Old working case**
```
                      boundary. | interior
boundary ->               x. x. |  0  0. 0
boundary ->.              0  x. |  x. 0. 0
interior ->               0. x. |. x. x. 0
interior ->.              0. 0. |  x. x. x   
```
becomes
```
                      boundary. | interior
boundary ->               x. x. |  0  0. 0
boundary ->.              0  x. |  x. 0. 0
interior ->               0. 0. |. Y. x. 0
interior ->.              0. 0. |  x. x. x   
```
after doing forward elimination from row 1 to row 2.  The interior rows are decoupled and we can use PCR.

**Case where `mxg=2` but `INVERT_BNDRY_ONE` is set**
```
                      boundary. | interior
boundary ->               x. x. |  0  0. 0
boundary ->.              x. x. |  x. 0. 0     <---- this row has three elements - it's really an internal row
interior ->               0. x. |. x. x. 0
interior ->.              0. 0. |  x. x. x   
```
But we can reduce this to the same form as above by doing forward elimination from row 0 to row 1, and then from row 1 to row 2:
```
                      boundary. | interior
boundary ->               x. x. |  0  0. 0
boundary ->.              0  Y. |  x. 0. 0
interior ->               0. 0. |. Z. x. 0
interior ->.              0. 0. |  x. x. x   
```

### DIFF
--- a/src/invert/laplace/impls/pcr/pcr.hxx
+++ b/src/invert/laplace/impls/pcr/pcr.hxx
@@ -118,6 +118,11 @@ private:
   /// MPI process x ID
   int xproc;
 
+  /// Number of inner boundary cells
+  int inbndry;
+  /// Number of outer boundary cells
+  int outbndry;
+
   /// Local private pointer for coefficient maxtix a
   Matrix<dcomplex> a, aa;
   /// Local private pointer for coefficient maxtix b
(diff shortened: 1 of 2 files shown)


JosephThomasParker:
Oh right, addresses #2350, at least for the `test-gyro` case.

src/invert/laplace/impls/pcr/pcr.hxx
```diff
@@ -118,6 +118,11 @@ private:
   /// MPI process x ID
   int xproc;
 
+  /// Number of inner boundary cells
+  int inbndry;
+  /// Number of outer boundary cells
+  int outbndry;
```
ZedThree, line 124: Is this change necessary? Looks like they could be kept as local variables?

ZedThree (approved):
Confirmed fix for `test-gyro`, but not enough to fix the PETSc Laplace tests

ZedThree:
I guess there's still an issue with `INVERT_AC_GRAD`

PR #2384, opened 2021-07-28
Fix RTD

dschwoerer:


ZedThree:
Maybe just take the `requirements.txt` entirely from the other branch?

dschwoerer:
4.1.2 is also not working. Thus I don't think taking the requirements.txt from the other branch would do the job. `<4.1.1` would be an option, but I am worried we will forget and keep using an old version.

Probably should get a proper fix -.-

dschwoerer:
My bad, I set to 4.0.1 in the other branch. Will do the same here :+1:

ZedThree:
Ok, I will have a quick look locally. I'm hoping locally it spits out a better error message than:

```
sphinx.util.cfamily.DefinitionError: Invalid C++ declaration: Expected identifier in nested name. [error at 8]
  typename
  --------^
```

Literally not even what file it's in!

ZedThree (approved):

ZedThree:
Ok, I'm merging this. Running locally and looking at the full log file, it's actually choking on `typename...` somewhere, but still doesn't give the exact location. I'm slowly commenting them out and see if it gets any further.

ZedThree:
Looks like a bug in either doxygen, breathe, or sphinx. Something doesn't like either `typename...` or `class...` in any form. I'm giving up now

PR #2396, opened 2021-08-09
Merge v4.4.0 release into next

ZedThree:
To fix the conflicts, I basically just took all changes from `next`, leaving only a few changes that are just in v4.4.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -88,18 +88,6 @@
 
 ## v4 known bugs
 
-- `FieldPerp` I/O has a couple of issues:
-  - the `global_yindex` attribute written in output files does not account for
-    boundary cells at the second (i.e.  upper) targets in double-null
-    configuration. This will be fixed in v5 using updated (but not
-    backward-compatible) indexing methods from `Mesh`.
-  - A `FieldPerp` may be read in on processors at different (but adjacent)
-    y-locations as it may be read in the guard cells. This is not the expected
-    behaviour (a given `FieldPerp`, if it represents a physical quantity rather
-    than just being a temporary variable for an x-z slice) should only be
-    defined and used at a single, global y-position. Should not be a problem
-    for `PhysicsModel` implementations that define a `FieldPerp` correctly, but
-    caused a problem for `test-io` and `test-io_hdf5`.
 
   See [\#2154](https://github.com/boutproject/BOUT-dev/pull/2154).
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,6 +86,117 @@
   `Options::as<T>()`
   [\#2341](https://github.com/boutproject/BOUT-dev/pull/2341)
 
+## v4 known bugs
+
+- `FieldPerp` I/O has a couple of issues:
+  - the `global_yindex` attribute written in output files does not account for
+    boundary cells at the second (i.e.  upper) targets in double-null
+    configuration. This will be fixed in v5 using updated (but not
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -86,6 +86,117 @@
...
+    boundary cells at the second (i.e.  upper) targets in double-null
+    configuration. This will be fixed in v5 using updated (but not
+    backward-compatible) indexing methods from `Mesh`.
+  - A `FieldPerp` may be read in on processors at different (but adjacent)
+    y-locations as it may be read in the guard cells. This is not the expected
+    behaviour (a given `FieldPerp`, if it represents a physical quantity rather
+    than just being a temporary variable for an x-z slice) should only be
+    defined and used at a single, global y-position. Should not be a problem
+    for `PhysicsModel` implementations that define a `FieldPerp` correctly, but
+    caused a problem for `test-io` and `test-io_hdf5`.
```
johnomotani, line 102: I've forgotten what the 'not backward-compatible' thing was now, but did #2395 fix this bug for 4.4?

johnomotani (approved):
Looks good to me.

CHANGELOG.md
ZedThree: Hmm, me either! I thought it was `getGlobalYIndex`/`getGlobalYIndexNoBoundaries`, but it looks like this did get merged into v4.3 at some point, so now I'm not sure.

johnomotani: I'm confused. When I added the note in #2154, I claimed that we didn't have the updated `getGlobalYIndex()` method in the v4 branch, but it looks very much like we did, because it was added in #1770 which went into v4.3.0. I don't remember at all why I wrote that part now. Probably best to delete it (I guess from `master` and here?)
```suggestion
- A `FieldPerp` may be read in on processors at different (but adjacent)
  y-locations as it may be read in the guard cells. This is not the expected
  behaviour (a given `FieldPerp`, if it represents a physical quantity rather
  than just being a temporary variable for an x-z slice) should only be
  defined and used at a single, global y-position. Should not be a problem
  for `PhysicsModel` implementations that define a `FieldPerp` correctly, but
  caused a problem for `test-io` and `test-io_hdf5`.
```
I'm not sure if the second issue is still correct or not, probably worth checking that too...

johnomotani: yeah, looks like the second one is fixed by the `FieldPerp::getGlobalIndex()` method from #2395, so we can delete that too.
```suggestion
```

johnomotani (dismissed):

CHANGELOG.md
```diff
@@ -86,6 +86,105 @@
...
+- Faster recompile [\#2294](https://github.com/boutproject/BOUT-dev/pull/2294) ([dschwoerer](https://github.com/users/dschwoerer))
+
+## [v4.3.3](https://github.com/boutproject/BOUT-dev/tree/v4.3.2) (2021-07-28)
+
+[Full Changelog](https://github.com/boutproject/BOUT-dev/compare/v4.3.2...master)
+
+**Merged pull requests:**
+
+- Fix documentation for examples/wave-slab [\#2369](https://github.com/boutproject/BOUT-dev/pull/2369) ([ZedThree](https://github.com/users/ZedThree))
+- Make include optional [\#2360](https://github.com/boutproject/BOUT-dev/pull/2360) ([dschwoerer](https://github.com/users/dschwoerer))
```
dschwoerer, line 171: This one is missing

dschwoerer (changes requested):

PR #2413, opened 2021-09-01
Configuration on Cori with nvcc

bendudson:
Updates to enable configuration and compilation on Cori.
Selectively merging some scripts from Holger's branch
Deleting a number of unneeded / unused files.


bendudson:
Library compiling on Cori (with nvcc) but doesn't link. Error is
```
make[2]: *** No rule to make target '/tmp/build/80754af9/libnetcdf_1545349563001/_build_env/x86_64-conda_cos6-linux-gnu/sysroot/usr/lib/librt.so', 
needed by 'examples/laplacexy/simple/CMakeFiles/laplacexy-simple.dir/cmake_device_link.o'
```
Suspect NetCDF configuration issue.

PR #2422, opened 2021-09-07
Change RAJA/Umpire/CUDA defines

bendudson:
Changes `BOUT_HAS_RAJA`, `BOUT_HAS_UMPIRE` and `BOUT_USE_CUDA` to always be defined, but either 1 or 0. The effect is that `#if` should be used rather than `#ifdef`. This makes these switches the same as other similar switches in BOUT++.


ZedThree (approved):

ZedThree (changes requested):
Tests are failing for a couple of reasons:

- `configure.ac`/`autoconf_build_defines.hxx.in` also need changing to always define these macros
- `array.hxx` still has some `#ifdef`s instead of `#if`

You could (temporarily) add these new macros to `bin/bout-v5-macro-upgrader.py` and use that to fix pretty much all the locations at once

PR #2423, opened 2021-09-07
Work on elm-pb-outerloop for RAJA

bendudson:
Updates to elm-pb-outerloop so that it uses and compiles with RAJA. Work in progress

1. RAJA kernels can't be used in classes with private or protected members. This seems to be a restriction coming from CUDA. 
    Note added to the manual.

2. All variables are captured by value in RAJA loops, and are const inside the kernel. The idea is that pointers are not modified, only the data those pointers point to.
`ddt(FieldAccessor)` therefore needs to take the `FieldAccessor` as const, but return a non-const `BoutRealArray` reference.   
  Somewhat gross `const_cast` used here.


ZedThree:
That's quite unfortunate that RAJA needs everything to be public. Is it related to this SO question do you think? https://stackoverflow.com/a/58491023/2043465

Alternatively, is it a side-effect of using `[=]` and not being more explicit about the captures?

bendudson:
Don't know why private/protected members are not allowed; it seems to be some CUDA related limitation. 
Perhaps even more unfortunately, member variables are captured via `this`, and so are not valid on the device. If you want to use a class member variable in the lambda function, first copy it to a local variable outside the lambda.

PR #2424, opened 2021-09-08
Next hypre outerloop cuda merged cori fix

ejm930:
Added fixes to address:

1. cmake HAS_RAJA_FLAG issue
2. const requirements for inline raja functions taking raja kernel captured pointers
3. Illegal memory accesses in the RAJA kernels


bendudson:
I believe all these changes are now included in PR #2397

PR #2427, opened 2021-09-14
Merge Array with/without Umpire

bendudson:
When Umpire and CUDA are enabled, needs to be able to run on GPU device (at least limited functionality).

Would like to keep memory pool feature, to minimise allocation overhead.

Working on possible performance issues. 

# elm-pb
Comparing `next`, `next-hypre-outerloop-cuda-merged` and this branch `merged-array`, with default input from this branch, `nout=1` and `timestep=0.2`

## next
2.000e-01         19       5.44e+02    99.6    0.3    0.0    0.0    0.1
Run time : 9 m 37 s

## next-hypre-outerloop-cuda-merged

2.000e-01         19       3.41e+02    99.4    0.3    0.0    0.0    0.2
Run time : 6 m 1 s

## merged-array

2.000e-01         19       5.51e+02    99.6    0.3    0.0    0.0    0.1
Run time : 9 m 37 s

With CMAKE_BUILD_TYPE=RELEASE:
2.000e-01         19       1.07e+01    96.2    1.7    0.1    0.4    1.6
Run time : 11 s

[elm-pb-outerloop gives
2.000e-01         19       7.14e+00    94.4    2.4    0.1    0.7    2.5
Run time : 8 ]

Note: elm-pb changed significantly between `master` and `next`, so can't compare directly. Instead use `blob2d`.

# blob2d

## master
2.000e+01         52       1.64e+00    61.6   17.4    0.1    1.2   19.7
Run time : 2 s

## merged-array
2.000e+01         52       1.60e+02    97.1    2.1    0.0    0.0    0.8
Run time : 2 m 43 s

(100x slower??)
Both branches using check level 2 and `-g` flag. `master` compiling with `-O2`.

With -DCMAKE_BUILD_TYPE=RELEASE  (adds -O3)
2.000e+01         52       8.69e+00    91.4    4.2    0.0    0.3    4.0
Run time : 9 s

** Make this the CMAKE_BUILD_TYPE default? **

Same result when `array.hxx` is copied from `master` into the `merged-array` branch.
=> Slowdown not related to changes to `Array`.

** Why slowdown in `next`? **


bendudson:
Now tentatively declaring victory: This version of `Array` compiles and runs on Cori GPUs, and is I think a minimal modification from what is in `master`. 
- (+) Enables pools to be used, so Umpire isn't called repeatedly. If multi-threaded these use a thread-local arena.
- (-) Arrays can be passed to or used inside RAJA CUDA loops

PR #2428, opened 2021-09-16
Set CMAKE_BUILD_TYPE default to RelWithDebInfo

bendudson:
Default build type is Debug, which results in no optimisation flags. In simple examples (e.g blob2d) this runs nearly 100 times slower than when compiled with Release build type (see PR #2427). This change aims to improve the out-of-the-box performance.

Following suggestion of @d7919, using BuildType.cmake from GS2:
https://bitbucket.org/gyrokinetics/utils/src/8.1-RC/cmake/BuildType.cmake

As suggested by @ZedThree, using RelWithDebInfo as the default.
This results in compiler flags "-O2 -g" (on my laptop).


ZedThree:
@dschwoerer The Fedora builds have stopped working, it looks like the `petsc-mpich-devel` package is missing. Please could you look into this?

ZedThree (approved):

PR #2429, opened 2021-09-16
CMake examples targets

bendudson:
If BOUT_BUILD_EXAMPLES is Off, add the examples subdirectory but with the EXCLUDE_FROM_ALL option. That enables examples to be built after the library is built, and specific examples can be built by using the `--target` argument to `cmake`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -825,6 +825,8 @@ endif()
 option(BOUT_BUILD_EXAMPLES "Build the examples" OFF)
 if(BOUT_BUILD_EXAMPLES)
   add_subdirectory(examples)
+else()
+  add_subdirectory(examples EXCLUDE_FROM_ALL)
 endif()
 
 ##################################################


ZedThree:
Ah sorry, I think my earlier comment was misleading -- you can build individual examples with `--target` but you would still need to configure with `BOUT_BUILD_EXAMPLES=on`.

dschwoerer (approved):
LGTM :+1:

bendudson:
I think this may be a little more complicated than I initially thought. There are I think three scenarios to cover:
1. BOUT-dev is being included as a sub-module of a model, so we just want the library and are not interested in examples
2. We are first compiling the library, then later may want to compile one or more examples (I think this should be the default)
3. We want to build executables for all examples. I don't think this will happen often, but should be an option at least for testing if all examples can build

Previously the `BOUT_BUILD_EXAMPLES` switch chose between (1) and (3); this PR changes it to choosing between (2) and (3). 

What about:
  `BOUT_CONFIGURE_EXAMPLES` is `On` by default, and adds the examples to the cmake build. Turning it `Off` also turns `BOUT_BUILD_EXAMPLES` to `Off`. This switches between (1) and (2 or 3)
  `BOUT_BUILD_EXAMPLES` is `Off` by default. It controls whether `EXCLUDE_FROM_ALL` is used when adding the examples. It switches between (2) and (3)
?

ZedThree:
An alternative: `BOUT_BUILD_EXAMPLES` configures them with
`EXCLUDE_FROM_ALL`, and we add a target `build-all-examples`?

We do a similar thing with the tests, and this feels like perhaps the
easiest thing to do from the user point of view.

PR #2430, opened 2021-09-17
Attempt to simplify RAJA loops

bendudson:
Wraps up RAJA index setup, and eliminates #ifdefs from user code. Including:
```
#include "bout/rajalib.hxx"
```
This enables the physics model/user to write:
```
BOUT_FOR_RAJA(i, region) {
   /* ... */
};   /// <- Note extra semicolon
```
If RAJA is disabled, `BOUT_FOR_RAJA` is just `BOUT_FOR`, and hopefully the redundant semicolon is ignored by the compiler.
If RAJA is enabled, the above expands to
```
RajaForAll(region) << [=] RAJA_DEVICE(int index) {
  /* ... */
};
```
The `RajaForAll` constructor does all the work to create an `Array<int>` of indices. `RajaForAll` defines a templated `<<` operator:
```
struct RajaForAll {
  ...
  template<typename F>
  const RajaForAll& operator<<(F f) const {
```
which enables the lambda function to be passed in from the right (avoiding a closing bracket). 

Currently implemented in `examples/blob2d-outerloop`, `hasegawa-wakatani-3d` and `elm-pb-outerloop`.

### DIFF
--- a/examples/blob2d-outerloop/blob2d.cxx
+++ b/examples/blob2d-outerloop/blob2d.cxx
@@ -18,6 +18,30 @@
 #include "RAJA/RAJA.hpp" // using RAJA lib
 #endif
 
+struct RajaWrapper {
+  RajaWrapper(const Region<Ind3D>& region) {
+    auto indices = region.getIndices(); // A std::vector of Ind3D objects
+    _ob_i_ind.reallocate(indices.size());
+    // Copy indices into Array
+    for(auto i = 0; i < indices.size(); i++) {
+      _ob_i_ind[i] = indices[i].ind;
+    }
+  }
+  template<typename F>
+  void operator<<(F f) {
+    // Get the raw pointer to use on the device
+    // Note: must be a local variable
+    int* _ob_i_ind_raw = &_ob_i_ind[0];
+    RAJA::forall<EXEC_POL>(RAJA::RangeSegment(0, _ob_i_ind.size()),
+			   [=] RAJA_DEVICE(int id) {
+			     // Look up index and call user function
+			     f(_ob_i_ind_raw[id]);
+			   });
+  }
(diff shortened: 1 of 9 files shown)


examples/blob2d-outerloop/blob2d.cxx
```diff
@@ -12,10 +12,47 @@
...
+    // Get the raw pointer to use on the device
+    // Note: must be a local variable
+    int* _ob_i_ind_raw = &_ob_i_ind[0];
+    RAJA::forall<EXEC_POL>(RAJA::RangeSegment(0, _ob_i_ind.size()),
+			   [=] RAJA_DEVICE(int id) {
+			     // Look up index and call user function
+			     f(_ob_i_ind_raw[id]);
+			   });
+  }
+  // Note: This is private, but keyword not used due to CUDA limitation
```
ZedThree, line 40: I've had a (very) quick attempt at reproducing this in plain CUDA and haven't managed it. I'd quite like to get to the bottom of this before it gets into `next` -- is it a particular version of CUDA, RAJA, or some interaction with BOUT++?

bendudson: It is odd. The error is `The enclosing parent function ("rhs") for an extended __device__ lambda cannot have private or protected access within its class`. It's something to do with "extended" (device) lambdas in CUDA:
https://docs.nvidia.com/cuda/cuda-c-programming-guide/#extended-lambda
There's a comment here which linked to the docs https://github.com/NVIDIA/thrust/issues/726#issuecomment-168733162

bendudson: It seems that the class *can* have private members, it's just that the function where the lambda function is defined can't be private or protected.

PR #2431, opened 2021-09-20
Fix hypre test

jonesholger:
The main purpose of the PR is to fix Hypre Vector Unit Test  "Assemble" when linked against Hypre v2.22.0 (Cuda-Compiled)
Hypre v2.22.0 is needed while compiling with CUDA 11.1.1
CUDA 11.1.1 is needed because BOUT introduced recursive lambdas in Options processing.

change your tpl_prefix to tpl_11h to pick up this new version of Hypre

Built/Tested with RAJA off and Umpire off on lassen and Cori-GPU: we are really only exercising Hypre CUDA

Secondarily we have made some CMake changes: 

Propagate Warning Flags to the Host Compiler through -Xcompiler when compiling with nvcc

See below for recommended script changes if you have trouble with either FindnetCDF or FindPETSc


jonesholger:
Also hw3d BOUT.inp needs edits as BOUT will complain that some options are not correct.

jonesholger:
to avoid "system" provided netcdf  consider the use of BOUT_NETCDF_VERSION (this works for me on Cori)
 -DBOUT_NETCDF_VERSION=4.7.4 \

if (BOUT_NETCDF_VERSION)
  find_package(netCDF ${BOUT_NETCDF_VERSION} EXACT QUIET CONFIG)
else()
  find_package(netCDF QUIET CONFIG)
endif()

jonesholger:
PETSc find uses small code to test viability of install but this generates several HDF5 link errors such as
libpetsc.so: undefined reference to `H5Pset_dxpl_mpio'  which now require RPATHS to resolve

jonesholger:
Just a reminder to build serial tests use : cmake --build . -j  --target serial_tests

PR #2434, opened 2021-09-22
Improve comments for Mesh functions

### LINKED ISSUE #2416: FV::Div_par() incorrect on double-null grids?
In `FV::Div_par()`
https://github.com/boutproject/BOUT-dev/blob/c3ca1ce21064e05999ac530d0d7bc081cf568392/include/bout/fv_ops.hxx#L181-L183
`firstY()` and `lastY()` are used in various places to apply special handling to y-boundaries. I think this would mean that the upper divertor in a double-null would not get the special handling, which presumably it needs. Suggest replacing them with `mesh->hasBndryLowerY()` and `mesh->hasBndryUpperY()` - that might also remove the need for some of the `periodicY()` checks (as long as `hasBndry*Y()` return `false` on periodic 'boundaries' - I'm not 100% sure they do).

bendudson:
firstY, lastY, hasBndryLowerY and hasBndryUpperY had comments which could be misleading. Hopefully now clearer.

Partly fixes #2416

### DIFF
--- a/include/bout/mesh.hxx
+++ b/include/bout/mesh.hxx
@@ -468,8 +468,16 @@ class Mesh {
   // Y communications
   virtual bool firstY() const = 0; ///< Is this processor first in Y? i.e. is there a boundary at lower Y?
   virtual bool lastY() const = 0; ///< Is this processor last in Y? i.e. is there a boundary at upper Y?
-  virtual bool firstY(int xpos) const = 0; ///< Is this processor first in Y? i.e. is there a boundary at lower Y?
-  virtual bool lastY(int xpos) const = 0; ///< Is this processor last in Y? i.e. is there a boundary at upper Y?
+
+  /// Is this processor first in Y?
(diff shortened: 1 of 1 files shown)


include/bout/mesh.hxx
```diff
@@ -468,8 +468,16 @@ class Mesh {
   // Y communications
   virtual bool firstY() const = 0; ///< Is this processor first in Y? i.e. is there a boundary at lower Y?
   virtual bool lastY() const = 0; ///< Is this processor last in Y? i.e. is there a boundary at upper Y?
```
johnomotani, line 470:
```suggestion

  ///< Is this processor first in Y?
  /// Note: First on the global grid, not necessarily at a boundary
  virtual bool firstY() const = 0;

  ///< Is this processor last in Y?
  /// Note: Last on the global grid, not necessarily at a boundary
  virtual bool lastY() const = 0;
```

PR #2441, opened 2021-10-01
Some cmake fixes

dschwoerer:


CMakeLists.txt
```diff
@@ -631,6 +631,7 @@ if (BOUT_USE_PETSC)
     target_link_libraries(bout++ PUBLIC PETSc::PETSc)
     string(JOIN " " CONFIG_PETSC_LIBRARIES ${PETSC_LIBRARIES})
     set(CONFIG_LDFLAGS "${CONFIG_PETSC_LIBRARIES} ${CONFIG_LDFLAGS}")
+    set(CONFIG_CFLAGS "${CONFIG_CFLAGS} -I${PETSC_INCLUDES}")
```
ZedThree, line 634: `PETSC_INCLUDES` is actually a list, so we'll need to iterate through and pretend each item with `-

I think we do something similar elsewhere in this file, so we might want to wrap it up in a function.

dschwoerer: It is only three lines - is it worth to add a function for that?

bendudson (approved):

PR #2443, opened 2021-10-01
Workaround for finding SUNDIALS

dschwoerer:
Using the default search can find the serial version, which is hard to recover
from.  It is saver to directly search for the parallel version.

### DIFF
--- a/cmake/FindSUNDIALS.cmake
+++ b/cmake/FindSUNDIALS.cmake
@@ -31,11 +31,6 @@
 
 include(FindPackageHandleStandardArgs)
 
-find_package(SUNDIALS CONFIG QUIET)
-if (SUNDIALS_FOUND)
-  return()
-endif()
-
 find_path(SUNDIALS_INCLUDE_DIR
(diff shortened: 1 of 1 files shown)


ZedThree:
I'm not very keen on this, because I think we should prefer to use the cmake config file if possible.

Could we not instead fail with a nice error message if `SUNDIALS_FOUND AND NOT TARGET SUNDIALS::nvecparallel`? (I think that's the parallel library we require?)

dschwoerer:
I think the issue is, I am not sure on how to actually get it configured without such a work around ...

So a nice error is potentially better then what we have now, but it still doesn't fix configuring.

I do agree, it is ugly, and I would prefer something nicer ...

ZedThree:
Try setting `-DSUNDIALS_ROOT`, and if that doesn't work, message me on slack and we'll work out where it's going wrong and how to fix it.

dschwoerer:
Thanks, I think it is working now - haven't checked in a while, so not sure what went wrong before ...

PR #2444, opened 2021-10-01
Only check error if function is called

dschwoerer:
While this shouldn't change anything, it avoids a warning, and is
potentially less confusing.


ZedThree (approved):
This is why we should _always_ use braces! :)

PR #2450, opened 2021-10-05
Fixes for compilation on Mac / Darwin

bendudson:
Compiling with Apple Clang 12.0.0 on MacOS. 

Compile flags to disable components which currently don't compile and/or link on MacOS have been put in the manual:
```
cmake . -B build -DBOUT_ENABLE_BACKTRACE=Off -DBUILD_SHARED_LIBS=Off -DBOUT_USE_NLS=Off -DBOUT_USE_UUID_SYSTEM_GENERATOR=Off
cd build
make
```

Clang has found another possible bug:
```
BOUT-dev/src/mesh/interpolation_xz.cxx:39:48: warning: returning address of local temporary object
      [-Wreturn-stack-address]
const char* strLocation(CELL_LOC loc) { return toString(loc).c_str(); }
```

### DIFF
--- a/manual/sphinx/user_docs/advanced_install.rst
+++ b/manual/sphinx/user_docs/advanced_install.rst
@@ -161,7 +161,7 @@ Compiling with Apple Clang 12, the following configuration has been known to wor
 
 .. code-block:: tcsh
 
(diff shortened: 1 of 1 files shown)


ZedThree:
I think probably `strLocation` is deprecated in `next`, replaced by `toString`? #2291 should remove it, if so.

bendudson:
Thanks @ZedThree that worked! I've updated the documentation rather than modify the CMakelists.txt file

PR #2454, opened 2021-10-15
Add matrix coloring to beuler solver (and fix imexbdf2)

bendudson:
Enables large reductions in the number of iterations needed to calculate the Jacobian elements. Having the Jacobian then enables good preconditioning.

Simple conduction example, nout=5 and timestep=10:

PVODE:
```
1.000e+01        644       1.25e-01    74.9    0.0    0.7   13.4   10.9
2.000e+01        161       4.41e-02    50.9    0.0    0.4   29.4   19.2
```

beuler, SNESType = anderson
```
1.000e+01        962       2.13e-01    65.8    0.0    0.7    8.2   25.4
2.000e+01        347       8.20e-02    61.3    0.0    0.6   14.0   24.0
```

beuler, SNESType = newtonls, matrix free (no Jacobian)
```
1.000e+01       3389       5.68e-01    89.2    0.0    0.8    3.3    6.6
2.000e+01       2325       3.55e-01    89.1    0.0    0.8    3.3    6.8
```

beuler, SNESType = newtonls, with Jacobian, no coloring
```
1.000e+01        833       1.53e-01    78.8    0.0    0.7   12.1    8.4
2.000e+01        521       9.81e-02    76.6    0.0    0.8   12.3   10.4
```

beuler, SNESType = newtonls, with Jacobian coloring
```
1.000e+01         85       3.78e-02    33.3    0.0    0.3   46.5   19.9
2.000e+01         57       3.16e-02    26.4    0.0    0.2   48.9   24.6
```

Also includes some reformatting and drive-by fixes for things flagged by Apple Clang. These are in separate commits (mostly) for reviewing.

PR #2459, opened 2021-11-11
Back-port beuler/snes improvements to master

bendudson:
Enables much quicker calculation of steady-state solutions than CVODE typically can.

Backported from #2454


ZedThree:
@bendudson Is this ready for review, or do you want to wait till #2454 goes in and backport then?

johnomotani:
I think this was put up so @hahahasan could try it out. Probably best to review after #2454 goes in.

ZedThree (approved):

PR #2460, opened 2021-11-19
Introduces changes from next

dschwoerer:
ea9cec499fb206b4cd943349ff8e7f28c8c6aaac is already in next, but got probably dropped by merging.

Also I would advocate for renaming `next-hypre-outerloop-cuda-merged` to `gpu`


bendudson (approved):
Thanks @dschwoerer good spot!

PR #2465, opened 2021-11-25
Replace 'config' directory with submodule

ZedThree:
As discussed in #2397, moves all the configuration scripts out into a standalone repo and replaces them with a submodule to that repo.


ZedThree:
Very good points @johnomotani -- I agree, having the configs remain separate is probably the best thing to do.

@bendudson Any thoughts on this?

If we're happy keeping them separate, I'll remove the submodule from this PR.

bendudson:
Thanks @ZedThree and @johnomotani yes I think this is a good plan. If anything I think the BOUT-dev repo could be a submodule of the configs repository: The configs are for a particular version of BOUT++.
We briefly discussed today having multiple config repositories e.g. NERSC, Livermore,... For now perhaps directories:

BOUT-config
   - Cori/
      cori-config.sh
      ...
      BOUT-dev   <- Submodule, pointing to known good commit?
   - Perlmutter/
   - Lassen/
   - Archer/
   - Marconi/

ZedThree:
Bundling BOUT-dev in the configs repo is a good idea. That sort of layout especially makes sense if we want to go even further and make a "superproject" where we bundled everything together, something like:

```
|- BOUT-dev
|- configs
| |- archer
| |- cori
| ...
\- externalpackages
  |- netCDF
  |- HYPRE
  |- SUNDIALS
  |- PETSc
  ...
```

dschwoerer (approved):

PR #2469, opened 2021-11-29
Raise more descriptive error if converting Option fails

### LINKED ISSUE #2466: Error messages from options parsing can be cryptic
When something goes wrong parsing options, the error messages can be less than helpful, e.g.
```
terminate called after throwing an instance of 'ParseException'
  what():  Cannot find variable 'true'
```
is shown because `true` was passed to something that expected an `int`.

Could we add information or maybe catch/rethrow so that the error can include (a) the key/expression that were being parsed and (b) the expected type?

ZedThree:
Before this change, setting an integer/BoutReal to, for example, 'true', would yield:

    Cannot find variable 'true'

With this change, now the error is:

    Couldn't get BoutReal from option conduction:chi = 'true': Cannot find variable 'true'

Closes #2466

I've done this off `next`, but it can probably be backported to `master` too

### DIFF
--- a/src/sys/options.cxx
+++ b/src/sys/options.cxx
@@ -273,6 +273,27 @@ template <> std::string Options::as<std::string>(const std::string& UNUSED(simil
   return result;
 }
 
+namespace {
+/// Use FieldFactory to evaluate expression
+double parseExpression(const Options::ValueType& value, const Options* options,
+                       const std::string& type, const std::string& full_name) {
+  try {
+    // Parse the string, giving this Option pointer for the context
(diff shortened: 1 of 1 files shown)


johnomotani (approved):
Nice! Looks good to me.

Is it worth using `parseExpression()` (or similar) to make error messages for Field3D/Field2D/FieldPerp nicer too? At the moment I think we just get the error from `FieldFactory::get()->create3D()`, etc.

bendudson (approved):

Issue #2470, opened 2021-11-30
Performance regression in next


ZedThree:
I noticed some quite alarming performance regressions between master and next, around a factor 5 slow down. I've looked at the conduction and blob2d examples. Built with CMake `-DCMAKE_BUILD_TYPE=Release -DCHECK=0`.

- Conduction: 2.2s (master) vs 9.7s (next)
- blob2d (delta_1): 52s (master) vs 287s (next)

Same number of iterations, but rhs massively slower. Probably going to need to bisect to find out when this happened.

If other people could verify this too, that would be useful.

bshanahan:
Can confirm for blob2d, but not for conduction which runs the same (same RHS, sometimes slightly slower, sometimes slightly faster depending on the timstep).

dschwoerer:
Might be related that the hermes-2 model run has 99.8 % of the double instructions performed done as scalar rather then SIMD?

d7919:
> Might be related that the hermes-2 model run has 99.8 % of the double instructions performed done as scalar rather then SIMD?

That's an interesting observation. The BOUT_FOR loops are meant to help things vectorise, so if this isn't happening that's a good place to start looking.

d7919:
For blob2d it looks like the difference is coming from `Delp2`

d7919:
OK I think I've tracked this down. Delp2 uses `Laplacian::tridagCoefs` with the overload in which we don't specify `kwave`. In master we have `kwave = jz * 2.0 * PI / coords->zlength();` whilst in next we have `kwave = jz * 2.0 * PI / coords->zlength()(jx,jy);`. It seems that indexing into zlength has a significant penalty here. I've confirmed that replacing this in next with the value of zlength from master I recover the performance of master.

d7919:
Zlength in next returns a Field2D, although without the 3D metrics the calculation is scalar and implicit cast/construction from scalar to Field2D must be happening. Maybe this is somewhat slow?

d7919:
Sorry I forgot that dz is now a Field2D/3D in next, so this is always do `Field2D * BoutReal` rather than a conversion.

d7919:
Hacking in the following recovers the performance of master on next (only implemented for the 2d metric branch).

~~~c++
  FieldMetric dx, dy, dz; ///< Mesh spacing in x, y and z                                                                                                                                     
  Field2D zlength_cache;
  Field2D zlength() {
    static bool visited = false;
[...]

PR #2471, opened 2021-12-06
Cache `Coordinates::zlength()` result

ZedThree:
At least a partial fix for #2470

We could use a `Field2D` directly for `zlength_cache` and check `zlength_cache.isAllocated()`, except that we don't expose a method for de-allocating a field's data array, and so can't (easily) invalidate the cache. Hence, using a `unique_ptr<Field2D>`.

I think the only thing to be aware of now is if you change `dz` and don't call `Coordinates::geometry` then `zlength` won't be updated.

### DIFF
--- a/src/mesh/coordinates.cxx
+++ b/src/mesh/coordinates.cxx
@@ -978,6 +978,20 @@ void Coordinates::outputVars(Datafile& file) {
   getParallelTransform().outputVars(file);
 }
 
+Field2D Coordinates::zlength() const {
+  if (not zlength_cache) {
+    zlength_cache = std::make_unique<Field2D>(0., localmesh);
+
+#if BOUT_USE_METRIC_3D
+    BOUT_FOR_SERIAL(i, dz.getRegion("RGN_ALL")) { zlength_cache[i] += dz[i]; }
+#else
+    (*zlength_cache) = dz * nz;
+#endif
+  }
+
+  return *zlength_cache;
+}
+
 int Coordinates::geometry(bool recalculate_staggered,
     bool force_interpolate_from_centre) {
   TRACE("Coordinates::geometry");
@@ -1259,6 +1273,10 @@ int Coordinates::geometry(bool recalculate_staggered,
     // Re-calculate interpolated Coordinates at staggered locations
     localmesh->recalculateStaggeredCoordinates();
   }
+
+  // Invalidate and recalculate cached variables
+  zlength_cache.reset();
+  zlength();
 
   return 0;
 }
--- a/src/mesh/coordinates.cxx
+++ b/src/mesh/coordinates.cxx
@@ -983,7 +983,7 @@ Field2D Coordinates::zlength() const {
     zlength_cache = std::make_unique<Field2D>(0., localmesh);
 
 #if BOUT_USE_METRIC_3D
-    BOUT_FOR_SERIAL(i, dz.getRegion("RGN_ALL")) { zlength_cache[i] += dz[i]; }
+    BOUT_FOR_SERIAL(i, dz.getRegion("RGN_ALL")) { (*zlength_cache)[i] += dz[i]; }
 #else
     (*zlength_cache) = dz * nz;
 #endif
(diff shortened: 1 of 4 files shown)


ZedThree:
Returning `const Field2D&` from `zlength` now seems to completely recover the performance of v4.4

src/mesh/coordinates.cxx
```diff
@@ -1260,6 +1274,10 @@ int Coordinates::geometry(bool recalculate_staggered,
     localmesh->recalculateStaggeredCoordinates();
   }
 
+  // Invalidate and recalculate cached variables
+  zlength_cache.reset();
+  zlength();
```
bendudson, line 1279: Do we need to recalculate here? Not critical, but perhaps it's not needed in some models?

ZedThree: Fair point. It is recalculated on demand, so we can drop it here. Will do that shortly!

ZedThree:
Looks like a potential race condition with the OpenMP build. I guess calculating `zlength` needs to be in a `BOUT_OMP(critical)` region?

d7919:
> Looks like a potential race condition with the OpenMP build. I guess calculating `zlength` needs to be in a `BOUT_OMP(critical)` region?

I guess when you were recalculating in geometry we didn't hit this as we'd always have the cached version available in our OMP regions later? I suspect it might make sense to pull the zlength call out of the OMP loops (presumably in the Lapace methods) as a minor optimisation anyway.

ZedThree:
> I guess when you were recalculating in geometry we didn't hit this as we'd always have the cached version available in our OMP regions later? 

Yes, I suspect so.

> I suspect it might make sense to pull the zlength call out of the OMP loops (presumably in the Lapace methods) as a minor optimisation anyway.

I'm not sure that will have much of an effect since we're now returning a reference, so `zlength()` should be pretty cheap (after it's been cached). It wouldn't hurt though.

Although, I think it's probably wise to make the check in `zlength` a `critical` region (so that it will always work correctly) and as we'd be hitting that on each iteration, that could be a significant hit.

src/mesh/coordinates.cxx
```diff
@@ -978,6 +978,21 @@ void Coordinates::outputVars(Datafile& file) {
...
+  BOUT_OMP(critical)
+  if (not zlength_cache) {
+    zlength_cache = std::make_unique<Field2D>(0., localmesh);
+
+#if BOUT_USE_METRIC_3D
+    BOUT_FOR_SERIAL(i, dz.getRegion("RGN_ALL")) { (*zlength_cache)[i] += dz[i]; }
+#else
+    (*zlength_cache) = dz * nz;
+#endif
+  }
```
dschwoerer, line 991:
```suggestion
  if (not zlength_cache) {
    BOUT_OMP(critical)
    if (not zlength_cache) {
      auto tmp = std::make_unique<Field2D>(0., localmesh);

#if BOUT_USE_METRIC_3D
    BOUT_FOR_SERIAL(i, dz.getRegion("RGN_ALL")) { (*tmp)[i] += dz[i]; }
#else
     (*tmp) = dz * nz;
#endif
      zlength_cache = tmp;
    }
  }
```

I haven't tested it, but I think something like that should avoid the locking in most cases - as the final update is (mostly?) atomic ...
However, not sure it is worth it - as it should be easy enough to pull out of OMP loops ...

d7919: Could `Delp2` be changed to use the tridag interface which takes `kwave` directly?

dschwoerer: `tmp` is needed so `zlength_cache` is immediately valid, and not first set to `0`.

ZedThree: @dschwoerer Ah, I see, ok

@d7919 In principle, but currently the `kwave` overload is `protected` in `Laplacian` and not public.

Looking further, `Delp2` doesn't use OpenMP, so this wouldn't affect it anyway.

ZedThree: I'm of the opinion this change probably isn't worth it, as all of the potentially affected library code is now fixed or not using OpenMP, and I suspect most/all user code that uses `zlength` won't be using OpenMP

bendudson (approved):

PR #2473, opened 2021-12-07
Print correct CMake build directory in configure message

### LINKED ISSUE #2472: CMake stdout build instructions bug?
If I configure with, for example,
```
cmake . -B build -DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_COMPILER=CC -DBOUT_DOWNLOAD_SUNDIALS=ON
```
(i.e., as suggested by the manual, without `cd`'ing to `build`), I get the instruction printed at the end of a successful configuration
```
*** Now run `cmake --build .` to compile BOUT++ ***

-- Configuring done
-- Generating done
-- Build files have been written to: ...
```
but `cmake --build .` will not work - I need `cmake --build build`.

I don't know how to make the suggestion always correct (I guess it would at least need to detect whether you'd done `cmake . -B build` or `mkdir build; cd build; cmake ..`, and the directory might be called something other than `build`...), but it's not great that the current instruction can be incorrect. One quick fix might be to just link back to the docs (https://bout-dev.readthedocs.io/en/latest/user_docs/installing.html#cmake) instead of giving a specific command?

ZedThree:
Fixes #2472 

Backported from next

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -973,5 +973,5 @@ message("
 
        export PYTHONPATH=${BOUT_PYTHONPATH}:\$PYTHONPATH
 
-*** Now run `cmake --build .` to compile BOUT++ ***
+*** Now run `cmake --build ${CMAKE_BINARY_DIR}` to compile BOUT++ ***
 ")


johnomotani (approved):
Works for me. Thanks @ZedThree!

PR #2474, opened 2021-12-07
Raise more descriptive error if converting Option fails

### LINKED ISSUE #2466: Error messages from options parsing can be cryptic

ZedThree:
Before this change, setting an integer/BoutReal to, for example, 'true', would yield:

    Cannot find variable 'true'

With this change, now the error is:

    Couldn't get BoutReal from option conduction:chi = 'true':  Cannot find variable 'true'

Backported from #2469 

Fixes #2466

### DIFF
--- a/src/sys/options.cxx
+++ b/src/sys/options.cxx
@@ -209,6 +209,27 @@ template <> std::string Options::as<std::string>(const std::string& UNUSED(simil
   return result;
 }
 
+namespace {
+/// Use FieldFactory to evaluate expression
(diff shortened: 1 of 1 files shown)


johnomotani (approved):
:+1:

PR #2475, opened 2021-12-08
Update "Running BOUT++" documentation

### LINKED ISSUE #2467: Misleading documentation about `archive`/restart files
Reported from slack:

> According to the user doc https://bout-dev.readthedocs.io/en/latest/user_docs/running_bout.html "Archived restart files have names like “BOUT.restart_0020.#.nc”, and are written every 20 outputs by default. To change this, set “archive” in the BOUT.inp file."

And reply from @johnomotani:

> It looks like the manual is out of date (sorry about that!). The archive option was removed in this commit e35f1f761ffa98911fd3efbf398ee4e00dcd4166.
I'd recommend using xBOUT (https://github.com/boutproject/xBOUT) to create restart files from the output data. There is a to_restart() method for this https://xbout.readthedocs.io/en/latest/xbout.html#xbout.boutdataset.BoutDatasetAccessor.to_restart. Not sure if it's obvious how to use, so please feel free to message again! The xBOUT docs are a bit sparse - this is making me think that we could use an FAQ section to include common recipes like 'how to make restart files'...

ZedThree:
Quite a bit of the manual section on running BOUT++ was out of date. There's certainly still lots more we should rewrite, but this fixes some glaring errors at least.

Fixes #2467

### DIFF
--- a/manual/sphinx/user_docs/running_bout.rst
+++ b/manual/sphinx/user_docs/running_bout.rst
@@ -82,7 +82,7 @@ run, and produce a bunch of files in the ``data/`` subdirectory.
    if needed. In some cases the options used have documentation, with a brief
    explanation of how they are used. In most cases the type the option is used
    as (e.g. ``int``, ``BoutReal`` or ``bool``) is given.
-   
+
 -  ``BOUT.restart.*.nc`` are the restart files for the last time point.
    Currently each processor saves its own state in a separate file, but
    there is experimental support for parallel I/O. For the settings, see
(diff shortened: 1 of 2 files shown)


manual/sphinx/user_docs/running_bout.rst
```diff
@@ -301,153 +245,149 @@ over this see if there are any important warnings or errors. Each
...
-            First       :  Fourth order central (C4)
-            Second      :  Fourth order central (C4)
-            Upwind      :  Third order WENO (W3)
-            Flux        :  Split into upwind and central (SPLIT)
-    Setting Z differencing methods
-            First       :  FFT (FFT)
-            Second      :  FFT (FFT)
-            Upwind      :  Third order WENO (W3)
-            Flux        :  Split into upwind and central (SPLIT)
-
```
ZedThree, line 375: It is a bit of a shame to have lost this output. Should we reinstate it?

johnomotani: :+1: I agree would be nice to reinstate it.

manual/sphinx/user_docs/running_bout.rst
```diff
@@ -301,153 +245,149 @@ over this see if there are any important warnings or errors. Each
...
+
+This will always be processor number ’0’ on screen as only the output
+from processor ’0’ is sent to the terminal.
+
+The process ID (pid) is also printed::
+
+    pid: 17835
+
+which is useful for distinguishing multiple simulations running at the
+same time and, for example, stop one run if it starts misbehaving.
```
johnomotani, line 279:
```suggestion
same time and, for example, to stop one run if it starts misbehaving.
```

johnomotani (approved):
This is a definite improvement! :+1:

johnomotani (approved):

PR #2481, opened 2022-01-06
Simplify specifying factory creator arguments

ZedThree:
Pulled out of #2336 as it's pretty orthogonal to the rest of those changes.

By moving the `Register[Unavailable]InFactory` types into the `Factory` itself, and then specifying the type's arguments as a template parameter pack, we can use the parameter pack directly in the the `Register` type. This cuts down on a lot of boilerplate for writing the factories, and they now all become type aliases instead of full definitions.

Technically a breaking change, but I very much doubt any user code has their own factories.

### DIFF
--- a/include/bout/generic_factory.hxx
+++ b/include/bout/generic_factory.hxx
@@ -45,7 +45,7 @@
 ///       static constexpr auto default_type = "derived_type";
 ///     };
 ///
-///     RegisterInFactory<Base, Derived, MyFactory> register("derived_type");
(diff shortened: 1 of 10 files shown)


johnomotani (approved):
Looks good to me!

ZedThree:
Readthedocs is currently failing because (for some reason?) it pulls in the latest `fmt` which has disabled implicit conversion of `enum class` to `int`. PR fixing this incoming

PR #2483, opened 2022-01-06
Command-line help for input options for factory created types

ZedThree:
Replaces #2336 with a _much_ simpler implementation. It does a very similar thing: pass an empty `Options` to a type to see what options it expects, and the associated default values and docstrings, and then print all that information.

#2336 tries to avoid actually constructing types, using a companion template class that just reads `Options`, which has some distinct advantages: we only need an `Options` to construct this companion class, so we don't need to set anything else up; we don't need to worry about any preconditions; we don't do any unnecessary work.

However, it does also have some glaring downsides: it requires either lots of duplication or ugly indirection; inheritance is painful to deal with; if default values depended on e.g. `Mesh` values, we have to be able to defer setting the real default value somehow.

So what we do here is just create a real object instead. Much simpler.

We need to make sure all the factories can take the explicit name of the type we want plus an `Options`, but this is not very much code. We also need to set up a global `Mesh`, but making a very small one is pretty cheap.

I was worried that there were some types that can only be constructed with >2 processors, but that doesn't seem to be the case.

Also, all the `Solver`s set most of their options in their `init` functions, which makes them a little bit pointless:

```
$  ./blob2d --help-solver arkode
Input options for Solver 'arkode':

solver:is_nonsplit_model_diffusive = true               # type: bool, doc: If not a split operator, treat RHS as diffusive?
solver:mms = false              # type: bool
solver:mms_initialise = false           # type: bool
solver:monitor_timestep = false         # type: bool
```

If we're happy with the direction of this PR, I'll change types to read the `Options` in their constructors instead.

### DIFF
--- a/src/bout++.cxx
+++ b/src/bout++.cxx
@@ -379,7 +379,7 @@ auto parseCommandLineArgs(int argc, char** argv) -> CommandLineArgs {
             "  --list-laplacians\t\tList the available Laplacian inversion solvers\n"
             "  --help-laplacian <laplacian>\tPrint help for the given Laplacian inversion solver\n"
             "  --list-laplacexz\t\tList the available LaplaceXZ inversion solvers\n"
-            "  --help-laplacex <laplacex>\tPrint help for the given LaplaceXZ inversion solver\n"
+            "  --help-laplacexz <laplacexz>\tPrint help for the given LaplaceXZ inversion solver\n"
             "  --list-invertpars\t\tList the available InvertPar solvers\n"
             "  --help-invertpar <invertpar>\tPrint help for the given InvertPar solver\n"
             "  --list-rkschemes\t\tList the available Runge-Kutta schemes\n"
--- a/src/bout++.cxx
+++ b/src/bout++.cxx
@@ -261,17 +261,92 @@ void setupGetText() {
 #endif // BOUT_HAS_GETTEXT
 }
 
-template <class T>
-void printAvailableImplementations(const T& factory) {
+/// Print all of the available implementations for a given `Factory`
+template <class Factory>
+[[noreturn]] void printAvailableImplementations() {
+  const auto factory = Factory::getInstance();
+
   for (const auto& implementation : factory.listAvailable()) {
     std::cout << implementation << "\n";
   }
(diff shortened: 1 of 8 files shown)


src/bout++.cxx
```diff
@@ -300,13 +375,21 @@ auto parseCommandLineArgs(int argc, char** argv) -> CommandLineArgs {
       output.write(
           _("  --print-config\t\tPrint the compile-time configuration\n"
             "  --list-solvers\t\tList the available time solvers\n"
+            "  --help-solver <solver>\tPrint help for the given time solver\n"
             "  --list-laplacians\t\tList the available Laplacian inversion solvers\n"
+            "  --help-laplacian <laplacian>\tPrint help for the given Laplacian inversion solver\n"
             "  --list-laplacexz\t\tList the available LaplaceXZ inversion solvers\n"
+            "  --help-laplacex <laplacex>\tPrint help for the given LaplaceXZ inversion solver\n"
```
johnomotani, line 382:
```suggestion
            "  --help-laplacexz <laplacexz>\tPrint help for the given LaplaceXZ inversion solver\n"
```

johnomotani (approved):
Neat!

bendudson:
Thanks @ZedThree (and happy New Year!). This looks like a great way to get this functionality in a simple way without lots of code changes. 

The two-stage construction of the solvers is a bit odd, and moving everything into the constructor would be good. I think `init` was needed to resolve some kind of circular dependency when adding PETSc support around 2013, but perhaps there is a way to remove it now.

johnomotani (approved):
:+1:

ZedThree:
Happy new year to you too @bendudson @johnomotani !

I'll have a look at squashing the solver `init` functions into the ctors in a separate PR. Might look at doing so for `Mesh` too? I'm less sure that will work though.

The other QoL missing from this PR is that the arguments must be separated with a space: it would be nice to split on equals too, but then we need to do more parsing of the command line, and at that point I almost want to find a library.
I'll investigate this more if people make noise about it.

ZedThree:
(Also, this PR is a factor 10 smaller than #2336 😶)

PR #2484, opened 2022-01-07
Move `Solver` input options into constructor

ZedThree:
Most of these changes are from moving input options from the various `Solver` implementation `init` methods into their respective constructors. Also now all of the solver inputs have docstrings. This makes #2483 much more useful for the solvers.

Some of those changes are a little bit larger than expected because the options were being read into local variables, and now they (usually) need to become members.

Other changes:

- While I was touching all this, I discovered that `Solver::max_dt` isn't used anywhere, so I removed it. Some solvers do have a maximum timestep, but they all use their own variables
- `nout`, `timestep` were stored by most of the solvers, so now we store them in the base class and provide getters
- With the `nout`, `timestep` now setter by the constructor, we can remove them entirely from the `init` signature
- The SLEPc solver does require a setter for `nout` when it's used without a separate advance solver

I wanted to also make `Solver::init` and `Solver::run` protected, as there's no reason for users to call them directly, but the SLEPc solver does make use of them, which is a bit of a shame.

---

Some things I haven't changed, but need thinking about:

- `PetscSolver` has two options, `interpolate` and `output_name`, that are set by `PetscOptionsGetBool/String` rather than our `Options`: should I change them to be normal options here?
- Some solvers have defaults for input options that are based on the number of fields (e.g. `mukeep` in the SUNDIALS solvers), which means this option can't really be set until we know all the fields, and so isn't set in the constructor and is therefore missing from the command line help :disappointed: 
- Some solvers have options that are set from the fields' sections (e.g. `use_vector_abstol` in SUNDIALS solvers adds a `atol` option to each field), which are not easily tracked
- Some solvers have "gated" options that are only used depending on the value of other options. I've moved most of these to the constructors, so they are now always possible to set -- but maybe we should note the dependencies in the docstring?

I'd really like to be able to completely remove the `init` method, but a separate method is needed by most (all?) solvers in order to allocate memory based on the total number of evolving fields.

One potential solution would be to hold off on constructing the solver until after the physics model, and use `SOLVE_FOR` to add fields to an intermediate container which is then passed wholesale to the solver constructor. We'd then have all the information at once, and the allocation etc could be done in the constructor, and we'd completely avoid the two-step construction.


bendudson:
Can we just combine the `init` functionality into `run`? It needs to be called before run, so can't we just call `init` the first time `run` is called? I think there's already a flag recording whether `init` has been run or not.

ZedThree:
We actually only call `run` once anyway, so that should work, yeah. Do you want me to make that change in this PR?

It doesn't get round the issue of some `options` not being able to be set (or rather have their default values set) until we've finished adding evolving variables, but that's a bigger refactor.

bendudson:
Hey @ZedThree Probably no need to make the change in this PR; we can think about refactoring later. I'll try to go through these changes and get this merged in first.

bendudson (approved):

PR #2485, opened 2022-01-12
Fix ambiguous variant use

bendudson:
Should use `bout::utils::visit`, rather than `std::visit`.

Without this fix an compiler error occurs with GCC 11.1.0


ZedThree (approved):

ZedThree:
Ah, fixed in master but not in next!

PR #2492, opened 2022-01-28
Fix bug in PetscLib that broke no-PETSc builds

johnomotani:
The no-PETSc branch needs to forward-declare `SNES` to avoid error.

Thanks to @bshanahan for finding the bug.

Are we missing a job with no PETSc in the CI testing? I'd have thought that this should have been caught automatically... If anyone wants to add on to this PR to tweak the CI, I'm not planning to, so feel free!

### DIFF
--- a/include/bout/petsclib.hxx
+++ b/include/bout/petsclib.hxx
@@ -126,9 +126,10 @@ private:
 
 #include "unused.hxx"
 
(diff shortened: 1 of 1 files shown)


ZedThree (approved):

ZedThree:
Thanks @johnomotani and @bshanahan !

We should probably remove the `configure_options` from the `Default options` job

PR #2494, opened 2022-02-02
CMake: option to ignore a conda environment; fix for PVODE

johnomotani:
Some systems have peculiar Python setups, so it can be easier to use conda to provide a totally self-contained Python install. However, conda installs its own versions of many tools and libraries. Sometimes it is desirable to be able to use conda, but ignore all conda-installed stuff when configuring BOUT++ or running BOUT++ executables. This PR provides some fixes so this is possible:
* ` BOUT_IGNORE_CONDA_ENV=ON` makes CMake ignore anything under the `$CONDA_PREFIX` when searching. Implelementation is hack-y because CMake does not make this easy at the moment (see comment in `CMakeLists.txt`). ea66a5e
* `SYMLINK_SYSTEM_UUID=ON` creates a symlink to the system `libuuid` under the BOUT++ build directory. This enables us to link to this version explicitly, instead of relying on `LD_LIBRARY_PATH` at run-time, which can lead to conflicts with a conda-installed version. See message db70e58 for more details.
   * also 4a7ba90 so external projects pick up the same version.

Other small fixes:
* 1e8fa51 If `NC_CONFIG` and/or `NCXX4_CONFIG` were set explicitly, always use them to find netCDF/netCDF-cxx4 instead of searching with `find_package()`. Previously (possibly just in external packages?), a system- or conda-installed netCDF could be picked up even when `NC_CONFIG` was passed explicitly.
* 9069728 make sure that paths to netCDF, not just netCDF-cxx4, are propagated to external packages.
* ceb33fe fix PVODE variables for external packages. When PVODE is built during BOUT++ configuration, only `PVODE_BINARY_DIR` and `PVODE_SOURCE_DIR` are set, not `PVODE_DIR`. It is necessary to export `PVODE_DIR` for external packages to find PVODE correctly. Previously if multiple versions of BOUT-dev were built (e.g. with different compilers, MPI, etc.), the automatic CMake search for PVODE could pick up a version from the wrong build, leading to linker errors.
* 11269c8 If a relative path is passed as `FFTW_ROOT` on the command line, convert it to an absolute path. This seems to happen for other variables automatically (at least the others that I was using).


johnomotani:
@ZedThree This set of updates gives me (touch wood) a workable CMake config for my Marconi setup. If/when everything seems OK, I propose to make another PR merging these changes into `next` and adding a paragraph to the docs about compiling BOUT++ while using `conda` - seems easier to add to the docs in `next` since there are some CMake-related changes to docs between `master` and `next`, and the `next` ones are the ones that appear online.

ZedThree (approved):

ZedThree:
An alternative to `SYMLINK_SYSTEM_UUID` might be to set `BOUT_USE_UUID_SYSTEM_GENERATOR=off`. Does that work for you @johnomotani ?

johnomotani:
> An alternative to SYMLINK_SYSTEM_UUID might be to set BOUT_USE_UUID_SYSTEM_GENERATOR=off. Does that work for you @johnomotani ?

Yes, that does work. I kept going with the symlink thing (i) because the system library has better random number generation (? if I remember right?) and (ii) as a prototype in case it's useful to do something similar for other libraries (e.g. system-installed netcdf) to get around conda conflicts. Actually, now might be a good time for me to go back to the 'generic' config and try to make that play nicely with conda...

johnomotani:
> A lot of stuff has gone into master recently, so I want to get another release out this month. Maybe next week even? At which point, this will get into next.

Cool, sounds good! :+1:

> Perhaps we should really set the default docs to be master?

Sounds like a good idea to me. People using `next` should be savvy enough to toggle the docs from master->next (or bookmark the page for next)! Could discuss at the BUG next week...

johnomotani:
...anticipating that we do as the discussion above suggests, I'll update the docs on this branch.

johnomotani:
Just setting `-DBOUT_IGNORE_CONDA_ENV=ON` for a setup building on https://github.com/boutproject/BOUT-configs/tree/generic-4.4 (on my laptop) seems to prevent all the errors and warnings that used to appear when configuring/compiling with an active conda environment, even though on my laptop quite a few libraries (netcdf, lapack, etc.) are installed under /usr. Not sure why this is OK, when libuuid wasn't before, possibly because `conda` seems not to be adding anything to `LD_LIBRARY_PATH` on my laptop (don't know why that is, maybe because I'm using the latest Python-3.9 version, while Marconi was a slightly older one??).

Anyway, the updates in here seem to be sufficient for a 'generic' install alongside conda on at least one system, so I think this PR can be merged, if @ZedThree's happy with my manual entry! We can iterate more if people report problems...

johnomotani:
I implemented `SYMLINK_SYSTEM_UUID` before `BOUT_IGNORE_CONDA_ENV`. Checking again now, it seems that setting `BOUT_IGNORE_CONDA_ENV=ON` fixes the warning about linking the wrong version at run-time that made me implement `SYMLINK_SYSTEM_UUID` in the first place. `SYMLINK_SYSTEM_UUID` is a not-very-nice hack too, so I'll revert it.

ZedThree (approved):

ZedThree:
Lovely, thanks @johnomotani !

PR #2495, opened 2022-02-10
Outerloop suggestions

ZedThree:
There's one breaking change in here: CMake will now look for `HYPRE_ROOT` instead of `HYPRE_DIR/HYPRE_SEARCH_DIR`, as `*_ROOT` is more idiomatic (and more importantly, built-in) CMake.
That probably means some of the config scripts will need updating.

This fixes various of @johnomotani, mine, and clang-tidy's comments on #2397, and manages to shave a few hundred lines off the PR.

There's still a few things I've not fixed. The hypre interface still uses `malloc` and I would very much like to get rid of that if possible. `next` needs to be merged in and conflicts fixed.

### DIFF
--- a/include/bout/hypre_interface.hxx
+++ b/include/bout/hypre_interface.hxx
@@ -855,7 +855,7 @@ public:
       break;
     }
     default: {
-      throw BoutException("Unsupported hypre_solver_type {}", solver_type);
+      throw BoutException("Unsupported hypre_solver_type {}", toString(solver_type));
     }
     }
 
--- a/src/invert/laplace/invert_laplace.cxx
+++ b/src/invert/laplace/invert_laplace.cxx
@@ -760,9 +760,7 @@ void laplace_tridag_coefs(int jx, int jy, int jz, dcomplex &a, dcomplex &b, dcom
                           const Field2D *ccoef, const Field2D *d, CELL_LOC loc) {
   Laplacian::defaultInstance()->tridagCoefs(jx,jy, jz, a, b, c, ccoef, d, loc);
 }
-#if !defined(__NVCC__)  // until we reconcile multiple definition errors under nvcc
 constexpr decltype(LaplaceFactory::type_name) LaplaceFactory::type_name;
 constexpr decltype(LaplaceFactory::section_name) LaplaceFactory::section_name;
 constexpr decltype(LaplaceFactory::option_name) LaplaceFactory::option_name;
 constexpr decltype(LaplaceFactory::default_type) LaplaceFactory::default_type;
-#endif
--- a/src/invert/parderiv/invert_parderiv.cxx
+++ b/src/invert/parderiv/invert_parderiv.cxx
@@ -39,9 +39,7 @@ const Field2D InvertPar::solve(const Field2D &f) {
 
 // DO NOT REMOVE: ensures linker keeps all symbols in this TU
 void InvertParFactory::ensureRegistered() {}
-#if !defined(__NVCC__) // until we reconcile multiple definition errors under nvcc
 constexpr decltype(InvertParFactory::type_name) InvertParFactory::type_name;
 constexpr decltype(InvertParFactory::section_name) InvertParFactory::section_name;
 constexpr decltype(InvertParFactory::option_name) InvertParFactory::option_name;
 constexpr decltype(InvertParFactory::default_type) InvertParFactory::default_type;
-#endif


ZedThree:
Hmm, merging in `next` brings in a black formatting commit, so the merge commit has `[skip ci]` in it, and I can't see how to trigger the CI manually

ZedThree:
I've reverted the `next` merge, as it was too much noise, and we can do that directly in #2397

ZedThree:
I think RTD is failing due to something we've fixed in `next`

cmake/FindnetCDF.cmake
```diff
@@ -28,11 +28,7 @@
 include(BOUT++functions)
 include(CMakePrintHelpers)
 
-if (BOUT_NETCDF_VERSION)
```
bendudson, line 31: It looks like this might be a case where someone wants to specify a particular version. Is there a way to achieve this without the EXACT flag to find_package?

examples/CMakeLists.txt
```diff
@@ -38,9 +38,7 @@ add_subdirectory(lapd-drift)
 add_subdirectory(laplacexy/alfven-wave)
 add_subdirectory(laplacexy/laplace_perp)
 add_subdirectory(laplacexy/simple)
-if(BOUT_HAS_HYPRE)
-  add_subdirectory(laplacexy/simple-hypre)
-endif ()
+add_subdirectory(laplacexy/simple-hypre)
```
bendudson, line 41: Will this example build without Hypre?

examples/blob2d-outerloop/CMakeLists.txt
```diff
@@ -6,9 +6,7 @@ if (NOT TARGET bout++::bout++)
   find_package(bout++ REQUIRED)
 endif()
 
-bout_add_example(blob2d-outerloop
-  SOURCES blob2d.cxx
-  EXTRA_FILES data/BOUT.inp)
+bout_add_example(blob2d-outerloop SOURCES blob2d.cxx)
```
bendudson, line 9: data/BOUT.inp not needed? Thought it should be there so it's copied into the build directory

examples/CMakeLists.txt
bendudson: Ah I see it's replaced by a `REQUIRES BOUT_HAS_HYPRE` clause to `bout_add_example`

bendudson:
These changes look great, thanks @ZedThree !
I'll have to try it out on a few different systems here and see if CMake still finds everything ok.

examples/blob2d-outerloop/CMakeLists.txt
ZedThree: `bout_add_example` copies it by default, along with `README.md` if that exists :)

examples/CMakeLists.txt
ZedThree: Yes, exactly, so the example gets skipped if Hypre isn't present

cmake/FindnetCDF.cmake
ZedThree: Hmm, I thought that `netCDF_FIND_VERSION` would basically do this out of the box, but it seems it might not work when it looks for `CONFIG` files.

The other, out of the box, method is to to use `netCDF_ROOT=/path/to/specific/version`

dschwoerer:
Not sure why this only fails on RTD, and not GHA:
```
.../externalpackages/fmt/include/fmt/core.h: In instantiation of ‘constexpr fmt::v8::detail::value<Context> fmt::v8::detail::make_arg(T&&) [with bool IS_PACKED = true; Context = fmt::v8::basic_format_context<fmt::v8::appender, char>; fmt::v8::detail::type <anonymous> = (fmt::v8::detail::type)15; T = const ZDirectionType&; typename std::enable_if<IS_PACKED, int>::type <anonymous> = 0]’:
.../externalpackages/fmt/include/fmt/core.h:1834:77:   required from ‘constexpr fmt::v8::format_arg_store<Context, Args>::format_arg_store(T&& ...) [with T = {const ZDirectionType&}; Context = fmt::v8::basic_format_context<fmt::v8::appender, char>; Args = {ZDirectionType}]’
.../externalpackages/fmt/include/fmt/core.h:3103:17:   required from ‘std::__cxx11::string fmt::v8::format(fmt::v8::format_string<T ...>, T&& ...) [with T = {const ZDirectionType&}; std::__cxx11::string = std::__cxx11::basic_string<char>; fmt::v8::format_string<T ...> = fmt::v8::basic_format_string<char, const ZDirectionType&>]’
.../include/boutexception.hxx:28:34:   required from ‘BoutException::BoutException(const S&, const Args& ...) [with S = char [23]; Args = {ZDirectionType}]’
.../src/sys/bout_types.cxx:12:11:   required from ‘const string& {anonymous}::safeAt(const std::map<T, std::__cxx11::basic_string<char> >&, T) [with T = ZDirectionType; std::__cxx11::string = std::__cxx11::basic_string<char>]’
.../src/sys/bout_types.cxx:164:42:   required from here
.../externalpackages/fmt/include/fmt/core.h:3103:44:   in constexpr expansion of ‘fmt::v8::make_format_args<>((* & args#0))’
.../externalpackages/fmt/include/fmt/core.h:1706:3: error: static assertion failed: Cannot format an argument. To make type T formattable provide a formatter<T> specialization: https://fmt.dev/latest/api.html#udt
```

ZedThree:
I have a feeling we bumped the version of `fmt` in `next`, which hasn't made into this branch yet -- but I've not actually checked that.

dschwoerer:
I fail to see how bumping in next would cause a difference between RTD and GHA.
They both run ubuntu, both gcc 7.5.0. One difference is that GHA has PETSc+SUNDIALS, different cmake version, openmpi vs mpich (and maybe some more).

Some (probably unrelated) cmake warnings:
```
CMake Warning (dev) at /home/docs/checkouts/readthedocs.org/user_builds/bout-dev/envs/2503/lib/python3.8/site-packages/cmake/data/share/cmake-3.22/Modules/CMakeDependentOption.cmake:84 (message):
  Policy CMP0127 is not set: cmake_dependent_option() supports full Condition
  Syntax.  Run "cmake --help-policy CMP0127" for policy details.  Use the
  cmake_policy command to set the policy and suppress this warning.
  cmake/SetupBOUTThirdParty.cmake:94 (cmake_dependent_option)
  CMakeLists.txt:406 (include)
This warning is for project developers.  Use -Wno-dev to suppress it.

-- Using mpark.variant submodule
-- Using fmt submodule
CMake Warning (dev) at /home/docs/checkouts/readthedocs.org/user_builds/bout-dev/envs/2503/lib/python3.8/site-packages/cmake/data/share/cmake-3.22/Modules/CMakeDependentOption.cmake:84 (message):
  Policy CMP0127 is not set: cmake_dependent_option() supports full Condition
  Syntax.  Run "cmake --help-policy CMP0127" for policy details.  Use the
  cmake_policy command to set the policy and suppress this warning.
  cmake/SetupBOUTThirdParty.cmake:113 (cmake_dependent_option)
  CMakeLists.txt:406 (include)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

ZedThree:
Ah, the issue is that on RTD we're doing `os.system("git clone https://github.com/fmtlib/fmt.git ../../externalpackages/fmt")` instead of `git submodule`, so we're pulling in the most recent version of `fmt`, and we needed a fix for 8.1.0: #2482 

I just merged `next` into the base branch, so I'll update this PR now

CMakeLists.txt
```diff
@@ -1,8 +1,8 @@
-cmake_minimum_required(VERSION 3.13...3.16)
-if(${CMAKE_VERSION} VERSION_LESS 3.13)
+cmake_minimum_required(VERSION 3.17...3.20)
```
dschwoerer, line 1: The clang-tidy-review CI has only cmake 3.16 - so that workflow needs to be adapted.

PR #2496, opened 2022-02-14
Update outer boundaries in PETSc 3D solver

bendudson:
Only affects boundary conditions if MXG > 1: First set of boundary cells are already set as part of the solve.

Changes from always zero-gradient to extrapolating. If zero-gradient boundary conditions are set then the result will not change.

### DIFF
--- a/src/invert/laplace/impls/petsc3damg/petsc3damg.cxx
+++ b/src/invert/laplace/impls/petsc3damg/petsc3damg.cxx
@@ -282,7 +282,7 @@ Field3D LaplacePetsc3dAmg::solve(const Field3D &b_in, const Field3D &x0) {
   //       (in the first boundary cell) so one boundary cell is already set
   BOUT_FOR(i, indexer->getRegionInnerX()) {
     for (int b = 1; b < localmesh->xstart; b++) {
-      solution[i.xm(b)] = 2.*solution[i.xm(b-1)] - solution[i.xm(b-2)];
+      solution[i.xm(b)] = 3.*solution[i.xm(b-1)] - 3.*solution[i.xm(b-2)] + solution[i.xm(b-3)];
     }
   }
 
--- a/src/invert/laplace/impls/petsc3damg/petsc3damg.cxx
+++ b/src/invert/laplace/impls/petsc3damg/petsc3damg.cxx
@@ -291,7 +291,7 @@ Field3D LaplacePetsc3dAmg::solve(const Field3D &b_in, const Field3D &x0) {
   //       (in the first boundary cell) so one boundary cell is already set
   BOUT_FOR(i, indexer->getRegionOuterX()) {
     for (int b = 1; b < localmesh->xstart; b++) {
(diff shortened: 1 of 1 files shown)


src/invert/laplace/impls/petsc3damg/petsc3damg.cxx
```diff
@@ -275,12 +275,23 @@ Field3D LaplacePetsc3dAmg::solve(const Field3D &b_in, const Field3D &x0) {
...
-    BOUT_FOR(i, indexer->getRegionInnerX()) {
-      solution[i.xm(b)] = solution[i];
+
+  // Set inner boundary cells by extrapolating
+  // from single boundary cell which is set by the solver
+  // Note: RegionInnerX is the set of points just outside the domain
+  //       (in the first boundary cell) so one boundary cell is already set
+  BOUT_FOR(i, indexer->getRegionInnerX()) {
+    for (int b = 1; b < localmesh->xstart; b++) {
+      solution[i.xm(b)] = 2.*solution[i.xm(b-1)] - solution[i.xm(b-2)];
```
johnomotani, line 285: To be consistent with the extrapolation in the boundary conditions, e.g. here (after rearranging to replace `delta * val` with values of `f` from the previous line)
https://github.com/boutproject/BOUT-dev/blob/ca49afe984ee93203ce0709d30c9219bbde17a68/src/mesh/boundary_standard.cxx#L2138-L2143
I think we need higher order extrapolation:
```suggestion
      solution[i.xm(b)] = 3.*solution[i.xm(b-1)] - 3.*solution[i.xm(b-2)] + solution[i.xm(b-3)];
```

src/invert/laplace/impls/petsc3damg/petsc3damg.cxx
```diff
@@ -275,12 +275,23 @@ Field3D LaplacePetsc3dAmg::solve(const Field3D &b_in, const Field3D &x0) {
...
-    BOUT_FOR(i, indexer->getRegionOuterX()) {
-      solution[i.xp(b)] = solution[i];
+  }
+
+  // Set outer boundary cells by extrapolating
+  // Note: RegionOuterX is the set of points just outside the domain
+  //       (in the first boundary cell) so one boundary cell is already set
+  BOUT_FOR(i, indexer->getRegionOuterX()) {
+    for (int b = 1; b < localmesh->xstart; b++) {
+      solution[i.xp(b)] = 2.*solution[i.xp(b-1)] - solution[i.xp(b-2)];
```
johnomotani, line 294: see above
```suggestion
      solution[i.xp(b)] = 3.*solution[i.xp(b-1)] - 3.*solution[i.xp(b-2)] + solution[i.xp(b-3)];
```

johnomotani (changes requested):
Thanks @bendudson! This looks totally general, and consistent with `boundary_standard.cxx` (with suggestions below). Nice how simple it is!

ZedThree:
Is this one of things that can only be effectively tested with MMS?

johnomotani (approved):
Looks good to me!

PR #2497, opened 2022-02-15
Various fixes and updates for CI

ZedThree:
- Have a build with as few dependencies as possible
- Remove the OpenMP build -- covered by other builds
- Automatically apply clang-format to PRs
- Only run certain workflows on changes to certain files
    - black on changes to Python files
    - clang-format and clang-tidy on changes to C++ files
- Bump clang-tidy-review version
    - Now makes suggestions from fixits that can be more easily applied
    - Plus cleaner config
- Cancel previous workflows on new pushes to a branch/PR


.github/workflows/black-fix.yml
```diff
@@ -1,6 +1,9 @@
 name: black
 
-on: [push, pull_request]
+on:
+  push:
+    paths:
+      - '**.py'
```
dschwoerer, line 6: Note we also ship executables without py extension ...

.github/workflows/clang-tidy-review.yml
```diff
@@ -1,54 +1,44 @@
...
+          cmake_command: |
+            cmake . -B build -DBUILD_SHARED_LIBS=ON \
+                             -DBOUT_ENABLE_OPENMP=ON \
+                             -DBOUT_USE_PETSC=ON \
+                             -DBOUT_USE_SLEPC=ON \
+                             -DBOUT_USE_SUNDIALS=ON \
+                             -DBOUT_BUILD_EXAMPLES=ON \
+                             -DBOUT_BUILD_DOCS=OFF \
+                             -DBOUT_ENABLE_PYTHON=OFF \
+                             -DCMAKE_EXPORT_COMPILE_COMMANDS=On
```
dschwoerer, line 44:
```suggestion
                             -DCMAKE_EXPORT_COMPILE_COMMANDS=ON
```

.github/workflows/black-fix.yml
ZedThree: Fair point, probably `bin/**` is sufficient here?

dschwoerer: Unfortunately not:
```
$ git grep '#!/.*/bin/.*python'|cut -d: -f1|grep -v \\\.py\$|grep -v runtest
bin/bout-add-mod-path
bin/bout-boutdata-restart-create
bin/bout-pylib-cmd-to-bin
bin/bout-pylib-cmd-to-bin
bin/bout-sonnet
examples/conduction/run
examples/elm-pb/runexample
examples/laplace-petsc3d/run
examples/staggered_grid/runandplot
manual/sphinx/user_docs/python_boutcore.rst
tests/integrated/test_suite
tests/requirements/metric_3d
tests/requirements/scipy
tools/tokamak_grids/elite/elite2nc
tools/tokamak_grids/gato/gato2nc
```

.github/workflows/clang-tidy-review.yml
```diff
@@ -1,54 +1,44 @@
...
-on: [pull_request]
+on:
+  pull_request:
+    paths:
+      - '**.cxx'
+      - '**.hxx'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
dschwoerer, line 11: Does it make sense to run the tests twice for push+PR? 
Or is that already sufficient to prevent this?

dschwoerer: Is there a way to check whether that changes anything? For most PRs they are based on the base branch, this is doing the same runs twice - if we could skip that, that would safe a lot of CPU time ...
But unless it is easy (which I worry it isn't) probably not worth to pursue ...

dschwoerer (approved):
Thanks for improving the CI system

.github/workflows/clang-tidy-review.yml
ZedThree: That's a really good question! I guess if there are commits to the base branch that are more recent than any commits in the current branch, then it's worth doing the PR jobs? It's probably worth looking to see if there's something already existing that could handle that for us

dschwoerer: It looks to me like this is what we want:
https://github.com/fkirc/skip-duplicate-actions#example-1-skip-entire-jobs

Issue #2365, opened 2021-07-04
cmake: Natural Language Support


dschwoerer:
It seems NLS isn't working with cmake yet.

dschwoerer:
There is support for that:
https://cmake.org/cmake/help/latest/module/FindGettext.html

We porbably want to start with the last step (`GETTEXT_PROCESS_PO_FILES`) to create `.mo` files, and also install them ...

Issue #2321, opened 2021-05-17
cmake: install incomplete


dschwoerer:
make install does not install
- [x] boutconfig
- [ ] boutcore
- [x] executables?
- [ ] examples?
- [x] cmake files?
- [ ] manual

dschwoerer:
`cmake` is maybe incomplete:
```
$ diff installed/lib64/cmake/bout++/ ../cmake
Only in installed/lib64/cmake/bout++/: bout++Config.cmake
Only in installed/lib64/cmake/bout++/: bout++ConfigVersion.cmake
Only in ../cmake: BOUT++functions.cmake
Only in installed/lib64/cmake/bout++/: bout++Targets.cmake
Only in installed/lib64/cmake/bout++/: bout++Targets-noconfig.cmake
Only in ../cmake: EnableCXXWarningIfSupport.cmake
Only in ../cmake: FindClangFormat.cmake
Only in ../cmake: FindCython.cmake
Only in ../cmake: FindCython.cmake~
Only in ../cmake: FindLibuuid.cmake
Only in ../cmake: FindnetCDFCxx.cmake
Only in ../cmake: FindNumpy.cmake
Only in ../cmake: FindNumpy.cmake~
Only in ../cmake: GetGitRevisionDescription.cmake
Only in ../cmake: GetGitRevisionDescription.cmake.in
```

ZedThree:
Good catch, definitely a few of those need to be installed -- the `Find*` modules, and I suspect `BOUT++functions.cmake` too. I'm fairly sure `EnableCXXWarningIfSupport` and `GetGitRevisionDescription` are only needed at configure time (for the library), and don't need to be installed.

I'm not sure about installing the examples -- they could be useful, but maybe not? It doesn't look like we install them with autotools, but we also don't have an option for compiling them all either.

Which executables do you mean? The programs installed by autotools are just `bout-config` and `bout-log-color`. I'm not sure we want to install any of the model executables.

bendudson:
For the examples, they're useful if someone can edit and run them so we could either:
1. Install them centrally, then have a script which copies them into the user's directory 
   ```
   $ bout-examples  conduction
   ```
   might copy the `conduction` example and compile it.
2. Make the examples a separate repository, which users can check out. This might be the most straightforward, if installing the BOUT++ library becomes the more standard way to use it.

dschwoerer:
[...]

Issue #2287, opened 2021-04-17
autotools 2.71


dschwoerer:
Autotools 2.71 was released a while ago, and there seem to be some issues with the new autotools.

I managed to run autoreconf successfully with:
```
autoupdate
touch build-aux/config.rpath
autoreconf -fiv
```

However, it seems to only work if I have various packages installed, just having `autotools`, `autotools-archives` and `/usr/bin/aclocal` is not sufficient - so there are still some issues remaining ...

Is this something to be bothered about? I am tempted to rather work on switching to cmake instead.

bendudson:
I think the cmake build is generally in pretty good shape (thanks @ZedThree), and with a bit more work could completely replace the autotools stuff.  If we can get the last bits done,  like boutcore,  then I'd be happy for cmake to be the recommended (only?)  build system for version 5

Issue #1921, opened 2020-02-26
cflags in `bout-config` broken


dschwoerer:
cflags in `bout-config` are wrong, as the quoting does not work reliably.
This is the case for configure. (cmake is missing `bout-config`)

I've been wondering for a while whether there is a nicer way to build against BOUT++, rather then depending on the `makefile` used by BOUT++ to get the right flags.

`cmake` is obviously one way, another would be `pkg-config`, that seems to be also (somewhat) supported by PETSc.

Another option would be to at put the defines in a file, then only the libs and normal flags need to be exposed via e.g. `pkg-config`, `bout-config` or similar.

edit: reproduce via:
```
./configure
bin/bout-config --cflags
```
result:
```
-I/home/dave/soft/bout-dev/tmp/include -I/usr/include -I/usr/include -I/home/dave/soft/bout-dev/tmp/externalpackages/mpark.variant/include -g -DHAS_PRETTY_FUN
CTION -Wall -Wextra -Wnull-dereference -Wno-cast-function-type -O2 -DCHECK=2 -DSIGHANDLE -DLOGCOLOR -DBOUT_VERSION_STRING=4.3.0 -DBOUT_VERSION_DOUBLE=4.30 -DB
ACKTRACE -DBOUT_HAS_FFTW -DNCDF4 -DHDF5 -DLAPACK -DBOUT_HAS_PVODE -DBOUT_HAS_GETTEXT -DBOUT_LOCALE_PATH=/usr/local/share/locale -DBOUT_FLAGS_STRING= -g -DHAS_
PRETTY_FUNCTION -Wall -Wextra -Wnull-dereference -Wno-cast-function-type -O2 -DCHECK=2 -DSIGHANDLE -DLOGCOLOR -DBOUT_VERSION_STRING=4.3.0 -DBOUT_VERSION_DOUBL
E=4.30 -DBACKTRACE -DBOUT_HAS_FFTW -DNCDF4 -DHDF5 -DLAPACK -DBOUT_HAS_PVODE -DBOUT_HAS_GETTEXT -DBOUT_LOCALE_PATH=/usr/local/share/locale
```
Expected output:
`-DBOUT_FLAGS_STRING=` is quoted, `-DBOUT_VERSION_STRING` is quoted, etc.

ZedThree:
[...]

PR #2502, opened 2022-02-22
Switch to CMake

dschwoerer:
Start removing configure stuff - supporting one build system is enough work.

Still todo:
- [ ] #2365 
- [ ] #2321 
- [ ] #1921 ?
- [ ] #1904 ?
- [ ] `bout-config` - not working, see failing test
- [ ] `make changelog LAST_VERSION=vA.B.C RELEASE_BRANCH=master|next` - needs replacement
- [ ] ~~`make -C locale update-all`~~ (still working)
- [ ] `make dist`
- [ ] Update manual/RELEASE_HOWTO.md
- [ ] Make sure `./configure` is removed from docs

Resolves:
- [x] #1488 
- [x] #2287 
- [x] #1921 (would resolve it, but is also broken)


.github/workflows/tests.yml
```diff
@@ -32,9 +32,12 @@ jobs:
             configure_options: "--with-petsc
                                 --with-slepc
                                 --with-sundials=/home/runner/local"
-            script_flags: "-uim"
+            cmake_options: "-DBOUT_USE_PETSC=ON
```
ZedThree, line 35: For the `default options` build, we should not specify any options really -- at least partly because we really need a build without PETSc/SLEPc

.github/workflows/tests.yml
```diff
@@ -32,9 +32,12 @@ jobs:
...
+                            -DBOUT_USE_SLEPC=ON
+                            -DBOUT_USE_SUNDIALS=ON
+                            -DSUNDIALS_ROOT=/home/runner/local"
 
-          - name: "Optimised, shared, Python"
+          - name: "Optimised, static"
             os: ubuntu-18.04
             configure_options: "--enable-shared
```
ZedThree, line 42: We can delete all the `configure_options` bits from the matrix too

cmake/FindSUNDIALS.cmake
```diff
@@ -31,10 +31,6 @@
 
 include(FindPackageHandleStandardArgs)
 
-find_package(SUNDIALS CONFIG QUIET)
```
ZedThree, line 34: Is this failing in some case?

ZedThree (commented):
There's also the `makefile`s in the various subdirectories that will need removing too

ZedThree:
Thanks @dschwoerer! I'm in favour of this, but I'm interested to know if anyone has any objections?

The `bout-config` problem seems to be that it's trying to link against `netcdf` rather than `-lnetcdf`. That's ultimately coming from `FindnetCDF.cmake`.

cmake/FindSUNDIALS.cmake
dschwoerer: Yes, I think this is for https://github.com/boutproject/BOUT-dev/issues/2381

I think I closed it because I had included this fix in my branch, thus thought it was fixed.

johnomotani:
I'm also in favour of this :+1:

dschwoerer:
Probably worth to discuss this in the BUG Meeting - to avoid the bias of 
only having answers from people working on cmake :-)

bendudson:
Thanks @dschwoerer ! I'm also in favour of this, with some reservations:

1. I'd like to put this into `next` after 5.0 is released, with a warning in the 5.0 release that Autotools will be removed in 5.1

2. We continue to generate and test the `bout-config` script. That allows people to use whatever build system they want to build their examples, including Makefiles, not only CMake.

PR #2507, opened 2022-02-24
xgettext fixes

dschwoerer:
Not sure why the other version isn't working ...


ZedThree (approved):

ZedThree:
Thanks! It's a pity gettext doesn't support C++ user literals, it would be much cleaner to use

dschwoerer:
They seem to support it:
https://git.savannah.gnu.org/gitweb/?p=gettext.git;a=blob;f=gettext-tools/tests/xgettext-c-c%2B%2B-1;h=09276c556608d7f9306f5927fa00139a44f5ce47;hb=HEAD#l62

No clue why it doesn't work ...

ZedThree:
They probably parse the repeated quote characters as the end of the string.

By user literals, I meant something like `"some text"t` which would avoid the need for wrapping up in a macro. I think that works for the translation bit, but not for the string extraction part

Issue #2186, opened 2020-12-31
dockerhub images out of date


johnomotani:
The images on https://hub.docker.com/u/boutproject/ are out of date. This may cause problems (https://github.com/boutproject/xBOUT/issues/178). Do we intend to provide these images? If so it would be nice to have one for the latest release at least, maybe update them as part of the release process? If not, maybe we should take down the out of date ones, and remove the instructions in the manual at https://bout-dev.readthedocs.io/en/latest/user_docs/installing.html?highlight=docker#docker-image to avoid confusion?

ZedThree:
It looks like it's possible to get them to auto-build, but we'd need to have the Dockerfile in this repo -- currently, there's a separate repo (https://github.com/boutproject/docker-bout) with all the build scripts in.

bendudson:
I'm trying to get them working for the latest release. Unfortunately Centos7 has an ancient compiler (gcc 4.8.5) which won't compile BOUT++ master. I've loaded `devtoolset-7` to get gcc 7, but then linking fails when `--enable-shared` is used.

bendudson:
I've built some up-to-date images, now on dockerhub: https://hub.docker.com/u/boutproject
- boutproject/bout-master:3b23201-arch     Latest `master` with Sundials (2.7.0) and PETSc (3.14.2)
- boutproject/bout-master:3b23201-sundials-arch   Latest `master` with just Sundials (2.7.0)
- boutproject/bout-next:4cee87c-arch   Latest `next` with Sundials (2.7.0) and PETSc (3.14.2)
- boutproject/bout-next:4cee87c-sundials-arch   Latest `next` with just Sundials (2.7.0)

All based on Arch linux. I don't mind which distribution we use, but Centos seems to be a dead end, and at least Arch is up to date.

@ZedThree Is it possible to have releases auto-built if the docker scripts are in a submodule? Having releases automatically built and pushed to Docker Hub would be handy.

johnomotani:
In principle I guess we don't even need the scripts in a submodule? A Github action to run on release could clone https://github.com/boutproject/docker-bout and run something in there. We use release actions for uploading the Python projects to PyPi, e.g. https://github.com/boutproject/xBOUT/blob/master/.github/workflows/pythonpublish.yml.

ZedThree:
Thanks @bendudson!

It looks like having the dockerfile in a submodule might be tricky, so a Github action might be the better way to go.

johnomotani:
[...]

PR #2511, opened 2022-02-28
Build containers in GHA CI

dschwoerer:
Working towards fixing #2186 

Next step would be to also build on releases.

And of course this needs to be adjusted for the boutproject repository, I have been pushing images [here](https://hub.docker.com/r/oi4ai/bout3d)

They are based on:
https://github.com/dschwoerer/bout-container-base
So petsc is included, but doesn't need to be rebuild each time.


.docker/fedora/Dockerfile.sh
```diff
@@ -0,0 +1,51 @@
+cat <<EOF
+FROM docker.io/oi4ai/bout-base:openmp-$MPI-mini
```
ZedThree, line 2: Rather than having this as a shell script, could we make use of `ARG`/`build_arg`? See e.g. https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact

dschwoerer: I should have read the official documentation - and not some random stuff on the internet :-(

I thought it wouldn't be possible, but as that seems to be the case, I can go back to the old version ...

dschwoerer:
Also, probably should note, for this to work, the password for the docker account need to be added.

ZedThree:
Maybe there's a dockerhub token we can add, similar to the PyPI one?

bendudson:
> Maybe there's a dockerhub token we can add, similar to the PyPI one?

That should be possible. I've got the dockerhub details somewhere and can add them to a Github secret or whatever it's called.

dschwoerer:
We should probably switch to quay.io to avoid the duckerhub rate limiting ...

ZedThree:
We could also just use Github's own package registry?

dschwoerer:
No, we cannot. They have a [limit of 500 MB](https://github.com/features/packages#pricing) for images - that is not even [1 image](https://hub.docker.com/r/boutproject/bout-next/tags)

dschwoerer:
Ah, that is for private repositories, so we should be fine :sweat: 
Always gets me -.-

In that case, sure, we could use ghcr.io :+1:

dschwoerer:
Now with ghcr.io :-)

Everybody can publish packages at https://github.com/boutproject/BOUT-dev/pkgs/container/bout-dev but deleting ones, e.g. with bad tags, can only be done by admins :-/

.docker/fedora/Dockerfile
```diff
@@ -0,0 +1,59 @@
...
+
+RUN sudo ls
+
+# Checkout submodules now so configure later is fast, and iterating on
+# it less painful
+RUN git clone $URL \
+ && chown -R boutuser /home/boutuser/BOUT-dev \
+ && cd BOUT-dev \
+ && git checkout $COMMIT \
+ && git -c submodule.recurse=false submodule update --init --recursive
```
ZedThree, line 28: This looks a bit curious -- does `--recursive` not imply `submodule.recurse=true`?

ZedThree: Yes, that _should_ be fine, but as you point out here, might be faster to do that first?

ZedThree:
Awesome, thanks @dschwoerer! @tomc271 has also been looking at reviving the docker images -- Tom, can you have a look over this please?

.docker/fedora/Dockerfile
dschwoerer: Oh, I missed that comment :+1:

tomc271:
Location of mpiexec is not in the PATH environment variable; Need to prefix it with $MPI_BIN/ in order to use it. 
Should it be added to the PATH?

dschwoerer:
It seems it is not possible to do that with ENV:

https://stackoverflow.com/q/46032023

However, as we would currently need to update it everytime a new python lands into fedora, it might be better to drop something in `/etc/profiles.d/`? I haven't tested whether that actually works, but I would assume that any shell should parse that at startup ...

dschwoerer:
`/etc/profiles.d/` is only read by login shells - so this is not useful
`/etc/environment` seems also ignored.
I now have something hopefully working with ENV:
https://github.com/dschwoerer/bout-container-base/blob/main/Dockerfile#L28

However, openmpi build of PETSc seems stuck ...

ZedThree:
Fedora builds now failing with:

```
 Problem: conflicting requests
  - nothing provides mpich-devel(x86-32) = 4.0.3 needed by petsc-mpich-devel-3.18.5-1.fc39.i686 from rawhide
  - nothing provides mpich-devel(x86-64) = 4.0.3 needed by petsc-mpich-devel-3.18.5-1.fc39.x86_64 from rawhide
```

dschwoerer:
Reported here: https://bugzilla.redhat.com/show_bug.cgi?id=2215261
Workaround build in progress: https://copr.fedorainfracloud.org/coprs/davidsch/fixes4bout/packages/

As soon as the build is finished we should be able to rerun the failed tests.

dschwoerer:
I noticed some unrelated cmake changes sneaked in - should I rebase to get them out?

dschwoerer:
@tomc271 this is now fixed:
```
podman run --rm ghcr.io/boutproject/bout-dev:mpich-3d-opt-docker-ci mpirun --version
HYDRA build details:
[...]
```

tomc271:
Should [bout-container-base](ghcr.io/dschwoerer/bout-container-base) become a boutproject repository?

dschwoerer:
> Should [bout-container-base](ghcr.io/dschwoerer/bout-container-base) become a boutproject repository?

Probably, yes. I can do it if this gets merged.

> (We already have https://github.com/boutproject/docker-bout)

This is probably not needed anymore, as it becomes part of BOUT++ with this PR.

tomc271:
Testing the docker container: 

The 'conduction' example built and ran fine (if a bit slow? 2m 12s)

However unable to find pip: 
```
bash-5.2$ pip install xbout
bash: pip: command not found
```

dschwoerer:
Which container did you use? Optimised?

Only some base images come with pip:
https://github.com/dschwoerer/bout-container-base/blob/main/Dockerfile#L13

You can get pip with either:
`sudo dnf install python3-pip`
or with
`python -m ensurepip`
either should work ...

tomc271:
Ah ok. It was mpich-3d-debug-docker-ci

.github/workflows/docker.yml
```diff
@@ -0,0 +1,84 @@
...
+
+          - name: "Optimised"
+            tag_postfix: "opt"
+            options: "-DCHECK=0"
+            base_postfix: "mini"
+
+          - name: "Full"
+            tag_postfix: "full"
+            options: "-DCHECK=3"
+            base_postfix: "full"
```
dschwoerer, line 51: The full would contain `pip`.
The `Debug` is compiled with `-DCHECK=3` as well, so that explains why it is a bit slow.
`opt` could be used for production runs ...

tomc271 (approved):

PR #2514, opened 2022-02-28
Switch to CMake in CI and docs

dschwoerer:
Rebase of #2502 

Now without removing configure files - to be done for v5.1.0


dschwoerer:
Is it intentionally that we now require `fastcov`?

ZedThree:
I think `fastcov` was just easier/faster to use

dschwoerer:
If you get fastcov to work in CI ...

ZedThree (approved):
LGTM. Couple of suggestions

ZedThree:
Todo before merge:

- [ ] Add deprecation notice to `./configure`
- [ ] Make sure the docs all use cmake by default

dschwoerer:
I updated the docs, however, I haven't tested all commands - so checking would be greatly appreciated.

ZedThree (approved):

ZedThree:
The failing tests are just Fedora, and that was already fixed in next

Issue #2501, opened 2022-02-15
Configure error when using  Sundials 6.1.1


sagitter:
Hi all.

`configure` is failing when i try to compile `bout++-4.4.1` against  `sundials-6.1.1` (GCC 12.0.1 20220212):

```
configure:11945: Searching for SUNDIALS IDA library
configure:12079: Checking for IDA header files
configure:12084: checking for /usr/include/mpich-x86_64/ida/ida.h
configure:12099: result: yes
configure:12113: checking for /usr/include/mpich-x86_64/ida/ida_spgmr.h
configure:12128: result: no
configure:12142: checking for /usr/include/mpich-x86_64/ida/ida_bbdpre.h
configure:12157: result: yes
configure:12171: checking for /usr/include/mpich-x86_64/nvector/nvector_parallel.h
configure:12186: result: yes
configure:12200: checking for /usr/include/mpich-x86_64/sundials/sundials_types.h
configure:12215: result: yes
configure:12390: Found IDA include path: /usr/include/mpich-x86_64
configure:12409: checking if SUNDIALS IDA library path is /usr/lib64/mpich/lib
[...]

PR #2517, opened 2022-03-01
Add support for SUNDIALS 6

dschwoerer:
Enable support for sundials 6

Does not address #2501 as this is cmake-only


src/solver/impls/arkode/arkode.cxx
```diff
@@ -164,6 +165,15 @@ constexpr auto& ARKStepSetPreconditioner = ARKSpilsSetPreconditioner;
 constexpr auto& ARKStepSetUserData = ARKodeSetUserData;
 #endif
 
+#if SUNDIALS_VERSION_MAJOR >= 6
+#define SUNCTX_PLACEHOLDER , suncontext
+#else
+#define SUNCTX_PLACEHOLDER
+#define SUN_PREC_RIGHT PREC_RIGHT
```
ZedThree, line 172: This is a good one to fix, it's a pretty simple change:

```suggestion
constexpr auto SUN_PREC_RIGHT = PREC_RIGHT;
```

src/solver/impls/arkode/arkode.cxx
```diff
@@ -164,6 +165,15 @@ constexpr auto& ARKStepSetPreconditioner = ARKSpilsSetPreconditioner;
 constexpr auto& ARKStepSetUserData = ARKodeSetUserData;
 #endif
 
+#if SUNDIALS_VERSION_MAJOR >= 6
+#define SUNCTX_PLACEHOLDER , suncontext
```
ZedThree, line 169: This is definitely clever, but it makes the source look _really_ strange. I think it might be nicer to just provide overloads/shims for previous versions (see earlier in this file for examples)

ZedThree:
Bah, Fedora still failing -- some issue with their mirrors?

Also, I forgot `clang-tidy-review` doesn't seem to like OpenMP for some reason. Might be easiest to just turn off OpenMP for that

ZedThree (approved):

PR #2521, opened 2022-03-03
Add backports and shims for SUNDIALS 6

ZedThree:
Adds a header for common backports shared between the different SUNDIALS solvers such that the main code only needs to differentiate between SUNDIALS pre/post 3.0.0

Tested locally on SUNDIALS 2.7.0, 3.2.1, 4.1.0, 5.8.0, and 6.1.1

### DIFF
--- a/include/bout/sundials_backports.hxx
+++ b/include/bout/sundials_backports.hxx
@@ -23,6 +23,8 @@
 #include <sunnonlinsol/sunnonlinsol_newton.h>
 #endif
 
+#include "unused.hxx"
+
 #if SUNDIALS_VERSION_MAJOR < 3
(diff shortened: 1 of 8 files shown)


dschwoerer (approved):
Nice, looks much cleaner :+1:

ZedThree:
clang-tidy-review broke but did actually catch a bug: I had `#include "unused.hxx"` in the wrong place, it just coincidentally worked because other headers were bringing it in first. Also applied some clang-tidy fixes for braces around single lines

ZedThree:
Failing test is because Fedora failed to update. This PR is into another PR so I'll merge and let the tests run again there

PR #2523, opened 2022-03-07
Do not document mpich from source

dschwoerer:
- on hpc systems there will be an MPI installed + configured
- on user systems, your distro should provide MPI


bendudson (approved):
I think installing from source was necessary on some systems where the modules were poorly configured. It is hopefully less needed now

PR #2533, opened 2022-04-01
Document parallel boundary conditions and 'aligned trasform' parallel derivatives

### LINKED ISSUE #2531: Missing (?) documentation on how to set parallel boundary conditions
We do not seem to have any documentation on how to set parallel boundary conditions when using `ShiftedMetric` or `FCI` parallel transforms. I have not used methods requiring `yup`/`ydown` fields myself, so please correct/confirm (@ZedThree?). I think the possibilities are:

For `ShiftedMetric`
1. Use the boundary conditions `parallel_dirichlet`, `parallel_dirichlet_O3` or `parallel_neumann`. These can be passed as the values for `bndry_par_yup`, `bndry_par_ydown` or `bndry_par_all`. These will set values in `yup()` and `ydown()` (although only at the first boundary point). Needs to be applied by the user after yup/ydown fields are calculated (i.e. after `mesh->communicate()` is called).
2. Use the modifier `fromFieldAligned`, e.g. `bndry_yup = fromFieldAligned(dirichlet)`. This transforms the field from 'orthogonal' to 'field aligned' coordinates, applies the boundary condition, and then transforms back. Needs to be applied before calculating yup/ydown fields, so is applied automatically for variables added to the time-solver. Not as efficient as it could be (at least at present) as it transforms the entire field, not just the needed part near the boundary.

For `FCI`
1. Use the boundary condition `parallel_dirichlet_interp` (again only sets first boundary point) as the value for `bndry_par_yup`, `bndry_par_ydown` or `bndry_par_all`. Needs to be applied by the user after yup/ydown fields are calculated (i.e. after `mesh->communicate()` is called).

When explicitly using separate field-aligned versions of variables:
1. Apply normal boundary conditions (`dirichlet`, etc.) in the y-direction (using `bndry_yup`/`bndry_ydown`, or included in `bndry_all`) to the field-aligned variables. `none` can be used for the y-direction of the non-aligned versions (to be as efficient as possible, since the y-boundary values of non-aligned versions are never used). Need to be applied by the user after calculating field-aligned version (after e.g. `n_aligned = toFieldAligned(n)`).

Also, I think the section on 'Shifted boundaries'
https://bout-dev.readthedocs.io/en/latest/user_docs/boundary_options.html#shifted-boundaries
applied to BOUT++ v3 and earlier, and is now out of date.

johnomotani:
Add documentation for parallel boundary conditions. Fixes #2531.

Add documentation for 'aligned transform' variant of shifted metric parallel derivatives (as used by STORM, Hermes). 'Aligned transform' is the version that calculates parallel derivatives by transforming to a globally field aligned grid, calculating the operation, and transforming the result back to the toroidal grid. I've called it 'aligned transform' to get a short name - any suggestions for a better name very welcome!

* [ ] @dschwoerer, @bshanahan, @ZedThree please check the section on FCI bcs - feel free to edit/expand.

### DIFF
--- a/manual/sphinx/user_docs/boundary_options.rst
+++ b/manual/sphinx/user_docs/boundary_options.rst
@@ -3,6 +3,8 @@
 Boundary conditions
 ===================
 
+See also :ref:`sec-physicsmodel-boundary-conditions`.
+
 Like the variable initialisation, boundary conditions can be set for
 each variable in individual sections, with default values in a section
 ``[All]``. Boundary conditions are specified for each variable, being
--- a/manual/sphinx/user_docs/boundary_options.rst
+++ b/manual/sphinx/user_docs/boundary_options.rst
@@ -84,9 +84,6 @@ following are available:
 -  ``relax`` - Relaxing boundaries. Evolve the variable towards the
    given boundary condition at a given rate
 
--  ``shifted`` - Apply boundary conditions in orthogonal X-Z
-   coordinates, rather than field-aligned
-
 -  ``width`` - Modifies the width of the region over which the boundary
    condition is applied
 
@@ -119,22 +116,6 @@ By default, the relaxation rate is set to :math:`10` (i.e. a time-scale
 of :math:`\tau=0.1`). To change this, give the rate as the second
 argument e.g. ``relax(dirichlet, 2)`` would relax to a Dirichlet
 boundary condition at a rate of :math:`2`.
-
-Shifted boundaries
-------------------
-
-By default boundary conditions are applied in field-aligned coordinates,
-where :math:`y` is along field-lines but :math:`x` has a discontinuity
-at the twist-shift location. If radial derivatives are being done in
-shifted coordinates where :math:`x` and :math:`z` are orthogonal, then
-boundary conditions should also be applied in shifted coordinates. To do
-this, the ``shifted`` boundary modifier applies a :math:`z` shift,
-applies the boundary condition, then shifts back. For example::
-
-    bndry_core = shifted( neumann )
-
-would ensure that radial derivatives were zero in shifted coordinates on
-the core boundary.
 
 Changing the width of boundaries
 --------------------------------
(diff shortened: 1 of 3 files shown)


ZedThree (approved):
LGTM, thanks @johnomotani !

I think these docs highlight several pain points for the parallel derivatives, not least the boundary conditions. I'm not sure what could be done, but it would be really nice to have something without so many sharp edges (many of which are least partly my fault!)

manual/sphinx/user_docs/boundary_options.rst
```diff
@@ -82,14 +84,185 @@ following are available:
...
+.. _sec-parallel-bc-fci:
+
+FCI boundary conditions
+^^^^^^^^^^^^^^^^^^^^^^^
+
+When using the FCI method (:ref:`sec-fci`), parallel boundary
+conditions must be applied to the parallel slices using
+`bndry_par_yup` and `bndry_par_ydown`, or `bndry_par_all` to set both
+together. The only option suitable for FCI currently implemented is
+``parallel_dirichlet_interp``. It is suggested, at least if there are
```
ZedThree, line 251: Actually, I don't think this is correct, I thought that they were all suitable for FCI? Certainly when I originally wrote the parallel boundary conditions, they were all designed for FCI. Maybe @dschwoerer or @bshanahan have had more recent experience with them and can weigh in

johnomotani: This is entirely my ignorance. `parallel_dirichlet`, `parallel_dirichlet_O3` and `parallel_neumann` looked to me like they should work for `ShiftedMetric`, and I just assumed `FCI` would need something special!

Someone who knows how `FCI` works - please edit this to say which boundary conditions make sense and when!

dschwoerer: I think they all should work. At least `parallel_neumann` is used a lot in hermes-2.

I have removed the sentence and added a bit FCI specific things.

bendudson (approved):

PR #2534, opened 2022-04-11
Fix `hist_hi` and `iteration` outputs

johnomotani:
This PR fixes two issues that I think are bugs:
1. `hist_hi` counts the number of steps of the most frequent `Monitor`, which is not necessarily the output monitor. I think we always want to count the number of output steps. Fixed in this PR by moving the increment of `Solver::iteration` to a `SolverMonitor`, which is called with the same frequency as the output monitor. The `SolverMonitor` is called before writing dump and output files, so `hist_hi` behaves the same as at present when the output monitor is the most frequent monitor.
2. `iteration`, as saved to the dump files, is too low by 1. I think the value should be: the same as `hist_hi` when the simulation was started from iteration 0 (i.e. not restarted); the same as the value printed to stdout at the same time as it is being saved. Fixed by 6b8eff8.

Question: is issue 2 actually slightly more general, and we should actually call `call_monitors` with the `iter` argument one greater (so that it counts the number of iterations that have been completed)? Proposed change to do this in #2535.

### DIFF
--- a/include/bout/solver.hxx
+++ b/include/bout/solver.hxx
@@ -426,6 +426,9 @@ protected:
   /// Calculate only the diffusive parts
   int run_diffusive(BoutReal t, bool linear = true);
 
+  /// Reset the iteration counter
+  void resetIterationCounter(int value = 0) { iteration = value; }
+
   /// Calls all monitor functions
   ///
   /// There are two important things to note about how \p iter is
--- a/include/bout/solver.hxx
+++ b/include/bout/solver.hxx
@@ -328,7 +328,7 @@ public:
   int getIterationCounter() const { return iteration; }
 
   /// Add one to the iteration count, used by BoutMonitor, but could be called by a
-  //user-defined monitor (if `bout_run()` is not used)
+  // user-defined monitor (if `bout_run()` is not used)
   int incrementIterationCounter() { return iteration++; }
 
 protected:
--- a/include/bout/solver.hxx
+++ b/include/bout/solver.hxx
@@ -329,7 +329,7 @@ public:
 
   /// Add one to the iteration count, used by BoutMonitor, but could be called by a
   // user-defined monitor (if `bout_run()` is not used)
-  int incrementIterationCounter() { return iteration++; }
+  int incrementIterationCounter() { return ++iteration; }
 
 protected:
   /// Number of command-line arguments
--- a/include/bout/solver.hxx
+++ b/include/bout/solver.hxx
@@ -324,6 +324,13 @@ public:
   /// The run from which this was restarted. Throws if the identifier hasn't been set yet.
   std::string getRunRestartFrom() const;
 
+  /// Get the number of completed output steps
+  int getIterationCounter() const { return iteration; }
+
+  /// Add one to the iteration count, used by BoutMonitor, but could be called by a
+  //user-defined monitor (if `bout_run()` is not used)
+  int incrementIterationCounter() { return iteration++; }
+
 protected:
   /// Number of command-line arguments
   static int* pargc;
@@ -411,8 +418,6 @@ protected:
 
   /// Current simulation time
   BoutReal simtime{0.0};
-  /// Current iteration (output time-step) number
-  int iteration{0};
 
   /// Run the user's RHS function
   int run_rhs(BoutReal t);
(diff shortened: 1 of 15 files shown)


include/bout/solver.hxx
```diff
@@ -474,6 +474,18 @@ protected:
...
+  public:
+    SolverMonitor() {}
+  private:
+    int call(Solver* solver, BoutReal UNUSED(t), int UNUSED(iter),
+             int UNUSED(NOUT)) override {
+      solver->iteration++;
+      return 0;
+    }
+  };
+  SolverMonitor solver_monitor;
```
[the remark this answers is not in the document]
ZedThree: Could maybe make `solver_monitor` private and add a protected getter?

ZedThree:
I like removing the duplicated `iteration++` from each solver!

Instead of adding a new monitor, does it make sense to do the iteration increment in the output monitor itself?

ZedThree:
We could also make `iteration` private then, which would be nice

johnomotani:
Should I rebase this branch to get rid of stuff that was added and then removed (e.g. `SolverMonitor`)?

include/bout/solver.hxx
```diff
@@ -324,6 +324,13 @@ public:
   /// The run from which this was restarted. Throws if the identifier hasn't been set yet.
   std::string getRunRestartFrom() const;
 
+  /// Get the number of completed output steps
+  int getIterationCounter() const { return iteration; }
+
+  /// Add one to the iteration count, used by BoutMonitor, but could be called by a
+  //user-defined monitor (if `bout_run()` is not used)
+  void incrementIterationCounter() { ++iteration; }
```
ZedThree, line 332: Small suggestion:

```suggestion
  /// Add one to the iteration count, used by BoutMonitor, but could be called by a
  /// user-defined monitor (if `bout_run()` is not used).
  /// Returns the incremented value
  int incrementIterationCounter() { return ++iteration; }
```

means below we can do:

```cpp
  // Get updated iteration count
  iteration = solver->incrementIterationCounter();
```

ZedThree:
> Should I rebase this branch to get rid of stuff that was added and then removed (e.g. `SolverMonitor`)?

Sure, that might be nice -- but it's not a blocker if you don't :)

include/bout/solver.hxx
johnomotani: Nice! I'll put that in.

ZedThree:
Tests are failing because there's a use of `++iteration` in `slepc.cxx` -- making it private was definitely a good idea, it caught a potential bug! :smile:

johnomotani:
> Tests are failing because there's a use of ++iteration in slepc.cxx -- making it private was definitely a good idea, it caught a potential bug!

It looked to me as though the uses of `iteration` in `SlepcSolver` would be alright, because `SlepcSolver` doesn't actually call the monitors (i.e. `BoutMonitor` is not called, except once before `SlepcSolver::run()` starts) - `SlepcSolver` is just in charge of `iteration`. So I've kept `Solver::iteration` private, but added a protected `Solver::resetIterationCounter()` for `SlepcSolver` to use. If someone who understands `SlepcSolver` could check that this PR doesn't break it (I hope we've got tests!), that would be nice :smile:

include/bout/solver.hxx
```diff
@@ -324,6 +324,13 @@ public:
   /// The run from which this was restarted. Throws if the identifier hasn't been set yet.
   std::string getRunRestartFrom() const;
 
+  /// Get the number of completed output steps
+  int getIterationCounter() const { return iteration; }
+
+  /// Add one to the iteration count, used by BoutMonitor, but could be called by a
+  // user-defined monitor (if `bout_run()` is not used)
+  int incrementIterationCounter() { return iteration++; }
```
ZedThree, line 332: This returns the value before the increment, but I think we probably want the value after increment:

```suggestion
  int incrementIterationCounter() { return ++iteration; }
```

Otherwise we're back to the original situation? Is it worth adding a test, perhaps to `test-restarting`, that checks `hist_hi` and `iteration`?

It is _incredibly_ unfortunate that `++` _after_ the variable returns the value _before_ the increment, and vice versa.

ZedThree:
> If someone who understands `SlepcSolver` could check that this PR doesn't break it (I hope we've got tests!), that would be nice

We do indeed have a `SlepcSolver` test, and that's passing so I assume it's correct :smile:

include/bout/solver.hxx
johnomotani: Yes, absolutely! Sorry, I mis-remembered and thought `++i` and `i++` were the other way around. Thank you for spotting my error!

ZedThree: I have to check every time, it's confusing! :smile:

PR #2540, opened 2022-04-29
Fix some issues with parallel BCs

dschwoerer:


ZedThree (approved):

src/field/field3d.cxx
```diff
@@ -514,17 +514,12 @@ void Field3D::applyParallelBoundary() {
   TRACE("Field3D::applyParallelBoundary()");
 
   checkData(*this);
+  ASSERT1(!yup_fields.empty());
```
johnomotani, line 517: Is it worth checking `ydown_fields` too? They should always be initialised together, but you never know...

ZedThree: `hasParallelSlices()` does this, that might also be more explicit:

```suggestion
  ASSERT1(hasParallelSlices());
```

include/field3d.hxx
```diff
@@ -257,7 +257,18 @@ class Field3D : public Field {
...
+      throw BoutException(
+          "Field3D::splitParallelSlices: forward/backward parallel slices not in sync.\n"
+          "    This is an internal library error");
+    }
+#endif
+#if CHECK
     return !yup_fields.empty() and !ydown_fields.empty();
+#else
+    return !yup_fields.empty();
+#endif
```
ZedThree, line 271: Moving the check that the sizes equal here is good idea :+1:

The optimisation if `CHECK` is off feels a little excessive, but I don't think it hurts :smile:

ZedThree (approved):

PR #2541, opened 2022-04-29
Only destroy MatP in LaplaceXZ if it exists

dschwoerer:
Avoids segfault if the instance gets destroyed before it is used.


ZedThree:
Please could you tweak the PR title? It's not clear what this refers to! :slightly_smiling_face: 

I'm also a little concerned as I thought `MatDestroy` was safe to call on uninitialised matrices. Looking at the source code though, it seems that's not the case. That feels like a bug in PETSc :(

dschwoerer:
Is the new title ok?

As long as it leads to a segfault it is fairly easy to spot (however might be ignored if it is only at the end of a simulation :disappointed: )

ZedThree (approved):

PR #2546, opened 2022-05-13
Output BoutException.what() in PhysicsModel main()

bendudson:
Some errors don't have anything useful in the backtrace, so the only message on error is an unhelpful "Error encountered".

Now outputs `e.what()` which is often more useful.


ZedThree:
Should this perhaps go into master instead?

clang-tidy-review is failing due to some issue with the Github `checkout` action I think

Issue #2548, opened 2022-05-18
Does CyclicLaplace actually work with periodicX=true?


johnomotani:
I just tried running the Hasegawa-Wakatani example with `periodicX = true` (also using `hw:modified=false`, but I don't think that's relevant here). It did run, but very badly until I also set the `INVERT_ZERO_DC` flag. That makes me suspect that there is an issue with the DC component when `periodicX = true`. I would have thought it needed some special handling (e.g. using Dirichlet boundary conditions for the DC component, so that the kz=0, kx=0 piece is fixed), but don't see any in `src/invert/laplace/impls/cyclic/cyclic_laplace.cxx` or `include/cyclic_reduction.hxx`. There is some handling of 'periodic' that is the same for all kz's. Does anyone know whether Laplace solves with `periodicX = true` ought to work?

bendudson:
@johnomotani I think it works in the sense that the code does calculate what it's designed to. Unfortunately the design has a problem: If both X and Z are periodic then the (0,0) mode is unconstrained, as it is when Neumann boundary conditions are applied to both boundaries, so the matrix being inverted for the `kz=0` component is singular. It seems to have a go at a solution, but runs poorly because I guess the solution is very numerically unstable.

Setting `INVERT_ZERO_DC` removes the troublesome (0,0) mode, but unfortunately also removes all (kx, 0) modes too, which are the interesting zonal modes.

To fix this we would need to remove the null-space (0,0) mode. I think we probably want to remove the average phi, rather than pin an arbitrary point to zero, but doing that in the matrix inversion breaks the tridiagonal structure. For the `kz = 0` mode we can probably fix an arbitrary X location to zero in the tridiagonal inversion, calculate phi, and then remove the X average...?

johnomotani:
@bendudson sounds reasonable to me! :+1: @spamela was asking me about Hasegawa-Wakatani, so I'll leave it up to the interested people whether this is worth pushing on. Should be a fairly simple fix though.

spamela:
Hey all,
Yes, there is definitely interest in having properly periodic boundary conditions for the HW model.
Let me know!

johnomotani:
So if someone wants to work on this (i.e. make a branch off `next` and submit a PR), I think the main thing is that there needs to be a special case for `localmesh->periodicX == true` here
[...]

PR #2557, opened 2022-07-06
PeriodicX option in cyclic Laplace solver

bendudson:
If `periodicX = true` then the (kx,kz) = (0,0) mode is unconstrained and the Laplacian matrix inversion of vorticity to get potential becomes singular.

The solution is to force solutions to have a zero average potential.
This fix retains the tridiagonal form of the matrix by treating the kz=0 mode specially:
- In the tridiagonal solve fix one location (x=xstart) to zero
- After the solve, calculate and subtract the X average of the kz=0
  modes i.e. set the (0,0) mode to zero.

For the Hasegawa-wakatani model this results in solutions which are periodic in X, have zonal (kz=0) modes, and run at least as quickly as the non-periodic simulations.

Fixes issue #2548

### DIFF
--- a/src/invert/laplace/impls/cyclic/cyclic_laplace.cxx
+++ b/src/invert/laplace/impls/cyclic/cyclic_laplace.cxx
@@ -240,11 +240,41 @@ FieldPerp LaplaceCyclic::solve(const FieldPerp& rhs, const FieldPerp& x0) {
                      &C1coef, &C2coef, &Dcoef,
                      false); // Don't include guard cells in arrays
       }
+      if (localmesh->periodicX and localmesh->firstX()) {
+        // Special case for kz = 0
+        // Need to exclude the kx=0,kz=0 mode
+        // - Set an arbitrary location to zero
+        // - Solve periodic tridiagonal system
+        // - Subtract average of result
+
+        a(0,0) = 0.0;
+        b(0,0) = 1.0;
+        c(0,0) = 0.0;
+        bcmplx(0, 0) = 0.0;
+      }
     }
 
     // Solve tridiagonal systems
     cr->setCoefs(a, b, c);
     cr->solve(bcmplx, xcmplx);
+
+    if (localmesh->periodicX) {
+      // Subtract X average of kz=0 mode
+      BoutReal local[2] = {
+        0.0, // index 0 = sum of coefficients
+        static_cast<BoutReal>(xe - xs + 1) // number of grid cells
+      };
+      for (int ix = xs; ix <= xe; ix++) {
+        local[0] += xcmplx(0, ix-xs).real();
+      }
+      BoutReal global[2];
+      MPI_Allreduce(local, global, 2, MPI_DOUBLE,
+                    MPI_SUM, localmesh->getXcomm());
+      BoutReal avg = global[0] / global[1];
+      for (int ix = xs; ix <= xe; ix++) {
+        xcmplx(0, ix-xs) -= avg;
+      }
+    }
 
     // FFT back to real space
     BOUT_OMP(parallel) {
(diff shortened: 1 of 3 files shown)


src/invert/laplace/impls/cyclic/cyclic_laplace.cxx
```diff
@@ -240,12 +240,41 @@ FieldPerp LaplaceCyclic::solve(const FieldPerp& rhs, const FieldPerp& x0) {
...
+        // Need to exclude the kx=0,kz=0 mode
+        // - Set an arbitrary location to zero
+        // - Solve periodic tridiagonal system
+        // - Subtract average of result
+
+        a(0, 0) = 0.0;
+        b(0, 0) = 1.0;
+        c(0, 0) = 0.0;
+        bcmplx(0, 0) = 0.0;
+      }
```
johnomotani, line 254: Should this go in `Laplace::tridagMatrix()`? That might mean other solvers (e.g. `pcr`) get a more sensible result if they're run with `periodicX`, although the code to subtract the x-average would have to go in each individual solver to fully support it.

src/invert/laplace/impls/cyclic/cyclic_laplace.cxx
```diff
@@ -466,13 +495,43 @@ Field3D LaplaceCyclic::solve(const Field3D& rhs, const Field3D& x0) {
...
                      &C1coef, &C2coef, &Dcoef,
                      false); // Don't include guard cells in arrays
+
+        if (localmesh->periodicX and localmesh->firstX() and kz == 0) {
+          // Special case kz = 0 and kx = 0
+          a3D(ind, 0) = 0.0;
+          b3D(ind, 0) = 1.0;
+          c3D(ind, 0) = 0.0;
+          bcmplx3D(ind, 0) = 0.0;
+        }
```
johnomotani, line 505: Should this go in `Laplace::tridagMatrix()`?

johnomotani (commented):
Implementation looks good to me (minor comment below).

I think we should add a test. It's unfortunate that `tests/integrated/test-laplace` tests against expected data saved in a binary file, so not ideal to add new cases to that test. Maybe we could test in `tests/MMS/laplace`?

Also would be good to add a section to the manual mentioning `periodicX` and how to use it. E.g. that you should only use `laplace:type = cyclic`.

src/invert/laplace/impls/cyclic/cyclic_laplace.cxx
bendudson: Thanks @johnomotani yes this is a good idea, now done.

johnomotani:
@ZedThree I think this PR still needs a test, and `periodicX` needs some mention in the docs. Do we want to fix those here, or get this merged now and open issues for them?

jonycastagna:
Hi there, I just tried to compile this branch on Marconi, but it fails (strangely the master works). I then tried to merge on my local repo the branch into the master but I get several conflicts (I guess due to "next" being very different from "master"). Could you please look into it? Do you have intention to merge this anytime soon?

johnomotani:
@jonycastagna:
* have you checked whether your build issue is related to this PR (e.g. can you compile the current `next` branch?)? I suspect it is not related to this PR specifically.
* you are more likely to get a helpful response if you provide more detail than 'it fails'! For example at what stage did compilation fail, and what was the error message...

So I'm having to guess, but if your error was
```
-- Submodule update
You need to run this command from the toplevel of the working tree.
CMake Error at CMakeLists.txt:52 (message):
  git submodule update --init failed with 1, please checkout submodules
  tests/unit/CMakeLists.txt:1 (bout_update_submodules)


-- Configuring incomplete, errors occurred!
```
then the problem is the ancient version of CMake that's available on Marconi. A fix/workaround has been added to `master` in #2490, but that has not been merged into `next` yet. The easiest workaround for now is to pass `-DBOUT_UPDATE_GIT_SUBMODULE=OFF` to your `cmake` configuration command (and make sure that if submodules need updating that you run `git submodule update --init --recursive` in the BOUT-dev repo manually).

d7919:
> Hi @johnomotani, thanks a lot for your help. The quick fix suggested works (despite my poor instructions, sorry!). However, testing now the hasegawa-wakatani example and getting the following error after "gmake" Compiling hw.cxx In file included from ../../include/bout.hxx:39:0, from ../../include/bout/physicsmodel.hxx:40, from hw.cxx:2: ../../include/bout/build_config.hxx:4:10: fatal error: bout/build_defines.hxx: No such file or directory #include "bout/build_defines.hxx" ^~~~~~~~~~~~~~~~~~~~~~~~ compilation terminated. gmake: *** [hw.o] Error 1
> 
> Any hint?

I suspect you may need to try using cmake to build the example if you have configured BOUT++ with cmake as well. @ZedThree  may be able to help?

PR #2559, opened 2022-07-20
implement global FCI indexing

bshanahan:
Attempts to index FCI X-indexes globally, allowing for splitting in X. In my limited testing, it seems to work.

@dschwoerer @bendudson @ZedThree  -- I would really appreciate a glance over this.

PR against David's branch as it is the current workhorse for FCI/3D metrics.


dschwoerer:
Unfortunately that does not work in general :-(

PR #2560, opened 2022-07-21
Use MPI_COMM_NULL rather than nullptr

bendudson:
Some MPI implementations use pointers for MPI_Comm types, but others use integers. MPI_COMM_NULL should always be defined.


ZedThree (approved):

ZedThree:
clang-tidy-review is failing because we need to bump the version of cmake that it uses

PR #2568, opened 2022-07-29
Add tool to move headers under `include/bout` and fix `#include`s

ZedThree:
Replaces similar bash tool, but tries to preserve history by first moving headers into a single commit, then adding compatibility shim headers (that issue a `#warning` and `#include` the new location) in a second commit. Also when updating `#include`s, it prints a diff of the changes, to allow the user to review before applying (just in case something goes wrong!).

Should be essentially last PR into 5.0 to avoid massive conflicts.

To apply, run in two steps:

1. `./bin/bout-v5-header-upgrader.py --files $(find . -type f -name "*.[hc]xx")` to make sure all the `#include`s point to the new location
2. `./bin/bout-v5-header-upgrader.py --move-deprecated-headers` to do the actual moving

The first command can then be used to fix user code and models. The second command should only need to be run once, ever.


bin/bout-v5-header-upgrader.py
```diff
@@ -0,0 +1,217 @@
...
+    # Now we can write the compatibility shim
+    for header in unmoved_headers:
+        write_header_shim(header)
+
+    run(f"git add {' '.join(map(str, unmoved_headers))}", shell=True, check=True)
+    run(
+        r"git commit -m 'Add compatibility shims for old header locations'",
+        shell=True,
+        check=True,
+    )
```
dschwoerer, line 80: Why not add as a single commit? That makes git-bisect less useful ...

ZedThree: This is to try and make sure that git tracks the first commit as a rename. I found that doing this in a single commit (at least sometimes, in my limited testing) meant that git couldn't track changes across the commit.

You're right that it potentially makes bisects a bit more painful, but that's easier to work around with `git bisect skip`. Bisects are probably also (hopefully!) rarer than `git blame` for instance

bendudson:
Trouble running `./bin/bout-v5-header-upgrader.py` here: The regex doesn't seem to match most `#include` lines that need changing. This is on a Mac, will try under Linux...

bin/bout-v5-header-upgrader.py
```diff
@@ -0,0 +1,217 @@
...
+
+
+def make_header_regex(deprecated_headers: List[str]) -> re.Pattern:
+    """Create a regular expression to match deprecated header locations"""
+    deprecated_header_joined = "|".join((header.name for header in deprecated_headers))
+    return re.compile(rf'(#include\s+<|")(?:\.\./)({deprecated_header_joined})(>|")')
```
bendudson, line 86:
```suggestion
    return re.compile(rf'(#include\s+<|")(?:\.\./)?({deprecated_header_joined})(>|")')
```
Without the `?` after the `(?:\.\./)` group this only matches includes containing "../".

ZedThree: Thanks @bendudson! I'm not sure how I missed this, especially as I definitely had this working

ZedThree:
Turns out there's a few other files that need the paths updating. Fix incoming.

Issue #2576, opened 2022-08-25
Documentation recommends broken setup


dschwoerer:
`    $ cmake -S BOUT-dev -B build_bout -DBOUT_DOWNLOAD_NETCDF_CXX4=ON`
https://github.com/boutproject/BOUT-dev/blame/next/manual/sphinx/user_docs/quickstart.rst#L42

This results in only 4 tests being run, and none of them working:

```
0% tests passed, 4 tests failed out of 4

Total Test time (real) =   4.31 sec

The following tests FAILED:
         39 - MMS-diffusion (Failed)
         40 - MMS-spatial-fci (Failed)
         41 - MMS-wave-1d (Failed)
         42 - MMS-wave-1d-y (Failed)
```

dschwoerer:
It seems this fftw related, for some reason the requires: fftw was droppen when switching to cmake. @ZedThree is this intentionally?

dschwoerer:
Sorry, seems I checked the single `./runtest` which included `# requires: fftw` - all the other ones where missing.
Will add to the cmake files ...

dschwoerer:
Was FFTW related so should be fixed now

PR #2579, opened 2022-08-29
Add requires for FFTW to tests

dschwoerer:
Several test require (optional) FFTW. Skip those if FFTW is not present.

Resolves: #2576


ZedThree (approved):
Nice catch, thanks @dschwoerer!

PR #2589, opened 2022-09-13
Improve boutpp compatibility with numpy

dschwoerer:
Improve agreement with numpy.

Next would probably to support views, so that modification of slices can be done, so something like `f3d[:2]` += 2`.

Also adds some tests and enables tests for cmake.

Should I split into several PRs?

Todo:
- [x] fix segfault (so far I cannot reproduce)
- [x] fix cmake dependency


tests/integrated/test-boutcore/slicing/notworking
```diff
@@ -0,0 +1,9 @@
+x[:, np.newaxis, :, :].shape
+x[:, None, :, :].shape
+x[:, np.newaxis] + x[np.newaxis, :]
+x[~np.isnan(x)]
+x[x < 0]
+x[x < 0] += 20
+x[2:7] = 1
+x[2:7] = np.arange(5)
+x[np.array([1, 1, 3, 1])] += 1
```
ZedThree, line 9: Are these examples of slices that won't work? Could you instead add a `README.md` to this directory and list them there?

tests/integrated/test-boutcore/slicing/runtest
```diff
@@ -0,0 +1,45 @@
...
+print(nout.shape)
+fout = {ex.replace('x', 'field')}
+assert fout.shape == nout.shape , f"Field3D returned {{ fout.shape }} but numpy {{ nout.shape }}"
+assert np.all(fout == nout) , f"data mismatch, {{ fout == nout }}"
+"""
+        # for l in toeval.split("\n"):
+        #     if not l:
+        #         continue
+        #     print(l)
+        #     eval(l)
```
ZedThree, line 37: Some commented out code

tests/integrated/test-boutcore/slicing/runtest
```diff
@@ -0,0 +1,45 @@
...
+        try:
+            print("running", ex)
+            exec(toeval)
```
ZedThree, line 40: Another option instead of `exec` and `str.replace` might be a list of `lambda`s. We could then just call each one with `ndat`/`field`

tools/pylib/_boutcore_build/boutcore.pyx.jinja
```diff
@@ -13,17 +13,21 @@ import copy
 from libcpp.memory cimport unique_ptr
 from cython.operator import dereference as deref
 
+size_t = np.uint
```
ZedThree, line 16: This looks a little surprising! Is it just standing in for the real `size_t`, or is it that the `indices` below do get passed as `np.uint`?

tools/pylib/_boutcore_build/boutcore.pyx.jinja
```diff
@@ -380,18 +400,29 @@ cdef class {{ field.field_type }}:
...
-        cdef np.ndarray[double, mode="c", ndim={{ field.ndims }}] data_ = np.ascontiguousarray(np.zeros(dims))
-        c_get_{{ field.fdd }}_part(self.cobj,&data_[{{ zeros }}], {{ field.makelist("inds[$i][0],inds[$i][1],inds[$i][2]") }});
-        return data_
-
-    def __setitem__(self,slices,data_):
+        cdef np.ndarray[double, mode="c", ndim=1] data_flat
+        cdef np.ndarray[size_t, mode="c", ndim=1] indices
+        slices = self._sanitise_slices(slices)
+        if isinstance(slices, tuple) and all([isinstance(x, slice) for x in slices]):
+            print("use custom code for", *slices)
```
ZedThree, line 407: Left in from debugging?

tools/pylib/_boutcore_build/boutcore.pyx.jinja
```diff
@@ -400,29 +431,34 @@ cdef class {{ field.field_type }}:
...
-        if isinstance(data_,{{ field.field_type }}):
-            # get all to get a numpy array
-            data_ = data_[{{ field.makelist(":") }}]
-        elif isinstance(data_,numbers.Number):
-            c_set_{{ field.fdd }}_part_(self.cobj,<double>data_,{{ field.makelist("inds[$i][0],inds[$i][1],inds[$i][2]") }})
+        cdef np.ndarray[double, mode="c", ndim=1] data_flat
+        slices = self._sanitise_slices(slices)
+        if isinstance(slices, tuple) and all([isinstance(x, slice) for x in slices]):
+            dims = []
+            import numbers
```
ZedThree, line 438: This should probably be at the top of the file, rather than in a method

tools/pylib/_boutcore_build/boutcore.pyx.jinja
```diff
@@ -442,10 +478,27 @@ cdef class {{ field.field_type }}:
...
+            # convert to numpy and return
+            dat = self.get()
+            return dat.__array_ufunc__(ufunc, method, (dat,), **kwargs)
+        return NotImplemented
+
 {% for name, op in ops %}
     def __{{ name }}__(self,other):
-        fu = {{ field.field_type }}()
         import numbers
+        fu = {{ field.field_type }}()
```
ZedThree, line 501: `fu` is not a great name. Is this really `result`?

tools/pylib/_boutcore_build/boutcore.pyx.jinja
```diff
@@ -457,13 +510,12 @@ cdef class {{ field.field_type }}:
...
         else:
-            print("{{ name }}",type(self),type(other),isinstance(self,numbers.Number))
-            return NotImplemented
+            return other.__r{{ name }}__(self)
         return fu
 
     def __r{{ name }}__(self,lhs):
         fu = {{ field.field_type }}()
-        if isinstance(lhs,float) or isinstance(lhs,int):
+        if isinstance(lhs, float) or isinstance(lhs, int):
```
ZedThree, line 518:
```suggestion
        if isinstance(lhs, (float, int)):
```

tools/pylib/_boutcore_build/boutcore.pyx.jinja
```diff
@@ -442,10 +478,27 @@ cdef class {{ field.field_type }}:
...
+    def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
+{% for fun in "sqrt", "exp", "sin", "cos", "log", "abs" %}
+        if ufunc == np.{{ fun }}:
+            assert method == "__call__"
+            assert len(inputs) == 1
+            assert kwargs == {}
+            assert self == inputs[0]
+            return {{ field.fdd }}FromObj(c.{{ fun }}(self.cobj[0]))
+{% endfor %}
+        print("Debug output:", ufunc, method, inputs, kwargs)
```
ZedThree, line 490:
```suggestion
```

tools/pylib/_boutcore_build/boutcore.pyx.jinja
dschwoerer: I couldn't find the real `size_t` in cython. `np.uint` is `size_t` - at least on `x86_64` - when we later build for fedora we will verify this for some other architectures as well. `np.uint` seems to be 64-bit on 64-bit platforms, and I guess `size_t` is always unsigned, so that should cover many architectures. Luckily the compiler should raise an error if there is a mismatch, so I guess this is good enough for now.

tests/integrated/test-boutcore/slicing/runtest
dschwoerer: removed

tools/pylib/_boutcore_build/boutcore.pyx.jinja
dschwoerer: :+1: removed

tools/pylib/_boutcore_build/boutcore.pyx.jinja
dschwoerer: I noticed there are a few more imports within methods:
`boutdata.collect`, `traceback`, `sys`, `gc`.

I would like to keep `boutdata` as a late import, as I think it is rather slow to import (250 ms on warm cache), and not often needed. `traceback` also seems to add ~10 ms, while the other ones seem to be negligible. Currently `import boutcore` takes around 160 ms, so `boutdata` would more than double that ...

(tested with python 3.11.0rc2, with `time python -c "import <x>"`)

tools/pylib/_boutcore_build/boutcore.pyx.jinja
dschwoerer: updated :+1:

tools/pylib/_boutcore_build/boutcore.pyx.jinja
dschwoerer: Good point, updated all of those.

tests/integrated/test-boutcore/slicing/runtest
dschwoerer: Initially I generated the list from the website, but I don't think I have the command to easily do it again, so probably no point in keeping it like this.

tools/pylib/_boutcore_build/boutcore.pyx.jinja
dschwoerer: Any suggestion for a good comment to explain that shortly?

dschwoerer (commented):
Thanks a lot for the valuable feedback :+1:

ZedThree (dismissed):

ZedThree:
Merging this; failing test is black workflow, to be fixed in #2596

PR #2595, opened 2022-10-04
Better versioning for BOUT++

dschwoerer:
Inspired by `setuptools_scm`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,12 +17,11 @@ endif()
 # CMake currently doesn't support proper semver
 # Set the version here, strip any extra tags to use in `project`
 # We try to use git to get a full description, inspired by setuptools_scm
-# Note we use distance from 4.0.0 so it does not reset on minor releases.
-# string(TIMESTAMP date "%Y%m%d" UTC)
+set(_bout_previous_version "v4.0.0")
+set(_bout_next_version "5.0.0.alpha")
 execute_process(
-  COMMAND "git" describe --tags --match=v4.0.0
-  COMMAND sed -e s/v4.0.0-/5.0.0.alpha.dev/ -e s/-/+/
-  # -e "s/$/.d${TIMESTAMP}/"
+  COMMAND "git" describe --tags --match=${_bout_previous_version}
+  COMMAND sed -e s/${_bout_previous_version}-/${_bout_next_version}.dev/ -e s/-/+/
   RESULTS_VARIABLE error_codes
   OUTPUT_VARIABLE BOUT_FULL_VERSION
   OUTPUT_STRIP_TRAILING_WHITESPACE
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -16,9 +16,24 @@ endif()
 
 # CMake currently doesn't support proper semver
 # Set the version here, strip any extra tags to use in `project`
-set(BOUT_FULL_VERSION 5.0.0-alpha)
-string(REGEX REPLACE "^([0-9]+\.[0-9]+\.[0-9]+)-.*" "\\1" BOUT_CMAKE_ACCEPTABLE_VERSION ${BOUT_FULL_VERSION})
-string(REGEX REPLACE "^[0-9]+\.[0-9]+\.[0-9]+-(.*)" "\\1" BOUT_VERSION_TAG ${BOUT_FULL_VERSION})
+# We try to use git to git a full description, inspired by setuptools_scm
```
d7919, line 19:
```suggestion
# We try to use git to get a full description, inspired by setuptools_scm
```

CMakeLists.txt
```diff
@@ -16,9 +16,24 @@ endif()
...
+# Not we use distance from 4.0.0 so it does not reset on minor releases.
```
d7919, line 20:
```suggestion
# Note we use distance from 4.0.0 so it does not reset on minor releases.
```

dschwoerer:
Thanks @d7919

bendudson (dismissed):

dschwoerer:
I actually tried `setuptools_scm` - but that would only bump to 4.4.$(n+1).dev - and there is no option to get 5.0.0.dev - only 5.1.0 or 5.0.1.dev, haven't tested, not sure. The issue is here: https://github.com/pypa/setuptools_scm/issues/523
But that wouldn't work nice, because if we than merge 4.5 into next, it would use that instead of 5.0.0dev, Anyway, I wouldn't know how to programmatically decide on whether this is going to 5.0.0 or 4.5.0 or 4.4.x ...

So, yes, unfortunately, it needs manual bumping.

I think setting BOUT_PREVIOUS_VERSION and BOUT_NEXT_VERSION might be nicer :+1:

ZedThree:
Ah, good point. That is a bit of a shame. Also very true that we just decide what the next version will be, and don't really use an algorithm. Nevermind, this is still nice!

ZedThree:
I've changed the hardcoded versions to variables, and updated the release checklist to reflect this.

I did consider making the variables settable from the command line (by making them cache variables), which would make it easier to test and change the variables for maintainers, but would probably cause more problems than it would solve when the version doesn't update automatically.

ZedThree:
This can go in as soon as the tests complete

PR #2596, opened 2022-10-04
cmake install fixes

dschwoerer:
Installs ~~boutcore~~ boutpp

Resolves one of the points of: https://github.com/boutproject/BOUT-dev/issues/2321


CMakeLists.txt
```diff
@@ -354,6 +354,15 @@ This warning is only important for BOUT++ developers and can otherwise be \
...
 
+execute_process(COMMAND ${Python3_EXECUTABLE} -c "import site ; print('/'.join(site.getusersitepackages().split('/')[-2:]))"
+  RESULT_VARIABLE PYTHON_WORKING
+  OUTPUT_VARIABLE PYTHON_SITEPATH_SUFFIX
+  OUTPUT_STRIP_TRAILING_WHITESPACE
+)
+set(CMAKE_INSTALL_PYTHON_SITEARCH lib/${PYTHON_SITEPATH_SUFFIX} CACHE STRING "Location to install python arch-specific modules")
+
+# message(FATAL_ERROR ${CMAKE_INSTALL_PYTHON_SITEARCH})
+
```
ZedThree, line 365:
```suggestion
```

ZedThree (changes requested):
LGTM, thanks @dschwoerer !

dschwoerer:
rpath is broken, so still WIP

dschwoerer:
Should be working now :+1: 

Unlike before the cython library is not directly imported, but a wrapper imports everything (except symbols starting with `_`)

dschwoerer:
Nice to have:
- [ ] Tests:
   - [ ] build sdist
   - [ ] build wheel
   - [ ] pip install sdist
- [ ] Documentation?
- [ ] a modulefile, so we can do

```
pip install boutpp
module use $(python -c 'import site; print(site.getusersitepackages())'/boutpp/share/module
module load bout++/python
```

Should this be split into several PRs?
Renaming as part of the PR was potentially a bad idea, I worry that rebasing will lead to many conflicts :see_no_evil:

CMakeLists.txt
```diff
@@ -598,9 +605,11 @@ if(BOUT_TESTS)
   add_subdirectory(tests/MMS EXCLUDE_FROM_ALL)
 
   # Targets for running the tests
-  add_custom_target(check-unit-tests
-    COMMAND ctest -R serial_tests --output-on-failure)
-  add_dependencies(check-unit-tests build-check-unit-tests)
+  if (BOUT_ENABLE_UNIT_TESTS)
```
ZedThree, line 608: I don't see where the option for this is?

tools/pylib/_boutpp_build/backend.py
```diff
@@ -0,0 +1,216 @@
...
+        try:
+            tmp = run2("git describe --tags --match=v4.0.0")
+            if useLocalVersion:
+                tmp = run2(
+                    f"echo {tmp.strip()} | sed -e s/v4.0.0-/v5.0.0.dev/ -e s/-/+/"
+                )
+            else:
+                tmp = run2(
+                    f"echo {tmp.strip()} | sed -e s/v4.0.0-/v5.0.0.dev/ -e s/-.*//"
+                )
```
ZedThree, line 34: Why not `re.sub` rather than starting another process?

tools/pylib/_boutpp_build/backend.py
```diff
@@ -0,0 +1,216 @@
...
+    ret = os.system(cmd)
+    assert ret == 0, f"{cmd} failed with {ret}"
+
+
+useLocalVersion = True
+
+
+def getversion(_cache={}):
+    if "r" not in _cache:
+        try:
```
ZedThree, line 25: Flip conditional:

```suggestion
    if "r" in _cache:
        return _cache["r"].strip()
```

then rest of function doesn't need to be indented.

Also, what is `r`? Is it really `version`?

tools/pylib/_boutpp_build/backend.py
```diff
@@ -0,0 +1,216 @@
...
+def run(cmd):
+    print(f"running `{cmd}`")
+    ret = os.system(cmd)
+    assert ret == 0, f"{cmd} failed with {ret}"
+
+
+useLocalVersion = True
+
+
+def getversion(_cache={}):
```
ZedThree, line 23: Normally mutable objects as default arguments is not what you want, but perhaps it is useful here?

tools/pylib/_boutpp_build/backend.py
```diff
@@ -0,0 +1,216 @@
...
+    return _cache["r"].strip()
+
+
+def run2(cmd):
+    child = subprocess.Popen(
+        cmd, stderr=subprocess.STDOUT, stdout=subprocess.PIPE, shell=True
+    )
+    output = child.stdout.read().decode("utf-8", "ignore")
+    child.communicate()
+    assert child.returncode == 0, f"{cmd} failed with {child.returncode}"
```
ZedThree, line 49: [`subprocess.run`](https://docs.python.org/3/library/subprocess.html#subprocess.run) has a much simpler interface:

```suggestion
    output = subprocess.run(cmd, capture_output=True, shell=True, check=True).stdout
```

tools/pylib/_boutpp_build/backend.py
```diff
@@ -0,0 +1,216 @@
...
+    )
+    run(f"cmake --build  _wheel_build/ -j {os.cpu_count()}")
+    run("cmake --install _wheel_build/")
+    distinfo = f"_wheel_install"
+    prepare_metadata_for_build_wheel("_wheel_install", record=True)
+
+    # Do not add --symlink as python's does not extract that as symlinks
+    run(f"cd {trueprefix} ; zip  {wheel_directory}/{whlname} . -rq")
+    # cmd = f"git archive HEAD -o {wheel_directory}/{whlname}"
+    # run(cmd)
```
ZedThree, line 104:
```suggestion
```

tools/pylib/_boutpp_build/backend.py
```diff
@@ -0,0 +1,216 @@
...
+        run(f"rm {tmp}")
+
+    # _, tmpd = tempfile.mkdtemp()
+    with open(tmp, "w") as f:
+        f.write(
+            f"""Metadata-Version: 2.1
+Name: boutpp
+Version: {getversion()}
+License-File: COPYING
+"""
```
ZedThree, line 143: No need to change, but you might like to know about [`textwrap.dedent`](https://docs.python.org/3/library/textwrap.html#textwrap.dedent) :)

tools/pylib/_boutpp_build/backend.py
```diff
@@ -0,0 +1,216 @@
...
+
+
+def mkdir_p(path):
+    full = ""
+    for k in path.split("/"):
+        full += k + "/"
+        try:
+            os.mkdir(full)
+        except FileExistsError:
+            pass
```
ZedThree, line 178: [`pathlib.Path.mkdir(path, parents=True, exist_ok=True)`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.mkdir)

ZedThree (changes requested):
Using a custom backend is very clever and really nice! I have a few suggestions, and it could do with some docstrings.

Could `pyproject.toml` have the project metadata too, or does the backend handle all that?

I suspect this will need conflicts fixing after the boutpp numpy PR goes in.

Also, the black workflow is broken: https://github.com/boutproject/BOUT-dev/actions/runs/3377858435/jobs/5607304385#step:5:39
So that will need double-checking

.github/workflows/black-fix.yml
```diff
@@ -39,7 +39,7 @@ jobs:
       run: |
         pwd
         ls
-        $HOME/.local/bin/black tests/ tools/ $(grep '^#.*python.*$' bin/ tests/ tools/ src/ -Ir|cut -d: -f1|grep -v _boutcore_build)
+        $HOME/.local/bin/black tests/ tools/ $(grep '^#.*python.*$' bin/ tests/ tools/ src/ -Ir|cut -d: -f1|grep -v _boutpp_build)
```
ZedThree, line 42: This picks up `# Install python packages` in `bin/bout-build-deps.sh`

I tried using `file`, which mostly gets it correct, but there are still a few false positives and negatives

Maybe:

```suggestion
        $HOME/.local/bin/black tests/ tools/ $(grep '^#.*python.*$' bin/ tests/ tools/ src/ -Ilr | grep -v _boutpp_build)
```

ZedThree:
Oh dear, #2589 and this PR both renamed a bunch of files. I had a quick go, but @dschwoerer you're probably going to have to deal with this, I'm afraid!

CMakeLists.txt
dschwoerer: It is set in the unit tests folder:
https://github.com/boutproject/BOUT-dev/pull/2596/files#diff-18b716ba95f75d9677945e5d932f2ec46a2c0e919a2123b154420422f94bd059

tools/pylib/_boutpp_build/backend.py
dschwoerer: It is a "nice" way to have a function local cache. But maybe we can have a module-level cache?

tools/pylib/_boutpp_build/backend.py
dschwoerer: It is not have have the return in two spots ...

`r` is for `return`. Will change.

tools/pylib/_boutpp_build/backend.py
dschwoerer: Copy-pased from the cmake version. But yes, `re.sub` is the way to go.

tools/pylib/_boutpp_build/backend.py
dschwoerer: Great :tada:

tools/pylib/_boutpp_build/backend.py
dschwoerer: :+1:

dschwoerer (commented):
Thanks for the feedback :+1:

dschwoerer:
> Using a custom backend is very clever and really nice! I have a few suggestions, and it could do with some docstrings.
> 
> Could `pyproject.toml` have the project metadata too, or does the backend handle all that?

It could, but then it needs to be parsed + reformatted and printed. Might be easier to add it directly. But more metadata is one of the missing things.

> I suspect this will need conflicts fixing after the boutpp numpy PR goes in.

Yes. Should be fixed now.

> Also, the black workflow is broken: https://github.com/boutproject/BOUT-dev/actions/runs/3377858435/jobs/5607304385#step:5:39 So that will need double-checking

black seems to be fixed. The `!` did the job :-)

ZedThree (approved):

ZedThree:
Some tests failing like:

```
  File "/home/runner/work/BOUT-dev/BOUT-dev/build/tests/MMS/bracket/runtest", line 17, in <module>
    import boutpp as bc
ValueError: source code string cannot contain null bytes
```

https://github.com/boutproject/BOUT-dev/actions/runs/3385119775/jobs/5622897973#step:9:823

ZedThree (approved):

PR #2598, opened 2022-10-13
Insert Options attributes with initializer_list

bendudson:
This is probably more readable if setting multiple attributes:
```
Options options;
options["value"].setAttributes({
    {"units", "m/s"},
    {"conversion", 10.2},
    {"long_name", "important value"}
  });
```

~~Uses `std::map::emplace` internally, so doesn't replace or update values.~~
Replaces any values that were previously set.

Includes tests and some documentation.

### DIFF
--- a/include/options.hxx
+++ b/include/options.hxx
@@ -229,6 +229,13 @@ public:
       return *this;
     }
 
+    /// Initialise with a value
+    /// This enables AttributeTypes to be constructed using initializer lists
+    template <typename T>
+    AttributeType(T value) {
+      operator=(value);
+    }
+
     /// Cast operator, which allows this class to be
     /// assigned to type T
     /// This will throw std::bad_cast if it can't be done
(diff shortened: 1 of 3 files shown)


ZedThree (dismissed):
LGTM! I wonder if we should go even further and have something like:

```cpp
namespace bout {
struct Metadata {
  std::string units;
  std::string conversion;
  std::string long_name;
}
}
```

Maybe that actually adds too much complication for not enough benefit.

> Uses `std::map::emplace` internally, so doesn't replace or update values.

It would be good to note this in the docstring and the manual, might be surprising otherwise!

bendudson:
Thanks @ZedThree ! The semantics of `map::emplace` are quite weird, so I've changed it to `Options.setAttributes()` and it always updates/replaces any value already set.

ZedThree (approved):

PR #2600, opened 2022-10-25
Upload boutpp to pypi

dschwoerer:
Uploads releases to boutpp and nightly "builds" to [boutpp-nightly](https://pypi.org/project/boutpp-nightly).

You can now `pip install boutpp-nightly` and than just `import boutpp` from python - libraries are there and install.

You may want to do install with 3D metrics to simulate your favorite stellarator with `pip install boutpp-nightly --config-settings=-DBOUT_ENABLE_METRIC_3D=ON`

Based on #2596 (I didn't want that PR to keep growing)


ZedThree:
Are the different jobs intended to be tests, with a side effect of uploading the built artefacts on master/next?

dschwoerer:
Yes. I thought having separate jobs for upload would only waste CPU time

PR #2602, opened 2022-11-01
Utility to convert UEDGE to BOUT++ grids

bendudson:
Uses a combination of code adapted from INGRID (MIT license) and Hypnotoad (GPL-3)

Tested on one SN grid and two DN grids.

### DIFF
--- a/tools/tokamak_grids/gridue_to_bout
+++ b/tools/tokamak_grids/gridue_to_bout
@@ -547,7 +547,11 @@ Note that in most cases these grids are non-orthogonal."""
     dx = np.zeros((nx, ny))
     for i in range(nx):
         for j in range(ny):
-            dx[i, j] = 0.5 * (psi[j, i, 3] + psi[j, i, 4] - psi[j, i, 1] - psi[j, i, 2])
+            if i > 1 and i < nx-2:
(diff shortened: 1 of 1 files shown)


johnomotani:
Is this a thing we could add to hypnotoad? Even if just as a standalone script? Then we could use the CI setup to test that it at least keeps running without errors.

[I have no idea about licences, but maybe that's a thing we'd have to worry about??]

bendudson:
Hey @johnomotani I was unsure where the best place to put this would be. Talking to Mike it sounds like B2 grids are very similar to UEDGE grids, so it may be possible to make something more general that works for both. If put into hypnotoad, might it possible to load in a mesh and reuse the hypnotoad code to generate outputs? This script currently duplicates some of that code e.g metric tensor calculation.

ZedThree:
Getting it into hypnotoad sounds like a good idea. Would that make it possible to do the reverse conversion too?

bendudson:
Hey @ZedThree ! The reverse conversion is probably easier, as long as we have the cell corner locations and B field components at those corners.

johnomotani:
hypnotoad does have cell corner positions. Not sure if Bp, Bt are calculated at the cell corners already, but that should be easy to do.

bendudson:
Closing, putting script into Hypnotoad instead https://github.com/boutproject/hypnotoad/pull/136

PR #2603, opened 2022-11-02
CI: Fix black workflow

ZedThree:
It was picking up non-python files


ZedThree:
Using admin powers to merge this

PR #2605, opened 2022-11-02
Clang format everything

ZedThree:
Applies clang-format to the entire repo, including the new option "insert braces" (this requires clang-format 15, CI only has 14, so it's currently disabled in the config file)

.git-blame-ignore-revs contains the commit hashes of the two main commits that touch everything. If you add

```ini
[blame]
	ignoreRevsFile = .git-blame-ignore-revs
```

to your `~/.gitconfig`, then these commits should be ignored when running `git blame`


dschwoerer:
Can we use clang 15 in CI? We could use e.g. the fedora images, they come with clang 15 ...

PR #2612, opened 2022-11-23
Fix input file expressions docs

johnomotani:
@mikekryjak pointed out there was an error in the expressions for `y` in tokamak geometry in the docs added in #2578.

I've also taken the chance to make a few other minor updates:
* note that section headings are case sensitive in one place
* add a definition of `z` in input file expressions
* add a note on how expressions are evaluated on staggered grids


bendudson (approved):
Thanks @johnomotani

PR #2626, opened 2023-01-06
Fix FV operators in double null configurations

bendudson:
The `Mesh::firstY()` and `Mesh::lastY()` functions can be used in slab and single-null tokamak geometry to determine if a cell is at a Y boundary. In double null geometries this fails because the upper target boundaries are in the middle of the Y domain.

The versions of these functions taking an X index use the communicators to determine if a point is first or last on its flux surface.

This fix replaces calls to no-argument `firstY()` and `lastY()` with calls the versions taking an X index. It also deprecates the no-argument functions as they are likely to cause similar confusion in future.

### DIFF
--- a/include/bout/fv_ops.hxx
+++ b/include/bout/fv_ops.hxx
@@ -27,7 +27,7 @@ Div_a_Laplace_perp(const Field3D& a, const Field3D& x) {
    * Divergence of a parallel diffusion Div( k * Grad_par(f) )
    */
   const Field3D Div_par_K_Grad_par(const Field3D &k, const Field3D &f, bool bndry_flux=true);
-  
+
   /*!
    * 4th-order derivative in Y, using derivatives
    * on cell boundaries.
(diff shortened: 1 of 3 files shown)


johnomotani (commented):
When using `firstY(ix)` and `lastY(ix)`, is it also necessary to check `periodicY(ix)` the way `D4DY4_Index()` does?
https://github.com/boutproject/BOUT-dev/blob/3bf597b130118658bf3344f7eb199cc2b9c8fcfd/src/mesh/fv_ops.cxx#L363-L366

`FV::Div_par` does something similar
https://github.com/boutproject/BOUT-dev/blob/3bf597b130118658bf3344f7eb199cc2b9c8fcfd/include/bout/fv_ops.hxx#L220

There was something in the comments for `firstY(ix)`/`lastY(ix)` about just being the first/last in the communicator?

johnomotani (approved):
Looks good to me!

Issue #2624, opened 2022-12-29
Steady-state solver under-relaxation


bendudson:
Since the aim of the `beuler` solver is to find steady states, accurate time evolution is not needed and the time derivatives can be modified to improve convergence.

From @mikekryjak : 
> In ANSYS Fluent we have under-relaxation factors, which basically artificially reduce the derivatives by a factor on a per-equation basis:
> [https://www.afs.enea.it/project/neptunius/docs/fluent/html/th/node374.htm#sec-underrelax-theory](https://urldefense.us/v3/__https:/www.afs.enea.it/project/neptunius/docs/fluent/html/th/node374.htm*sec-underrelax-theory__;Iw!!G2kpM7uM-TzIFchu!j5EsSiyB_SqExDFsC6IRNbrCLMhzh-xRfq0kBNIiuOQfA6qhl9itYFd3LW6iZI7l$)

Scaling the time derivative of each equation may be related to the row scaling that UEDGE uses to accelerate convergence. Under-relaxation and more general pseudo-timestepping methods should be implemented in the `beuler` solver.

bendudson:
The Pseudo Transient Continuation (PTC) method is now implemented in #3181.

PR #2631, opened 2023-01-12
Improving backward Euler solver

bendudson:
Attempting to improve the steady state (beuler / snes) solver and address #2624

So far:
- If `scale_rhs = true` then the approximate Jacobian calculated by PETSc coloring is used to scale the time derivatives (RHS) by row scaling the Jacobian. This results in pseudo-timestepping, since every equation is effectively using a different timestep. Default is off (`false`)


bendudson:
Replaced by #3009

PR #2633, opened 2023-01-19
Fix `SAVE_ONCE` for `Vector2/3D`

### LINKED ISSUE #2632: `DataFileFacade` missing overloads for `Vector2/3D`
In v5, we currently can't easily save `Vector2/3D` due to missing the `DataFileFacade::add` overloads.

I'm not sure if we want to be able to store `Vector`s in `Options`, and whether that requires more work to be able to do so.

ZedThree:
Add missing `DataFileFacade::add(Vector2/3D)` overloads, along with a test

Fixes #2632 

I couldn't see a decent existing place to add a test, so I added a whole new one


tests/integrated/test-datafilefacade/runtest
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env python3
+#
+# Test file I/O by loading from restart files and writing to dump files
+#
+# requires: netcdf
+# cores: 4
```
dschwoerer, line 6:
```suggestion
# cores: 2
```

dschwoerer (dismissed):

tests/integrated/test-datafilefacade/runtest
ZedThree: Oh yes, and probably should fix that comment above too

PR #2636, opened 2023-02-03
Fix use of removed `Datafile` in `Laplacian` constructor; update docs

ZedThree:
There was still a use of `Datafile` in the `Laplacian` constructor which one of the examples (`hw-3D`) was using, which meant it didn't compile.

I (finally!) added some docs on the new file IO system that relies on `OptionsNetCDF`, and how to update physics models to use it.

Also fix the predefined macros for doxygen so it preprocesses everything correctly.

Lastly, just for fun, I've updated the docs to use [sphinx-book-theme](https://sphinx-book-theme.readthedocs.io/en/stable/) as a refresh for v5. It's cleaner and easier to navigate than the default readthedocs theme.

This is (hopefully) truly the last PR for v5 -- everything else is getting bumped to 5.1, unless there are new build-blocking bugs discovered in the next week.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -87,11 +87,12 @@
   removed. This implementation had some quite significant drawbacks
   that made it not terribly useful in practice
   [\#3566](https://github.com/boutproject/BOUT-dev/pull/3566)
-- `DataFile` and `bout::globals::dump` have been removed in favour of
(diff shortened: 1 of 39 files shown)


ZedThree:
[Here's how the new docs theme looks](https://bout-dev--2636.org.readthedocs.build/en/2636/)

Any major objections to the new theme?

PR #2646, opened 2023-02-08
Improve invSg calculation

dschwoerer:
Do some caching and fixes for FCI


src/mesh/coordinates.cxx
```diff
@@ -1695,11 +1715,20 @@ Coordinates::FieldMetric Coordinates::Grad2_par2(const Field2D& f, CELL_LOC outl
...
+    (*invSg) = 1.0 / sqrt(g_22);
+  }
+  if (Grad2_par2_DDY_invSg == nullptr) {
+    // Communicate to get parallel slices
+    localmesh->communicate(*invSg);
+    invSg->applyParallelBoundary("parallel_neumann");
+    // cache
+    Grad2_par2_DDY_invSg = std::make_unique<FieldMetric>();
+    (*Grad2_par2_DDY_invSg) = DDY(*invSg, outloc, method) * (*invSg);
+  }
```
ZedThree, line 1729: Probably worth pulling these two bits out into separate helper functions, maybe something like:

```
const FieldMetric& invSg() {
  if (invSg == nullptr) {
    invSg = std::make_unique<FieldMetric>();
    (*invSg) = 1.0 / sqrt(g_22);
  }
  return *invSg;
}
```
and then it can be used immediately as `DDY(invSg()) * invSg()`

src/mesh/coordinates.cxx
```diff
@@ -1695,11 +1715,20 @@ Coordinates::FieldMetric Coordinates::Grad2_par2(const Field2D& f, CELL_LOC outl
...
-  auto invSg = 1.0 / sqrt(g_22);
-  // Communicate to get parallel slices
-  localmesh->communicate(invSg);
-  auto result = DDY(invSg, outloc, method) * DDY(f, outloc, method) * invSg
-                + D2DY2(f, outloc, method) / g_22;
+  if (invSg == nullptr) {
+    invSg = std::make_unique<FieldMetric>();
+    (*invSg) = 1.0 / sqrt(g_22);
+  }
+  if (Grad2_par2_DDY_invSg == nullptr) {
```
ZedThree, line 1722: We presumably actually need a cached version for each possible `outloc` (and possibly `method` too?) -- unless `this` is already at `outloc`, in which case this happens automatically?

dschwoerer:
Thanks @ZedThree for your valuable feedback and suggestions :+1: 

I think I have addressed everything.

include/bout/coordinates.hxx
```diff
@@ -149,7 +149,7 @@ public:
 
   FieldMetric DDY(const Field2D& f, CELL_LOC outloc = CELL_DEFAULT,
                   const std::string& method = "DEFAULT",
-                  const std::string& region = "RGN_NOBNDRY");
+                  const std::string& region = "RGN_NOBNDRY") const;
```
ZedThree, line 152: Thanks! Bit by bit, we'll be `const` correct eventually

ZedThree (approved):

PR #2647, opened 2023-02-08
fix boutconfig's `has`

dschwoerer:


ZedThree (approved):

ZedThree:
Do we not have a test using this already? I'm not sure what one would look like exactly, maybe comparing `./some_model --print-config` to the config dict?

PR #2657, opened 2023-02-17
Update bundled boututils and boutdata

dschwoerer:
I think that may be the reason builds for fedora are failing, as there not the bundled copies are used and one test failure seems to be related to that ...

### DIFF
--- a/.ci_fedora.sh
+++ b/.ci_fedora.sh
@@ -49,7 +49,7 @@ then
     sudo -u test ${0/\/tmp/\/home\/test} $mpi
 ## If we are called as normal user, run test
 else
-    pip install --user zoidberg
+    pip install --user zoidberg natsort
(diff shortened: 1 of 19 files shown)


dschwoerer:
So the difference between old and new boututils/boutdata is this
```
		fperp_lower:yindex_global = 0 ;
```
changed to this:
```
		fperp_lower:yindex_global = 0LL ;
```

The question is, should this be reverted in the python packages?
Why does bout++ not handle reading the `LL` version?

Should we stop supporting evolving fperp?

ZedThree:
Is the issue that boutdata is now writing `int64` and we're trying to read `int32`?  Easiest thing is probably to enforce `int32` in the test. We _could_ add `int64` to BOUT++, but I rather suspect it'll be quite a lot of effort.

ZedThree (approved):

PR #2660, opened 2023-02-17
Error early with in-source-builds

dschwoerer:
* Previously the second cmake invocation failed, e.g. after some source changes or after changing a configure time option
* It also generated many files, some of which overwrite git-tracked files from autotools, and others where not git-ignored.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,7 +16,10 @@ if(${CMAKE_VERSION} VERSION_GREATER_EQUAL 3.22)
 endif()
 
 if ("${CMAKE_CURRENT_BINARY_DIR}" STREQUAL "${CMAKE_CURRENT_SOURCE_DIR}")
-  message(FATAL_ERROR "BOUT++ does not currently supports in source buildds. Try building out of source, e.g. with `cmake -S . -B build`")
+  option(BOUT_ALLOW_INSOURCE_BUILD "Whether BOUT++ should really allow to build in source." OFF)
+  if (NOT ${BOUT_ALLOW_INSOURCE_BUILD})
+    message(FATAL_ERROR "BOUT++ does not recommends in source builds. Try building out of source, e.g. with `cmake -S . -B build` or set -DBOUT_ALLOW_INSOURCE_BUILD=ON - but thinks may break!")
+  endif()
 endif()
 
 # CMake currently doesn't support proper semver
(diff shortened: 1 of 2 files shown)


dschwoerer:
It seems there is one use case for in-souce-build - RTD :sweat: 
There we never run cmake twice and don't care about a polluted source-tree ...

I personally find it annoying when I accidentally do an in-source build, but maybe there are legitimate use cases?

ZedThree:
I'm not sure about this, it doesn't seem to be something people usually do in CMakeLists and potentially just makes things more fragile.

dschwoerer:
As the second invocation of cmake fails, I would call the current case more fragile. It essentially means you need to manually delete some file, before you can run cmake again. I think this is fragile, as that might be much later, at which point I don't neccessarily remember how BOUT++ was configured, some dependent project has been configured, so just moving to a subdir is a pain.
Also, we do not support out-of-source-builds as soon as there is an in-source build, so that is another reason to hint the user to do out of source builds.

SO does not seem to object to disabling in-source-build [0], and others are recommending as well [1]. E.g. llvm and openCV have it disabled [2, 3]

[0] https://stackoverflow.com/questions/1208681/with-cmake-how-would-you-disable-in-source-builds
[1] https://medium.com/@onur.dundar1/cmake-tutorial-585dd180109b
[2] https://github.com/root-project/cling/issues/20
[3] https://stackoverflow.com/questions/45518317/in-source-builds-are-not-allowed-in-cmake

ZedThree:
Ok, I'm convinced

CMakeLists.txt
```diff
@@ -15,6 +15,13 @@ if(${CMAKE_VERSION} VERSION_GREATER_EQUAL 3.22)
   cmake_policy(SET CMP0127 NEW)
 endif()
 
+if ("${CMAKE_CURRENT_BINARY_DIR}" STREQUAL "${CMAKE_CURRENT_SOURCE_DIR}")
+  option(BOUT_ALLOW_INSOURCE_BUILD "Whether BOUT++ should really allow to build in source." OFF)
+  if (NOT ${BOUT_ALLOW_INSOURCE_BUILD})
+    message(FATAL_ERROR "BOUT++ does not recommends in source builds. Try building out of source, e.g. with `cmake -S . -B build` or set -DBOUT_ALLOW_INSOURCE_BUILD=ON - but thinks may break!")
```
ZedThree, line 21: Couple of typos:

```suggestion
    message(FATAL_ERROR "BOUT++ does not recommend in source builds. Try building out of source, e.g. with `cmake -S . -B build` or set -DBOUT_ALLOW_INSOURCE_BUILD=ON - but things may break!")
```

ZedThree (changes requested):

bendudson (approved):

PR #2664, opened 2023-02-22
Improve docs + fix boutpp docs (master)

dschwoerer:


dschwoerer:
tl;dr: the boutpp API documentation was missing, and is now back.

I should have added, this is the old state:
https://bout-dev.readthedocs.io/en/latest/_apidoc/boutpp.html
this is with the fixes:
https://bout-dev--2664.org.readthedocs.build/en/2664/_apidoc/boutpp.html

PR #2668, opened 2023-02-22
Some fixes for tokamak-2fluid example

ZedThree:
Not comprehensive, still fails to run successfully on basic input file (zero pivot in cyclic reduce)

Fixes:

- unused variables
- wrong variable names (including case)
- creating laplacians from wrong input sections
- some issues with staggering

Currently fails with:

```
====== Back trace ======
 -> CyclicReduce::solve on line 172 of '/home/peter/Codes/BOUT++/BOUT-dev/include/bout/cyclic_reduction.hxx'
 -> LaplaceCyclic::solve(Field3D, Field3D) on line 305 of '/home/peter/Codes/BOUT++/BOUT-dev/src/invert/laplace/impls/cyclic/cyclic_laplace.cxx'
 -> Solving for phi on line 546 of '/home/peter/Codes/BOUT++/BOUT-dev/examples/tokamak-2fluid/2fluid.cxx'
 -> Running RHS: PvodeSolver::rhs(0) on line 307 of '/home/peter/Codes/BOUT++/BOUT-dev/src/solver/impls/pvode/pvode.cxx'
 -> Initialising PVODE solver on line 93 of '/home/peter/Codes/BOUT++/BOUT-dev/src/solver/impls/pvode/pvode.cxx'

Zero pivot in CyclicReduce::reduce
```

The `d3d` cases are missing `ShiftAngle` from the grid file and so don't run for that reason.

---

Maybe we should have a clean out of the examples, and delete everything that doesn't successfully run?

### DIFF
--- a/examples/tokamak-2fluid/2fluid.cxx
+++ b/examples/tokamak-2fluid/2fluid.cxx
@@ -4,6 +4,7 @@
  * for cross-benchmarking etc.
  *******************************************************************************/
 
+#include "bout/bout_types.hxx"
 #include <bout/physicsmodel.hxx>
 
 #include <bout/derivs.hxx>
@@ -98,6 +99,9 @@ private:
   std::unique_ptr<Laplacian> aparSolver;
   Field2D acoef; // Coefficient in the Helmholtz equation
 
+  /// Location of possibly staggered fields
+  CELL_LOC stagger_loc = CELL_LOC::deflt;
+
   int init(bool UNUSED(restarting)) override {
     TRACE("int init(bool) ");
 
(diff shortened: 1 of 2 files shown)


examples/tokamak-2fluid/2fluid.cxx
```diff
@@ -206,79 +209,92 @@ class TwoFluid : public PhysicsModel {
...
     vort_include_pi = options["vort_include_pi"].withDefault(false);
 
-    lowPass_z = options["lowPass_z"].withDefault(-1);
+    lowPass_z = options["low_pass_z"].withDefault(-1);
+
+    ////////////////////////////////////////////////////////
+    // Equation terms
+
+    auto& ni_options = globalOptions["Ni"];
+    ni_options.setConditionallyUsed();
```
dschwoerer, line 218: Wouldn't it be nicer to only set that if the option is not used?

bendudson:
Merging for now as better than nothing. I think the correct thing to do is to remove examples that don't work or are not useful for e.g. the manual. If people want full physics models then those are in other repositories.

ThomasMellor:
Hello, I have recently installed bout and have been trying the examples. This is the error I get for the tokamak-2fluid example. Was this fixed since the issue was brought up?

PR #2676, opened 2023-03-14
More forgiving restart inputs

bendudson:
Convert errors to warnings when restarting.  It's sometimes useful to restart simulations with new equations enabled (e.g. turn on electromagnetic fields). In those cases restart files don't have all fields expected. This now results in warnings rather than errors.


dschwoerer:
Awesome - I just cherry picked this to enable drifts after the simulation had already settled down a bit :tada: 

Not sure the simulation will run, but this patch is working as expected :+1:

dschwoerer (approved):

Issue #2663, opened 2023-02-22
Latest boutdata and boututils are incompatible with BOUT++ v5.0.0


dschwoerer:
This issue is related to PR #2657

While trying to update them, I noticed they are incompatible.

So the difference between old and new boututils/boutdata is this
```
		fperp_lower:yindex_global = 0 ;
```
changed to this:
```
		fperp_lower:yindex_global = 0LL ;
```

The question is, should this be reverted in the python packages?
Why does bout++ not handle reading the `LL` version?

Should we stop supporting evolving fperp?

ZedThree:
I think this is fixed now

PR #2679, opened 2023-03-15
Update boututils and boutdata to working versions

dschwoerer:
boututils changed the default for datafile to netcdf4, which breaks reading the y position of the FieldPerp slices.

Workaround for #2663 
Alternative for https://github.com/boutproject/boututils/pull/46
This does not fix the issue though.

### DIFF
--- a/tests/integrated/test-restart-io/runtest
+++ b/tests/integrated/test-restart-io/runtest
@@ -51,7 +51,7 @@ except FileExistsError:
     pass
 
 with DataFile(
-    os.path.join(restartdir, "BOUT.restart.0.nc"), create=True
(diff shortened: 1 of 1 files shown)


tests/integrated/test-restart-io/runtest
```diff
@@ -51,7 +51,7 @@ except FileExistsError:
     pass
 
 with DataFile(
-    os.path.join(restartdir, "BOUT.restart.0.nc"), create=True
+    os.path.join(restartdir, "BOUT.restart.0.nc"), create=True, format="NETCDF3_64BIT"
```
ZedThree, line 54: Do we still need this with the update to boutdata?

dschwoerer: I think I fixed the issue: https://github.com/boutproject/boututils/pull/47

PR #2682, opened 2023-03-22
CI: switch to openmpi for fedora GHA

dschwoerer:
It seems mpich is muuuuch slower in CI then openmpi - should we switch the fedora one to also use openmpi?

There is some advantage of testing both mpich and openmpi, so I am not sure it makes sense to switch everything to openmpi, but faster tests are also nice ...

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -187,7 +187,7 @@ jobs:
         with:
           submodules: true
       - name: Build Fedora rawhide
-        run: ./.ci_fedora.sh setup mpich rawhide
+        run: ./.ci_fedora.sh setup openmpi rawhide
         shell: bash
         env:
           TRAVIS_BUILD_DIR: ${{ github.workspace }}


johnomotani:
I remember looking at some point at how to oversubscribe MPI tasks to processes. OpenMPI and MPICH had different arguments, and MPICH (if I remember right) said it really needed a couple of flags setting for performance to not be awful.
Don't remember exactly what I was looking at at the time. Maybe this: https://github.com/pmodels/mpich/blob/main/doc/wiki/faq/Frequently_Asked_Questions.md
> Q: Why does my MPI program run much slower when I use more processes?
A: The default channel in MPICH (starting with the 1.1 series) is ch3:nemesis. This channel uses busy polling in order to improve intranode shared-memory communication performance. The downside to this is that performance will generally take a dramatic hit if you oversubscribe your nodes. Oversusbscription is the case where you run more processes on a node than there are cores on the node. In this scenario, you have a few choices:
>
>Just don't run with more processes than you have cores available. This may not be an option depending on what you are trying to accomplish.
>Run your code on more nodes or on the same number of nodes but with larger per-node core counts. That is, your job size should not exceed the total core count for the system on which you are running your job. Again, this may not be an option for you, since you might not have access to additional computers.
>Configure your MPICH installation with --with-device=ch3:sock. This will use the older ch3:sock channel that does not busy poll. This channel will be slower for intra-node communication, but it will perform much better in the oversubscription scenario.

dschwoerer:
The first two options are useless, and the third one needs to be changed at compile time.
Of course I could try to compile mpich myself and use that, but I do not think that is worth my time ...

johnomotani:
@dschwoerer That rings a bell, I think it's why I gave up on MPICH at the time!
Sorry, that FAQ didn't help at all.

ZedThree:
I think it's fine to only test with OpenMPI, we can't test every combination of every library.

ZedThree (approved):

PR #2683, opened 2023-03-22
Python: If we are on a tag, we should use that as version

dschwoerer:


ZedThree (approved):
I'm a bit worried this getting complex -- should we just require `setuptools_scm` and use that?

Issue #2654, opened 2023-02-10
warning: format specifies type 'long' but the argument has type 'pvode::integer' and other warnings


yurivict:
clang complains:
```
/usr/ports/science/bout++/work/BOUT-dev-5.0.0/externalpackages/PVODE/source/cvode.cpp:496:28: warning: format specifies type 'long' but the argument has type 'pvode::integer' (aka 'int') [-Wformat]
    fprintf(fp, MSG_BAD_N, N);
                ~~~~~~~~~  ^
/usr/ports/science/bout++/work/BOUT-dev-5.0.0/externalpackages/PVODE/source/cvode.cpp:1829:1: warning: non-void function does not return a value in all control paths [-Wreturn-type]
}
^
/usr/ports/science/bout++/work/BOUT-dev-5.0.0/externalpackages/PVODE/source/cvode.cpp:2395:1: warning: non-void function does not return a value in all control paths [-Wreturn-type]
}
^
/usr/ports/science/bout++/work/BOUT-dev-5.0.0/externalpackages/PVODE/precon/band.cpp:189:14: warning: using the result of an assignment as a condition without parentheses [-Wparentheses]
    if (swap = (l != k)) {
        ~~~~~^~~~~~~~~~
/usr/ports/science/bout++/work/BOUT-dev-5.0.0/externalpackages/PVODE/precon/band.cpp:189:14: note: place parentheses around the assignment to silence this warning
    if (swap = (l != k)) {
             ^
        (              )
/usr/ports/science/bout++/work/BOUT-dev-5.0.0/externalpackages/PVODE/precon/band.cpp:189:14: note: use '==' to turn this assignment into an equality comparison
    if (swap = (l != k)) {
             ^
             ==

```

PR #2684, opened 2023-03-23
Fix pvode warnings

dschwoerer:
Resolves: #2654


dschwoerer:
The string is only used once - so it would only trigger if the int type is changed

PR #2689, opened 2023-03-23
Recommend `requirements.txt` for dependencies

dschwoerer:


ZedThree (approved):

ZedThree:
We should probably change most of the CI to only run on changes to the C++ and CMake files, no need to run all the tests for doc changes :/

PR #2694, opened 2023-04-18
Review of PR #2514 (Switch to CMake)

tomc271:
Some corrections and minor improvements to PR #2514.


tomc271:
Just checked and it seems not, so changing to 3.17.5.

manual/sphinx/user_docs/advanced_install.rst
```diff
@@ -468,9 +468,9 @@ BOUT++ can use PETSc https://www.mcs.anl.gov/petsc/ for time-integration
 and for solving elliptic problems, such as inverting Poisson and
 Helmholtz equations.
 
-Currently, BOUT++ supports PETSc versions 3.7 - 3.14. More recent versions may
+Currently, BOUT++ supports PETSc versions 3.7 - 3.17. More recent versions may
 well work, but the PETSc API does sometimes change in backward-incompatible
-ways, so this is not guaranteed. To install PETSc version 3.13, use the
+ways, so this is not guaranteed. To install PETSc version 3.19, use the
```
dschwoerer, line 473:
```suggestion
ways, so this is not guaranteed. To install PETSc version 3.17, use the
```

dschwoerer:
I am wondering whether we should recommend running cmake from the build directory.

While it requires an additional `cd` call but:
 - it allows to copy-paste the code without modification
 - tab completion for cmake only works from the build directory (after one successful invocation)

PR #2702, opened 2023-04-27
mesh:paralleltransform is a section

dschwoerer:


johnomotani:
> In this section it also states you can set ShiftAngle in the [mesh] section of the input file. This doesn't work for me -- is it still possible?

I'm not sure this is documented at all, but I think the grid variables are loaded *either* from a grid file (if that exsists) *or* from the `BOUT.inp`, but not both. If the grid file is used, any 'grid' variables in the `[mesh]` section of `BOUT.inp` are ignored (probably, based on a quick glance at `src/mesh/data/` showing separate classes for loading grid data from grid file or options).

PR #2708, opened 2023-05-02
Generate BOUT++ tar via boutpp backend

dschwoerer:
Allows to drop a bash script for a python script, that we anyway need to maintain (and regularly use)


dschwoerer:
https://github.com/boutproject/BOUT-dev/pull/2709 fixes the failures

tools/pylib/_boutpp_build/backend.py
```diff
@@ -166,6 +173,10 @@ def build_sdist(sdist_directory, config_settings=None):
                     enable_xz = True
                 else:
                     raise ValueError(f"unknown option {v} for {k}")
+            if k == "dist":
```
ZedThree, line 176: Might make sense to use something like `argparse` for this in a future version?

ZedThree (approved):

dschwoerer:
Yes, it seems they use different CPUs, and with a bit of bad luck you 
get the old ones, so the cached SLEPc (I assume they add -march=native) 
fail sometimes. Seen that before, triggering a rerun solves the issue.

ZedThree:
Any idea if we can turn that flag off?

tools/pylib/_boutpp_build/backend.py
```diff
@@ -27,8 +27,14 @@ def getversion():
     """
     global version
     if version is None:
+        try:
+            version = os.environ["BOUT_PRETEND_VERSION"]
+            return version
+        except KeyError:
+            pass
```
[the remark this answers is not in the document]
dschwoerer: That is very handy, thanks for letting me know :rocket:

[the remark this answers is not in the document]
dschwoerer: NB: In case anybody wonders, this is [available since 3.4](https://docs.python.org/3/library/contextlib.html#contextlib.suppress), so can be used for essentially all codes.

dschwoerer:
Fixed - I copy pasted from our instructions, but `-march=native` is apparently a bad idea for CI

.build_petsc_for_ci.sh
```diff
@@ -20,7 +20,7 @@ if test $BUILD_PETSC ; then
 	    --with-scalar-type=real \
 	    --with-shared-libraries=1 \
 	    --with-debugging=0 \
-	    {C,CXX,F}OPTFLAGS="-O3 -march=native" \
```
d7919, line 23: One could use `mtune` if we 
1. Care about the PETSc CI performance (or think people might use this to build their own PETSc and care about performance).
2. Think that we're likely to get the same CPU type _most_ but not all of the time.

I doubt these are true, particularly 1, but perhaps this could inform the docs (as I think march came from the docs).

dschwoerer: I do agree :+1: 
`-mtune` would have been the better solution.

tools/pylib/_boutpp_build/backend.py
dschwoerer: probably, but I was lazy :see_no_evil:

ZedThree (approved):

ZedThree (approved):

PR #2709, opened 2023-05-03
Do not try to run SLEPc or PETSc in configure

dschwoerer:
Just set the variables to `YES` - as that may fail for a number of reasons.
It is e.g. expected to fail on login nodes of HPC systems.

### DIFF
--- a/cmake/FindPETSc.cmake
+++ b/cmake/FindPETSc.cmake
@@ -281,11 +281,8 @@ int main(int argc,char *argv[]) {
   return 0;
 }
 ")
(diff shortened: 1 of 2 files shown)


dschwoerer:
This PR should fix the current CI failures for fedora ...

ZedThree:
I was a bit worried about this, because it's potentially useful for making sure the correct flags are included. However, I don't think it has actually ever done that, and only ever reported false positives (that is, failed the configure step even though everything was fine).

So maybe let's just remove uses of `PETSC_TEST_RUNS` entirely?

dschwoerer:
Any idea why completely removing the TEST_RUNS fails?

dschwoerer:
A bit worrying: The runners use different CPUs, so sometimes cached jobs can fail with SIGILL if the code was compiled on a newer CPU than it is run.

bendudson (approved):

PR #2710, opened 2023-05-03
The additional check is not needed

dschwoerer:
The parallel transform already checks whether it is needed.

For the identity transform this is always needed, so the second check is harmful in that case, and otherwise not needed.


bshanahan (commented):
This fixes my issue, but I will leave it to @ZedThree, @johnomotani  or @bendudson to approve as it could also potentially affect shifted metrics.

PR #2716, opened 2023-05-25
Solver improvements

bendudson:
## cvode

Added input options to control tolerances on the inner nonlinear and linear solvers. Tightening the linear solver tolerance, for example, might improve the robustness of the nonlinear iterations. The optimal choice will be problem dependent, so the defaults are unchanged.

## beuler / snes

Improved / fixed the preallocation of Jacobian rows, and detection of boundary cells, so that no allocations are needed while setting the stencil pattern. Tested in 1D and 2D (tokamak X-point). 

Added an option `pc_hypre_type` to specify the type of hypre solver (choices are `boomeramg`, `euclid` and `pilut`). The default is `pilut` which seems to work best for a 1D test problem; `euclid` seems to be better in 2D. `boomeramg` doesn't seem to work very well for the transport problems tested.

Moved the PETSc options setting to before PETSc is initialized. This fixes the log_view option. In BOUT.inp:
```
[petsc]
log_view = true
```
or on the command line `petsc:log_view`

### DIFF
--- a/src/solver/impls/cvode/cvode.cxx
+++ b/src/solver/impls/cvode/cvode.cxx
@@ -174,12 +174,15 @@ CvodeSolver::CvodeSolver(Options* opts)
                     .doc("Use right preconditioner? Otherwise use left.")
                     .withDefault(false)),
       use_jacobian((*options)["use_jacobian"].withDefault(false)),
-      cvode_nonlinear_convergence_coef((*options)["cvode_nonlinear_convergence_coef"]
-                                       .doc("Safety factor used in the nonlinear convergence test")
-                                       .withDefault(0.1)),
-      cvode_linear_convergence_coef((*options)["cvode_linear_convergence_coef"]
-                                    .doc("Factor by which the Krylov linear solver’s convergence test constant is reduced from the nonlinear solver test constant.")
-                                    .withDefault(0.05)),
+      cvode_nonlinear_convergence_coef(
+          (*options)["cvode_nonlinear_convergence_coef"]
+              .doc("Safety factor used in the nonlinear convergence test")
+              .withDefault(0.1)),
+      cvode_linear_convergence_coef(
+          (*options)["cvode_linear_convergence_coef"]
+              .doc("Factor by which the Krylov linear solver’s convergence test constant "
+                   "is reduced from the nonlinear solver test constant.")
+              .withDefault(0.05)),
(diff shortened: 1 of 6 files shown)


mikekryjak:
This is exciting. Can we add some flags for diagnostics? `-log_view` gives a detailed profiling report after the completion of the simulation. It would be nice to have that in the input file. 

I remember that SNES had a separate tolerance setting for controlling when it's "steady state enough" to go fast and not worry about time accuracy. Could this be a factor that influences the tests? I don't remember whether we have it exposed in the input file or not. I seem to remember it was called `ptol` or something similar...

src/solver/impls/snes/snes.cxx
```diff
@@ -167,6 +172,7 @@ int SNESSolver::init() {
   }
 
   // Nonlinear solver interface (SNES)
+  output.write("Create SNES\n");
```
dschwoerer, line 175:
```suggestion
  output_info.write("Create SNES\n");
```

src/solver/impls/snes/snes.cxx
```diff
@@ -234,10 +240,26 @@ int SNESSolver::init() {
       std::vector<PetscInt> o_nnz(localN);
 
       // Set values for most points
+      const int ncells_x = (mesh->LocalNx > 1) ? 2 : 0;
+      const int ncells_y = (mesh->LocalNy > 1) ? 2 : 0;
```
dschwoerer, line 244:
```suggestion
      const int ncells_y = (mesh->ystart > 0) ? 2 : 0;
```

src/solver/impls/cvode/cvode.cxx
```diff
@@ -458,6 +467,10 @@ int CvodeSolver::init() {
 #endif
   }
 
+  // Set internal tolerance factors
+  CVodeSetNonlinConvCoef(cvode_mem, cvode_nonlinear_convergence_coef);
+  CVodeSetEpsLin(cvode_mem, cvode_linear_convergence_coef);
```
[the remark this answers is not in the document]
ZedThree: Ah, this is because we use the system version of SUNDIALS for the clang-tidy action, which turns out to be 3.1.2.

I think just updating `clang-tidy-review` will use a more recent version of Ubuntu and give us SUNDIALS 5.1.8, but I guess the question which versions of SUNDIALS do we want to support?

PR #2719, opened 2023-06-14
Bump required C++ standard to 17

ZedThree:
This only requires GCC 8, and that is available basically everywhere that we care about now.


dschwoerer (dismissed):
Would be happy to use c++17 features :+1: 
gcc 8 seems old enough - it is the oldest gcc provided on the mpcdf systems.

ZedThree:
Do the mpcdf systems have GCC 10? We might be able to start using some C++20 features as well

bendudson (approved):

PR #2721, opened 2023-06-15
Add Array<int> to options, additional limiters for grids

bendudson:
- Adds `Array<int>` to the Options type
- Reads `Array<int>` from NetCDF files (e.g. mesh input file), and makes them available through `Mesh::get`

The reason this was added is to be able to include additional limiters in the input file. The format is a short-term solution only, though these have a habit of becoming permanent.


bendudson:
This was branched from `v5.1.0-rc`, but probably should go into `next` after 5.1.0 is released.

dschwoerer:
@bendudson Do you mind rebasing to make the diff more clear?

bendudson:
Replaced by #2767

PR #2722, opened 2023-06-15
Fix FV::Div_a_Grad_perp Y alignment

bendudson:
Raises exception that Y alignment is incorrect when calling `fromFieldAligned` at the end of this function. Need to set aligned when constructing field.


dschwoerer (approved):
Is that covered by a CI test?
If not, it would be nice to add one ...

PR #2723, opened 2023-06-15
CI install natsort

bendudson:
In CI install natsort package.

Required from `boutdata.restart`, in `tests/integrated/test-restart-io`

### DIFF
--- a/.pip_install_for_ci.sh
+++ b/.pip_install_for_ci.sh
@@ -4,7 +4,7 @@ set -e
 
 export PATH=${HOME}/.local/bin:${PATH}
 pip3 install --user --upgrade pip~=20.0 setuptools~=46.1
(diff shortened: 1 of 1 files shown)


ZedThree:
Error is `g++: error: netcdf: No such file or directory`. I think I might have clobbered the fix when merging `next` into the `v5.1.0-rc` branch

dschwoerer:
That might also have been broken by me.

The tests were only merged, so it was not easy to detect this before.
I cannot reproduce this on my fedora, I end up with the full path.

Would it make sense to try to sanitise the LDFLAGS? And if it is neither starting with `-` nor a valid path prefix it with `-l`?

dschwoerer (approved):
Fixes an issue, the static failures are unrelated, so this can be merged :+1:

PR #2725, opened 2023-06-16
Revert sundials detection

dschwoerer:
Fixes the issue found by @bshanahan


bshanahan (approved):
This fixes the problem that nvecparallel could not be found when compiling a model.

PR #2726, opened 2023-06-19
Bump `clang-tidy-review` version

ZedThree:
The fixes now get uploaded as a build artefact, which can then be downloaded and applied locally


ZedThree:
Err, good question! Turns out we explicitly turned it off on RC branches:

```yaml
on:
  pull_request:
    paths:
      - '**.cxx'
      - '**.hxx'
    branches-ignore:
      # Release candidate branches tend to have big PRs which causes all sorts of problems
      - 'v*rc'
```

Could put it into master/next instead I suppose, along with some C++ changes -- might be better to just wait and see on some other PR? Not ideal

ZedThree:
Failing tests are the static build which was fixed in another PR

PR #2730, opened 2023-06-28
Improving compatibility with fmt 10

dschwoerer:
Todo:
- [x] check this resolves the FTBFS

### DIFF
--- a/include/bout/invertable_operator.hxx
+++ b/include/bout/invertable_operator.hxx
@@ -430,7 +430,8 @@ public:
     KSPConvergedReason reason;
     ierr = KSPGetConvergedReason(ksp, &reason);
     if (reason <= 0) {
-      throw BoutException("KSPSolve failed with reason {:d}.", reason);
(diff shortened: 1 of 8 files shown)


ZedThree:
Fedora is failing with:

```
Unknown argument "builddep" for command "dnf5". Add "--help" for more information about the arguments.
```

But there's also a compile error:

```
/home/runner/work/BOUT-dev/BOUT-dev/src/solver/impls/imex-bdf2/imex-bdf2.cxx:1244:25: error: ‘SNESGetConvergedReasons’ was not declared in this scope; did you mean ‘SNESGetConvergedReason’?
```

The suggestion is wrong, and I think, in classic PETSc fashion, there's a completely different thing to use here: [`SNESGetConvergedReasonString`](https://petsc.readthedocs.io/en/latest/docs/manualpages/SNES/SNESGetConvergedReasonString.html)

Also, we now use this pattern in quite a few places (I count at least 7), so might be worth wrapping up in a helper function?

ZedThree (approved):

PR #2720, opened 2023-06-14
5.1 release candidate

ZedThree:
We've got into a bit of mess between `master` and `next`, and I don't have a great deal of time to sort it out, so the simplest thing is to get everything into master and release 5.1.

New and currently open PRs into `next` should target this branch now.

# Making a New Release of BOUT++

This is checklist of things to do (in order) when making a new
release. This applies equally to both major/minor releases and bugfix
releases

- [ ] Check there are no open issues/PRs for this milestone
    - Fix or bump to next version
- [x] Make a branch named `vX.Y.Z-rc`
    - The GitHub repo is setup to protect branches named in this style
    - Major and minor release points (`X`/`Y`) should be off
      `next`. Bugfix releases (`Z`) should be off `master`
- [x] Make pull request into `master`
    - Any new bugfixes should be PRs into the RC branch, where
      "bugfixes" can include:
        - silencing warnings
        - improving documentation
        - adding tests
- [ ] Run `make check-all`
    - Raise issues for any tests that fail
- Possibly run `clang-tidy`, `clang-check`, `coverity`, etc.
- [ ] Review pinned pip package versions for Travis
    
Before merging PR:

- [ ] Update locale translation files
    - `make -C locale update-all`
    - Be aware that this *will* update the timestamps and *possibly*
      reorder file paths in the .po and .pot files
- [ ] Update [`CHANGELOG.md`][changelog]:
    - Run [bout-changelog-generator.py LAST_RELEASE NEXT_RELEASE][bin/bout-changelog-generator.py]
        - See the docs for how to get the token
- [ ] Get list of authors:
    - [ ] `git log --format='%aN' | sort | uniq`
    - [ ] Compare to list in [`CITATION.cff`][citation], add new authors
- [ ] Prep a new Zenodo release:
    - https://doi.org/10.5281/zenodo.1423212
    - "New Version"
    - "Reserve DOI" -> copy DOI
    - Add any new authors
    - Save draft
- [ ] Change DOI in [`CITATION.cff`][citation] to new DOI
- [ ] Change DOI in [`README.md`][README] to new DOI
- [ ] Change date-released in [`CITATION.cff`][citation]
- [ ] Check `abidiff` to see if `soname` needs bumping in `makefile`:
- [ ] Change version number, removing prerelease tag in:
    - [ ]  [`configure.ac`][configure]: `AC_INIT`
    - [ ]  [`CITATION.cff`][citation]: `version`
[...]

PR #2740, opened 2023-09-13
Allow static builds with Sundials

bendudson:
Previously if shared libs are not enabled, then neither static nor shared sundials libraries would be built.


dschwoerer:
Failing test is unrelated. It is because master needs to be merged into next; needs #2720

ZedThree:
Should this go into 5.1?

PR #2741, opened 2023-09-13
Update RELEASE_HOWTO.md

dschwoerer:
If we are on a tag, the python backend versioning should automatically do the right thing and not add a pre-release tag.
However, checking would be great - things could e.g. go wrong if the new tag is not locally present.


dschwoerer:
I guess the `make check-all` can be removed? I think most tests are now run anyway?

ZedThree (approved):

bendudson (approved):

PR #2742, opened 2023-09-13
Prefer dnf5 for fedora CI

dschwoerer:
dnf5 is faster, and is sufficiently feature complete for our needs

PR #2754, opened 2023-10-05
Add fallback if the version has been bumped bot not tagged

dschwoerer:
Should resolve at least part of the currenly observed errors

### DIFF
--- a/tools/pylib/_boutpp_build/backend.py
+++ b/tools/pylib/_boutpp_build/backend.py
@@ -152,7 +152,7 @@ def build_wheel(wheel_directory, config_settings=None, metadata_directory=None):
     run(
         "cmake -S . -B _wheel_build/ -DBOUT_ENABLE_PYTHON=ON"
         + f" -DCMAKE_INSTALL_PREFIX={prefix} -DCMAKE_INSTALL_LIBDIR={prefix}"
-        + f" -DCMAKE_INSTALL_PYTHON_SITEARCH={trueprefix} -DCMAKE_INSTALL_RPATH=\$ORIGIN"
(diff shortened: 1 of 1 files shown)


dschwoerer:
The upload still fails as there is no easy way to determine a proper version.

I could just detect this case and not upload, but I am worried that it might fail in other cases, so I would rather manually ignore this error ...

dschwoerer:
Remaining failures are #2755 and #2742

tools/pylib/_boutpp_build/backend.py
```diff
@@ -137,7 +154,7 @@ def build_wheel(wheel_directory, config_settings=None, metadata_directory=None):
     run(
         "cmake -S . -B _wheel_build/ -DBOUT_ENABLE_PYTHON=ON"
         + f" -DCMAKE_INSTALL_PREFIX={prefix} -DCMAKE_INSTALL_LIBDIR={prefix}"
-        + f" -DCMAKE_INSTALL_PYTHON_SITEARCH={trueprefix} -DCMAKE_INSTALL_RPATH=\$ORIGIN"
+        + f" -DCMAKE_INSTALL_PYTHON_SITEARCH={trueprefix} -DCMAKE_INSTALL_RPATH=$ORIGIN"
         + opts
```
ZedThree, line 158: No need to change now, but for future reference you can just do:

```python
  run(
        "cmake -S . -B _wheel_build/ -DBOUT_ENABLE_PYTHON=ON"
        f" -DCMAKE_INSTALL_PREFIX={prefix} -DCMAKE_INSTALL_LIBDIR={prefix}"
        f" -DCMAKE_INSTALL_PYTHON_SITEARCH={trueprefix} -DCMAKE_INSTALL_RPATH=$ORIGIN"
        + opts
    )
```

ZedThree (approved):

ZedThree:
Thanks @dschwoerer !

PR #2756, opened 2023-10-06
Merge v5.1.0 into next

ZedThree:


ZedThree:
We're expecting the fedora builds to fail, and possibly the python packages too

bendudson:
Looks like it's PETSc's own test that's failing
```
Performing Test MULTIPASS_TEST_2_petsc_works_allincludes - Failed
```

bendudson (approved):

dschwoerer:
I think we compile PETSc as shared object. Why it is required to link against all libraries - I do not understand ...
PETSc :shrug: 
```
-- Performing Test MULTIPASS_TEST_3_petsc_works_alllibraries
-- Performing Test MULTIPASS_TEST_3_petsc_works_alllibraries - Success
-- PETSc only need minimal includes, but requires explicit linking to all dependencies.  This is expected when PETSc is built with static libraries.
```

PR #2765, opened 2023-10-11
CI: Run on PETSc developement branch

dschwoerer:
Allows to detect issues with PETSc before the release. Should help with bisecting :-)

It would be nicer to only run if PETSc actually changed - but I don't know how to do that

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -3,7 +3,8 @@ on:
   push:
   pull_request:
   schedule:
-    - cron: '30 6 * * *'
+    # run sunday moring
+    - cron: '30 6 * * 0'
 
 defaults:
   run:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,5 +1,9 @@
 name: Tests
-on: [push, pull_request]
+on:
+  push:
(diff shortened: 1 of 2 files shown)


dschwoerer:
We should probably disable fail-fast if we have failing tests for a long time, to avoid hiding new errors in the other runs.

ZedThree:
I think we should only run this cron job once a week. That should still give us pretty good resolution for breaking changes without using too many resources.

.github/workflows/tests.yml
```diff
@@ -1,5 +1,9 @@
 name: Tests
-on: [push, pull_request]
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron: '30 6 * * *'
```
dschwoerer, line 6: @ZedThree It is only one CI job thus I thought daily would gives us very few PRs to look at.
But we can only run it weekly, at which case there are probably like ~10 PRs to go through.
I would rather use computer time but maintainer time.
```suggestion
    # run sunday moring
    - cron: '30 6 * * 0'
```

ZedThree (approved):

ZedThree:
We might even consider limiting the set of tests to those that actually use PETSc

PR #2739, opened 2023-09-11
WIP: Adios io

pnorbert:
Just showing diff for discussion.

PR #2766, opened 2023-10-12
ADIOS 2 I/O

bendudson:
Work in progress to enable BOUT++ to write and read data using ADIOS-2 (https://adios2.readthedocs.io/en/v2.9.1/). 

Rebased #2739 on the `next` branch.

BOUT++ can now be configured with `-DBOUT_DOWNLOAD_ADIOS=ON` to download, configure and build ADIOS.

Once configured, the output and restart file data format can be set with `output:type` and `restart_files:type` settings e.g. 
```
./conduction output:type=adios
```
outputs a `BOUT.dmp.bp` ADIOS2 dataset.


ZedThree:
I think the clang-tidy comments are mostly sensible, except to use `[[maybe_unused]]` instead of commenting out the parameter name for unused arguments.

src/sys/options/options_adios.cxx
```diff
@@ -0,0 +1,517 @@
...
+namespace bout {
+
+OptionsADIOS::OptionsADIOS() {}
+
+OptionsADIOS::OptionsADIOS(std::string filename, FileMode mode, bool singleWriteFile)
+    : OptionsIO(filename, mode, singleWriteFile) {}
+
+OptionsADIOS::~OptionsADIOS() = default;
+OptionsADIOS::OptionsADIOS(OptionsADIOS&&) noexcept = default;
+OptionsADIOS& OptionsADIOS::operator=(OptionsADIOS&&) noexcept = default;
```
ZedThree, line 27: Move to header?

ZedThree (commented):
I think my only two major comments are:

1. if `OptionsIO` is the new interface, then we could/should make `options_adios/netcdf.hxx` private
2. it would be good to switch to using the generic `Factory` to handle the creation of the specific type

bendudson:
After an extended game of whack-a-mole, I think this is now ready for merging. I don't know what's going on with clang-tidy-review, but I don't think it's related to the changes.
More documentation and tests would be nice, but for trying out in `next` I'm calling this good.

PR #2767, opened 2023-10-12
Add Array<int> to options, additional limiters for grids (rebase)

bendudson:
Rebase of #2721 on `next`

- Adds Array<int> to the Options type
- Reads Array<int> from NetCDF files (e.g. mesh input file), and makes them available through Mesh::get

The reason this was added is to be able to include additional limiters in the input file. The format is a short-term solution only, though these have a habit of becoming permanent.

### DIFF
--- a/include/bout/options.hxx
+++ b/include/bout/options.hxx
@@ -194,9 +194,9 @@ public:
   static void cleanup();
 
   /// The type used to store values
-  using ValueType =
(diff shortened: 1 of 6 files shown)


dschwoerer (dismissed):
Test would be nice :blush:

ZedThree (dismissed):

ZedThree (approved):

PR #2778, opened 2023-10-17
Increase shm size in container

dschwoerer:
It seems new libfabric requires more memory. This should resolve the issue.

Note that the workaround of old libfabric is still in place, so this should work either way.
I will delete the downgraded libfabric, once this has been merged, so that there are no regression for next and master.


dschwoerer:
I think it makes sense to get the fixes into master, and then merge that into next ...

PR #2781, opened 2023-10-17
Fix test-laplace-petsc3d

dschwoerer:
The essential bit from #2713 for fixing the test issues


ZedThree (approved):

ZedThree:
Thanks @dschwoerer, I didn't even think of this, definitely the best way of solving things!

PR #2794, opened 2023-10-30
CI: Replace pip script with requirements.txt

ZedThree:
- Don't pin pip and setuptools versions
- Bump netcdf4 and Cython
- Install boututils and boutdata

Most tests are currently failing because Python `3.x` on CI is now 3.12 and we were pinning `pip` to something too old.

I've replaced the `.pip_install_for_ci.sh` script with a plain `requirements.txt` as I don't think it gave any benefit -- it actually has some issues in that it doesn't necessarily use the same `python/pip` as setup in the main CI environment (at least, this is what happens in a local venv)


requirements.txt
```diff
@@ -1,6 +1,8 @@
 Jinja2>=2.11.3
 numpy>=1.14.1
```
dschwoerer, line 2: While you are at it, I think we can remove the version. I doubt we can run with such ancient versions ...
```suggestion
numpy
```

requirements.txt
```diff
@@ -1,6 +1,8 @@
...
 scipy>=1.0.0
-netcdf4>=1.3.1
+netcdf4~=1.6.0
 matplotlib>=2.0.0
```
dschwoerer, line 5:
```suggestion
matplotlib
```

dschwoerer (approved):
Looks good :-)

PR #2800, opened 2023-11-06
Set oversubscribe flags for openmpi 5

### LINKED ISSUE #2799: Fedora CI requires oversubscribe setting
Tests currently failing with:

```
There are not enough slots available in the system to satisfy the 4
slots that were requested by the application:

  ./testVec

Either request fewer procs for your application, or make more slots
available for use.

A "slot" is the PRRTE term for an allocatable unit where we can
launch a process.  The number of slots available are defined by the
environment in which PRRTE processes are run:

  1. Hostfile, via "slots=N" clauses (N defaults to number of
     processor cores if not provided)
  2. The --host command line parameter, via a ":N" suffix on the
     hostname (N defaults to 1 if not provided)
  3. Resource manager (e.g., SLURM, PBS/Torque, LSF, etc.)
  4. If none of a hostfile, the --host command line parameter, or an
     RM is present, PRRTE defaults to the number of processor cores

In all the above cases, if you want PRRTE to default to the number
of hardware threads instead of the number of processor cores, use the
--use-hwthread-cpus option.

Alternatively, you can use the --map-by :OVERSUBSCRIBE option to ignore the
number of available slots when deciding the number of processes to
launch.
```

I thought we set this? Or is that only for Github?

dschwoerer:
Resolves #2799 

Sorry, forgot to push this, thanks for the reminder @ZedThree 

I do not know where this is documented but Orion told me :-)

### DIFF
--- a/.ci_fedora.sh
+++ b/.ci_fedora.sh
@@ -56,6 +56,7 @@ else
     . /etc/profile.d/modules.sh
     module load mpi/${1}-x86_64
     export OMPI_MCA_rmaps_base_oversubscribe=yes
+    export PRTE_MCA_rmaps_default_mapping_policy=:oversubscribe
     export TRAVIS=true
     export FLEXIBLAS=NETLIB
     cd
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -28,6 +28,7 @@ jobs:
       OMP_NUM_THREADS: ${{ matrix.config.omp_num_threads }}
       PYTHONPATH: ${{ github.workspace }}/tools/pylib
       OMPI_MCA_rmaps_base_oversubscribe: yes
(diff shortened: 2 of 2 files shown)


dschwoerer:
We could have more jobs and [set them to be run manually](https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow). Is that what you are looking for?

ZedThree:
Not quite, it's that this PR only affects the Fedora build so it would be nice if we could choose to only run that instead of everything else. I guess we could manually cancel unneeded jobs, but then they appear as having failed. I don't suppose it matters that much!

dschwoerer:
Ah, indeed, that would be handy.

Technically, it also changes the other jobs, but that will only come 
into affect once openmpi 5 lands in ubuntu ...

PR #2823, opened 2023-12-20
WIP: FV::Div_a_Grad_perp operator

bendudson:
Going through this operator (again). Reviewing how metric elements are interpolated to cell edges.

To do:
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/src/mesh/fv_ops.cxx
+++ b/src/mesh/fv_ops.cxx
@@ -51,11 +51,15 @@ Field3D Div_a_Grad_perp(const Field3D& a, const Field3D& f) {
       for (int k = 0; k < mesh->LocalNz; k++) {
         // Calculate flux from i to i+1
 
-        BoutReal fout = 0.5 * (a(i, j, k) + a(i + 1, j, k))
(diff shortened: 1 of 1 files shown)


dschwoerer:
Failing test should be fixed by #2824

mikekryjak:
Thanks for this @bendudson. I've put this through our puff test. For everyone else, we are facing a very tough to debug problem where our neutrals are spontaneously increasing in temperature and decreasing in density around the X-point. We have ruled out a lot of things and our current suspect is the operator. We came up with a test where we take a simplified model with a fixed plasma background of constant temperature and density and impose a neutral source near the X-point. We then observe how the diffusive transport behaves, particularly how the density falloff evolves with distance from the puff source.

I am seeing no difference with this new operator. There is a bifurcation near the source which is identical in both cases. The bifurcation can be seen near X=0 in the RHS plot we see several points of similar distance but different density. It may be that this is due to cell size after all, i.e. because the cell value represents the average across the entire cell, meaning that a larger cell covering a greater distance from the puff source would have a lower density value (I think?).

Old results:
image

New results:
image

PR #2824, opened 2023-12-20
Work around pip changes 

dschwoerer:
I think I did this already for a different CI pipeline. As this is a CI environement, breaking it seems not so bad, but we might want to adopt an officially recommended approach, in case user copy instructions from there, although I guess we should first update our documentation to switch to venv ...


dschwoerer:
Fix is also in  #2324 ...

Can we get #2324 merged instead?

ZedThree (approved):

ZedThree:
Another option is something like: https://github.com/aminya/setup-cpp which is pretty popular.

PR #2828, opened 2023-12-21
More consistent parsing of boolean inputs (Breaking change)

bendudson:
TL;DR: Only "true" and "false", "True" and "False", 1 and 0 are allowed for booleans. This enables boolean options expressions like:
```
thing = true
another = false
third = thing & another
```
but breaks inputs using "yes", and other strings that are not "true" or "false", "True", "False", "1" or "0" for boolean settings.

When converting a string to a BoutReal, Options uses an expression parser that can use variables defined in the input options. The same parser is used when converting to an integer. When converting to a bool, however, Options was doing a simple case-insensitive match to "true", "false", "yes", "no" and a few similar strings.

This PR changes this behaviour, treating bools like a special case of an integer: The same parser is used, but the result must be either 1 or 0. This enables boolean quantities to behave consistently with int and BoutReal types.

We now define "true" to be equal to 1, and "false" to be 0, so can be used in expressions e.g `true * 2 -> 2`, and `false * 3 -> 0`. No other variations so e.g. "yes", "no", "True" are undefined.

We define operators `|` and `&` for logical `or`, and logical `and`; and `!` for logical `not`. Due to the simplicity of the parser binary operators must be single characters.  These both check that their left and right arguments are bools (i.e. either 1 or 0). Also adding operators `<` and `>` for comparing values, evaluating to 1 (true) or 0 (false).

Previously x,y,z and t were initialized to 0, but for booleans this would now have quite bad consequences: Inputs that used "y" to mean "true" would now evaluate to `0` which would be interpreted as `false`. With this change we get an exception ("y" is undefined)
rather than incorrect behaviour.

By removing these defaults, scalar quantities (e.g. integers, booleans) are evaluated without defining x,y,z or t. Trying to convert an expression like "x + y" to an integer or bool would previously evaluate to 0, but will now throw a BoutException.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 ### Breaking changes
 
 - The autotools `./configure` build system has been removed
+- Parsing of booleans has changed [\#2828][https://github.com/boutproject/BOUT-dev/pull/2828] ([bendudson][https://github.com/bendudson]).
+  See the [manual page](https://bout-dev.readthedocs.io/en/stable/user_docs/bout_options.html#boolean-expressions) for details.
 
 
 ## [v5.1.0](https://github.com/boutproject/BOUT-dev/tree/v5.1.0)
--- a/include/bout/sys/expressionparser.hxx
+++ b/include/bout/sys/expressionparser.hxx
@@ -3,7 +3,7 @@
  *
  * Parses strings containing expressions, returning a tree of generators
  *
- * Copyright 2010-2023 BOUT++ contributors
+ * Copyright 2010-2024 BOUT++ contributors
  *
  * Contact: Ben Dudson, dudson2@llnl.gov
  *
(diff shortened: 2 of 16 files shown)


dschwoerer:
I guess this also resolves https://github.com/boutproject/BOUT-dev/issues/1701

dschwoerer (dismissed):
I feel slightly uneasy about such a breaking change in a 5.2.0 release.

However:
 * The change should result in errors, not in subtle changes
 * The change might surface bugs
 * In general at least, it should be easy to update to the new syntax

Should this also be documented?

ZedThree:
LGTM! I think it's really nice to tighten up the input file format, as well as fixing other issues and adding new features!

It would be good to have this change documented somewhere -- there's a breaking changes section in the changelog for example.

Is there some way we could deprecate `yes/no/True/False` etc? Maybe we could add generators for them that print warnings if they get used?

PR #2832, opened 2024-01-05
Remove propietary data (master)

dschwoerer:
Here is a report with the various licenses found in the 5.1.0 release:
http://miroslav.suchy.cz/fedora/spdx-reports/bout++.html


dschwoerer:
All 4 files are binary IDL files, that have a text comment:

```
 NOTICE: 
 
IDL Save/Restore files embody unpublished proprietary information 
about the IDL program. Reverse engineering of this file is therefore
forbidden under the terms of the IDL End User License Agreement 
(IDL EULA). All IDL users are required to read and agree to the 
terms of the IDL EULA at the time that they install IDL. Non-RSI  
supplied software that reads or writes files in the IDL 
Save/Restore format must have a license from Research Systems 
explicitly granting the right to do so. In this case, the license 
will be included with the software for your inspection. Please 
report software that does not have such a license to 
Research Systems, Inc. 
```

I guess that would imply that any browser would need such a license to download files?
Anyway, we likely do not want those files ...

bendudson (approved):

PR #2836, opened 2024-01-12
Do not excessively bump googletest

dschwoerer:
I think it would still be nice to get other updates in a somewhat timely fashion ...


ZedThree:
I think it's excessively noisy because they have a "live at HEAD" policy and quite an active repo. If we could limit it to once a month or something, that would be fine. And also have it update PRs, rather than close and open a new one!

ZedThree (approved):
We'll see how we get on with weekly, I feel we might need to change it to monthly instead

PR #2840, opened 2024-01-19
Add warnings about SUNDIALS versions <4

Steven-Roberts:
Following discussion with @bendudson, this warns people using SUNDIALS <v4 so that support can be dropped in the next release. SUNDIALS <v4 appears incompatible anyway due to this call

https://github.com/boutproject/BOUT-dev/blob/aeb08976f3d277ef45b4bf3c30ea6a97925ed9d6/src/solver/impls/cvode/cvode.cxx#L472

I'll follow up with PRs to clean up code specific to SUNDIALS <v4 and to support v7.


dschwoerer:
If it is broken anyway, should we then not drop support immediately?
I much rather have cmake saying it is not supported then a compile time error, or even worse a misbehavior at runtime ...

I anyway doubt people are actually using ancient sundials with a recent bout++, otherwise we tend to require rather recent dependencies ...

Support for sundials 7 would be great :tada:

bendudson (approved):

PR #2843, opened 2024-01-22
Remove `tools/pylib/post_bout`

ZedThree:
Somehow I'd never bothered to look at this. Looks like it's completely supplanted by both `boututils` and `xbout`


dschwoerer (approved):
I doubt all of that has been replaced, but also a significant part seems to be very model / code specific and of limited general use ...

PR #2844, opened 2024-01-22
Fix various documentation issues

ZedThree:
This is a bid to try and get the number of Sphinx warnings down, to ideally start using error on warnings.

The major issue now is things like:

```
BOUT-dev/manual/sphinx/_breathe_autogen/file/pcr__thomas_8hxx.rst:4: WARNING: Duplicate C++ declaration, also defined at _breathe_autogen/file/pcr__thomas_8hxx:4.
Declaration is '.. cpp:function:: virtual void setCoefA (const Field2D &val)=0'.
```

which I think is complaining about overriding the pure virtual function. I don't know whether this is an issue with breathe or doxygen. Breathe is not super well maintained currently, so I might investigate other options.

I've fixed a massive chunk of these warnings by just not processing the `.cxx` files. I think this is probably reasonable, but it might involve going through and moving docstrings to the headers. That will take a little bit of time to do.

Making a draft now in case anyone has feedback or ideas on these changes.


ZedThree:
I've discovered that the vast majority of the warnings emitted from Sphinx are due to two separate bugs in doxygen and breathe.

Things that still need doing:

- [ ] `boutdata/boututils` docs are still built as part of this repo, should probably build them separately
- [ ] go through every single file and make sure the docstring is in the header

manual/sphinx/conf.py
```diff
@@ -31,7 +31,8 @@
 import subprocess
 import sys
 
-sys.path.append("../../tools/pylib")
+sys.path.append("../../tools/pylib/boutpp")
+sys.path.append("../../tools/pylib/boutconfig")
```
dschwoerer, line 35: Why did you do this?
That breaks the RTD build:
https://bout-dev--2844.org.readthedocs.build/en/2844/_apidoc/boutpp.html#boutpp-api
https://bout-dev.readthedocs.io/en/latest/_apidoc/boutpp.html#boutpp-api

The tools are directly in boutpp and boutconfig ...

dschwoerer (changes requested):

include/bout/generic_factory.hxx
```diff
@@ -1,5 +1,9 @@
+/// \file
 /// Base type for factories
 
+// Copyright 2018-2022 Peter Hill, BOUT++ Team
+// SPDX-License-Identifier: MIT
```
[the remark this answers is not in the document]
ZedThree: The code is heavily based on Modern C++ Design/Loki by A. Alexandrescu, which is MIT

manual/sphinx/conf.py
ZedThree: This was to stop generating the docs for the other directories under pylib, and to be explicit about what's built here. It looks like it built the docs for everything except these two, which is very odd. Not sure what's going on!

dschwoerer: I think the other ones get pip installed?
I haven't checked though!

Anyway, for zoidberg we currently don't build the docs but reference the bout++ docs, so if you want to drop zoidberg here, we should add zoidberg independently.
Also, it would be better to remove the sections, rather then to have them being empty because python cannot find the modules.

include/bout/generic_factory.hxx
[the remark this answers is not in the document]
dschwoerer: Could you please add then a copy of the license?
I think that is even required by the license ...

dschwoerer (dismissed):
Thanks for the cleanup!

dschwoerer (dismissed):

include/bout/fv_ops.hxx
```diff
@@ -19,59 +19,66 @@
 #include "bout/vector2d.hxx"
 
 namespace FV {
-/*!
- * Div ( a Grad_perp(f) ) -- ∇⊥ ( a ⋅ ∇⊥ f) -- Vorticity
- */
+/// \f[
```
dschwoerer, line 23:
```suggestion
/// Vorticity:
/// \f[
```

dschwoerer (approved):

PR #2848, opened 2024-01-25
Remove unused `format.hxx` header

ZedThree:
Previously used for annotating `printf` like functions, since replaced with `fmt` formatting


bendudson (approved):
Thanks @ZedThree ! I forgot that this existed.

PR #2849, opened 2024-01-25
Replace `MAYBE_UNUSED` macro with C++17 `[[maybe_unused]]`

ZedThree:


bendudson (approved):

ZedThree:
I think we probably want to turn off the identifier-length and easily-swappable-arguments checks, just too noisy

PR #2850, opened 2024-01-25
Use C++17 `inline` variables to remove some linker hacks

ZedThree:
With C++17, there's no need to have a separate declaration of `static constexpr` variables in a `.cxx` file in order for them to get an address.

Note that `constexpr` for namespace/file scope variables implies `static` but not `inline`, but `constexpr` class member variables implies `inline` but not `static` :+1: not confusing at all.


bendudson (approved):
Nice tidyup, thanks @ZedThree !

ZedThree:
`inline` isn't actually enough to save the `RegisterInFactory` symbols without some more rearranging of the code, and I don't fancy that right now, so I'm going to revert the last commit. I'd forgotten this was actually unrelated to the factory's static members

bendudson (approved):

PR #2851, opened 2024-01-26
Replace `mpark.variant` with `std::variant`

ZedThree:


dschwoerer:
It seems std::variant is broken with old gcc versions?

I guess that we should delay this move for now?

It might also be good to check with intel compilers by the time we switch ...

ZedThree:
Bah, there is a C++17DR for making `std::visit` work when inheriting from `std::variant`. The fix is in gcc 11 and clang 15. This works fine in Intel.

So options are:
1. hold off on this for now, which is pretty safe. I think `mpark::variant` even falls back to `std::variant` for C++17
2. bump minimum compiler version required to gcc 11. This is mostly safe, most OSes with earlier versions are either EOL or close to EOL.

dschwoerer:
I suppose we could revisit this now?

PR #2856, opened 2024-02-01
Update CUDA standard to cxx17, add CI build test

ggeorgakoudis:
Fixes compilation error for CUDA builds (CXX standard should be upped to 17) and adds a containerized build test for CUDA builds.

The container is pulled from my own github registry, I suggest moving it under the boutproject registry.

### DIFF
--- a/cmake/SetupBOUTThirdParty.cmake
+++ b/cmake/SetupBOUTThirdParty.cmake
@@ -19,9 +19,10 @@ if (BOUT_HAS_CUDA)
   set(BOUT_SOURCES_CXX ${BOUT_SOURCES})
   list(FILTER BOUT_SOURCES_CXX INCLUDE REGEX ".*\.cxx")
 
-  set_source_files_properties(${BOUT_SOURCES_CXX} PROPERTIES LANGUAGE CUDA )
+  # NOTE: CUDA inherits the CXX standard setting from the top-level
+  # compile features, set for the bout++ target.
+  set_source_files_properties(${BOUT_SOURCES_CXX} PROPERTIES LANGUAGE CUDA)
   find_package(CUDAToolkit)
-  set_target_properties(bout++ PROPERTIES CUDA_STANDARD 14)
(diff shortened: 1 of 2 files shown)


cmake/SetupBOUTThirdParty.cmake
```diff
@@ -19,9 +19,9 @@ if (BOUT_HAS_CUDA)
   set(BOUT_SOURCES_CXX ${BOUT_SOURCES})
   list(FILTER BOUT_SOURCES_CXX INCLUDE REGEX ".*\.cxx")
 
-  set_source_files_properties(${BOUT_SOURCES_CXX} PROPERTIES LANGUAGE CUDA )
+  set_source_files_properties(${BOUT_SOURCES_CXX} PROPERTIES LANGUAGE CUDA)
   find_package(CUDAToolkit)
-  set_target_properties(bout++ PROPERTIES CUDA_STANDARD 14)
+  set_target_properties(bout++ PROPERTIES CUDA_STANDARD 17)
```
ZedThree, line 24: Is it possible to set this from the value of `CMAKE_CXX_STANDARD`?

ggeorgakoudis: No, because we don't set `CMAKE_CXX_STANDARD` but instead use `target_compile_features` on the bout++ target. However, the point of avoiding duplication is a good one. We can remove setting the `CUDA_STANDARD` for CUDA compilation since `target_compile_features` for `cxx_std_17` implicitly enables `cuda_std_17`. How does that sound?

bendudson: Maybe a comment in the code that `CUDA_STANDARD` is implicitly set, so someone doesn't re-add it?

ggeorgakoudis:
@bendudson @ZedThree would you like to put the CUDA container under the boutproject packages? I will need to amend the github action for the new URL if you do so.

bendudson (approved):

PR #2651, opened 2023-02-08
Allow X splitting with FCI

dschwoerer:
Uses PETSc. Currently only implemented for hermitespline.

PR #2858, opened 2024-02-03
WIP: XZ interpolation using PETSc

bendudson:
Work in progress

Intend to adapt D.Bold's PETSc interpolation for yup/ydown, for merging into next.

### DIFF
--- a/src/mesh/interpolation_xz.cxx
+++ b/src/mesh/interpolation_xz.cxx
@@ -38,7 +38,9 @@ const Field3D interpolate(const Field3D& f, const Field3D& delta_x,
                           const Field3D& delta_z) {
   TRACE("Interpolating 3D field");
   XZLagrange4pt interpolateMethod{f.getMesh()};
-  return interpolateMethod.interpolate(f, delta_x, delta_z);
+  // Cast to base pointer so virtual function overload is resolved
+  return static_cast<XZInterpolation*>(&interpolateMethod)
+      ->interpolate(f, delta_x, delta_z);
 }
 
 const Field3D interpolate(const Field2D& f, const Field3D& delta_x,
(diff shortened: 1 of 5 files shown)


dschwoerer:
Hi Ben,
just to make sure you did not miss the PR, here is my PETSc implementation as PR for next:
https://github.com/boutproject/BOUT-dev/pull/2651

The FCI MMS tests are passing, thus I am not sure the issue I am seeing in hermes-2 with X-splitting is actually related to the PETSc part, or some other part of the code that has hard-coded the assumption that there is no splitting in X.

bendudson:
Hi @dschwoerer ! Thanks yes; I'm going through your PR in steps so I can (a) understand it myself; (b) implement the PETSc method as a separate implementation of XZInterpolate. As I'm going through it I'm trying to work out where the no-X-splitting assumptions are being made.

src/mesh/interpolation_xz.cxx
```diff
@@ -38,7 +38,9 @@ const Field3D interpolate(const Field3D& f, const Field3D& delta_x,
                           const Field3D& delta_z) {
   TRACE("Interpolating 3D field");
   XZLagrange4pt interpolateMethod{f.getMesh()};
-  return interpolateMethod.interpolate(f, delta_x, delta_z);
+  // Cast to base pointer so virtual function overload is resolved
+  return static_cast<XZInterpolation*>(&interpolateMethod)
+      ->interpolate(f, delta_x, delta_z);
```
ZedThree, line 43: This shouldn't be needed -- the issue is probably because we now need a `using XZInterpolation::interpolate` in the derived classes to bring in the missing overloads of the virtual function.

dschwoerer:
@bendudson I have recently changed the XZHermiteSpline to be a template, I could therefore change the `#if` into `if constexpr()` to have it generate all versions, which would allow have all implementations available at runtime. Do you think that would be helpful?

bendudson:
This PR no longer needed

PR #2861, opened 2024-02-07
Options: Delete copy constructor and copy assignment

bendudson:
Accidentally copying Options leads to unexpected bugs where changes (e.g documentation) changes a copy rather than the original. Usually what is intended is a reference, but the author omits an `&`.

Example: https://github.com/bendudson/hermes-3/pull/204#discussion_r1476518484

Copy assignment is also ambiguous: Is it the value, the attributes, or the children that should be copied?

Now if a copy is really intended, there is a `.copy()` method (thanks Rust!) that makes a deep copy.

### DIFF
--- a/include/bout/options.hxx
+++ b/include/bout/options.hxx
@@ -212,6 +212,9 @@ public:
     template <typename T>
     CopyableOptions(T value) : value(std::move(value)) {}
 
+    /// Special case for char*, which can otherwise become cast to bool
+    CopyableOptions(const char* value): value(std::string(value)) {}
+
     CopyableOptions(
         std::initializer_list<std::pair<std::string, CopyableOptions>> children)
         : children(children) {}
--- a/include/bout/options.hxx
+++ b/include/bout/options.hxx
@@ -214,7 +214,7 @@ public:
 
     CopyableOptions(
         std::initializer_list<std::pair<std::string, CopyableOptions>> children)
-        : children(std::move(children)) {}
+        : children(children) {}
     ValueType value;
     std::initializer_list<std::pair<std::string, CopyableOptions>> children;
   };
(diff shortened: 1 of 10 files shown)


ZedThree:
Looks like the failing tests are just because a fedora mirror 404'd, rerunning now.

LGTM, thanks @bendudson! In C++26 we might even be able to do `=delete("Use .copy() instead")`, so that'll be nice in about 10 years :smile: 

Please could you put a line under "breaking changes" in `CHANGELOG.md`?

ZedThree:
For some reason, only the Fedora build is failing with:

```
/home/test/BOUT-dev/src/solver/impls/cvode/cvode.cxx: In instantiation of 'std::vector<double> CvodeSolver::create_constraints(const std::vector<Solver::VarStr<T> >&) [with FieldType = Field3D]':
/home/test/BOUT-dev/src/solver/impls/cvode/cvode.cxx:358:46:   required from here
/home/test/BOUT-dev/src/solver/impls/cvode/cvode.cxx:488:25: note:   358 |     auto f3d_constraints = create_constraints(f3d);
/home/test/BOUT-dev/src/solver/impls/cvode/cvode.cxx:488:25: note:       |                            ~~~~~~~~~~~~~~~~~~^~~~~
/home/test/BOUT-dev/src/solver/impls/cvode/cvode.cxx:488:25: error: use of deleted function 'Options::Options(const Options&)'
```

Not sure why the other compilers aren't catching that! Will investigate locally

ZedThree:
Oh, this is because most of our builds use sundials 4.1.0, and we have a guard https://github.com/boutproject/BOUT-dev/blob/283917db5f18e5b52ac7acc66803c332890f9add/src/solver/impls/cvode/cvode.cxx#L350-L358

which is not correct. I'll make a separate PR for that fix (we also have some deprecations to fix, I'll have a stab at those too)

PR #2863, opened 2024-02-12
Fix SUNDIALS version guard and some deprecated function calls

ZedThree:
Fixes an issue with one version guard (caught versions < x.2 instead of < 3.2)

Also fixes some deprecated function calls. Turns out SUNDIALS v4 introduced a bunch of functions we should've been calling, although they were only recently actually deprecated.

There is one remaining call to a deprecated function (deprecated in 5.7):

```cpp
src/solver/impls/arkode/arkode.cxx:351:33: warning: ‘int ARKStepSetAdaptivityMethod(void*, int, int, int, realtype*)’ is deprecated: use SUNAdaptController instead [-Wdeprecated-declarations]
  351 |   if (ARKStepSetAdaptivityMethod(arkode_mem, adap_method, 1, 1, nullptr) != ARK_SUCCESS) {
      |       ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Unfortunately the fix is not exactly straightforward, and it's not really from the docs how to upgrade.

### DIFF
--- a/include/bout/sundials_backports.hxx
+++ b/include/bout/sundials_backports.hxx
@@ -23,8 +23,6 @@
 #include <sunnonlinsol/sunnonlinsol_newton.h>
 #endif
 
-#include "bout/unused.hxx"
-
 #if SUNDIALS_VERSION_MAJOR < 3
 using SUNLinearSolver = int*;
 inline void SUNLinSolFree([[maybe_unused]] SUNLinearSolver solver) {}
--- a/src/solver/impls/cvode/cvode.cxx
+++ b/src/solver/impls/cvode/cvode.cxx
@@ -347,7 +347,8 @@ int CvodeSolver::init() {
     CVodeSetMaxNonlinIters(cvode_mem, max_nonlinear_iterations);
   }
 
-#if not(SUNDIALS_VERSION_MAJOR >= 3 and SUNDIALS_VERSION_MINOR >= 2)
+#if (SUNDIALS_VERSION_MAJOR < 3) \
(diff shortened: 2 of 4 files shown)


bendudson (dismissed):
Thanks @ZedThree ! @Steven-Roberts is working on the upgrade, so we'll hopefully support SUNDIALS v7, and drop support for really old versions,

ZedThree:
Looking at what [versions of SUNDIALS are available across OSes](https://pkgs.org/search/?q=sundials), it seems most have >= 5.0, so we could use that as our minimum version?

There's still a couple of OSes (notably Ubuntu 20 LTS) that only have SUNDIALS 3, but we do have the option to download and build SUNDIALS as part of BOUT++, so this is probably not a massive problem. There's also a couple of OSes that have separate packages for SUNDIALS 2, but I doubt this is worth continuing to support.

Steven-Roberts:
If we support SUNDIALS 5.0, I don't think there are any extra shims or special cases we need to handle to also support 4.0. Considering that BOUT++ is currently [incompatible with SUNDIALS <4.0](https://github.com/boutproject/BOUT-dev/pull/2840) and presumably this has not caused issues with users, it seems natural to me to pick 4.0 as the new minimum version to support. Just let me know @bendudson and @ZedThree what you want to go with.

bendudson:
Thanks @Steven-Roberts ! It sounds like >= 4.0 is a good choice for now. I wouldn't spend a lot of time on supporting older versions though: We can recommend users configure BOUT++ to download and compile a compatible SUNDIALS version.

ZedThree:
Fixed by #2896

PR #2868, opened 2024-02-20
Add number of OpenMP threads to dump file

dschwoerer:
Much nicer than having to parse the log to get the number of threads for e.g. plotting ...

### DIFF
--- a/include/bout/array.hxx
+++ b/include/bout/array.hxx
@@ -382,7 +382,7 @@ private:
 
     // Clean by deleting all data -- possible that just stores.clear() is
     // sufficient rather than looping over each entry.
-    BOUT_OMP(single)
+    BOUT_OMP_SAFE(single)
     {
       for (auto& stores : arena) {
         for (auto& p : stores) {
--- a/include/bout/array.hxx
+++ b/include/bout/array.hxx
@@ -375,17 +375,9 @@ private:
    * @param[in] cleanup   If set to true, deletes all dataBlock and clears the store
    */
   static storeType& store(bool cleanup = false) {
-#ifdef _OPENMP
     static arenaType arena(omp_get_max_threads());
-#else
-    static arenaType arena(1);
-#endif
     if (!cleanup) {
-#ifdef _OPENMP
       return arena[omp_get_thread_num()];
-#else
-      return arena[0];
-#endif
     }
 
     // Clean by deleting all data -- possible that just stores.clear() is
--- a/include/bout/array.hxx
+++ b/include/bout/array.hxx
@@ -31,7 +31,7 @@
 #include <memory>
 #include <vector>
 
-#ifdef _OPENMP
+#if BOUT_USE_OPENMP
 #include <omp.h>
 #endif
 
--- a/include/bout/openmpwrap.hxx
+++ b/include/bout/openmpwrap.hxx
@@ -45,11 +45,11 @@
 #else
 #define BOUT_OMP(...)
 #ifdef _OPENMP
+#error OpenMP used but BOUT++ thinks it is disabled
+#else
 inline int constexpr omp_get_max_threads() { return 1; }
 inline int constexpr omp_get_num_threads() { return 1; }
 inline int constexpr omp_get_thread_num() { return 0; }
-#else
-#error OpenMP used but BOUT++ thinks it is disabled
 #endif
 #endif
 
--- a/include/bout/openmpwrap.hxx
+++ b/include/bout/openmpwrap.hxx
@@ -30,9 +30,10 @@
 #include "bout/build_defines.hxx"
 
 #if BOUT_USE_OPENMP
-
 #include "omp.h"
+#endif
 
+#ifdef _OPENMP
 //Some helpers for indirection -- required so that the _Pragma gets "omp <x>"
 //where <x> is any number of valid omp options/environments (e.g. atomic, critical etc.)
 #define INDIRECT0(a) #a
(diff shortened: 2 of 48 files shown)


src/bout++.cxx
```diff
@@ -715,6 +715,13 @@ void addBuildFlagsToOptions(Options& options) {
   options["use_backtrace"].force(bout::build::use_backtrace);
   options["use_color"].force(bout::build::use_color);
   options["use_openmp"].force(bout::build::use_openmp);
+  options["openmp_threads"].force(
+#ifdef _OPENMP
+      omp_get_max_threads()
```
d7919, line 720: It's a long time since I've worked with BOUT++ internals, but could the `ifdef` be replaced by a check of `options["use_openmp"]`?

d7919:
> On 2/20/24 10:01, David Dickinson wrote: ***@***.**** commented on this pull request. It's a long time since I've worked with BOUT++ internals, but could the |ifdef| be replaced by a check of |options["use_openmp"]|?
> I thought the omp_get_max_threads() function would only be definied if we use openmp, thus it might work with an if constexpr, but I am not sure whether that would be C++17? bout::build::use_openmp should be a constexpr, and thus usable in a if constexpr ...

Yes sorry, immediately after posting the comment I realised it was a silly suggestion so deleted it.

dschwoerer:
Sorry, I was confused why I couldn't see it on the web.

Looking at: https://en.cppreference.com/w/cpp/language/if it seems that would be possible if that was within a function, but does not seem to be possible in normal function.

We could add `int omp_get_max_threads() {return 1};` to a header, for the case without openmp. That would avoid `#ifdef` in 2 or 3 places ...

dschwoerer:
There is one case where it may be set to 0 without openmp, but in other cases it is 1.
I think also from a logical point of view, no openmp is mostly equivalent to 1 thread.

include/bout/openmpwrap.hxx
```diff
@@ -39,6 +39,9 @@
 #define BOUT_OMP(...) _Pragma(INDIRECT2(__VA_ARGS__))
 #else
 #define BOUT_OMP(...)
+int constexpr omp_get_max_threads() { return 1; }
+int constexpr omp_get_num_threads() { return 1; }
+int constexpr omp_get_thread_num() { return 0; }
```
ZedThree, line 44:
```suggestion
inline int constexpr omp_get_max_threads() { return 1; }
inline int constexpr omp_get_num_threads() { return 1; }
inline int constexpr omp_get_thread_num() { return 0; }
```

ZedThree (approved):
Lovely, thanks @dschwoerer ! Always nice to remove some `#ifdef`s

ZedThree:
> There is one case where it may be set to 0 without openmp, but in other cases it is 1. I think also from a logical point of view, no openmp is mostly equivalent to 1 thread.

I'm on the fence about this. In some sense, without OpenMP we have no threads -- there's definitely a measurable difference between 1 thread and no OpenMP, but I guess we do also store whether or not OpenMP is enabled, so we can use that in post-processing if necessary.

include/bout/array.hxx
```diff
@@ -375,17 +375,9 @@ private:
    * @param[in] cleanup   If set to true, deletes all dataBlock and clears the store
    */
   static storeType& store(bool cleanup = false) {
-#ifdef _OPENMP
     static arenaType arena(omp_get_max_threads());
-#else
-    static arenaType arena(1);
-#endif
```
dschwoerer, line 382: we can certainly change no OpenMP -> `0` - but than this needs to be reverted.

It is also very convinient to calculate the total number of threads, just multiply `omp_get_max_threads() * NXPE * NYPE`.

What would be the use case for using `0`?

bendudson: Is there another way that we can distinguish between no OpenMP and OpenMP with one thread? If there's another flag to say whether we are using OpenMP, perhaps renaming this from `omp_get_max_threads` to just `max_threads()`, since the threads may not be OpenMP threads?

ZedThree:
The OpenMP builds are failing because we're missing

```cpp
#include "bout/build_defines.hxx"
```

in `openmpwrap.hxx` in order to bring in the `BOUT_USE_OPENMP` macro. I'm a bit worried this means the `BOUT_OMP` macro has been a no-op in several files for some time!

dschwoerer:
The cuda builds are worrying me. They seem to have openmp enabled, but BOUT_USE_OPENMP is disabled. That means that all the thread-safety feature we can enable are not enabled. That might work most of the time (I have not tried) but can lead to race conditions, data corruption (two threads using the same data block) and crashes (random and annoying, but still better then simply wrong results).

I have added a check now, which is why the cuda build fails. One solution would be to just enable openmp for BOUT++ for cuda.
Is that a viable approach? Do we need to options, one for ensuring thread safety, and one for parallelising? Why is openmp enabled for cuda?

The other builds with openmp are failing in the test for `bout-config` - for some reason `-fopenmp` does not get added to it, I am confused as to why that fails ...

ZedThree:
Verbose CUDA build here: https://github.com/boutproject/BOUT-dev/actions/runs/8002349081/job/21855360746

ZedThree:
Also, I noticed `has-openmp` doesn't appear in the output of `bout-config --all`

dschwoerer:
So something adds `-fopenmp`. @ggeorgakoudis did you add that intentionally? Is that needed? If so, I think we should just enable openmp for bout++ as well. Then we get the thread safety. If it turns out for performance reasons we do want to enable openmp but not have bout++ do parallelisation, we can add that option later. This PR is already enormous for adding 1 single integer to the dump file ...

ZedThree:
It might be coming from RAJA or UMPIRE, it doesn't appear to be turned on after we detect CUDA: https://github.com/boutproject/BOUT-dev/actions/runs/8002552988/job/21856012784#step:4:64

ZedThree:
Thanks for fixing this @dschwoerer, but it feels a bit like overkill now, as you say just to add one flag to the output!

It sort of sidesteps the issue of why the CUDA build is getting OpenMP turned on when we haven't requested it. @ggeorgakoudis Is this coming from the RAJA dependency? Should we actually be forcing OpenMP on in BOUT++ if RAJA is built with OpenMP?

Just because a dependency uses OpenMP doesn't necessarily mean we need to enable it in BOUT++ to ensure thread safety -- that depends on whether we're passing callbacks that include thread unsafe code.

ggeorgakoudis:
@ZedThree No, I haven't made any changes to the build system so I do not explicitly set openmp flags in compilation. It may be picked up by including RAJA (since its spack installation in the CI container includes the openmp variant), although I haven't found the exact spot in the cmake file hierarchy where this happens (let me know if you spot it). We can enable openmp in the CUDA CI configuration (https://github.com/boutproject/BOUT-dev/actions/runs/8002552988/job/21856012784#step:4:15) to avoid BOUT++ thinking it builds without openmp when RAJA pulls that in. Makes sense?

dschwoerer:
> Thanks for fixing this @dschwoerer, but it feels a bit like overkill now, as you say just to add one flag to the output!

I did not mean overkill, I just meant feature creep. I think there was a bug that hasn't been noticed before and this fixes it.
There was some consideration for parallelising Hermes-3 with OpenMP, and for this we would also need to be thread safe, but not parallelise the for loops. Thus this PR might be interesting for @bendudson as well ...

> Just because a dependency uses OpenMP doesn't necessarily mean we need to enable it in BOUT++ to ensure thread safety -- that depends on whether we're passing callbacks that include thread unsafe code.

Why would this be limited to callbacks?
If the function calls with different threats into some function that allocations some memory, and we use our cached malloc pool, that could lead to issues without any callbacks.

I think BOUT++ should always try to be on the safe side. Otherwise we could go back to commit [62f3d2c](https://github.com/boutproject/BOUT-dev/pull/2868/commits/62f3d2c5b9a0b49273382554e76ae9eec16e86e7) and ignore all of the other changes ... however I think it is not so bad, compared to getting sometimes wrong results because some memory block is double used and contains the wrong data.

bendudson (approved):

PR #2869, opened 2024-02-22
Do not specify redundant class in ctor

dschwoerer:
This is an error in C++20 and gcc is warning by default.

### DIFF
--- a/include/bout/region.hxx
+++ b/include/bout/region.hxx
@@ -516,9 +516,9 @@ public:
 
   // Want to make this private to disable but think it may be needed as we put Regions
   // into maps which seems to need to be able to make "empty" objects.
-  Region<T>() = default;
+  Region() = default;
 
-  Region<T>(int xstart, int xend, int ystart, int yend, int zstart, int zend, int ny,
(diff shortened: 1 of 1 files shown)


ZedThree (approved):

dschwoerer:
@ZedThree reordering was a bad idea. We first need to set the blocks, and only after that we can calculate the indices based on the blocks. I have added a comment to prevent that in the future.

ZedThree:
Ah, that means we should reorder the member declarations in the class then. Looks like there's a few other things worth fixing here too, I'll do all that in a separate PR

dschwoerer:
On 2/23/24 10:09, Peter Hill wrote:
> Ah, that means we should reorder the member declarations in the class 
> then. Looks like there's a few other things worth fixing here too, I'll 
> do all that in a separate PR

That is what I thought initially, too.
But just above this function is a different one, where it is computed 
just the other way round. I switched that to member initialisation, thus 
swapping the order would break that and lead to crashes.
It is just not possible, because depending on the constructor the one or 
the other is first initialised.

It seems fine to me as it is right now.

PR #2870, opened 2024-02-22
Make check for zoidberg much faster

dschwoerer:
Importing zoidberg takes for me around 5 seconds. This only takes 0.1 :-)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -366,7 +366,7 @@ else()
   set(BOUT_GENERATE_FIELDOPS_DEFAULT OFF)
 endif()
 
-execute_process(COMMAND ${Python3_EXECUTABLE} -c "import zoidberg"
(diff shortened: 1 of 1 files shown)


ZedThree (approved):

ZedThree:
Readthedocs is failing due to:

> Command killed due to timeout or excessive memory consumption
> Command time: 1775s Return: 137

1775s is 29.5 minutes. 

The last passing build took 1227 seconds. Not sure why this one is so much slower!

dschwoerer:
I had that in PR #2648 as well, one was slow and timed out. The commits later, it was fast again:
1172 s: https://readthedocs.org/projects/bout-dev/builds/23525051/
1909 s: https://readthedocs.org/projects/bout-dev/builds/23520098/

Not sure what RTD is doing, but sometimes it is slow :-(
Can we just re-run the build? I don't think it makes sense for this one, it should have no impact on RTD. Or at least, the cmake part did not fail.

PR #2888, opened 2024-03-19
Add .zpm for sign-independent coding

dschwoerer:


dschwoerer:
I think it is not so bad - if you know the sign of the offset, you can pick a specific one, otherwise the other.
Even if it is not an issue right now with a specific compiler, it might become one for a different compiler or architecture (e.g. GPUs)

If we allow z splitting, that anyway changes :sweat_smile:

ZedThree (approved):

PR #2889, opened 2024-03-19
Pvcode + Cvode improvements

dschwoerer:
* Exposes more options to the user.
* Allows to dump if the solver fails. This can be useful for figuring out why the solver is slow. Currently only for pvode, but should also be possible to implement this for cvode (for cvode we can use a documented API)

### DIFF
--- a/include/bout/field.hxx
+++ b/include/bout/field.hxx
@@ -683,4 +683,12 @@ inline T floor(const T& var, BoutReal f, const std::string& rgn = "RGN_ALL") {
 
 #undef FIELD_FUNC
 
+template <typename T, typename = bout::utils::EnableIfField<T>, class... Types>
+inline T setName(T&& f, const std::string& name, Types... args) {
+#if BOUT_USE_TRACK
+  f.name = fmt::format(name, args...);
+#endif
+  return f;
+}
+
 #endif /* FIELD_H */
--- a/include/bout/field3d.hxx
+++ b/include/bout/field3d.hxx
@@ -514,6 +514,13 @@ private:
 
   /// RegionID over which the field is valid
   std::optional<size_t> regionID;
+
+  int tracking_state{0};
+  Options* tracking{nullptr};
+  std::string selfname{""};
+  template <class T>
+  Options* track(const T& change, std::string op);
+  Options* track(const BoutReal& change, std::string op);
 };
 
 // Non-member overloaded operators
--- a/include/bout/field3d.hxx
+++ b/include/bout/field3d.hxx
@@ -295,6 +295,10 @@ public:
   /// cuts on closed field lines?
   bool requiresTwistShift(bool twist_shift_enabled);
 
+  /// Enable a special tracking mode for debugging
+  /// Save all changes that, are done to the field, to tracking
+  Field3D& enableTracking(const std::string& name, Options& tracking);
+
   /////////////////////////////////////////////////////////
   // Data access
 
--- a/src/mesh/coordinates.cxx
+++ b/src/mesh/coordinates.cxx
@@ -1542,7 +1542,11 @@ Field3D Coordinates::Grad_par(const Field3D& var, CELL_LOC outloc,
   TRACE("Coordinates::Grad_par( Field3D )");
   ASSERT1(location == outloc || outloc == CELL_DEFAULT);
 
-  return ::DDY(var, outloc, method) * invSg();
+  if (invSg == nullptr) {
+    invSg = std::make_unique<FieldMetric>();
+    (*invSg) = 1.0 / sqrt(g_22);
+  }
+  return setName(::DDY(var, outloc, method) * invSg(), "Grad_par({:s})", var.name);
 }
 
 /////////////////////////////////////////////////////////
@@ -1601,7 +1605,7 @@ Field3D Coordinates::Div_par(const Field3D& f, CELL_LOC outloc,
     f_B.yup(i) = f.yup(i) / Bxy_floc.yup(i);
     f_B.ydown(i) = f.ydown(i) / Bxy_floc.ydown(i);
   }
-  return Bxy * Grad_par(f_B, outloc, method);
+  return setName(Bxy * Grad_par(f_B, outloc, method), "Div_par({:s})", f.name);
 }
 
 /////////////////////////////////////////////////////////
--- a/src/solver/impls/pvode/pvode.cxx
+++ b/src/solver/impls/pvode/pvode.cxx
@@ -42,11 +42,38 @@
 #include <pvode/iterativ.h> // contains the enum for types of preconditioning
 #include <pvode/pvbbdpre.h> // band preconditioner function prototypes
 
+#include <string>
+
 using namespace pvode;
 
 void solver_f(integer N, BoutReal t, N_Vector u, N_Vector udot, void* f_data);
 void solver_gloc(integer N, BoutReal t, BoutReal* u, BoutReal* udot, void* f_data);
 void solver_cfn(integer N, BoutReal t, N_Vector u, void* f_data);
+
+namespace {
+// local only
+void pvode_load_data_f3d(const std::vector<bool>& evolve_bndrys,
+                         std::vector<Field3D>& ffs, BoutReal* udata) {
+  int p = 0;
+  Mesh* mesh = ffs[0].getMesh();
+  const int nz = mesh->LocalNz;
+  for (const auto& bndry : {true, false}) {
+    for (const auto& i2d : mesh->getRegion2D(bndry ? "RGN_BNDRY" : "RGN_NOBNDRY")) {
+      for (int jz = 0; jz < nz; jz++) {
+        // Loop over 3D variables
+        std::vector<bool>::const_iterator evolve_bndry = evolve_bndrys.begin();
+        for (std::vector<Field3D>::iterator ff = ffs.begin(); ff != ffs.end(); ++ff) {
+          if (bndry && !*evolve_bndry)
+            continue;
+          (*ff)[mesh->ind2Dto3D(i2d, jz)] = udata[p];
+          p++;
+        }
+        ++evolve_bndry;
+      }
+    }
+  }
+}
+} // namespace
 
 const BoutReal ZERO = 0.0;
 
--- a/src/solver/impls/pvode/pvode.cxx
+++ b/src/solver/impls/pvode/pvode.cxx
@@ -212,6 +212,9 @@ int PvodeSolver::init() {
   for (i = 0; i < OPT_SIZE; i++) {
     ropt[i] = ZERO;
   }
+  /* iopt[MXSTEP] : maximum number of internal steps to be taken by *
+   *                the solver in its attempt to reach tout.        *
+   *                Optional input. (Default = 500).                */
   iopt[MXSTEP] = pvode_mxstep;
 
   {
--- a/src/solver/impls/pvode/pvode.cxx
+++ b/src/solver/impls/pvode/pvode.cxx
@@ -214,7 +214,34 @@ int PvodeSolver::init() {
   }
   iopt[MXSTEP] = pvode_mxstep;
 
-  cvode_mem = CVodeMalloc(neq, solver_f, simtime, u, BDF, NEWTON, SS, &reltol, &abstol,
+  {
+    /* ropt[H0]      : initial step size. Optional input.             */
+
+    /* ropt[HMAX]    : maximum absolute value of step size allowed.   *
+     *                 Optional input. (Default is infinity).         */
+    const BoutReal hmax(
+        (*options)["max_timestep"].doc("Maximum internal timestep").withDefault(-1.));
+    if (hmax > 0) {
+      ropt[HMAX] = hmax;
+    }
+    /* ropt[HMIN]    : minimum absolute value of step size allowed.   *
+     *                 Optional input. (Default is 0.0).              */
+    const BoutReal hmin(
+        (*options)["min_timestep"].doc("Minimum internal timestep").withDefault(-1.));
+    if (hmin > 0) {
+      ropt[HMIN] = hmin;
+    }
+    /* iopt[MAXORD] : maximum lmm order to be used by the solver.     *
+     *                Optional input. (Default = 12 for ADAMS, 5 for  *
+     *                BDF).                                           */
+    const int maxOrder((*options)["max_order"].doc("Maximum order").withDefault(-1));
+    if (maxOrder > 0) {
+      iopt[MAXORD] = maxOrder;
+    }
+  }
+  const bool use_adam((*options)["adams_moulton"].doc("Use Adams Moulton solver instead of BDF").withDefault(false));
+
+  cvode_mem = CVodeMalloc(neq, solver_f, simtime, u, use_adam ? ADAMS : BDF, NEWTON, SS, &reltol, &abstol,
                           this, nullptr, optIn, iopt, ropt, machEnv);
 
   if (cvode_mem == nullptr) {
--- a/src/solver/impls/pvode/pvode.cxx
+++ b/src/solver/impls/pvode/pvode.cxx
@@ -293,6 +293,49 @@ BoutReal PvodeSolver::run(BoutReal tout) {
   // Check return flag
   if (flag != SUCCESS) {
     output_error.write("ERROR CVODE step failed, flag = {:d}\n", flag);
+    CVodeMemRec* cv_mem = (CVodeMem)cvode_mem;
+    if (f2d.empty() and v2d.empty() and v3d.empty()) {
+      Options debug{};
+      using namespace std::string_literals;
+      Mesh* mesh{};
+      for (const auto& prefix : {"pre_"s, "residuum_"s}) {
+        std::vector<Field3D> ffs{};
+        std::vector<bool> evolve_bndrys{};
+        for (const auto& f : f3d) {
+          Field3D ff{0.};
+          ff.allocate();
+          ff.setLocation(f.location);
+          mesh = ff.getMesh();
+          debug[fmt::format("{:s}{:s}", prefix, f.name)] = ff;
+          ffs.push_back(ff);
+          evolve_bndrys.push_back(f.evolve_bndry);
+        }
+        pvode_load_data_f3d(evolve_bndrys, ffs,
+                            prefix == "pre_"s ? udata : N_VDATA(cv_mem->cv_acor));
+      }
+
+      for (auto& f : f3d) {
+        f.F_var->enableTracking(fmt::format("ddt_{:s}", f.name), debug);
+	setName(f.var, f.name);
+      }
+      run_rhs(simtime);
+
+      for (auto& f : f3d) {
+	debug[f.name] = *f.var;
+      }
+
+      if (mesh) {
+        mesh->outputVars(debug);
+        debug["BOUT_VERSION"].force(bout::version::as_double);
+      }
+
+      std::string outname = fmt::format(
+          "{}/BOUT.debug.{}.nc",
+          Options::root()["datadir"].withDefault<std::string>("data"), BoutComm::rank());
+
+      bout::OptionsNetCDF(outname).write(debug);
+      MPI_Barrier(BoutComm::get());
+    }
     return (-1.0);
   }
 
--- a/src/field/field3d.cxx
+++ b/src/field/field3d.cxx
@@ -243,6 +243,7 @@ Field3D& Field3D::operator=(const Field3D& rhs) {
   }
 
   TRACE("Field3D: Assignment from Field3D");
+  track(rhs, "operator=");
 
   // Copy base slice
   Field::operator=(rhs);
@@ -263,6 +264,7 @@ Field3D& Field3D::operator=(const Field3D& rhs) {
 
 Field3D& Field3D::operator=(Field3D&& rhs) {
   TRACE("Field3D: Assignment from Field3D");
+  track(rhs, "operator=");
 
   // Move parallel slices or delete existing ones.
   yup_fields = std::move(rhs.yup_fields);
@@ -283,6 +285,7 @@ Field3D& Field3D::operator=(Field3D&& rhs) {
 
 Field3D& Field3D::operator=(const Field2D& rhs) {
   TRACE("Field3D = Field2D");
+  track(rhs, "operator=");
 
   /// Check that the data is allocated
   ASSERT1(rhs.isAllocated());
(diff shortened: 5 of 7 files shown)


include/bout/field.hxx
```diff
@@ -683,4 +683,12 @@ inline T floor(const T& var, BoutReal f, const std::string& rgn = "RGN_ALL") {
 
 #undef FIELD_FUNC
 
+template <typename T, typename = bout::utils::EnableIfField<T>, class... Types>
+inline T setName(T&& f, const std::string& name, Types... args) {
```
ZedThree, line 687: I think most other setters like this are member functions

ZedThree:
My worry here is the tracking storing a *lot* of data if it's storing the full `Field3D` on every single change. Should it also be a compile-time option so that it completely disappears for production runs?

dschwoerer:
> My worry here is the tracking storing a _lot_ of data if it's storing the full `Field3D` on every single change. Should it also be a compile-time option so that it completely disappears for production runs?

It could be, but I don't think it is needed. I would personally keep it enabled for production runs, as it has a runtime switch, that is disabled by default. And I would also keep it like this, at least by default.

If a simulation crashes, it can be very useful to immediately get why it failed (this is currently the only use of the name tracking). Having to restart + queue again, just to get to the point where it crashes can be very time consuming, and the cost should be really negligable, if disabled.

I did notice a slow down in my simulations with this enabled, but that was when it was enabled all the time and the names did not get reset, thus the names have been growing exponentially.

dschwoerer:
On 3/19/24 16:34, Peter Hill wrote:
> ***@***.**** commented on this pull request.
> 
> ------------------------------------------------------------------------
> 
> In include/bout/field.hxx 
> <https://github.com/boutproject/BOUT-dev/pull/2889#discussion_r1530625642>:
> 
>> @@ -683,4 +683,12 @@ inline T floor(const T& var, BoutReal f, const std::string& rgn = "RGN_ALL") {
>   
>   #undef FIELD_FUNC
>   
> +template <typename T, typename = bout::utils::EnableIfField<T>, class... Types>
> +inline T setName(T&& f, const std::string& name, Types... args) {
> 
> I think most other setters like this are member functions

But then they return either Field, or need to be overwritten.

dschwoerer:
The cuda build fails with:
```log
/__w/BOUT-dev/BOUT-dev/src/field/field3d.cxx: In member function 'Options* Field3D::track(const T&, std::string)':
/__w/BOUT-dev/BOUT-dev/src/field/field3d.cxx:851:63: error: expected ';' before '}' token
  851 |     (*tracking)[outname].setAttributes({
      |                                                               ^
      |                                                               ;
/__w/BOUT-dev/BOUT-dev/src/field/field3d.cxx:851:64: error: expected primary-expression before ',' token
  851 |     (*tracking)[outname].setAttributes({
      |                                                                ^
/__w/BOUT-dev/BOUT-dev/src/field/field3d.cxx:851:66: error: expected primary-expression before '{' token
  851 |     (*tracking)[outname].setAttributes({
      |                                                                  ^
/__w/BOUT-dev/BOUT-dev/src/field/field3d.cxx:851:95: error: invalid use of void expression
  851 |     (*tracking)[outname].setAttributes({
      |                                                                                               ^
[ 16%] Building CUDA object CMakeFiles/bout++.dir/src/field/field_data.cxx.o
```

The relevant code is:
```c++
    (*tracking)[outname].setAttributes({
      {"operation", operation},
#if BOUT_USE_TRACK
      {"rhs.name", change.name},
#endif
    });
```

Is gcc9.4 not able to use setAttributes?

ZedThree:
> It could be, but I don't think it is needed. I would personally keep it enabled for production runs, as it has a runtime switch, that is disabled by default. And I would also keep it like this, at least by default.

We do try to completely remove debugging features for production runs, see `TRACE` for example. Being able to remove the function call entirely at lower `CHECK` levels would make me much happier about this.

> If a simulation crashes, it can be very useful to immediately get why it failed (this is currently the only use of the name tracking). Having to restart + queue again, just to get to the point where it crashes can be very time consuming, and the cost should be really negligable, if disabled.

This is true, but I would really like to understand the costs. It looks like it's storing potentially hundreds or thousands of copies of full fields, so although it looks like a really interesting and useful feature, it would be good to have a good idea of the costs.

> I did notice a slow down in my simulations with this enabled, but that was when it was enabled all the time and the names did not get reset, thus the names have been growing exponentially.

Is that with the current form?

> But then they return either Field, or need to be overwritten.

It should almost certainly return `Field&` and then this wouldn't be an issue

dschwoerer:
> > It could be, but I don't think it is needed. I would personally keep it enabled for production runs, as it has a runtime switch, that is disabled by default. And I would also keep it like this, at least by default.
> 
> We do try to completely remove debugging features for production runs, see `TRACE` for example. Being able to remove the function call entirely at lower `CHECK` levels would make me much happier about this.

If at all, I would be in favor of adding a new configure flag. I use this in runs with CHECK=0.

> > If a simulation crashes, it can be very useful to immediately get why it failed (this is currently the only use of the name tracking). Having to restart + queue again, just to get to the point where it crashes can be very time consuming, and the cost should be really negligable, if disabled.
> 
> This is true, but I would really like to understand the costs. It looks like it's storing potentially hundreds or thousands of copies of full fields, so although it looks like a really interesting and useful feature, it would be good to have a good idea of the costs.

It is only storing fields, if the solver has failed. Unless that happens, no additional storage is used.
There are some calls, but they are all no-ops, if you think that might be a problem, I can certainly guard them with `#if` statements, but I think that should be orthogonal to any of the flags we have.

In the case the solver crashes, I think the cost is really negligible. This is only done once, if the solver decided it cannot continue. For hermes-2, the dump file contains 53 `Field3D`s, while the normal dump file contains 45 `Field3D`s.
And part of that is just the metric, so I guess the overhead is less then 100% of the storage requirement of the run.

But if BOUT++ runs out of memory here, it would not be much worse then if not - the simulation finishes anyway.

> > I did notice a slow down in my simulations with this enabled, but that was when it was enabled all the time and the names did not get reset, thus the names have been growing exponentially.
> 
> Is that with the current form?

No, currently it is not noticeable. I activate it for all my runs.

> > But then they return either Field, or need to be overwritten.
> 
> It should almost certainly return `Field&` and then this wouldn't be an issue

That is exactly what I tried:

```patch
diff --git a/include/bout/field.hxx b/include/bout/field.hxx
index 04035f5b7..488aecaab 100644
--- a/include/bout/field.hxx
+++ b/include/bout/field.hxx
@@ -84,6 +84,14 @@ public:
     return *this;
   }
 
+  template <class... Types>
+  inline Field& setName(const std::string& name, Types... args) {
+#if BOUT_USE_TRACK
+    this->name = fmt::format(name, args...);
+#endif
+    return *this;
+  }
+
   std::string name;
 
 #if CHECK > 0
@@ -683,19 +691,4 @@ inline T floor(const T& var, BoutReal f, const std::string& rgn = "RGN_ALL") {
 
 #undef FIELD_FUNC
 
-template <typename T, typename = bout::utils::EnableIfField<T>, class... Types>
-inline void setName(T& f, const std::string& name, Types... args) {
-#if BOUT_USE_TRACK
-  f.name = fmt::format(name, args...);
-#endif
-}
-
-template <typename T, typename = bout::utils::EnableIfField<T>, class... Types>
-inline T setName(T&& f, const std::string& name, Types... args) {
-#if BOUT_USE_TRACK
-  f.name = fmt::format(name, args...);
-#endif
-  return f;
-}
-
 #endif /* FIELD_H */
diff --git a/src/mesh/coordinates.cxx b/src/mesh/coordinates.cxx
index 32774d622..643e148b1 100644
--- a/src/mesh/coordinates.cxx
+++ b/src/mesh/coordinates.cxx
@@ -1542,7 +1542,7 @@ Field3D Coordinates::Grad_par(const Field3D& var, CELL_LOC outloc,
   TRACE("Coordinates::Grad_par( Field3D )");
   ASSERT1(location == outloc || outloc == CELL_DEFAULT);
 
-  return setName(::DDY(var, outloc, method) * invSg(), "Grad_par({:s})", var.name);
+  return (::DDY(var, outloc, method) * invSg()).setName("Grad_par({:s})", var.name);
 }
 
 /////////////////////////////////////////////////////////
@@ -1601,7 +1601,7 @@ Field3D Coordinates::Div_par(const Field3D& f, CELL_LOC outloc,
     f_B.yup(i) = f.yup(i) / Bxy_floc.yup(i);
     f_B.ydown(i) = f.ydown(i) / Bxy_floc.ydown(i);
   }
-  return setName(Bxy * Grad_par(f_B, outloc, method), "Div_par({:s})", f.name);
+  return (Bxy * Grad_par(f_B, outloc, method)).setName("Div_par({:s})", f.name);
 }
 
 /////////////////////////////////////////////////////////
diff --git a/src/solver/impls/pvode/pvode.cxx b/src/solver/impls/pvode/pvode.cxx
index db28f64d8..c36985ae0 100644
--- a/src/solver/impls/pvode/pvode.cxx
+++ b/src/solver/impls/pvode/pvode.cxx
@@ -376,7 +376,7 @@ BoutReal PvodeSolver::run(BoutReal tout) {
 
       for (auto& f : f3d) {
         f.F_var->enableTracking(fmt::format("ddt_{:s}", f.name), debug);
-        setName(*f.var, f.name);
+        f.var->setName(f.name);
       }
       run_rhs(simtime);
```

gcc14 complains with:
```
/u/dave/soft/BOUT-dev/merge/src/mesh/coordinates.cxx: In member function ‘Field3D Coordinates::Grad_par(const Field3D&, CELL_LOC, const std::string&)’:                                                                                                                       /u/dave/soft/BOUT-dev/merge/src/mesh/coordinates.cxx:1545:56: error: could not convert ‘operator*(const Field3D&, const Field2D&)((* &((Coordinates*)this)->Coordinates::invSg())).Field3D::Field.Field::setName<std::__cxx11::basic_string<char, std::char_traits<char>, std:
:allocator<char> > >(std::__cxx11::basic_string<char>(((const char*)"Grad_par({:s})"), std::allocator<char>()), std::__cxx11::basic_string<char>(var.Field3D::Field.Field::name))’ from ‘Field’ to ‘Field3D’
 1545 |   return (::DDY(var, outloc, method) * invSg()).setName("Grad_par({:s})", var.name);      
      |          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                                        |
      |                                                        Field                       
/u/dave/soft/BOUT-dev/merge/src/mesh/coordinates.cxx: In member function ‘Field3D Coordinates::Div_par(const Field3D&, CELL_LOC, const std::string&)’:
/u/dave/soft/BOUT-dev/merge/src/mesh/coordinates.cxx:1604:55: error: could not convert ‘operator*(const Field2D&, const Field3D&)(Coordinates::Grad_par(const Field3D&, CELL_LOC, const std::string&)(f_B, outloc, (* & method))).Field3D::Field.Field::setName<std::__cxx11::
basic_string<char, std::char_traits<char>, std::allocator<char> > >(std::__cxx11::basic_string<char>(((const char*)"Div_par({:s})"), std::allocator<char>()), std::__cxx11::basic_string<char>(f.Field3D::Field.Field::name))’ from ‘Field’ to ‘Field3D’
 1604 |   return (Bxy * Grad_par(f_B, outloc, method)).setName("Div_par({:s})", f.name);                                                                                                                                                                                            |          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~
      |                                                       |
      |                                                       Field
```

ZedThree:
@dschwoerer After discussing this with @bendudson, we're quite concerned about the potential overhead of the temporary field dumping mechanism, both in terms of memory and performance, for what seems like a small benefit when debugging. We'd really like to see some performance analysis and scaling to see it doesn't have an affect when disabled. I'm also worried that we'd need to see `setName` everywhere in the code for this to be really useful.

Could something similar not be achieved with a custom monitor instead? 

Exposing more CVODE options to the user seems very straightforward and useful, so maybe that could be split out?

dschwoerer:
> @dschwoerer After discussing this with @bendudson, we're quite concerned about the potential overhead of the temporary field dumping mechanism, both in terms of memory and performance, for what seems like a small benefit when debugging.

But the memory overhead is only if the solver fails. Do you still think at that point that is an issue?
I know of @totork using BOUT++ in a inner-loop of an optimization script, where just failing vs crashing would matter. Thus having an option to disable would make sense.

> We'd really like to see some performance analysis and scaling to see it doesn't have an affect when disabled.

I can certainly run blob2d for this branch and next. But I am certain there are no significant differences, thus I haven't done it. Would you like to see something else? Would just the BOUT++ internal timings sufficient? Would you like mpi runs?

> I'm also worried that we'd need to see `setName` everywhere in the code for this to be really useful.

Sure, having them makes things more easily readable. But that is optional, and I can certainly add some if you think this is worth merging an wanted before merging. It would make `BOUT_USE_TRACK` also more useful. Right now 

> Could something similar not be achieved with a custom monitor instead?

I would not know how. If you can explain how to, I am happy to change the design.

> Exposing more CVODE options to the user seems very straightforward and useful, so maybe that could be split out?

I can split them out :+1:

src/solver/impls/pvode/pvode.cxx
```diff
@@ -185,10 +213,47 @@ int PvodeSolver::init() {
...
+    }
+  }
+  const bool use_adam((*options)["adams_moulton"]
+                          .doc("Use Adams Moulton solver instead of BDF")
+                          .withDefault(false));
+
+  debug_on_failure =
+      (*options)["debug_on_failure"]
+          .doc("Run an aditional rhs if the solver fails with extra tracking")
+          .withDefault(false);
```
dschwoerer, line 253: It is now disabled by default, and needs to be enabled with `solver:debug_on_failure=true`

ZedThree:
Ok, I think we understand this a bit better now. I'd still be much happier if this was a macro and only turned on for `CHECK > 0`. I'm quite worried about this proliferating everywhere and performance dying a death by a thousand cuts.

dschwoerer:
> Ok, I think we understand this a bit better now. I'd still be much happier if this was a macro and only turned on for `CHECK > 0`. I'm quite worried about this proliferating everywhere and performance dying a death by a thousand cuts.

That is a fair point. I do understand that worry, but at the same time I am convinced it is orthogonal to `CHECK=0`.
I am happy to add a new configure option. `-DBOUT_EXTRA_DEBUG_ON_FAILURE` ? I guess it make sense to have it disabled by default, as unless you know about the option and will look at the debug files, it is rather useless.

dschwoerer:
@ZedThree anything I can do to move this forward?

src/solver/impls/pvode/pvode.cxx
```diff
@@ -293,6 +358,57 @@ BoutReal PvodeSolver::run(BoutReal tout) {
...
+        }
+
+        for (auto& f : f3d) {
+          f.F_var->enableTracking(fmt::format("ddt_{:s}", f.name), debug);
+          setName(*f.var, f.name);
+        }
+        run_rhs(simtime);
+
+        for (auto& f : f3d) {
+          debug[f.name] = *f.var;
```
ZedThree, line 390: `saveParallel` already includes this

src/solver/impls/pvode/pvode.cxx
```diff
@@ -293,6 +358,57 @@ BoutReal PvodeSolver::run(BoutReal tout) {
...
+        const std::string outname =
+            fmt::format("{}/BOUT.debug.{}.nc",
+                        Options::root()["datadir"].withDefault<std::string>("data"),
+                        BoutComm::rank());
+
+        bout::OptionsIO::create(outname)->write(debug);
+        MPI_Barrier(BoutComm::get());
+      } else {
+        output_warn.write("debug_on_failure is currently only supported for Field3Ds");
+      }
```
ZedThree, line 410: Invert the conditional and move this first, so we can immediately `return` and then unindent the rest of the block

include/bout/field3d.hxx
```diff
@@ -514,6 +527,13 @@ private:
 
   /// RegionID over which the field is valid
   std::optional<size_t> regionID;
+
+  int tracking_state{0};
+  Options* tracking{nullptr};
```
ZedThree, line 532: I really don't like storing raw pointers. Probably this makes most sense as `weak_ptr`, a non-owning reference to a `shared_ptr` allocated and owned elsewhere. That way it's very obvious we don't own it and we can rely on it living long enough

ZedThree:
Exposing the additional PVODE options is just obvious and should probably just be it's own PR as well

mikekryjak:
@ZedThree just wanted to chime in and say that having this capability available would be extremely useful for me in optimising Hermes-3. I had already played with it quite a bit, but unfortunately I had to put it on the back burner for now because of the DLS paper and other work. I fully intend to do proper testing of this to optimise Hermes-3 performance sometime in Jan-Feb, or earlier if Michael Hardman can have a go.

I understand that there could be performance concerns - just wanted to say that this is much more than a small debugging benefit for me.

include/bout/options.hxx
```diff
@@ -949,6 +949,9 @@ Tensor<BoutReal> Options::as<Tensor<BoutReal>>(const Tensor<BoutReal>& similar_t
 /// Convert \p value to string
 std::string toString(const Options& value);
 
+/// Save the parallel fields
+void saveParallel(Options& opt, const std::string name, const Field3D& tosave);
```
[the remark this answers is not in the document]
dschwoerer: I do not think that works, as we assign to `name` as well as `f"{name}_y+1"` etc ...

dschwoerer (commented):
Thanks a lot for your review @ZedThree 
I forgot that this was pending on me :-(

I have addressed everything and this should be ready for re-review. Again, sorry for the delay.

src/sys/options.cxx
```diff
@@ -358,6 +358,23 @@ Options& Options::assign<>(Tensor<int> val, std::string source) {
   return *this;
 }
 
+void saveParallel(Options& opt, const std::string& name, const Field3D& tosave) {
+  opt[name] = tosave;
+  const size_t numberParallelSlices =
+      tosave.hasParallelSlices() ? 0 : tosave.getMesh()->ystart;
+  for (size_t i0 = 1; i0 <= numberParallelSlices; ++i0) {
+    for (int i : {i0, -i0}) {
```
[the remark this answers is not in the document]
dschwoerer: :shrug: Unfortunately there is no nice way to avoid this ...

src/mesh/coordinates.cxx
```diff
@@ -1590,7 +1590,7 @@ Field3D Coordinates::Div_par(const Field3D& f, CELL_LOC outloc,
     f_B.yup(i) = f.yup(i) / Bxy_floc.yup(i);
     f_B.ydown(i) = f.ydown(i) / Bxy_floc.ydown(i);
   }
-  return Bxy * Grad_par(f_B, outloc, method);
+  return xy * Grad_par(f_B, outloc, method);
```
dschwoerer:
```suggestion
  return Bxy * Grad_par(f_B, outloc, method);
```

src/solver/impls/euler/euler.cxx
```diff
@@ -134,7 +142,46 @@ void EulerSolver::take_step(BoutReal curtime, BoutReal dt, Array<BoutReal>& star
...
+    }
+
+    const std::string outnumber =
+        dump_at_time < -3 ? fmt::format(".{}", debug_counter++) : "";
+    const std::string outname =
+        fmt::format("{}/BOUT.debug{}.{}.nc",
+                    Options::root()["datadir"].withDefault<std::string>("data"),
+                    outnumber, BoutComm::rank());
+
+    bout::OptionsIO::create(outname)->write(debug);
```
ZedThree: I think we should be able to use `OptionsIO::createOutput()` here to get this to also work with ADIOS2. Something like:

```cpp
Options file_options = {{"prefix", fmt::format("BOUT.debug{}", outnumber)}};
bout::OptionsIO::createOutput(&file_options)->write(debug);
```

I haven't tested this!

src/solver/impls/euler/euler.cxx
```diff
@@ -134,7 +142,46 @@ void EulerSolver::take_step(BoutReal curtime, BoutReal dt, Array<BoutReal>& star
...
+    Options& debug = *debug_ptr;
+    Mesh* mesh{nullptr};
+    for (auto& f : f3d) {
+      saveParallel(debug, f.name, *f.var);
+      mesh = f.var->getMesh();
+    }
+
+    if (mesh != nullptr) {
+      mesh->outputVars(debug);
+      debug["BOUT_VERSION"].force(bout::version::as_double);
```
ZedThree: Could do:
```suggestion
      bout::experimental::addBuildFlagsToOptions(debug);
```
but the extra bits probably not necessary?

src/solver/impls/pvode/pvode.cxx
```diff
@@ -291,7 +365,54 @@ BoutReal PvodeSolver::run(BoutReal tout) {
...
+      using namespace std::string_literals;
+      Mesh* mesh{};
+      for (const auto& prefix : {"pre_"s, "residuum_"s}) {
+        std::vector<Field3D> list_of_fields{};
+        std::vector<bool> evolve_bndrys{};
+        for (const auto& f : f3d) {
+          mesh = f.var->getMesh();
+          Field3D to_load{0., mesh};
+          to_load.allocate();
+          to_load.setLocation(f.location);
```
ZedThree, line 383: Providing an initial value should also allocate. I had also thought that:

```suggestion
          Field3D to_load{0., mesh}.setLocation(f.location);
```

should also work, but that's not really important

src/solver/impls/pvode/pvode.cxx
```diff
@@ -291,7 +365,54 @@ BoutReal PvodeSolver::run(BoutReal tout) {
...
+      if (mesh != nullptr) {
+        mesh->outputVars(debug);
+        debug["BOUT_VERSION"].force(bout::version::as_double);
+      }
+
+      const std::string outname = fmt::format(
+          "{}/BOUT.debug.{}.nc",
+          Options::root()["datadir"].withDefault<std::string>("data"), BoutComm::rank());
+
+      bout::OptionsIO::create(outname)->write(debug);
```
ZedThree: Same comment as for the euler solver. In fact, if it's possible we might want to add this in more places, maybe worth a helper function to create these files?

dschwoerer: I think it should be possible. The CVode solver even has a documented API to get the residua :-)

src/solver/impls/euler/euler.cxx
dschwoerer: I do not think they are needed. I have moved it to the function, so we can change it later all in one place :+1:

src/sys/options/options_io.cxx
```diff
@@ -55,4 +56,12 @@ void writeDefaultOutputFile(Options& data) {
   OptionsIOFactory::getInstance().createOutput()->write(data);
 }
 
+void OptionsIO::write(const std::string& prefix, Options& data, Mesh* mesh) {
```
ZedThree, line 59: Nice little helper function, thanks!

PR #2891, opened 2024-03-21
Reintroduce removed copy function marked deprecated

dschwoerer:
This gives dependent codes more time to adopt.

It gives a warning now, with the intent to revert later to get the error again.

We can revert this commit later, to finally remove it.
Sorry that I did not think of this earlier, but it is not yet released, so might be still be useful ...


bendudson (approved):
Thanks @dschwoerer ! This is a good idea that I didn't think of. I guess this is why we have a `next` branch :)

PR #2916, opened 2024-05-24
Add getters for Laplacian flags

ZedThree:
These methods allow us to make the `{global, inner_boundary, outer_boundary}_flags` members private in `Laplacian`. This combo silences a whole bunch of clang-tidy warnings, mostly around implicit comparison between `int` and `bool`. One extra change was to stop passing the flags into `tridagMatrix` -- this was a bit redundant as it's a private method on `Laplacian` anyway, so we were passing members to a method.

While checking all this, I found that the `LaplaceSPT` dtor crashed clang-tidy, mostly likely because it was doing some pointer address fiddling. I've also fixed that, which also fixes a couple of other issues at the same time.

### DIFF
--- a/include/bout/invert_laplace.hxx
+++ b/include/bout/invert_laplace.hxx
@@ -238,6 +238,10 @@ public:
   virtual void setInnerBoundaryFlags(int f) { inner_boundary_flags = f; }
   virtual void setOuterBoundaryFlags(int f) { outer_boundary_flags = f; }
 
+  virtual int getGlobalFlags() const { return global_flags; }
+  virtual int getInnerBoundaryFlags() const { return inner_boundary_flags; }
+  virtual int getOuterBoundaryFlags() const { return outer_boundary_flags; }
+
   /// Does this solver use Field3D coefficients (true) or only their DC component (false)
   virtual bool uses3DCoefs() const { return false; }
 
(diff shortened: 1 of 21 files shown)


include/bout/invert_laplace.hxx
```diff
@@ -322,15 +340,13 @@ protected:
                    CELL_LOC loc = CELL_DEFAULT);
 
   void tridagMatrix(dcomplex* avec, dcomplex* bvec, dcomplex* cvec, dcomplex* bk, int jy,
-                    int kz, BoutReal kwave, int flags, int inner_boundary_flags,
-                    int outer_boundary_flags, const Field2D* a, const Field2D* ccoef,
+                    int kz, BoutReal kwave, const Field2D* a, const Field2D* ccoef,
```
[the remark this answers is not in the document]
ZedThree: Could also fix this at the same time I suppose, `a` -> `acoef` and `d` -> `dcoef`

include/bout/invert_laplace.hxx
```diff
@@ -308,9 +312,23 @@ protected:
...
+  bool isOuterBoundaryFlagSet(int flag) const {
+    return (outer_boundary_flags & flag) != 0;
+  }
+
+  /// Return true if \p flag is set for the inner boundary condition
+  /// on the inner processor(s)
+  bool isInnerBoundaryFlagSetOnFirstX(int flag) const;
+  /// Return true if \p flag is set for the outer boundary condition
+  /// on the outer processor(s)
+  bool isOuterBoundaryFlagSetOnLastX(int flag) const;
```
dschwoerer, line 331: It sounds like this gets the flag from the first X proc, and uses it here.
```suggestion
  /// Return true if \p flag is set for the inner boundary condition
  /// and this is the first proc in X direction
  bool isInnerBoundaryFlagSetOnFirstX(int flag) const;
  /// Return true if \p flag is set for the outer boundary condition
  /// and this the last proc in X direction
  bool isOuterBoundaryFlagSetOnLastX(int flag) const;
```

dschwoerer (approved):

include/bout/invert_laplace.hxx
ZedThree: Thanks!

dschwoerer (approved):

PR #2917, opened 2024-05-24
WIP: PETSc XZHermiteSpline

ZedThree:
Continues #2858 pulling out a separate class for `PetscXZHermiteSpline` that was started in #2651

This does have a working implementation, but I've tried to switch it the existing `PetscMatrix` wrapper, and unfortunately I've gone wrong somewhere -- probably at the boundaries.

Double unfortunately, I really don't have much time to spend digging into this. @dschwoerer Maybe you want to take a look?


ZedThree:
Ok, I have managed to convert this to use the `PetscMatrix` interface. I need to spend a lot of time faffing about with `GlobalIndexer` and realised the source of my pain: it can only convert indices that are both local to and *owned by* the current process. For the interpolation here, we want to deal with indices that are on other processes. And because `PetscMatrix` wants to convert local indices to global ones, we need to first convert the global index of the (x, z) offsets into local ones, and then convert *those* to global indices.

So I've implemented a new `GlobalIndexer` subclass that will probably only work for very simple parallel decompositions, but can convert e.g. negative local indices into global ones. That gets an index converter that is actually pretty straightforward and can be used with `PetscMatrix`, which then allows a pretty clean implementation in our new interpolation!

include/bout/interpolation_xz.hxx
```diff
@@ -251,19 +248,68 @@ public:
...
+  // and the h10 and h11 basis functions are applied to its derivative
+  // along the interpolation direction.
+  Field3D h00_x;
+  Field3D h01_x;
+  Field3D h10_x;
+  Field3D h11_x;
+  Field3D h00_z;
+  Field3D h01_z;
+  Field3D h10_z;
+  Field3D h11_z;
```
dschwoerer, line 280: They should only be used in the constructor, so no need to keep them around

src/mesh/interpolation/petsc_hermite_spline_xz.cxx
```diff
@@ -0,0 +1,240 @@
...
+    h10_x[i] = t_x * (1. - t_x) * (1. - t_x);
+    h10_z[i] = t_z * (1. - t_z) * (1. - t_z);
+
+    h11_x[i] = (t_x * t_x * t_x) - (t_x * t_x);
+    h11_z[i] = (t_z * t_z * t_z) - (t_z * t_z);
+
+    // Need to convert from global indices to local
+    const auto i_c = Field3D::ind_type(
+        (((localmesh->getLocalXIndex(i_corn) * ny) + (y + y_offset)) * nz
+         + localmesh->getLocalZIndex(k_corner(x, y, z))),
```
dschwoerer, line 144: Note that this seems like it would work with Z-splitting, but it does not!

This is inherently incompatible with z-splitting. It only works if ny and nz are the same as the local ...

src/mesh/interpolation/petsc_hermite_spline_xz.cxx
```diff
@@ -0,0 +1,240 @@
...
+Field3D
+PetscXZHermiteSpline::interpolate(const Field3D& f,
+                                  [[maybe_unused]] const std::string& region) const {
+
+  ASSERT1(f.getMesh() == localmesh);
+  const PetscVector<Field3D> f_petsc(f, indexer);
+  const PetscVector<Field3D> f_result = weights * f_petsc;
+  // TODO: we should only consider the passed-in region
+  // const auto region2 = y_offset == 0 ? region : fmt::format("RGN_YPAR_{:+d}", y_offset);
+  return f_result.toField();
```
dschwoerer, line 223: This does not set the region - is this set in fci.cxx/fci.hxx?
Otherwise this will not work with high check level and 3d metrics, afaik.

dschwoerer (commented):
A few things are indeed much nicer then with the abstraction.

I am not sure whether z-splitting will be tricky. Right now it works because only x is split, and x is the slowest index.

One advantage with the alternative implementation is, that the monotonic version is implementation for all hermite spline interpolation, although we probably want to keep at most the non-petsc version, although I am not sure of that, we might just say: Get PETSc if you want to do FCI ...

Have you by chance did a performance measurement? This implementation should copy slightly less data ...

PR #2922, opened 2024-06-11
Fix version checking for SUNDIALS

Steven-Roberts:
`find_package` required version SUNDIALS 4.x.x as opposed to >=4.0.0.


Steven-Roberts:
It seems the variables added to `CONFIG_LDFLAGS` here are never set. Is this line needed?

https://github.com/boutproject/BOUT-dev/blob/ed69fd674bef813f01656bd99962fad880b922f9/cmake/SetupBOUTThirdParty.cmake#L306

cmake/SetupBOUTThirdParty.cmake
```diff
@@ -297,7 +297,11 @@ if (BOUT_USE_SUNDIALS)
     FetchContent_MakeAvailable(sundials)
     message(STATUS "SUNDIALS done configuring")
   else()
-    find_package(SUNDIALS 4 REQUIRED)
+    ENABLE_LANGUAGE(C)
+    find_package(SUNDIALS REQUIRED)
```
ZedThree, line 301: I take it:

```suggestion
    find_package(SUNDIALS 4.0.0 REQUIRED)
```

doesn't work?

ZedThree:
That line is used to set the linking flags that `bout-config` outputs. Those variables were probably set in previous versions (of BOUT++). They are a bit tricky to get robustly and generically from CMake, so we probably need to revisit how to set these.

cmake/SetupBOUTThirdParty.cmake
Steven-Roberts: No, that only seems to work for 4.x.x. For a newer version, I get a warning like
```
CMake Warning at cmake/FindSUNDIALS.cmake:34 (find_package):
  Could not find a configuration file for package "SUNDIALS" that is
  compatible with requested version "4".

  The following configuration files were considered but not accepted:

    .../SUNDIALSConfig.cmake, version: 7.0.0

  cmake/SetupBOUTThirdParty.cmake:301 (find_package)
  CMakeLists.txt:479 (include)
```
from [here](https://github.com/boutproject/BOUT-dev/blob/ed69fd674bef813f01656bd99962fad880b922f9/cmake/FindSUNDIALS.cmake#L34) (with QUIET removed). This leads to a linking error later.

With cmake 3.19, I think `find_package(SUNDIALS 4...7 REQUIRED)` might work.

ZedThree: Oh yes, pretty sure `4...7` is the correct syntax

Steven-Roberts: Turns out that the `4...7` syntax doesn't work with
https://github.com/LLNL/sundials/blob/b685654c6958699821fc68a822b66076fec8cae7/CMakeLists.txt#L267
so the current manual check is probably the best option.

ZedThree: Ah, yes, I see now, that's very unfortunate. `SameMajorVersion` means "the requested version must have exactly the same major version", and makes no allowances for requesting "any of these major versions".

Agreed, the manual check is the best option.

cmake/SetupBOUTThirdParty.cmake
```diff
@@ -297,7 +297,11 @@ if (BOUT_USE_SUNDIALS)
     FetchContent_MakeAvailable(sundials)
     message(STATUS "SUNDIALS done configuring")
   else()
-    find_package(SUNDIALS 4 REQUIRED)
+    ENABLE_LANGUAGE(C)
```
ZedThree, line 300:
```suggestion
    enable_language(C)
```

ZedThree (approved):

ZedThree:
Thanks @Steven-Roberts !

ZedThree:
Numpy 2.0 is out, and we're getting some incompatible versions in CI which is causing the job failures

bendudson (approved):

PR #2924, opened 2024-06-17
Remove versions from pip dependencies

dschwoerer:
The lower bounds are mostly outdated

Alternative to #2923

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,8 +1,8 @@
-Jinja2>=2.11.3
-numpy>=1.14.1
-scipy>=1.0.0
-netcdf4~=1.6.0
-matplotlib>=2.0.0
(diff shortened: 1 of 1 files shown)


ZedThree:
I think these are probably sensible versions:

```
Jinja2~=3.1.0
numpy~=1.26.0
scipy>=1.11.0
netcdf4~=1.7.1
matplotlib>=3.7.0
Cython~=3.0.0
boututils~=0.2.1
boutdata~=0.2.1
```

dschwoerer:
Are they just for CI or are they also for users?

I am fine if they are for CI, but if they are for users, I would like to 
be less strict and also allow generally newer versions.

bendudson:
Versions updated in #2929

PR #2930, opened 2024-06-24
Next IMEX issue

maggul:
`solver:imex=false` flag was causing a potential error when `solver:type=arkode` was used,  this is now fixed. Also, now omits creation of implicit algebraic solver when running with an explicit solver.  

When `solver:imex=false` and `solver:implicit=true` and `solver:explicit=true`, then the module now sets `solver:imex=true` and proceeds with this option. If you prefer returning an error instead, just let me know.


src/solver/impls/arkode/arkode.cxx
```diff
@@ -195,30 +195,29 @@ int ArkodeSolver::init() {
...
+  else if (solve_explicit)
+  {
+    arkode_mem = callWithSUNContext(ARKStepCreate, suncontext, arkode_rhs, nullptr,
+                                    simtime, uvec);    
+  }
+  else
+  {
+    arkode_mem = callWithSUNContext(ARKStepCreate, suncontext, nullptr, arkode_rhs,
+                                    simtime, uvec);
+  }
```
Steven-Roberts, line 215:
```suggestion
  else if (solve_explicit && !imex)
  {
    arkode_mem = callWithSUNContext(ARKStepCreate, suncontext, arkode_rhs, nullptr,
                                    simtime, uvec);    
  }
  else if (solve_implicit && !imex)
  {
    arkode_mem = callWithSUNContext(ARKStepCreate, suncontext, nullptr, arkode_rhs,
                                    simtime, uvec);
  } else {
    throw BoutException("Incompatible imex, explicit, and implicit options\n");
  }
```
The assert above can be removed with this

Steven-Roberts (commented):
I think it would make sense to have `solve_explicit` and `solve_implicit` be `false` by default now.

https://github.com/boutproject/BOUT-dev/blob/9376d3553a91301f258568026c80889fe30b7444/src/solver/impls/arkode/arkode.cxx#L78-L81

Perhaps an even simpler option here is to consolidate the imex, explicit, and implicit options into a single option that accepts the string `"imex"`, `"explicit"`, or `"implicit"` as possible values.

src/solver/impls/arkode/arkode.cxx
```diff
@@ -194,36 +197,42 @@ int ArkodeSolver::init() {
...
+  else if (treatment == "explicit")
+  {
+    arkode_mem = callWithSUNContext(ARKStepCreate, suncontext, arkode_rhs, nullptr,
+                                    simtime, uvec);    
+  }
+  else
+  {
+    arkode_mem = callWithSUNContext(ARKStepCreate, suncontext, nullptr, arkode_rhs,
+                                    simtime, uvec);
+  }
```
Steven-Roberts, line 225: With this the `ASSERT1` can be removed
```suggestion
  else if (treatment == "implicit")
  {
    arkode_mem = callWithSUNContext(ARKStepCreate, suncontext, nullptr, arkode_rhs,
                                    simtime, uvec);
  } else {
    throw BoutException("Invalid treatment: {}\n", treatment);
  }
```

src/solver/impls/arkode/arkode.cxx
maggul: ASSERT1 is removed

bendudson:
One way to handle the `treatment` and `adap_method` options is to use an enum: The BOUT_ENUM_CLASS (https://github.com/boutproject/BOUT-dev/blob/master/include/bout/bout_enum_class.hxx#L69) creates an enum class and handles the conversion to/from strings. It's used in the CVODE interface for the positivity constraint option https://github.com/boutproject/BOUT-dev/blob/master/src/solver/impls/cvode/cvode.cxx#L75 . 
An issue is that `explicit` is a reserved keyword in C++, so I don't think we can use it in an enum. For now the string approach seems to work ok.

bendudson (approved):

Steven-Roberts:
If an option value is assigned to an enum, does it check if the user-specified value is a valid enum entry? If so, that might be cleaner than what we have now.

bendudson:
Yes: It will try to convert the string to the enum, and throw an exception if the value is invalid (https://github.com/boutproject/BOUT-dev/blob/master/include/bout/bout_enum_class.hxx#L89). That code could be improved to print a list of acceptable values.

Steven-Roberts:
Ah nice. That seems preferable to what we have now. Do you mind trying that @maggul?

bendudson:
We'll have to choose a name that's not "explicit". I guess we could use uppercase: `IMEX`, `Implicit` and `Explicit`

maggul:
I have been working on this couple of days now and I was not able to get it work yet. I had no trouble having an enum in the class but couldn't get the string from the command line to convert. I appreciate your help if you can.

bendudson:
Hi @maggul Please give this a try. I removed the "imex", "implicit" and "explicit" input booleans, and changed both `treatment` and `adap_method` inputs to enum classes. If the user provides an invalid value then the error message will contain a list of valid values.

maggul:
Thank you for your quick response. I had tried a similar approach and encountered the same error message, which still appears even after making this modification.

`./conduction solver:type=arkode solver:treatment = Implicit`

gives the following error:

```
====== Back trace ======
 -> Treatment TreatmentFromString(const std::string&) on line 66 of '/home/maggul/ABOUND/BOUT-dev/src/solver/impls/arkode/arkode.hxx'

Did not find enum Implicit. Valid values:  explicit imex implicit

Abort(1) on node 0 (rank 0 in comm 496): application called MPI_Abort(comm=0x84000000, 1) - process 0
```
On the other hand, the treatment seems to be inputted already.

`Option solver:treatment = Implicit (Command line)
`

bendudson:
That error message is supposed to happen: The valid values are `Valid values:  explicit imex implicit` so "Implicit" is not valid. The input is case sensitive, so "implicit" will work.

maggul:
Thanks for letting me know. My modification works now as well. It seems like case sensitivity was the issue all the time.

bendudson (approved):

PR #2896, opened 2024-03-28
Update SUNDIALS support (rebase onto next)

ZedThree:
Replaces #2885 (rebased that branch onto `next`)

PR #2937, opened 2024-07-04
SUNDIALS v7 support (backport to v5)

ZedThree:
Backport of #2896 and #2922 into v5.1.1

This drops support for SUNDIALS < v4, so maybe this means this release should be 5.2.0 rather than 5.1.1? Everything else is a pretty minor change though


dschwoerer:
It seems < v4 was anyway not working, so making this more obvious is not a breaking change, but rather a bug fix, I would argue.
https://github.com/boutproject/BOUT-dev/pull/2840

Needing a newer c++ might however a breaking change, justifying 5.2.0 ...

ZedThree:
Ah, ok, I hadn't realised < v4 was broken. Ok, probably fine for 5.1.1 then. `next` is v6 and includes upgrading to C++17.

Looks like I should've attempted to compile this locally, there's some C++17 bits that need downgrading to 14

ZedThree:
I think Fedora mirrors are maybe having some issues

dschwoerer:
It seems just one of them is bad. We probably should set a higher minimum bandwidth for dnf, to pick a faster one. The default is quite low, as it should also work for poor mobile connections.

dschwoerer:
Slow mirrors should be fixed by #2940

dschwoerer (approved):

PR #2950, opened 2024-07-29
A few more fixes to the examples

tomc271:
A couple of lines are just commented-out rather than deleted -- please delete them if appropriate (or let me know so that I will).

### DIFF
--- a/examples/subsampling/monitor.cxx
+++ b/examples/subsampling/monitor.cxx
@@ -16,10 +16,7 @@ class Monitor1dDump : public Monitor {
 public:
   Monitor1dDump(BoutReal timestep, std::string section_name)
       : Monitor(timestep),
-        output_file(getCustomOutputName(Options::root()[section_name]),
-                    Options::root()[section_name]["append"].withDefault(false)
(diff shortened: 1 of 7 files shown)


examples/subsampling/monitor.cxx
```diff
@@ -31,7 +28,7 @@ class Monitor1dDump : public Monitor {
         output[item.name].attributes["time_dimension"] = "t";
       }
     }
-    output_file.write(output);
+//    output_file.write(output);
```
dschwoerer, line 31: Again, the data should be dumped, otherwise it is rather useless.

Or did you test that?

include/options_netcdf.hxx
```diff
@@ -1,4 +0,0 @@
-#pragma once
-// BOUT++ header shim
-#warning Header "options_netcdf.hxx" has moved to "bout/options_netcdf.hxx". Run `bin/bout-v5-header-upgrader.py` to fix
```
dschwoerer, line 3: I think we should either remove all compat headers, or none.

I would be in favor of deleting, as we probably have by now anyway breaking changes that need adoption ...

PR #2959, opened 2024-08-12
Update boutdata and boututils submodules

dschwoerer:
boututils has been merged into the boutdata project. As such we can remove the boututuils submodule and update the compat symlinks.

Replaces #2956

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -13,6 +13,3 @@
 [submodule "externalpackages/boutdata"]
 	path = externalpackages/boutdata
 	url = https://github.com/boutproject/boutdata.git
-[submodule "externalpackages/boututils"]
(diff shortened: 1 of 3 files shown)


dschwoerer:
Do we want these updates also in v5.1.1?
I think the .tar.gz we release contains the submodules, so that would make sense, but I am not sure whether we should test the merge a bit more?

ZedThree:
Yes, I think 5.1.1 is appropriate. Assuming the tests actually pass, I think that's good enough

dschwoerer:
I have rebased to the rc and force-pushed, now all tests are passing :+1:

ZedThree (approved):

ZedThree:
Thanks @dschwoerer !

PR #2942, opened 2024-07-14
mark fmt::format() const

tchaikov:
fmt v11 preserves and enforces the constness of the formatted argument. without this change, the tree fails to build with fmt v11, like:

```
In file included from /usr/include/fmt/format.h:41,
                 from /usr/include/fmt/core.h:5,
                 from /builddir/build/BUILD/bout++-5.1.0-build/BOUT++-v5.1.0/include/bout/sys/expressionparser.hxx:33:
/usr/include/fmt/base.h: In instantiation of â€˜static void fmt::v11::detail::value<Context>::format_custom_arg(void*, typename Context::parse_context_type&, Context&) [with T = Options; Formatter = fmt::v11::formatter<Options>; Context = fmt::v11::context; typename Context::parse_context_type = fmt::v11::basic_format_parse_context<char>]â€™:
/usr/include/fmt/base.h:1373:19:   required from â€˜fmt::v11::detail::value<Context>::value(T&) [with T = Options; Context = fmt::v11::context]â€™
 1373 |     custom.format = format_custom_arg<
      |     ~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~
 1374 |         value_type, typename Context::template formatter_type<value_type>>;
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/fmt/base.h:1631:41:   required from â€˜constexpr fmt::v11::detail::value<Context> fmt::v11::detail::make_arg(T&) [with bool PACKED = true; Context = fmt::v11::context; T = Options; typename std::enable_if<PACKED, int>::type <anonymous> = 0]â€™
 1631 |   return {arg_mapper<Context>().map(val)};
      |                                         ^
/usr/include/fmt/base.h:2002:74:   required from â€˜constexpr fmt::v11::detail::format_arg_store<Context, NUM_ARGS, 0, DESC> fmt::v11::make_format_args(T& ...) [with Context = context; T = {Options}; long unsigned int NUM_ARGS = 1; long unsigned int NUM_NAMED_ARGS = 0; long long unsigned int DESC = 15; typename std::enable_if<(NUM_NAMED_ARGS == 0), int>::type <anonymous> = 0]â€™
 2002 |   return {{detail::make_arg<NUM_ARGS <= detail::max_packed_args, Context>(
[...]

PR #2961, opened 2024-08-13
Fmt update

dschwoerer:
Rebase of #2942 for next
So that we do not have that many test failures, and it is easier to spot regressions.


bendudson (approved):
Thanks @dschwoerer !

PR #2962, opened 2024-08-17
Fix Field3D::setBoundaryTo parallel boundary conditions

bendudson:
Didn't handle parallel boundaries correctly. Now handles input arguments with and without parallelslices. If the input doesn't have parallel slices then they should be cleared in the target field, and to/fromFieldAligned used to set Y boundaries in field-aligned coordinates.

See discussion here of issue caused by previous implementation: https://github.com/bendudson/hermes-3/pull/248/commits/97a21fdfaeb5a1672a57d5dbb4261145bb195566

### DIFF
--- a/src/field/field3d.cxx
+++ b/src/field/field3d.cxx
@@ -451,19 +451,71 @@ void Field3D::setBoundaryTo(const Field3D& f3d) {
 
   allocate(); // Make sure data allocated
 
-  /// Loop over boundary regions
-  for (const auto& reg : fieldmesh->getBoundaries()) {
-    /// Loop within each region
-    for (reg->first(); !reg->isDone(); reg->next()) {
-      for (int z = 0; z < nz; z++) {
-        // Get value half-way between cells
-        BoutReal val =
-            0.5 * (f3d(reg->x, reg->y, z) + f3d(reg->x - reg->bx, reg->y - reg->by, z));
-        // Set to this value
-        (*this)(reg->x, reg->y, z) =
-            2. * val - (*this)(reg->x - reg->bx, reg->y - reg->by, z);
+  if (!fieldmesh->periodicX) {
+    // X boundaries
+    if (fieldmesh->firstX()) {
+      for (int jy = fieldmesh->ystart; jy <= fieldmesh->yend; ++jy) {
+        for (int jz = 0; jz < fieldmesh->LocalNz; jz++) {
+          BoutReal val = 0.5 * (f3d(fieldmesh->xstart, jy, jz) + f3d(fieldmesh->xstart - 1, jy, jz));
+          (*this)(fieldmesh->xstart - 1, jy, jz) = 2. * val - (*this)(fieldmesh->xstart, jy, jz);
+        }
       }
     }
+    if (fieldmesh->lastX()) {
+      for (int jy = fieldmesh->ystart; jy <= fieldmesh->yend; ++jy) {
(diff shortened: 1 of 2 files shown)


src/field/field3d.cxx
```diff
@@ -451,19 +451,81 @@ void Field3D::setBoundaryTo(const Field3D& f3d) {
...
+
+    for (auto& bndry : fieldmesh->getBoundariesPar()) {
+      const Field3D& f3d_next = f3d.ynext(bndry->dir);
+      Field3D& this_next = ynext(bndry->dir);
+
+      for (bndry->first(); !bndry->isDone(); bndry->next()) {
+        const int x = bndry->x;
+        const int y = bndry->y;
+        const int z = bndry->z;
+        BoutReal val = 0.5 * (f3d(x, y, z) + f3d_next(x, y, z));
```
dschwoerer, line 492:
```suggestion
        BoutReal val = 0.5 * (f3d(x, y, z) + f3d_next(x, y+bndr->dir, z));
```

src/field/field3d.cxx
```diff
@@ -451,19 +451,81 @@ void Field3D::setBoundaryTo(const Field3D& f3d) {
...
+        this_next(x, y, z) = 2. * val - (*this)(x, y, z);
```
dschwoerer, line 493:
```suggestion
        this_next(x, y+bndry->dir, z) = 2. * val - (*this)(x, y, z);
```

dschwoerer (changes requested):
I thought the yboundary iterator had changed, but if it compiles it seems that has not yet been merged.

On a more general question, what is the purpose of this function? I do understand what it does, but why would one want to use it? Why not simply copy the boundaries? In what cases is this useful?

src/field/field3d.cxx
```diff
@@ -451,19 +451,81 @@ void Field3D::setBoundaryTo(const Field3D& f3d) {
...
+      }
+    }
```
dschwoerer, line 495: I think this should be roughly the new syntax:
```suggestion
    for (auto& region : fieldmesh->getBoundariesPar()) {
      for (auto& pnt: region) {
        pnt.dirichlet_o1(*this, (f3d.ynext(pnt.dir)[pnt.ind().yp(pnt.dir)] + f3d[pnt.ind()] ) * 0.5);
      }
    }
```

bendudson:
Thanks @dschwoerer ! What is being done is:
- A field (P) comes from the solver
- It is modified, removing negative values and handling low density regions
- The modified field has a boundary condition applied
- The original field now should be modified to have the same boundary condition as the modified field.

One option is to just copy the boundary cells, but then the boundary condition would be slightly different. I'm not sure which would be better behaved numerically.

dschwoerer:
I see, so the assumption is that both fields are very similar.
One thing that could happen with this method, if you want to ensure positive also of the boundary cells, and you modify them in this operation, you lose that property, and there is no way around this. Also note that this only copies Dirichlet boundary conditions, that should probably be documented.

I have pushed a fix, either as an additional commit (https://github.com/boutproject/BOUT-dev/tree/fix-setboundaryto-fixup) or squashed to remove the broken commits from history, for easier bisecting in the future (https://github.com/boutproject/BOUT-dev/tree/fix-setboundaryto-rebase)

dschwoerer:
One test is failing as it triggers the assertion about parallel slices:
```
26/50 Test #28: test-invertable-operator ............***Failed    4.89 sec
```

PR #2964, opened 2024-08-21
Resolve merge conflict for v5.1.1-rc

dschwoerer:
@ZedThree Is merging master into the RC the correct way to resolve the conflict?


ZedThree:
Yes that's right, although the PR is empty? I guess we'll see if Github works out that this fixes the conflict?

ZedThree (approved):

dschwoerer:
Yes, I think the correct version was already in the rc, so we only need to make sure git is aware that this is the correct version :+1:

PR #2975, opened 2024-09-17
Read 2D variables into Field3D

bendudson:
If the variable in the mesh file is 2D (x,y) but read into a Field3D then it would previously fail. This reads as 2D and assumes constant in z.

### DIFF
--- a/src/mesh/data/gridfromfile.cxx
+++ b/src/mesh/data/gridfromfile.cxx
@@ -392,6 +392,25 @@ void GridFile::readField(Mesh* m, const std::string& name, int ys, int yd, int n
 
   var.allocate();
 
+  if (size.size() == 2) {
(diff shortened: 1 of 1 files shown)


src/mesh/data/gridfromfile.cxx
```diff
@@ -392,6 +392,25 @@ void GridFile::readField(Mesh* m, const std::string& name, int ys, int yd, int n
 
   var.allocate();
 
+  if (size.size() == 2) {
+    // A 2D field in the grid file
+    readFieldAttributes(data[name], var);
+
+    var.allocate();
```
ZedThree, line 399: Duplicated from just above?

```suggestion
```

ZedThree (approved):

PR #2976, opened 2024-09-18
CI: Disable Fedora builds due to problem with missing packages

ZedThree:


dschwoerer:
Could you clarify what the error is?
https://github.com/boutproject/BOUT-dev/actions/runs/10925719586/job/30327869312 seems to have run without error, recently.
There may have been an intermediate error in resolving some dependencies, but that should be resolved by now.

ZedThree:
That's the first one that's succeeded in about two weeks. If it keeps passing, we can close this. Otherwise, we need to disable the build or find a way of only soft-failing in the case of package problems.

PR #2977, opened 2024-09-18
CI: Bump all ubuntu images

ZedThree:


ZedThree:
CI failing because `-lnetCDF::netcdf` is getting into `bout-config` via `SetupBOUTThirdParty.cmake`, probably: https://github.com/boutproject/BOUT-dev/blob/a628ec012859db7de7d5b7743ad9a8760044e0fe/cmake/SetupBOUTThirdParty.cmake#L174

Likely because newer Ubuntu version is supplying `netCDFConfig.cmake` and setting things differently than we expect.

Not a major issue really, but we should fix

bendudson (approved):

PR #2995, opened 2024-10-21
Try to do the right thing for linking netCDF::netcdf

dschwoerer:
This will certainly fail in some cases, but I think there is in general no easy way to get the needed flags for linking.
If anybody has better ideas: great :-)


ZedThree:
Seems like a reasonable attempt!

That whole bit is getting quite long now, but be nice to pull it out into a `configure_bout_config_script` function somewhere else to keep the main `CMakeFiles.txt` a bit tidier

PR #3002, opened 2024-10-22
Ensure pointer is checked before dereferencing

dschwoerer:
dynamic_cast can return a nullptr, thus we should check. Otherwise gcc raises a warning.


dschwoerer:
RTD is randomly failing, other tests pass

PR #3007, opened 2024-10-23
Fix lots of clang-tidy issues in fv_ops

ZedThree:
Mostly:

- include required headers
- rename some short identifiers
- make some local variables `const`
   - in a few cases, this involved renaming them to avoid reuse of the
     same variable for different cases (e.g. left/right fluxes)

This is mostly to avoid getting the clang-tidy warnings in #2873

### DIFF
--- a/include/bout/fv_ops.hxx
+++ b/include/bout/fv_ops.hxx
@@ -83,10 +83,17 @@ Field3D D4DY4_index(const Field3D& f_in, bool bndry_flux = true);
    */
 struct Stencil1D {
   // Cell centre values
+#if CHECK > 0
   BoutReal c = BoutNaN, m = BoutNaN, p = BoutNaN, mm = BoutNaN, pp = BoutNaN;
 
   // Left and right cell face values
   BoutReal L = BoutNaN, R = BoutNaN;
+#else
+  BoutReal c, m, p, mm, pp;
+
+  // Left and right cell face values
+  BoutReal L, R;
+#endif
 };
 
 /*!
--- a/include/bout/fv_ops.hxx
+++ b/include/bout/fv_ops.hxx
@@ -443,14 +443,12 @@ Field3D Div_f_v(const Field3D& n_in, const Vector3D& v, bool bndry_flux) {
     if ((i.x() == mesh->xend) && (mesh->lastX())) {
       // At right boundary in X
       if (bndry_flux) {
-        BoutReal flux = BoutNaN;
-        if (v_R > 0.0) {
-          // Flux to boundary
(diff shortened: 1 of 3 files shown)


include/bout/fv_ops.hxx
```diff
@@ -286,45 +301,45 @@ const Field3D Div_par(const Field3D& f_in, const Field3D& v_in,
...
 
-        cellboundary(s); // Calculate s.R and s.L
+        cellboundary(stencil); // Calculate s.R and s.L
 
         ////////////////////////////////////////////
         // Right boundary
 
         // Calculate velocity at right boundary (y+1/2)
-        BoutReal vpar = 0.5 * (v(i, j, k) + v(i, j + 1, k));
-        BoutReal flux;
```
dschwoerer, line 301: performance implications?

include/bout/fv_ops.hxx
```diff
@@ -383,63 +398,60 @@ const Field3D Div_par(const Field3D& f_in, const Field3D& v_in,
...
 
-    cellboundary(s);
+    cellboundary(stencil);
 
     if ((i.x() == mesh->xend) && (mesh->lastX())) {
       // At right boundary in X
       if (bndry_flux) {
-        BoutReal flux;
-        if (vR > 0.0) {
+        BoutReal flux = BoutNaN;
```
dschwoerer, line 439: performance? Apparently the compiler cannot deduct that flux is always written, and thus cannot optimize the not-needed write ...

include/bout/fv_ops.hxx
```diff
@@ -451,21 +463,21 @@ const Field3D Div_f_v(const Field3D& n_in, const Vector3D& v, bool bndry_flux) {
       // At left boundary in X
 
       if (bndry_flux) {
-        BoutReal flux;
-        if (vL < 0.0) {
+        BoutReal flux = BoutNaN;
```
dschwoerer, line 466: performance?

dschwoerer (commented):
I am a bit worried about adding not-needed writes by default.
I would be fine if they would only be present with `-DCHECK>0` as I do not expect that they add significant overhead, but it still does not feels like something we should be doing for `-DCHECK=0`

dschwoerer:
@ZedThree sorry I did not approve / merge this after I got my suggestions merged. Do we still need this? I think we should be mostly clang-tidy warning-free, with #3335 ?

dschwoerer:
I think they are mostly fixed, as #3335 touched that code.

ZedThree:
There's still a few, but I'll open a separate PR

PR #3010, opened 2024-10-24
ARKODE_MRI interface

maggul:
Initial draft for the interface of the ARKODE_MRI interface and an integrated test with MRI


ZedThree:
@maggul Please could you reword or squash these commits? Most of them just have "some progress" as the message which is not terribly useful I'm afraid. Commit messages should be descriptive, for example "Add ARKODE multi-rate solver". This makes them useful for others looking through the history. Thanks!

maggul:
Unfortunately, I was under the impression that these commits would disappear after the merge, so I didn’t pay close attention to them. Is there anything you would recommend I do to change the commit messages? @ZedThree

ZedThree:
Please either reword them or squash them together with a nice commit message, and then either force-push to this branch or a new branch and PR.

PR #3016, opened 2024-10-30
CI: clang-tidy-review tweaks

ZedThree:
- Use annotations instead of comments
- Don't post a "LGTM" comment
- Also don't check identifier length, just way too noisy


dschwoerer (approved):
LGTM

the short identifiers are indeed annoying.

PR #3021, opened 2024-11-07
Fix recompilation cascade

ZedThree:
I noticed that making a new commit triggered recompilation of many more files than required. This was due to `version.hxx` being included in `bout.hxx`, instead of just in the implementation file.

This prompted me to reorganise some of the unit tests and fix some other includes to try and cut down on unnecessary rebuilds.


bendudson (approved):
Wow! This was quite the rabbit warren. Thanks @ZedThree for draining this swamp.

Issue #2701, opened 2023-04-27
Add CI with HYPRE


dschwoerer:
Helps to catch error in hypre-only code, e.g. #2667

PR #3022, opened 2024-11-07
Fix minor HYPRE and ADIOS2 compilation issues

ZedThree:
It turns out we're not building HYPRE on CI at all, and we're silently not _actually_ using ADIOS2, and that's let a couple of compilation issues creep in.

This fixes the compilation and CI issues (hopefully).


ZedThree:
Hmm, more problems with 3D metrics and the Hypre laplacians. Guess I'll have to build that config locally to fix them all.

ZedThree:
I think it's mostly just making some types `auto` instead of `Field2D` or whatever, but I need to find some time to look into it

dschwoerer:
Fixes #2701

ZedThree:
Failing test is probably because ADIOS2 is missing from `bout++Config.cmake.in` -- need a `find_dependency` and set `ADIOS2_ROOT` or similar

ZedThree:
The test turns out to be failing because the solver wasn't converging. Switching `laplacexy:hypre_solver_type` to `gmres` fixes this.

The default HYPRE solver is `bicgstab` -- is this test an indication to change the default to `gmres`, or are we happy to just change the type for this test?

ZedThree:
Still having a weird issue with the coverage build

ZedThree:
Think this has finally fixed the issues I was originally trying to fix!

cmake/SetupBOUTThirdParty.cmake
```diff
@@ -190,9 +190,12 @@ endif()
 message(STATUS "NetCDF support: ${BOUT_USE_NETCDF}")
 set(BOUT_HAS_NETCDF ${BOUT_USE_NETCDF})
 
-option(BOUT_USE_ADIOS2 "Enable support for ADIOS output" ON)
+option(BOUT_USE_ADIOS2 "Enable support for ADIOS output" OFF)
```
dschwoerer, line 193: I would prefer if such feature would be automatically enabled if they are found.

dschwoerer (approved):
Looks good, thanks :+1:

PR #3024, opened 2024-11-08
Ensure PETSc headers are included after `bout/petsclib.hxx`

ZedThree:
From that header:

> PETSc "helpfully" defines macros for MPI functions that clobber the
> real names, and short of `#undef`-ing all of them in every file
> that includes any PETSc header, we can define the following macro
> which should disable them, which I'm sure will work forever. This
> means we _must_ `#include` this header _before_ any PETSc header!

A recent commit reordered the PETSc headers in this file


dschwoerer (approved):
Thanks, looks good!

PR #3027, opened 2024-11-16
Coordinates: Replacing sqrt(g_22) with JB

bendudson:
Aiming to consistently handle left-handed coordinate systems, where J is negative. These happen in the standard BOUT++ field-aligned coordinates when the poloidal field is negative.

Implemented on top of `refactor-coordinates` #2873

### DIFF
--- a/src/mesh/difops.cxx
+++ b/src/mesh/difops.cxx
@@ -479,7 +479,8 @@ Coordinates::FieldMetric b0xGrad_dot_Grad(const Field2D& phi, const Field2D& A,
 
   // Upwind A using these velocities
   Coordinates::FieldMetric result = VDDX(vx, A, outloc) + VDDY(vy, A, outloc);
-  result /= SQ(metric->J()) * metric->Bxy(); // J^2 B same sign for left & right handed coordinates
+  result /= SQ(metric->J())
+            * metric->Bxy(); // J^2 B same sign for left & right handed coordinates
 
   ASSERT1(result.getLocation() == outloc);
 
--- a/src/mesh/difops.cxx
+++ b/src/mesh/difops.cxx
@@ -2,9 +2,9 @@
  * Various differential operators defined on BOUT grid
  *
  **************************************************************************
- * Copyright 2010 B.D.Dudson, S.Farley, M.V.Umansky, X.Q.Xu
+ * Copyright 2010 - 2024 BOUT++ contributors
  *
- * Contact: Ben Dudson, bd512@york.ac.uk
+ * Contact: Ben Dudson, dudson2@llnl.gov
  *
  * This file is part of BOUT++.
  *
@@ -104,7 +104,9 @@ Field3D Grad_parP(const Field3D& apar, const Field3D& f) {
   for (int x = 1; x <= mesh->LocalNx - 2; x++) {
     for (int y = mesh->ystart; y <= mesh->yend; y++) {
       for (int z = 0; z < ncz; z++) {
-        BoutReal by = 1. / sqrt(metric->g_22(x, y, z));
+
+        // Note: by is negative in a left-handed coordinate system
+        BoutReal by = 1. / (metric->J(x, y, z) * metric->Bxy(x, y, z));
         // Z indices zm and zp
         int const zm = (z - 1 + ncz) % ncz;
         int const zp = (z + 1) % ncz;
@@ -258,14 +260,13 @@ Field3D Div_par(const Field3D& f, const Field3D& v) {
         BoutReal const vR = 0.5 * (v(i, j, k) + v.yup()(i, j + 1, k));
 
         // Calculate flux at right boundary (y+1/2)
+        // Note: Magnitude of B at the midpoint used rather than J/sqrt(g_22)
         BoutReal const fluxRight =
-            fR * vR * (coord->J(i, j, k) + coord->J(i, j + 1, k))
-            / (sqrt(coord->g_22(i, j, k)) + sqrt(coord->g_22(i, j + 1, k)));
+            fR * vR * 2 / (coord->Bxy(i, j, k) + coord->Bxy(i, j + 1, k));
 
         // Calculate at left boundary (y-1/2)
         BoutReal const fluxLeft =
-            fL * vL * (coord->J(i, j, k) + coord->J(i, j - 1, k))
-            / (sqrt(coord->g_22(i, j, k)) + sqrt(coord->g_22(i, j - 1, k)));
+            fL * vL * 2 / (coord->Bxy(i, j, k) + coord->Bxy(i, j - 1, k));
 
         result(i, j, k) =
             (fluxRight - fluxLeft) / (coord->dy(i, j, k) * coord->J(i, j, k));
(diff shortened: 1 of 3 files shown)


ZedThree (approved):

ZedThree:
Ah, this currently doesn't work because we've not added the overload for `Bxy(x, y, z)`

dschwoerer:
Does this fix things for FCI where JB is not equal to sqrt(g22)?
I wanted to discuss this with Ben on Wednesday.

bendudson:
> Does this fix things for FCI where JB is not equal to sqrt(g22)? I wanted to discuss this with Ben on Wednesday.

Hey @dschwoerer I hope this will help with FCI too: I'm trying to remove all use of `sqrt(g_22)` from the code, and assumptions that J is positive. B is always positive (as it's the magnitude of B), but J can be positive or negative. In the tokamak case J < 0 when Bp < 0. 

I haven't yet looked at the boundary conditions, or how to handle parallel components of vector quantities: When J < 0 it means that the y coordinate is in the opposite direction to B. Hence the 'y' component of a vector `v` is in the opposite direction to `b dot v`. It'll take some time to work out the implications for the operators e.g. `Grad_par()` takes a scalar argument, whereas `Div_par()` takes a vector component (`b dot v`).

ZedThree:
We also have some checks that `J` is positive that I guess need to be loosened?

bendudson:
> We also have some checks that `J` is positive that I guess need to be loosened?

Ah yes. I think we should check that all elements of J have the same sign: All positive or all negative.

src/mesh/difops.cxx
```diff
@@ -285,7 +286,7 @@ Field3D Div_par_flux(const Field3D& v, const Field3D& f, CELL_LOC outloc,
   auto Bxy_floc = f.getCoordinates()->Bxy();
 
   if (!f.hasParallelSlices()) {
-    return metric->Bxy() * FDDY(v, f / Bxy_floc, outloc, method) / sqrt(metric->g_22());
+    return FDDY(v, f / Bxy_floc, outloc, method) / metric->J();
```
dschwoerer, line 289: Should it really be `Bxy_floc`? I thought this term comes from the curvi-linear geometry, and thus should be independent of B (for non clebsch)

src/mesh/difops.cxx
```diff
@@ -294,7 +295,7 @@ Field3D Div_par_flux(const Field3D& v, const Field3D& f, CELL_LOC outloc,
   f_B.splitParallelSlices();
   f_B.yup() = f.yup() / Bxy_floc;
   f_B.ydown() = f.ydown() / Bxy_floc;
-  return metric->Bxy() * FDDY(v, f_B, outloc, method) / sqrt(metric->g_22());
+  return FDDY(v, f_B, outloc, method) / metric->J();
```
dschwoerer, line 298: Same here

src/mesh/difops.cxx
```diff
@@ -478,7 +479,8 @@ Coordinates::FieldMetric b0xGrad_dot_Grad(const Field2D& phi, const Field2D& A,
 
   // Upwind A using these velocities
   Coordinates::FieldMetric result = VDDX(vx, A, outloc) + VDDY(vy, A, outloc);
-  result /= metric->J() * sqrt(metric->g_22());
+  result /= SQ(metric->J())
+            * metric->Bxy(); // J^2 B same sign for left & right handed coordinates
```
dschwoerer, line 483: move comment above to avoid strange formatting?

src/mesh/difops.cxx
```diff
@@ -595,7 +597,7 @@ Field3D b0xGrad_dot_Grad(const Field3D& phi, const Field3D& A, CELL_LOC outloc)
 
   Field3D result = VDDX(vx, A, outloc) + VDDY(vy, A, outloc) + VDDZ(vz, A, outloc);
 
-  result /= (metric->J() * sqrt(metric->g_22()));
+  result /= SQ(metric->J()) * metric->Bxy(); // J^2 B
```
dschwoerer, line 600: Is that not only true for Clebsch?

dschwoerer (commented):
It seems some of the cases are (still) only valid for Clebsch.
I have not looked in too much detail, so if you disagree I can look in more detail ...

src/mesh/difops.cxx
```diff
@@ -104,7 +104,9 @@ Field3D Grad_parP(const Field3D& apar, const Field3D& f) {
   for (int x = 1; x <= mesh->LocalNx - 2; x++) {
     for (int y = mesh->ystart; y <= mesh->yend; y++) {
       for (int z = 0; z < ncz; z++) {
-        BoutReal by = 1. / sqrt(metric->g_22(x, y, z));
+
+        // Note: by is negative in a left-handed coordinate system
+        BoutReal by = 1. / (metric->J(x, y, z) * metric->Bxy(x, y, z));
```
dschwoerer, line 109:
```suggestion
        BoutReal by = 1. / (metric->J(x, y, z) * metric->Bxy()(x, y, z));
```

src/mesh/difops.cxx
```diff
@@ -258,14 +260,13 @@ Field3D Div_par(const Field3D& f, const Field3D& v) {
         BoutReal const vR = 0.5 * (v(i, j, k) + v.yup()(i, j + 1, k));
 
         // Calculate flux at right boundary (y+1/2)
+        // Note: Magnitude of B at the midpoint used rather than J/sqrt(g_22)
         BoutReal const fluxRight =
-            fR * vR * (coord->J(i, j, k) + coord->J(i, j + 1, k))
-            / (sqrt(coord->g_22(i, j, k)) + sqrt(coord->g_22(i, j + 1, k)));
+            fR * vR * 2 / (coord->Bxy(i, j, k) + coord->Bxy(i, j + 1, k));
```
dschwoerer, line 265:
```suggestion
            fR * vR * 2 / (coord->Bxy()(i, j, k) + coord->Bxy()(i, j + 1, k));
```

src/mesh/difops.cxx
```diff
@@ -258,14 +260,13 @@ Field3D Div_par(const Field3D& f, const Field3D& v) {
...
 
         // Calculate at left boundary (y-1/2)
         BoutReal const fluxLeft =
-            fL * vL * (coord->J(i, j, k) + coord->J(i, j - 1, k))
-            / (sqrt(coord->g_22(i, j, k)) + sqrt(coord->g_22(i, j - 1, k)));
+            fL * vL * 2 / (coord->Bxy(i, j, k) + coord->Bxy(i, j - 1, k));
```
dschwoerer, line 269:
```suggestion
            fL * vL * 2 / (coord->Bxy()(i, j, k) + coord->Bxy()(i, j - 1, k));
```

dschwoerer (commented):
Here are some fixes for the compilation errors.

Alternative is to add the function to `metric`, like `J` and `g_22` ...

bendudson:
> It seems some of the cases are (still) only valid for Clebsch. I have not looked in too much detail, so if you disagree I can look in more detail ...

@dschwoerer I agree: The identity `sqrt(g_22) = JB` is only for Clebsch coordinates. In general how should the 'parallel' direction be defined? Perhaps we use `sign(J) * sqrt(g_22)` as the parallel metric, which would be consistent with the Clebsch case but I think more general. That would mean that if `J > 0` then `b` is in the same direction as `e_y`, but if `J < 0` then they are in opposite directions.

bendudson:
Note: Mesh checks that `J > 0`. Change to check that `J` has the same sign (+ or -) in all cells.

bendudson:
Closing as this needs more careful thought on handling non-Clebsch coordinates and left-handed coordinates.

PR #3030, opened 2024-11-21
Fix Field3D::setBoundaryTo for FCI methods

bendudson:
Without this fix, boundary conditions set on yup/down fields are not applied when a boundary is copied from one field to another.

@dschwoerer This is needed in Hermes-3 with FCI so that this line correctly applies the sheath boundary conditions to the density field:
https://github.com/bendudson/hermes-3/blob/master/src/evolve_density.cxx#L217

### DIFF
--- a/src/field/field3d.cxx
+++ b/src/field/field3d.cxx
@@ -504,13 +504,15 @@ void Field3D::setBoundaryTo(const Field3D& f3d) {
         pnt.dirichlet_o2(*this, val);
       }
     }
-    return;
   }
 
   // Non-FCI.
   // Transform to field-aligned coordinates?
   // Loop over boundary regions
   for (const auto& reg : fieldmesh->getBoundaries()) {
+    if (isFci() && reg->by != 0) {
+      continue;
+    }
     /// Loop within each region
     for (reg->first(); !reg->isDone(); reg->next()) {
       for (int z = 0; z < nz; z++) {
--- a/src/field/field3d.cxx
+++ b/src/field/field3d.cxx
@@ -491,7 +491,25 @@ void Field3D::setBoundaryTo(const Field3D& f3d) {
 
   allocate(); // Make sure data allocated
 
-  /// Loop over boundary regions
+  if (isFci()) {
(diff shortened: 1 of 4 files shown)


src/field/field3d.cxx
```diff
@@ -491,7 +491,25 @@ void Field3D::setBoundaryTo(const Field3D& f3d) {
...
+    // Set yup/ydown using midpoint values from f3d
+    ASSERT1(f3d.hasParallelSlices());
+    ASSERT1(hasParallelSlices());
+
+    for (auto& region : fieldmesh->getBoundariesPar()) {
+      for (const auto& pnt : *region) {
+        // Interpolate midpoint value in f3d
+        const BoutReal val = pnt.interpolate_sheath_o1(f3d);
+        // Set the same boundary value in this field
+        pnt.dirichlet_o2(*this, val);
```
dschwoerer, line 504: We should either take o2 twice, or o1 twice:
```suggestion
        pnt.dirichlet_o1(*this, val);
```

I think `o1` should be sufficient, as `f3d` and `*this` should be close. But I am not sure what is more stable ...

src/field/field3d.cxx
```diff
@@ -491,7 +491,25 @@ void Field3D::setBoundaryTo(const Field3D& f3d) {
...
+      }
+    }
+    return;
```
dschwoerer, line 507: For hermes that might work, but should we not also copy the x-boundary conditions with FCI?
```suggestion
```

dschwoerer (changes requested):

src/field/field3d.cxx
dschwoerer: The FA code also uses first order, so I have pushed a fixed based on that ...

dschwoerer:
@bendudson Are you ok with these changes?

dschwoerer:
I just noticed #2962 is also open - is the other still needed with this one?

bendudson (commented):
Thanks @dschwoerer ! The changes look good to me.
I think #2962 is not needed once this is merged.

dschwoerer (approved):

PR #3036, opened 2024-11-29
Avoid using the wrong grid by accident

dschwoerer:
I sometimes have `[mesh:file]` set in the input file, and specify `grid` on the command line, only to be confused why the wrong grid was picked.

### DIFF
--- a/src/mesh/mesh.cxx
+++ b/src/mesh/mesh.cxx
@@ -29,8 +29,16 @@ MeshFactory::ReturnType MeshFactory::create(const std::string& type, Options* op
 
   if (options->isSet("file") or Options::root().isSet("grid")) {
     // Specified mesh file
-    const auto grid_name =
-        (*options)["file"].withDefault(Options::root()["grid"].withDefault(""));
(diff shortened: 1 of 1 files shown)


ZedThree:
We could just completely remove one of these names in 6.0? The [upgrader](https://github.com/boutproject/boutdata/tree/master/src/boutupgrader) would make fixing it easy.

dschwoerer:
There is one use - you could have several meshes, and they all could rely on `grid`. However, that seems rather unlikely ...

If we remove things, then `grid` needs to go, as otherwise it is not possible to have different meshes based on different grid files ...

Note that there are a few thinks like this, where we have a name in the global namespace, that then is used as default for a value in a section, I think e.g. nout.

Having the global options is nice if you want to change things quickly from the commandline. That can probably not be updated with the updater, as that would require parsing scripts - or worse, command history and human memory ...

ZedThree (approved):

ZedThree:
`--mesh:file=filename` is no great hardship though :)

It's true that the upgrader can't check command line usage, but we could have an explicit deprecation warning/deleted exception

PR #3038, opened 2024-12-02
Deprecate options that are only used as default for other options

dschwoerer:
See #3036 for motivation


bendudson (approved):
Thanks, @dschwoerer !

PR #3055, opened 2025-01-09
Hypre3D Laplacian diagnostics, parallel boundary default fix

bendudson:
- Hypre3D Laplacian solver now saves diagnostic data if `Laplacian::savePerformance()` is called. 
- The default parallel boundary condition "parallel_dirichlet" is not implemented. Changed to "parallel_dirichlet_o2".
- Updated and fixed the elm-pb example: Added `phiSolver->savePerformance` call, and fixed the `data-hypre` input.


bendudson:
Merged into #3022

PR #3061, opened 2025-01-17
Make iteration more robust and give more options in LaplaceNaulin

johnomotani:
I found this old branch. Maybe it's still useful? I have no time to test/update now though, so if it needs work, volunteers needed!

Previously a cycle decreasing the underrelax_factor was triggered if the error increased on an iteration. However sometimes it can be more robust to just continue the iteration with moderate increases in the error - there are cases when decreasing the underrelax_factor never results in a decreasing error, but continuing the main iteration loop for a few steps would. This commit tries to handle this kind of case, with several new parameters which can be adjusted to control how and when underrelax_factor is decreased.

Also documents more options.


bendudson:
Thanks, @johnomotani ! I'm using the Naulin solver for electromagnetic Hermes-3 simulations so will try this out. I added #2983 a while ago that does something a bit similar: It tries to reach a given atol/rtol, but if they are not reached after `maxits` iterations then it will accept a looser tolerance. Often the error will be go back down on the next call / time step.

bendudson (approved):

Issue #3060, opened 2025-01-15
Build failing on MacOS 15.2 (M1) 


sun51027:
Dear developers, 

I am installing Hermes-3 (along with its automatic download of BOUT-dev) on my personal laptop (M1 Mac) running macOS 15.2.
I have executed the cmake command:
`$ cmake . -B build -DBOUT_DOWNLOAD_SUNDIALS=ON -DBOUT_ENABLE_BACKTRACE=Off -DCMAKE_OSX_ARCHITECTURES=arm64` 
then built it by `$ cmake --build build`
However, it is quite tricky for me to solve the issue below:
```
[ 13%] Built target pvode
[ 14%] Building CXX object external/BOUT-dev/externalpackages/PVODE/CMakeFiles/pvpre.dir/precon/pvbbdpre.cpp.o
[ 14%] Building CXX object external/BOUT-dev/externalpackages/PVODE/CMakeFiles/pvpre.dir/precon/band.cpp.o
[ 14%] Linking CXX shared library ../../lib/libpvpre.dylib
Undefined symbols for architecture arm64:
  "pvode::N_VWrmsNorm(pvode::nvec_type*, pvode::nvec_type*)", referenced from:
      pvode::PVBBDPrecon(int, double, pvode::nvec_type*, pvode::nvec_type*, int, int*, double, pvode::nvec_type*, double, double, long*, void*, pvode::nvec_type*, pvode::nvec_type*, pvode::nvec_type*) in pvbbdpre.cpp.o
  "pvode::RSqrt(double)", referenced from:
      pvode::PVBBDPrecon(int, double, pvode::nvec_type*, pvode::nvec_type*, int, int*, double, pvode::nvec_type*, double, double, long*, void*, pvode::nvec_type*, pvode::nvec_type*, pvode::nvec_type*) in pvbbdpre.cpp.o
  "pvode::N_VScale(double, pvode::nvec_type*, pvode::nvec_type*)", referenced from:
      pvode::PVBBDPrecon(int, double, pvode::nvec_type*, pvode::nvec_type*, int, int*, double, pvode::nvec_type*, double, double, long*, void*, pvode::nvec_type*, pvode::nvec_type*, pvode::nvec_type*) in pvbbdpre.cpp.o
[...]

PR #3063, opened 2025-01-17
Use isfinite, fix pvode linking

bendudson:
Should fix issue #3060 , building on MacOS

- Use `std::isfinite` rather than the non-standard `finite` function
- Link PVODE pvpre library against pvode library

also:
- Update copyright notices in touched files


dschwoerer (approved):
Thanks @bendudson

PR #3075, opened 2025-02-19
Fixes zoidberg

dschwoerer:
#3054 for master

Should resolve the current CI failures for master on fedora, as that comes with the new zoidberg.

I also included the no-upper-bounds patch for the python packages, as I think also in master we want to be able to run on up-to-date systems. If we want to also test for a fixed, lower version, that should probably be just one of the CI runs, not all but fedora.


dschwoerer:
#3075 and #3076 are both needed to fix the CI failure ...

bendudson (approved):

PR #3078, opened 2025-02-25
Support BOUT_FOR_RAJA GPU field operators

ggeorgakoudis:
* Generate BOUT_FOR_RAJA code for GPU execution of field operators


dschwoerer (changes requested):
This PR roughly doubles the lines of `gen_fieldops.jinja` - making it very hard to read.

I think we should either manage to move more to the python file, or split it into two files, which of course makes it more tricky to keep in sync, but might be the lesser of two evils ...

Also look at the clang-tidy comments.

bendudson:
This was replaced by #3385

PR #2887, opened 2024-03-19
Add isFci for fields

dschwoerer:

PR #3079, opened 2025-03-03
Changes for FCI with hermes-3

dschwoerer:
Opened a PR to get the documentation build.

The PR needs likely be split up, in e.g. #2651 #2887 #3004 and once they are merged likely more ...


src/field/gen_fieldops.jinja
```diff
@@ -12,10 +12,41 @@
     {% if lhs == rhs == "Field3D" %}
       {{out.name}}.setRegion({{lhs.name}}.getMesh()->getCommonRegion({{lhs.name}}.getRegionID(),
                                                                      {{rhs.name}}.getRegionID()));
+#if BOUT_USE_FCI_AUTOMAGIC
```
bendudson, line 15: I think `BOUT_USE_FCI_AUTOMAGIC` may be useful to get something working, but I don't think it's a good way to go: We will have code that either breaks when this is turned off, or performs too many operations when it is turned on. I think the options are:

a) Have two functions / operators: One that discards yup/ydown, and the other that keeps them. 

b) Always operate on yup/down if they are present, but add a method to fields e.g `withoutParallelSlices()` that returns a copy of the field without the yup/down slices. That could be used e.g in assignment
```
Field3D f = g; // f has copies of g.yup and g.ydown
Field3D f = g.withoutParallelSlices(); // f has no parallel slices
```
I suspect that operating on two fields, one with slices and one without, should be an error: Either they should both have slices, or neither have.

PR #3085, opened 2025-03-05
Small updates for FCI output

bendudson:
FCITransform reads cell center coordinates from the mesh, and saves to the output.
This is to facilitate code coupling and visualisation.

### DIFF
--- a/include/bout/coordinates.hxx
+++ b/include/bout/coordinates.hxx
@@ -9,9 +9,9 @@
  *
  * 
  **************************************************************************
- * Copyright 2014 B.D.Dudson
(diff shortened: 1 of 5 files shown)


ZedThree (approved):
Should it also save $\phi$ as well?

dschwoerer:
The toroidal angle? Unfortunately, I think zoidberg does not store this ...

In general it would be nice to provide an option for the user, what to store ...

For visualisation, it might also be useful to `{for,back}ward_{R,Z}`, so that the data can be interpolated in `y` direction ...

PR #3090, opened 2025-03-12
CI: Avoid issues with special characters

dschwoerer:
This time I tested it :innocent: :
https://github.com/dschwoerer/BOUT-dev/actions/runs/13806780796

It did run twice, and if you run it several times, it slows starts expanding the indention of code, there are now two commits, and the second commit does that:
https://github.com/dschwoerer/BOUT-dev/pull/45/commits/fda08d2a276303fc1179c26bd110cda6249a98ad

So git-clang-format only formats the lines just before and after the change, but if they are changed, that will cause the next lines to be now the adjacent lines ...
If the formatting commit triggers a new git-clang-format run, that might cause some load on the system ...

One option would be to do it in one go:
```bash
while ! git-clang-format ...;
do
   :
done
```

### DIFF
--- a/.github/workflows/clang-format.yml
+++ b/.github/workflows/clang-format.yml
@@ -23,7 +23,7 @@ jobs:
       - name: Run clang-format
         id: format
         run:
-          while ! git clang-format origin/${{ github.base_ref }} ; do true ; done
+          while ! git clang-format origin/${{ github.base_ref }} ; do git add . ; done
 
       - name: Commit to the PR branch
         uses: stefanzweifel/git-auto-commit-action@v5


ZedThree:
Ah, that could indeed be a problem. Maybe it would be better to run `clang-format` on _everything_ -- if we do it once, that should avoid too much issue in the PRs?

dschwoerer:
Here now with the loop-ed approach:

https://github.com/dschwoerer/BOUT-dev/actions/runs/13808906584/job/38625749054

It still at most formats the whole file, so that might be nicer than run it once on everything ...

I do not have the time to update the PR, feel free to to do so ...

dschwoerer:
Now it at most formats the whole file, and does so in only one commit.
If it turns out to be excessive in some cases, we can consider changing it to not run again if the last commit was a formatting commit ...

dschwoerer:
> Ah, that could indeed be a problem. Maybe it would be better to run `clang-format` on _everything_ -- if we do it once, that should avoid too much issue in the PRs?

I think we did that for v5 - and I still have nightmares about the resulting merge conflicts ...

Many commits I just rewrote, because I was not able to get it to format exactly the same way as it was done in next.

dschwoerer:
This CI is currently broken in next, causing PRs to net getting formatted at all:
https://github.com/boutproject/BOUT-dev/actions/runs/15414194422

@ZedThree @bendudson  can we merge this?

ZedThree (approved):

PR #3092, opened 2025-03-15
WIP: Tidying FCI implementation

bendudson:
- Fix Div_par(f, v) for FCI

### DIFF
--- a/include/bout/field.hxx
+++ b/include/bout/field.hxx
@@ -527,6 +527,10 @@ T pow(BoutReal lhs, const T& rhs, const std::string& rgn = "RGN_ALL") {
  * and uses checkData() to, if CHECK >= 3, check
  * result for non-finite numbers
  *
+ * If the input field has parallel slices, then those will also be
+ * operated on. To avoid this, use `withoutParallelSlices()` to
+ * discard slices before calling.
+ *
  */
 #ifdef FIELD_FUNC
 #error This macro has already been defined
@@ -540,6 +544,20 @@ T pow(BoutReal lhs, const T& rhs, const std::string& rgn = "RGN_ALL") {
     /* Define and allocate the output result */                        \
     T result{emptyFrom(f)};                                            \
     BOUT_FOR(d, result.getRegion(rgn)) { result[d] = func(f[d]); }     \
+    if (f.hasParallelSlices()) {                                       \
+      /* Operate on parallel slices */                                 \
+      result.splitParallelSlicesAndAllocate();                         \
+      for (size_t i{0}; i != f.numberParallelSlices(); ++i) {          \
+        BOUT_FOR(d, result.getRegion(rgn)) {                           \
+          result.yup(i)[d] = func(f.yup(i)[d]);                        \
+        }                                                              \
+        checkData(result.yup(i));                                      \
+        BOUT_FOR(d, result.getRegion(rgn)) {                           \
+          result.ydown(i)[d] = func(f.ydown(i)[d]);                    \
+        }                                                              \
+        checkData(result.ydown(i));                                    \
+      }                                                                \
+    }                                                                  \
     result.name = std::string(#_name "(") + f.name + std::string(")"); \
     checkData(result);                                                 \
     return result;                                                     \
--- a/include/bout/field.hxx
+++ b/include/bout/field.hxx
@@ -535,32 +535,28 @@ T pow(BoutReal lhs, const T& rhs, const std::string& rgn = "RGN_ALL") {
 #ifdef FIELD_FUNC
 #error This macro has already been defined
 #else
-#define FIELD_FUNC(_name, func)                                        \
-  template <typename T, typename = bout::utils::EnableIfField<T>>      \
-  inline T _name(const T& f, const std::string& rgn = "RGN_ALL") {     \
-    AUTO_TRACE();                                                      \
-    /* Check if the input is allocated */                              \
-    checkData(f);                                                      \
-    /* Define and allocate the output result */                        \
-    T result{emptyFrom(f)};                                            \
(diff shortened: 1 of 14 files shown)


include/bout/field.hxx
```diff
@@ -527,22 +527,36 @@ T pow(BoutReal lhs, const T& rhs, const std::string& rgn = "RGN_ALL") {
...
+    /* Check if the input is allocated */                                                \
+    checkData(f);                                                                        \
+    /* Define and allocate the output result */                                          \
+    T result{emptyFrom(f)};                                                              \
+    BOUT_FOR(d, result.getRegion(rgn)) { result[d] = func(f[d]); }                       \
+    if (f.hasParallelSlices()) {                                                         \
+      /* Operate on parallel slices */                                                   \
+      result.splitParallelSlicesAndAllocate();                                           \
+      for (size_t i{0}; i != f.numberParallelSlices(); ++i) {                            \
+        BOUT_FOR(d, result.getRegion(rgn)) { result.yup(i)[d] = func(f.yup(i)[d]); }     \
```
dschwoerer, line 551: but here we need a shifted region (at least for FCI) ...

include/bout/field3d.hxx
```diff
@@ -237,13 +237,23 @@ public:
...
   void splitParallelSlicesAndAllocate();
 
   /*!
    * Clear the parallel slices, yup and ydown
    */
   void clearParallelSlices();
 
+  /// Returns a shallow copy without parallel slices
+  Field3D withoutParallelSlices() const {
+    Field3D result{getMesh(), getLocation(), getDirections(), getRegionID()};
```
dschwoerer, line 252: Would that not be more general:
```suggestion
    auto result{emptyFrom(*this)};
```

src/mesh/coordinates.cxx
```diff
@@ -1600,16 +1600,14 @@ Field3D Coordinates::Div_par(const Field3D& f, CELL_LOC outloc,
...
   for (int i = 0; i < f.getMesh()->ystart; ++i) {
-    f_B.yup(i) = f.yup(i) / coords->J.yup(i) * sqrt(coords->g_22.yup(i));
-    f_B.ydown(i) = f.ydown(i) / coords->J.ydown(i) * sqrt(coords->g_22.ydown(i));
+    f_B.yup(i) = f.yup(i) / Bxy_floc.yup(i);
+    f_B.ydown(i) = f.ydown(i) / Bxy_floc.ydown(i);
   }
-  return setName(coords->J / sqrt(coords->g_22) * Grad_par(f_B, outloc, method),
-                 "Div_par({:s})", f.name);
+  return setName(Bxy * Grad_par(f_B, outloc, method), "Div_par({:s})", f.name);
 }
```
dschwoerer, line 1611: But this about the metric coefficients, not about the strength about the metric field.
Only for Clebsch coordinates this is the same, but not for FCI.

src/mesh/difops.cxx
```diff
@@ -234,57 +234,50 @@ Field3D Div_par(const Field3D& f, const std::string& method, CELL_LOC outloc) {
...
+  auto dy = coord->dy;
 
-        result(i, j, k) =
-            (fluxRight - fluxLeft) / (coord->dy(i, j, k) * coord->J(i, j, k));
-      }
+  Field3D result{emptyFrom(f)};
+  BOUT_FOR(i, f.getRegion("RGN_NOBNDRY")) {
+    result[i] = B[i]
+                * ((f_up[i] * v_up[i] / B_up[i]) - (f_down[i] * v_down[i] / B_down[i]))
+                / (dy[i] * sqrt(g_22[i]));
```
dschwoerer, line 272: This is only valid in clebsch

src/mesh/difops.cxx
```diff
@@ -234,57 +234,50 @@ Field3D Div_par(const Field3D& f, const std::string& method, CELL_LOC outloc) {
...
+
+#if CHECK > 0
+    if (!std::isfinite(result[i])) {
+      output.write("{} {} {} {}\n", f_up[i], v_up[i], f_down[i], v_down[i]);
+      output.write("{} {} {} {} {}\n", B[i], B_up[i], B_down[i], dy[i], sqrt(g_22[i]));
```
dschwoerer, line 277: We probably want to move that to the exception message below, and also add `i` to make it easier to figure out why that is non-finite ...

dschwoerer (commented):
I think we should talk about this, I do not think we want to remove the part about the metric coefficients?

src/mesh/difops.cxx
bendudson: Ah no. This is always valid for all coordinate systems, not just Clebsch. It's just a vector identity from Div(B) = 0

Div(b v) = Div(B v / |B|) = Div(B) v/|B| + B dot Grad(v / |B|)

so the divergence of a parallel vector Div_par(v) is always B * Grad_par( v / B )

src/mesh/coordinates.cxx
bendudson: It works because the area of a flux tube is inversely proportional to B. If we're mapping along flux tubes from one plane to the next, then the ratio of the flux tube areas is (inverse) the ratio of B on the two planes.

src/mesh/difops.cxx
dschwoerer: Because the mesh is locally field aligned - I see :+1:

include/bout/field.hxx
bendudson: My thinking with this was that if we are operating over a given region of a field, then we also want to operate over all the yup/down fields connected to those points. Hence the region indices should be the same for yup/down. I guess you have a different model for how this should work?

dschwoerer:
Closing this, as the work on target branch has stopped.

PR #3094, opened 2025-03-18
Automatically select apropriate parallel slice

dschwoerer:


dschwoerer:
Adding a new type would require  a lot of new code. Do you have some benchmarks that suggest that this a significant overhead?

dschwoerer:
I just realised with this and `Field3DParallel`
```
void floor_all(Field3D &f, BoutReal val) {
  alloc_all(f);
  BOUT_FOR(i, f.getRegion("RGN_ALL")) {
    f[i] = floor(f,val);
    f.yup()[i] = floor(f,val);
    f.ydown()[i] = floor(f,val);
  }
}
```
can be simplified to
```
void floor(Field3DParallel &f, BoutReal val) {
  BOUT_FOR(i, f.getRegion("RGN_ALL")) {
    f[i] = floor(f,val);
  }
}
```
if `getRegion` contains also shifted indices.
That can then just be
```
template <class F>
void floor(F &f, BoutReal val) {
  BOUT_FOR(i, f.getRegion("RGN_ALL")) {
    f[i] = floor(f,val);
  }
}
```
which we probably have already :-)

tomc271:
Profiled with hyperfine: 
(Minimum time from 50 runs)

Unit tests in Release with CHECKS=0: 
|Test|Min time (s)|% of next|
|-|-|-|
|next|1.03|100%|
|yup-ydown-auto|1.05|102%|
|master|1.10|107%|

Unit tests in Debug with CHECKS=2: 
|Test|Min time (s)|% of next|
|-|-|-|
|next|5.40|100%|
|yup-ydown-auto|5.53|102%|
|master|5.68|105%|

test-drift-instability integration test in Release with CHECKS=0: 
|Test|Min time (s)|% of next|
|-|-|-|
|next|1.26|100%|
|yup-ydown-auto|1.36|108%|
|master|1.19|95%|

test-drift-instability integration test in Debug with CHECKS=2: 
|Test|Min time (s)|% of next|
|-|-|-|
|next|11.08|100%|
|yup-ydown-auto|12.88|116%|
|master|10.81|98%|

dschwoerer:
Amazing, thanks @tomc271! So it seems there is quite some overhead (8%) without checks. Looking at the absolute numbers for `CHECK=2`, I would argue that the 16% is perfectly fine.

I am slightly worried about the 8% - that is probably not great.
So probably worth to hold of the merging, and see whether we/I can come up with a better implementation. Templates could be used, then some of the checks could be done at compile time, but that requires more time to think about it ...

It would make the code however more readable, which also a benefit for non-FCI users. Maybe we could make make a compile time switch to enable parallel slices, so that users that do not use it, can keep it disabled. However, then we either cannot start using this feature and clean up the code, or we are kind of forced to stay with it, even if we cannot come up with a faster implementation ...

include/bout/field3d.hxx
```diff
@@ -351,8 +358,27 @@ public:
     return std::end(getRegion("RGN_ALL"));
   };
 
-  BoutReal& operator[](const Ind3D& d) { return data[d.ind]; }
-  const BoutReal& operator[](const Ind3D& d) const { return data[d.ind]; }
+  BoutReal& operator[](const Ind3D& d) {
```
daviddickinson, line 361: It has been a long time since I've looked at this code but could this be `return ynext(d.yoffset - yoffset)[d]`? One could then still include some of the consistency checking but it would move some of the logic into `CHECK` regions that disappear  in production runs?

include/bout/field3d.hxx
```diff
@@ -351,8 +358,27 @@ public:
...
+    if (d.yoffset != 0) {
+      if (yoffset == 0) {
+        if (hasParallelSlices()) {
+          return ynext(d.yoffset)[d];
+        }
+#if CHECK >= 2
+        if (isFci()) { // We probably should assert here that this is field aligned
```
daviddickinson, line 368: Could this be `if ( ! hasParallelSlices() && isFci())` and placed before the return line so the return doesn't need to be guarded?

include/bout/field3d.hxx
dschwoerer: `return ynext(d.yoffset - yoffset)[d]` would always call ynext - even if this already the right field, and thus would fail.

The issue is that both the main field, which contains parallel fields, as well as the parallel fields, are of type `Field3D`.

We could (together with PR #3149) only enable this for Field3DParallel. There might be subtle bugs, where it depends on the type (Field3DParallel/Field3D) on whether it uses the correct indexing. But we could add those with higher CHECK level, so that should be fine.

include/bout/field3d.hxx
dschwoerer: I think it would still need to be guarded - because for FA there are no parallel slices, and thus we need to use the return at the end of the function.

PR #3098, opened 2025-04-13
Add attributes to ADIOS2 output to "define" dimensions as names. We n…

pnorbert:
…eed this for xarray support that is designed for the NetCDF model, where dimensions are separately defined entities.


bendudson (approved):
Thanks @pnorbert !

PR #3116, opened 2025-06-04
Handle PETSc signature change for FormFunctionForColoring

tomc271:
Use a wrapper function to handle FormFunctionForColoring signature mismatch, as using reinterpret_cast no longer works. 

Expected signature is `PetscErrorCode (*)()`
but is `PetscErrorCode (*)(SNES, Vec, Vec, void*)` in newer PETSc versions (on main branch, but no releases yet). 

For the newer version the current solver instance must be passed as the 'cxt' argument,
so store 'this' instance in a global variable (before calling `MatFDColoringSetFunction`).

Have to do it this way because the function passed as the (second) argument to `MatFDColoringSetFunction` must be a free function, so cannot make the function wrapper an instance method and pass 'this' from there.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -523,10 +523,8 @@ if (BOUT_ENABLE_WARNINGS)
    )
 
  include(EnableCXXWarningIfSupport)
- # Note we explicitly turn off -Wcast-function-type as PETSc *requires*
- # we cast a function to the wrong type in MatFDColoringSetFunction
  target_enable_cxx_warning_if_supported(bout++
-   FLAGS -Wnull-dereference -Wno-cast-function-type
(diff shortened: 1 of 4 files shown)


ZedThree:
Thanks @tomc271, unfortunately this approach won't work -- although `MatFDColoringSetFunction` takes a function `PetscErrorCode (*f)(void)`, it's actually lying about the signature. `FormFunctionForColoring` already has the correct signature for when it's called, it's just that `MatFDColoringSetFunction` expects it to be _passed_ with the wrong signature, hence the requirement for `reinterpret_cast`. Yes this is disgusting and wrong, no we can't do much about it.

ZedThree:
Hmm, I did think we had tests on this, maybe they're not on by default

PR #3117, opened 2025-06-05
Fix cast for new petsc

dschwoerer:
Alternative to https://github.com/boutproject/BOUT-dev/pull/3116

### DIFF
--- a/src/solver/impls/imex-bdf2/imex-bdf2.cxx
+++ b/src/solver/impls/imex-bdf2/imex-bdf2.cxx
@@ -652,11 +652,11 @@ void IMEXBDF2::constructSNES(SNES* snesIn) {
       // Create data structure for SNESComputeJacobianDefaultColor
       MatFDColoringCreate(Jmf, iscoloring, &fdcoloring);
       // Set the function to difference
-#if PETSC_VERSION_LE(3, 22, 0) \
-    || (PETSC_VERSION_GE(3, 23, 0) && PETSC_VERSION_RELEASE == 0)
-#define FUNC_MAYBE_CAST(func) reinterpret_cast<MatFDColoringFn>(func)
-#else
+#if PETSC_VERSION_GE(3, 24, 0) \
+    || (PETSC_VERSION_GE(3, 23, 0) && PETSC_VERSION_RELEASE != 0)
 #define FUNC_MAYBE_CAST(func) func
+#else
+#define FUNC_MAYBE_CAST(func) reinterpret_cast<MatFDColoringFn>(func)
 #endif
       MatFDColoringSetFunction(fdcoloring, FUNC_MAYBE_CAST(FormFunctionForColoring),
                                this);
--- a/src/solver/impls/imex-bdf2/imex-bdf2.cxx
+++ b/src/solver/impls/imex-bdf2/imex-bdf2.cxx
@@ -651,15 +651,15 @@ void IMEXBDF2::constructSNES(SNES* snesIn) {
       // Create data structure for SNESComputeJacobianDefaultColor
       MatFDColoringCreate(Jmf, iscoloring, &fdcoloring);
       // Set the function to difference
-      MatFDColoringSetFunction(fdcoloring,
-#if PETSC_VERSION_GE(3, 24, 0) \
+#if PETSC_VERSION_LE(3, 22, 0) \
     || (PETSC_VERSION_GE(3, 23, 0) && PETSC_VERSION_RELEASE == 0)
-                               FormFunctionForColoring,
+#define FUNC_MAYBE_CAST(func) reinterpret_cast<MatFDColoringFn>(func)
 #else
-                               reinterpret_cast<PetscErrorCode (*)()>(
-                                   FormFunctionForColoring),
+#define FUNC_MAYBE_CAST(func) func
 #endif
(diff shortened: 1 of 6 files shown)


src/solver/impls/snes/snes.cxx
```diff
@@ -1291,8 +1291,16 @@ void SNESSolver::updateColoring() {
...
-      fdcoloring, reinterpret_cast<PetscErrorCode (*)()>(FormFunctionForColoring), this);
+  MatFDColoringSetFunction(fdcoloring,
+#if PETSC_VERSION_GE(3, 24, 0) \
+    || (PETSC_VERSION_GE(3, 23, 0) && PETSC_VERSION_RELEASE == 0)
+                           reinterpret_cast<PetscErrorCode (*)(void*, Vec, Vec, void*)>(
+                               FormFunctionForColoring),
+#else
+                           reinterpret_cast<PetscErrorCode (*)()>(
+                               FormFunctionForColoring),
+#endif
```
ZedThree: I'd really _very_ much prefer it if we could keep preprocessor guards out of expressions. It looks like PETSc 3.23 provides a `MatFDColoringFn` typedef, so instead, we could provide a backport of this in our petsc shim header:

```cpp
// Or whatever the correct version is
#if PETSC_VERSION_LE(3, 23, 0)
using MatFDColoringFn = PetscErrorCode (*)();
#endif
```

and then just use that here (and elsewhere):

```suggestion
                           reinterpret_cast<MatFDColoringFn>(
                               FormFunctionForColoring
                           ),
```

ZedThree:
Possibly the `petsc_interface.hxx` header needs including somewhere?

src/solver/impls/imex-bdf2/imex-bdf2.cxx
```diff
@@ -651,9 +652,15 @@ void IMEXBDF2::constructSNES(SNES* snesIn) {
...
-          this);
+#if PETSC_VERSION_GE(3, 24, 0) \
+    || (PETSC_VERSION_GE(3, 23, 0) && PETSC_VERSION_RELEASE == 0)
+#define FUNC_MAYBE_CAST(func) func
+#else
+#define FUNC_MAYBE_CAST(func) reinterpret_cast<MatFDColoringFn>(func)
+#endif
+      MatFDColoringSetFunction(fdcoloring, FUNC_MAYBE_CAST(FormFunctionForColoring),
+                               this);
+#undef FUNC_MAYBE_CAST
```
ZedThree: Do we need all this? Could we at least put it in the petsc interface header?

src/solver/impls/petsc/petsc.cxx
```diff
@@ -514,9 +514,7 @@ int PetscSolver::init() {
   CHKERRQ(ierr);
   ierr = ISColoringDestroy(&iscoloring);
   CHKERRQ(ierr);
-  ierr = MatFDColoringSetFunction(matfdcoloring,
-                                  reinterpret_cast<PetscErrorCode (*)()>(solver_f), this);
-  CHKERRQ(ierr);
+  throw BoutException("Coloring is not working");
```
ZedThree, line 517: What's the issue here? If we know this doesn't work, can we remove it entirely? Or at least error when we set the option?

src/solver/impls/imex-bdf2/imex-bdf2.cxx
dschwoerer: We could always call the cast, but that might hide issues, so I would like to avoid that.

I put it here, as it is only used here, but we can of course also move it to a header ...

src/solver/impls/petsc/petsc.cxx
dschwoerer: We probably can, but I thought I will do a follow up PR, to just remove the solver completely. Thus I did not want to try figuring out when this is used.

The issue is that the function signature does not match what is expected by PETSc (and as far as I can tell it newer matched) thus I assume it is broken, but have not tested. I would assume that it fails at runtime or produces garbage, but have not tested.

dschwoerer:
@ZedThree do you see the issue, why `MatFDColoringFn` is not found for the "new PETSc" (released) CI?

src/solver/impls/petsc/petsc.cxx
ZedThree: Ah yes, I see. It almost certainly produces nonsense, but I guess if it's only for the colouring it doesn't impact the accuracy very much?

Yes, fair enough, removal in separate PR is sensible :+1:

src/solver/impls/imex-bdf2/imex-bdf2.cxx
ZedThree: That's true. If it's in the header, we can use it in `snes.cxx` too. Maybe rename to `BOUT_FUNC_MAYBE_CAST`?

ZedThree:
I think @tomc271 was looking at the test failure too -- did you manage to get anywhere Tom?

ZedThree:
Ah yes, thanks @tomc271! The version checks are inconsistent. Ok, I have a fix, and one for that new error too.

ZedThree (approved):

ZedThree:
Failing test is just Fedora due to a broken package

PR #3134, opened 2025-06-24
CI: Switch to released fedora

dschwoerer:
Can we try fedora:latest instead?

You are right, testing of rawhide should probably be moved. In this case it is the switch to python3.14 as the system python. It will take some time to get all dependencies working with python3.14 and I do not have as much time as I used to.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -208,8 +208,8 @@ jobs:
       - uses: actions/checkout@v4
         with:
           submodules: true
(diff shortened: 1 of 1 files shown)


ZedThree (approved):

ZedThree:
Bah, now hypre tests are failing. It doesn't seem to be a new version, so I don't know what the issue is. Unrelated to this change at least. I'll try rerunning them and then merge

Issue #3127, opened 2025-06-17
Fedora build failing yet again


ZedThree:
This time:

```
Problem: conflicting requests
  - nothing provides python3.13dist(natsort) needed by python3-boutdata-0.2.1-8.fc42.noarch from rawhide
  - nothing provides python(abi) = 3.13 needed by python3-boutdata-0.2.1-8.fc42.noarch from rawhide
  - nothing provides python3.13dist(numpy) >= 1.22 needed by python3-boutdata-0.2.1-8.fc42.noarch from rawhide
  - nothing provides python3.13dist(matplotlib) >= 3.2.1 needed by python3-boutdata-0.2.1-8.fc42.noarch from rawhide
  - nothing provides python3.13dist(boututils) >= 0.1.9 needed by python3-boutdata-0.2.1-8.fc42.noarch from rawhide
  - nothing provides python3.13dist(scipy) >= 1.4.1 needed by python3-boutdata-0.2.1-8.fc42.noarch from rawhide
  - nothing provides python3.13dist(sympy) >= 1.5.1 needed by python3-boutdata-0.2.1-8.fc42.noarch from rawhide
```

We have a cron job running tests on `master` every week, and they fail most weeks due to some issue with the Fedora build, and almost always it's an issue upstream.

@dschwoerer Can we just move this build completely into the Fedora infrastructure? It's hard to see what benefit it gives us when it's so flaky. While it's nice to test on a different OS, I would much rather have it be something less volatile than rawhide.

PR #3017, opened 2024-10-30
Replace deprecated `boututils.file_import`

ZedThree:

PR #3135, opened 2025-06-24
CI: Switch to released fedora (master)

dschwoerer:
Resolves: #3127


ZedThree:
```
  File "/home/test/BOUT-dev/build/tests/integrated/test-drift-instability-staggered/runtest", line 10, in <module>
    from boututils.file_import import file_import
ModuleNotFoundError: No module named 'boututils.file_import'
```

Didn't we fix this already? Or was that only in `next`?

dschwoerer:
I think it was only in next: #3017

ZedThree:
So many different things are broken on `master`, I'll merge this assuming the Fedora build passes

PR #3082, opened 2025-03-05
Eliminate boundary equations to improve HYPRE solves

rfalgout:
This PR adds code to eliminate boundary equations to get better behavior in HYPRE solvers.

This is not fully tested yet and likely requires some reorganization and cleaning up.

PR #3136, opened 2025-06-24
Revert eliminate hypre boundary equations

ZedThree:
Reverts #3082 as it breaks the hypre tests.


bendudson (approved):
Thanks @ZedThree ! Sorry I was too hasty in merging.

PR #3138, opened 2025-06-25
CI: Update runners

dschwoerer:


ZedThree (approved):
On `next`, we even switched some of these to `latest`, but that'll get sorted when we (eventually) get a release

PR #3142, opened 2025-06-29
Handle absolute paths to options file

bendudson:
The command line switch -f specifies the path to the BOUT.inp file. Previously this was always concatenated with the data directory. Now treats paths with leading '/' as absolute paths.

With C++17 this can be replaced by `std::filesystem` paths.


bendudson:
Thanks @ZedThree! Now using `std::filesystem` :)

ZedThree (approved):
Nice, thanks @bendudson!

PR #3166, opened 2025-09-04
Better pypi package

dschwoerer:
This fixes the recent failure (e.g. https://github.com/boutproject/BOUT-dev/actions/runs/16992233636/job/48174226873)
and adds the description.

Preview can be found here:

https://test.pypi.org/project/boutpp-nightly/

### DIFF
--- a/tools/pylib/_boutpp_build/backend.py
+++ b/tools/pylib/_boutpp_build/backend.py
@@ -223,7 +223,7 @@ Version: {getversion()}
             for l in src:
                 f.write(f"{pre}{l}")
                 pre = "         "
-        f.write("\nDescription-Content-Type: text/markdown\n\n")
+        f.write("Description-Content-Type: text/markdown\n\n")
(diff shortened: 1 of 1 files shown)


dschwoerer:
Also, there have been over 900 commits since 5.1.0 - maybe it is time for 5.2.0?

dschwoerer:
> I think it would be worth getting an RSE to looking to see if we can replace this custom backend with [scikit-build-core](https://scikit-build-core.readthedocs.io/en/latest/). I'm certain we can't be doing anything that special that requires us to maintain a whole custom backend.

Probably a good idea, however I am not sure it is on the top of the priority list. I also wanted to look into this, but did not have enough time ...

> > Also, there have been over 900 commits since 5.1.0 - maybe it is time for 5.2.0?
> 
> Yes, I just need to find some time to actually do this. I want to merge the coordinates refactor and some other breaking changes for instance.

:+1:

PR #3189, opened 2025-10-29
CI: switch Fedora to a pre-built container

dschwoerer:
This currently uses https://github.com/dschwoerer/bout-container-base/tree/ci-fedora - (which should probably be moved under the boutproject).

Using this approach it takes < 30 sec to have all dependencies installed.
It is intended as a PoC, as other workflows could benefit even more from this.

Note, this switches back to rawhide, rather then a released fedora. This reverts #3134 - but as we are using the last successful build of the base image, we are not (directly) impacted, and just keep using a slightly older rawhide version.
I am happy to revert this change, if you are not happy with this, @ZedThree

### DIFF
--- a/.ci_fedora.sh
+++ b/.ci_fedora.sh
@@ -21,11 +21,10 @@ then
 	cmd="sudo docker"
     fi
     test . != ".$2" && mpi="$2" || mpi=openmpi
-    test . != ".$3" && version="$3" || version=rawhide
-    time $cmd pull registry.fedoraproject.org/fedora:$version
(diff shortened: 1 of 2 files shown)


dschwoerer:
But that would subvert the idea of testing a change before it lands?

Or are you suggesting to also run the BOUT++ tests in the container repo, so that we only deploy images that do not introduce regressions on the fedora side? That might actually be a nice idea.

What do you think of converting the adios2 and PETSc-unreleased run to a pre-build container?

PR #3194, opened 2025-11-04
Configure cleanup

dschwoerer:


ZedThree (approved):
Oof, didn't need all these hanging around, thanks @dschwoerer!

PR #3196, opened 2025-11-04
Add more tests for FCI operators

ZedThree:
- Add MMS tests for all the main FCI derivative operators
- A bunch of drive-by fixes for clang-tidy warnings


src/mesh/parallel/fci.cxx
```diff
@@ -299,19 +321,69 @@ Field3D FCIMap::integrate(Field3D& f) const {
...
       throw BoutException(
           "Incorrect parallel transform type '" + parallel_transform
           + "' used "
             "to generate metric components for FCITransform. Should be 'fci'.");
     }
   } // else: parallel_transform variable not found in grid input, indicates older input
-    //       file or grid from options so must rely on the user having ensured the type is
-    //       correct
+  //       file or grid from options so must rely on the user having ensured the type is
+  //       correct
```
dschwoerer, line 387:
```suggestion
  //         file or grid from options so must rely on the user having ensured the type is
  //         correct
```

src/mesh/interpolation/hermite_spline_xz.cxx
```diff
@@ -177,7 +177,19 @@ void XZHermiteSpline::calcWeights(const Field3D& delta_x, const Field3D& delta_z
...
+    // know the value in the boundaries, so _any_ interpolation in the
+    // last interior cell can't be done. Instead, we fudge the
+    // interpolation in the last cell to be at the extreme right-hand
+    // edge of the previous cell (that is, exactly on the last
+    // interior point). However, this doesn't work with only one
+    // interior point, because we have to do something similar to the
+    // _first_ cell, and these two fudges cancel out and we end up
+    // indexing into the boundary anyway.
+    // TODO(peter): Can we remove this if we apply (dirichlet?) BCs to
+    // the X derivatives? Note that we need at least _2_
```
dschwoerer, line 186: I do not think that applying dirichlet BCs will be in general a good idea.
I think in general this is not really an issue - anything that takes the case below should be `i_corn == xend` and `t_x = 0`. Other cases should be boundaries, and should not take this path.

src/mesh/interpolation/hermite_spline_xz.cxx
```diff
@@ -346,6 +358,8 @@ Field3D XZHermiteSpline::interpolate(const Field3D& f, const std::string& region
   ASSERT1(f.getMesh() == localmesh);
   Field3D f_interp{emptyFrom(f)};
 
+  // TODO(peter): Should we apply dirichlet BCs to derivatives?
+
```
dschwoerer: I do not think so. What would be the motivation? So we do not need two x-guard cells?

src/mesh/parallel/fci.cxx
```diff
@@ -37,79 +37,99 @@
...
+  const auto parallel_slice_field_name = [&](std::string_view field) -> std::string {
+    const auto direction = (offset_ > 0) ? "forward"sv : "backward"sv;
     // We only have a suffix for parallel slices beyond the first
     // This is for backwards compatibility
-    const std::string slice_suffix =
-        (std::abs(offset) > 1) ? "_" + std::to_string(std::abs(offset)) : "";
-    return direction + "_" + field + slice_suffix;
+    const auto slice_suffix =
+        (std::abs(offset_) > 1) ? fmt::format("_{}", std::abs(offset_)) : "";
+    return fmt::format("{}_{}{}", direction, field, slice_suffix);
```
dschwoerer: We could also do this with just one call to `fmt::format`
```suggestion
    if (std:abs(offset_) == 1){
       return fmt::format("{}_{}", direction, field);
     }
    return fmt::format("{}_{}_{}", direction, field, std::abs(offset_));
```

tests/MMS/spatial/fci/fci_mms.cxx
```diff
@@ -1,37 +1,61 @@
...
-  Field3D result{Grad_par(input)};
-  Field3D error{result - solution};
+  Field3D K{FieldFactory::get()->create3D("K", Options::getRoot(), mesh)};
+
+  // Communicate to calculate parallel transform.
+  if constexpr (bout::build::use_metric_3d) {
+    // Div_par operators require B parallel slices:
+    // Coordinates::geometry doesn't ensure this (yet)
+    auto& Bxy = mesh->getCoordinates()->Bxy;
+    mesh->communicate(Bxy);
```
dschwoerer, line 53: In general you would need to apply parallel BCs, but I guess the tested grids do not have this?

dschwoerer (dismissed):
Looks good!
Thanks @ZedThree

tests/MMS/spatial/fci/fci_mms.cxx
ZedThree: Correct, no BCs set because all the input fields are created from expressions

src/mesh/interpolation/hermite_spline_xz.cxx
ZedThree: Ah, that's true about the boundaries. The pain here is for grids where there's only one point in either x or z, so this fudge ends up cancelling out and breaking things. That's probably not super common, so maybe we just need `ASSERT4(mesh.Nx() >= 2)` or something

tests/MMS/spatial/fci/fci_mms.cxx
dschwoerer: I think B should be loaded also for the parallel slices, in which case we do not need to communicate at all ...

src/mesh/interpolation/hermite_spline_xz.cxx
dschwoerer: Yes, I think that is only an issue for tests, that try to have a minimal grid. I do not think we should to much effort in for them ...

tests/MMS/spatial/fci/fci_mms.cxx
ZedThree: It doesn't seem to be the case -- the `communicate` was necessary here to get any parallel slices on `Bxy`

dschwoerer:
Note that this is not https://en.wikipedia.org/wiki/Monotone_cubic_interpolation - but hermitespline - with the value cropped to the bound of the 4 surrounding values.

tests/MMS/spatial/fci/fci_mms.cxx
dschwoerer: Yes, these changes are not in next, only in f3dwy (PR #3197 )

dschwoerer:
Sorry, I failed to see that this was still waiting on my review.

PR #3200, opened 2025-11-07
Port some Hermes-3 finite volume operators

ZedThree:
Includes #3196 

Ports the following from [Hermes-3 `fci-more-fixes-update-more`](https://github.com/boutproject/hermes-3/tree/fci-more-fixes-update-more):

- `FV::Div_par_mod`
- `FV::Div_par_fvv`
- `Div_par_K_Grad_par_mod` (not strictly FV!)
- `FV::Superbee`

Plus adds MMS tests for these and the existing FV operators:

- `FV::Div_par`
- `FV::Div_par_K_Grad_par`

with all the available slope limiters, as appropriate.

----

Because the slope limiters kind of necessarily introduce inaccuracies, I've reduced the expected order of the operators, with `Upwind` having order 1 and the rest 1.5. This feels like a bit of a fudge, but it's not clear to me how to treat this better.

----

The ported operators will obviously conflict with the Hermes-3 ones. So 2/3 things to do:

- Remove the ported ones from Hermes-3 when updating the bundled BOUT++
- Move all the operators here into `bout::FV::` namespace
- Move all the operators there into `hermes::FV::` namespace


tests/MMS/spatial/finite-volume/fv_mms.cxx
```diff
@@ -42,6 +42,15 @@ int main(int argc, char** argv) {
   Field3D v{FieldFactory::get()->create3D("v", Options::getRoot(), mesh)};
 
   // Communicate to calculate parallel transform.
+  if constexpr (bout::build::use_metric_3d) {
+    // Div_par operators require B parallel slices:
+    // Coordinates::geometry doesn't ensure this (yet)
+    auto& Bxy = mesh->getCoordinates()->Bxy;
+    auto& J = mesh->getCoordinates()->J;
```
dschwoerer, line 50: Note that communicating J is likely wrong for any realistic grid. You need #3197 
For a simple slab test, that may well work, but not for real grids.

ZedThree: What's the issue? And what's required to fix it?

dschwoerer: The parallel slices can have a completely different geometry. Thus one needs J from the traced parallel slices. The communication interpolates J, which works for lab-based quantities, but not for mesh based quantities.

One solution is to also write out the parallel metric components in zoidberg, and read them in in BOUT++, and prevent them from being overwritten:
https://github.com/boutproject/zoidberg/blob/more-6/zoidberg/zoidberg.py#L421
https://github.com/boutproject/BOUT-dev/blob/f3dwy/src/mesh/parallel/fci.cxx#L134

See also:
https://bout.zulipchat.com/#narrow/channel/535664-BSTING/topic/Connection.20between.20J.20and.20B.20in.20FCI/with/554530388

mrhardman:
@ZedThree Is there a possibility that this PR could also cover the operators and MMS tests of operators featured https://github.com/boutproject/hermes-3/pull/285 and https://github.com/boutproject/hermes-3/pull/327? The latter PR also has model-level tests that could not be ported to BOUT-dev. I imagine that the functions in https://github.com/boutproject/hermes-3/blob/nonorthogonal-debug/tests/mms/perpendicular_laplacian.py could be factored into https://github.com/boutproject/boutdata/blob/master/src/boutdata/mms.py.

mrhardman:
> Yes, certainly possible, will investigate!

Thank you! I made a start on trying to make the existing tests in my latest PR https://github.com/boutproject/hermes-3/pull/327 more similar to `boutdata.mms` here https://github.com/boutproject/hermes-3/pull/327/commits/32c8536f98a144329b1fe56515fd3e3d7c8cc792. It likely better to view the files directly. I note that the fact that the operators use an optional argument for `metric` by convention tripped me up as I tried to propagate my `GeneralMetric` to all functions, as it was possible to mistakenly not pass a consistent metric to all differential operators.

dschwoerer (dismissed):
Looks good!

I did not notice that this PR was still pending, and rebased it on top of the recent changes, including #3320 and #3334.

Do you prefer resolving the conflicts here and merge this, or use #3335 instead?

bendudson:
Replaced by #3335

PR #3206, opened 2025-11-18
Add some asserts for non parallelised XZ interpolation

dschwoerer:


src/mesh/interpolation/bilinear_xz.cxx
```diff
@@ -31,6 +31,10 @@ XZBilinear::XZBilinear(int y_offset, Mesh* mesh)
     : XZInterpolation(y_offset, mesh), w0(localmesh), w1(localmesh), w2(localmesh),
       w3(localmesh) {
 
+  if (localmesh->getNXPE() > 1) {
+    throw BoutException("Do not support MPI splitting in X");
```
ZedThree: Let's make it a little more descriptive:
```suggestion
    throw BoutException("bilinearxz interpolation does not support MPI splitting in X");
```

Please also add this fact to the class docstring.

Same for the lagrange class

dschwoerer: Which doc string?
I could not find any documentation, either in the header nor in the cxx file. Did I miss something?
It feels a bit weird just having that as only bit of documentation.

ZedThree: On the `XZBilinear` class. Yes, there currently isn't a docstring, but having something is better than nothing :)

dschwoerer:
`test-cyclic` was stuck in CI. A restart solved that. Just in case this happens again, we maybe should investigate. It would not be great to have a solver get stuck in a production run ...

PR #3207, opened 2025-11-21
Support reading `Array/Matrix/Tensor<int>` in `OptionsAdios2`

ZedThree:
Although writing was supported, we didn't actually support reading these types. I've refactored the implementation to reduce a lot of duplication in calculating/setting the Adios2 starts/counts etc.

Also fixes a bug with reading/writing `FieldPerp`

I've also replaced the integrated test with a more comprehensive unit test. We still don't test in parallel, but this is better than nothing.

Some of this would be nicer with C++26's `template for`. Maybe we can use that in a few years.

### DIFF
--- a/src/sys/adios_object.cxx
+++ b/src/sys/adios_object.cxx
@@ -65,8 +65,8 @@ ADIOSStream& ADIOSStream::ADIOSGetStream(const std::string& fname, adios2::Mode 
   return it->second;
 }
 
-void ADIOSSetParameters(const std::string& input, char delimKeyValue,
-                        char delimItem, adios2::IO& io) {
+void ADIOSSetParameters(const std::string& input, char delimKeyValue, char delimItem,
+                        adios2::IO& io) {
   auto lf_Trim = [](std::string& input) {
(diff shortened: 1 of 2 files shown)


src/sys/options/options_adios.cxx
```diff
@@ -55,207 +177,183 @@ Options readVariable(adios2::Engine& reader, adios2::IO& io, const std::string&
...
+  }
+  if (type == adios2::GetType<float>()) {
+    return readVariable<float>(reader, io, name, type);
+  }
+  if (type == adios2::GetType<double>()) {
+    return readVariable<double>(reader, io, name, type);
+  }
+  if (type == adios2::GetType<long double>()) {
+    return readVariable<long double>(reader, io, name, type);
+  }
```
bendudson, line 298: Why no macro magic? This feels like there must be a better solution e.g. visitor pattern, but this will work.

ZedThree: Oh, I didn't really mean to leave this in, I was just searching for a neater non-macro solution, but didn't find one. Here's where `template for` would be really useful!

PR #3212, opened 2025-11-24
Remove `Mesh::load`

ZedThree:
`Mesh` should be entirely created in constructor

This is a little bit experimental -- the reason we have `Mesh::load` in the first place is to break the circular dependency between `Mesh` and `Coordinates`, and the reason we construct `Coordinates` in `BoutMesh::BoutMesh` is to ensure any communication required in `Coordinates` is properly synchronised. So we might just end up replacing calls to `Mesh::load` with calls to `Mesh::getCoordinates`, but at least the `Mesh` constructor will be atomic

### DIFF
--- a/src/mesh/impls/bout/boutmesh.cxx
+++ b/src/mesh/impls/bout/boutmesh.cxx
@@ -259,9 +259,6 @@ If you want the old setting, you have to specify mesh:symmetricGlobalY=false in 
     has_boundary_upper_y = static_cast<bool>(allbndry);
   }
 
-  // Initialize default coordinates
-  getCoordinates();
-
   output_info.write(_("\tdone\n"));
 }
 


ZedThree:
Ah, clang-tidy has spotted a bunch of virtual functions that will need qualifying with `BoutMesh::`

Yeah, and some tests fail, so this trivial change is definitely not enough. Shame

ZedThree:
> Perhaps we could remove the `Mesh*` argument to the `FieldFactory` constructor. It sets a default `Mesh*`, overridden by optional `Mesh*` arguments to `create2D` and `create3D` methods. Those arguments would now be required.

That could work, I think we'd also need to add a `Mesh*` argument to `Options::as` (at least for the field versions?)

EDIT: I have added the `static FieldFactory::setMesh(Mesh*)` here and it seems to work -- the tests all pass at least! I think this experiment mostly highlights where we have circular dependencies between classes

PR #3216, opened 2025-12-01
Remove deprecated `FieldGenerator::generate`

ZedThree:
The only use was in boundary conditions at `CELL_ZLOW`. It look a minute for me
to understand what was happening here and why we couldn't just use `Context()`
already -- we need to set a custom `x` OR `y` depending on where the boundary
is, although at first glance it looks like we're setting both, the trick is that
only one of `bndry->bx/by` is actually non-zero.

An alternative to this would be to change `Context(BoundaryRegion)` to use the
staggered location like a bitfield and pass e.g. `CELL_XLOW | CELL_ZLOW`. This
would avoid some duplicated work.

The only remaining use of a deprecated function is now `Mesh::firstY()/lastY()`,
again in the boundaries, although `BoutMesh` also uses it without passing
`xpos`.

I suppose this also serves as our annual reminder that the boundaries need a
complete refactor.

### DIFF
--- a/include/bout/sys/expressionparser.hxx
+++ b/include/bout/sys/expressionparser.hxx
@@ -64,12 +64,6 @@ public:
     return nullptr;
   }
 
-  [[deprecated("This will be removed in a future version. Implementations should "
-               "override the Context version of this function.")]] virtual double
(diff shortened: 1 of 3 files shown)


dschwoerer (approved):
Thanks for the cleanup!

PR #3222, opened 2025-12-02
Add ability to use variables from grid file in input file expressions

ZedThree:
This adds the ability to read fields (2D and 3D) and doubles\* from the grid file (`mesh:file`) and use them in expressions in the input file, for example to use coordinates directly from the grid generator:

```ini
[input:grid_variables]
rho = field2d
theta = field2d
scale = boutreal

[mesh]
B = (scale / rho) * cos(theta)
```

_\* ints can also be read, but the expression parser only handles doubles_

---

This could also be used for something like region labels to the input file by having hypnotoad make a field which is 1.0 in that region and 0.0 elsewhere:

```python
region_pfr = 1.0 if is_in_pfr else 0.0
```

and using like

```ini
[input:grid_variables]
region_pfr = field2d

[some_source]
function = region_pfr * cos(x) * whatever
```

We could probably also do this in BOUT++ directly, making fields for each boundary region.


dschwoerer (dismissed):
Looks good, thanks for cleaning up 👍

dschwoerer (dismissed):

dschwoerer (dismissed):

bendudson (dismissed):
Thanks @ZedThree ! This looks really useful. I think this is good as-is, but for future if this becomes widely used:
1. These expressions are in the global scope rather than being in the `mesh` namespace.
2. A nice-to-have might be renaming imports analogous to "import x as y" to avoid name conflicts.
These may over-complicate things, so I think this simplest implementation is good.

bendudson:
Hey @ZedThree ! This functionality is really useful :) I had to make the loading lazy to avoid issues where a `FieldFactory` is created during `BoutMesh::load` and before `mesh->get` works.

bendudson:
Sorry, I force-pushed to remove a bad merge with `next`. It introduced all kinds of exciting memory errors.

bendudson:
Probably the easiest way to namespace variables is to append something like "mesh:" (or "input:"?) to the start of the first argument to `addGenerator` here: https://github.com/boutproject/BOUT-dev/blob/field-variable-expression/src/field/field_factory.cxx#L97
i.e.
```
factory.addGenerator(std::string("mesh:" + name, std::make_shared<GridVariable<T>>(&mesh, name));
```
The generator lookup is a simple dictionary key search that happens before any splitting into section names: They only have one namespace.

An alternative is more complex but might simplify code elsewhere: To store fields, or lazily loaded fields, in the `Options` tree and allow them to be used in expressions. Currently only strings can be resolved by `findOption` (https://github.com/boutproject/BOUT-dev/blob/field-variable-expression/src/field/field_factory.cxx#L452) but perhaps that could be extended. Seems like a bigger refactor with more thinking needed.

mikekryjak:
I would support accessing the variables by `region::midplane` for example. People may have custom variables in the input file that could clash with the names, and having a prefix makes it easier to identify these variables as well.

ZedThree:
I've rebased onto `next` and namespaced grid variables under `gridvar:` -- I figure a new section name like that should prevent accidental collisions with existing variables

bendudson (approved):
Thanks @ZedThree !

PR #3223, opened 2025-12-03
Split out `XZPetscHermiteSpline`

ZedThree:
Includes #3200 for testing

- Split out `XZPetscHermiteSpline` as a separate implementation, this allows us to remove all the `#ifdef` mess
- Move all the `XZInterpolations` into the more usual `impls/` directory layout

### DIFF
--- a/src/mesh/interpolation/xz/impls/hermite_spline_xz.hxx
+++ b/src/mesh/interpolation/xz/impls/hermite_spline_xz.hxx
@@ -50,8 +50,8 @@ protected:
   //
   // Protected so this can be reused between `XZHermiteSpline` and `XZMonotonicHermiteSpline`
   auto interpolate_point(const Field3D& f, const Field3D& fx, const Field3D& fz,
-                         const Field3D& fxz, const Ind3D& i) const
-      -> InterpolatePointResult {
+                         const Field3D& fxz,
+                         const Ind3D& i) const -> InterpolatePointResult {
     const auto ic = i_corner[i];
     const auto iczp = ic.zp();
     const auto icxp = ic.xp();


src/mesh/interpolation/xz/impls/petsc_hermite_spline_xz.cxx
```diff
@@ -322,31 +321,30 @@ void XZHermiteSpline::calcWeights(const Field3D& delta_x, const Field3D& delta_z
...
-          {i, j + yoffset, k_mod, h00_z(i, j, k) - 0.5 * h11_z(i, j, k)},
-          {i, j + yoffset, k_mod_p1, h01_z(i, j, k) + 0.5 * h10_z(i, j, k)},
+          {i, j + yoffset, k_mod, h00_z(i, j, k) - (0.5 * h11_z(i, j, k))},
+          {i, j + yoffset, k_mod_p1, h01_z(i, j, k) + (0.5 * h10_z(i, j, k))},
           {i, j + yoffset, k_mod_p2, 0.5 * h11_z(i, j, k)}};
 }
 
-Field3D XZHermiteSpline::interpolate(const Field3D& f, const std::string& region) const {
+Field3D XZPetscHermiteSpline::interpolate(const Field3D& f,
+                                          const std::string& region) const {
```
ZedThree, line 338: @dschwoerer `region` isn't actually used here and I don't really understand why not. This technically makes this version behave differently to the other implementations. Should they actually all use `RGN_NOY/YPAR_<N>`?

It doesn't look like we ever actually call `interpolate` with a region anywhere, so perhaps it would be better to remove the argument and fix all the implementations?

dschwoerer: > @dschwoerer `region` isn't actually used here and I don't really understand why not.

Because I could not understand why any other region should ever be used.

> This technically makes this version behave differently to the other implementations. Should they actually all use `RGN_NOY/YPAR_<N>`?
> 
> It doesn't look like we ever actually call `interpolate` with a region anywhere, so perhaps it would be better to remove the argument and fix all the implementations?

Yes, removing the argument is probably best.

dschwoerer:
I unified all the HermiteSpline interpolations, because they have a fair bit in common, it is much easier to keep the different versions aligned.

Maybe we should move some of the `#ifdef` to `if constexpr` + templates?

src/mesh/interpolation/xz/impls/petsc_hermite_spline_xz.cxx
ZedThree: I think the intention was we might be able to use the regions for things like embedded boundaries, but at the minute that isn't possible out of the box without changes to parallel transforms and `communicate`, so probably best to just remove the argument for now.

ZedThree:
Ok, I can make `XZHermiteSpline` a base class of `XZPetscHermiteSpline` too to avoid the duplication of the basis calculations. We could also add `XZPetscMonotonicHermiteSpline` by simply doing the interpolation and clipping afterwards

PR #3224, opened 2025-12-04
Forward arguments to `BoutException` and `Output::write`

ZedThree:
This enables passing views like `fmt::join` which otherwise fail with:
    
```
error: static assertion failed: passing views as lvalues is disallowed
```

Very simple fix: change the argument types to `Args&&...` (universal forwarding
reference) and call `std::forward`.

Plus some drive-by clang-tidy fixes.

### DIFF
--- a/include/bout/output.hxx
+++ b/include/bout/output.hxx
@@ -160,13 +160,13 @@ class ConditionalOutput : public Output {
 public:
   /// @param[in] base    The Output object which will be written to if enabled
   /// @param[in] enabled Should this be enabled by default?
-  ConditionalOutput(Output* base, bool enabled = true) : base(base), enabled(enabled) {};
+  ConditionalOutput(Output* base, bool enabled = true) : base(base), enabled(enabled){};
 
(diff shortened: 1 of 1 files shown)


src/sys/boutexception.cxx
```diff
@@ -82,27 +92,21 @@ std::string BoutException::getBacktrace() const {
...
     }
   }
 #else
   backtrace_message = "Stacktrace not enabled.\n";
 #endif
 
-  return backtrace_message + msg_stack.getDump() + "\n" + header + message + "\n";
-}
-
-void BoutException::makeBacktrace() {
```
dschwoerer, line 103: This is a breaking ABI change. Should we just keep it as a no-op?

ZedThree: No, I don't think that's necessary, I'm not sure it ever really made sense to make this part of the public interface as we always called it in the constructor anyway.

Also, next release will have more meaningful breaking changes :)

PR #3225, opened 2025-12-05
Use `cpptrace` for prettier backtraces from exceptions

ZedThree:
We have historically been reluctant to add too many external dependencies, but with CMake and spack making it easier to pull them in and build them, I thought I'd try this library for generating stack traces.

Two useful things:
1) it's very portable out of the box, so we don't need to rely on external utilities like `addr2line` and instead can always enable backtraces
2) we can print snippets:

```
#3  in checkData(double)
    at /home/peter/Codes/BOUT++/BOUT-dev/include/bout/utils.hxx:493:65
     491: inline void checkData(BoutReal f) {
     492:   if (!std::isfinite(f)) {
   > 493:     throw BoutException("BoutReal: Operation on non-finite data");
                                                                          ^
     494:   }
     495: }
#4  in operator*(Field3D const&, double)
    at /home/peter/Codes/BOUT++/BOUT-dev/src/field/generated_fieldops.cxx:480:12
     478:   Field3D result{emptyFrom(lhs)};
     479:   checkData(lhs);
   > 480:   checkData(rhs);
                     ^
     481: 
     482:   result.setRegion(lhs.getRegionID());
#5  in Blob2D::rhs(double)
    at /home/peter/Codes/BOUT++/BOUT-dev/examples/blob2d/blob2d.cxx:177:41
     176:     ddt(n) = -bracket(phi, n, BRACKET_ARAKAWA) // ExB term
   > 177:              + 2 * DDZ(n) * (rho_s / R_c)      // Curvature term
                                                  ^
     178:              + D_n * Delp2(n);                 // Diffusion term
     179:     if (compressible) {
```

and can even enable colour:

image

It also has its own signal handler we could perhaps use instead of ours (though I've not tested this out).

Lastly, this is maybe a complete enough solution that we could also remove `MsgStack/TRACE/AUTO_TRACE`?

### DIFF
--- a/bin/bout-add-mod-path
+++ b/bin/bout-add-mod-path
@@ -104,8 +104,7 @@ def create_mod(modulepath, name, top, build):
     else:
         prereq = ""
     with open(filename, "w") as f:
-        f.write(
-            f"""#%Module 1.0
+        f.write(f"""#%Module 1.0
 #
 #  BOUT++ module for use with 'environment-modules' package
 #  Created by bout-add-mod-path v0.9
@@ -119,17 +118,14 @@ setenv        BOUT_TOP         {top}
 prepend-path  PATH             {top}/bin
 prepend-path  PYTHONPATH       {top}/tools/pylib
 prepend-path  IDL_PATH        +{top}/tools/idllib:'<IDL_DEFAULT>'
-"""
-        )
+""")
         if build != top:
-            f.write(
-                f"""#%Module 1.0
+            f.write(f"""#%Module 1.0
 setenv        BOUT_BUILD       {build}
 prepend-path  PATH             {build}/bin
 prepend-path  LD_LIBRARY_PATH  {build}/lib
 prepend-path  PYTHONPATH       {build}/tools/pylib
-"""
-            )
+""")
     print(f"created `{filename}`")
 
 
--- a/bin/bout-pylib-cmd-to-bin
+++ b/bin/bout-pylib-cmd-to-bin
@@ -126,8 +126,7 @@ def createwrapper(mod, func_name, func, name):
             out += end
             f.write(out)
 
-        fprint(
-            """#!/usr/bin/env python3
+        fprint("""#!/usr/bin/env python3
 # PYTHON_ARGCOMPLETE_OK
 
 import argparse
@@ -136,8 +135,7 @@ try:
 except ImportError:
     argcomplete=None
 
-"""
-        )
+""")
         doc = True
         para = False
         ret = False
@@ -183,19 +181,16 @@ except ImportError:
                         arg_help[curarg].append(esc(blas))
         # Print functions that are needed
         if "str_to_slice" in arg_type.values():
-            fprint(
-                """
+            fprint("""
 def str_to_slice(sstr):
     args=[]
     for s in sstr.split(','):
         args.append(int(s))
     print(args)
     return slice(*args)
-"""
-            )
+""")
         if "str_to_bool" in arg_type.values():
-            fprint(
-                """
+            fprint("""
 def str_to_bool(sstr):
     low=sstr.lower()
     # no or false
@@ -206,8 +201,7 @@ def str_to_bool(sstr):
         return True
     else:
         raise ArgumentTypeError("Cannot parse %s to bool type"%sstr)
-"""
-            )
+""")
         # Create the parser
         fprint("parser = argparse.ArgumentParser(%s)" % (esc(docs)))
         spec = inspect.signature(func)
@@ -247,24 +241,19 @@ def str_to_bool(sstr):
                     pre = "\n    "
             fprint(")")
 
-        fprint(
-            """
+        fprint("""
 if argcomplete:
     argcomplete.autocomplete(parser)
 
-# late import for faster auto-complete"""
-        )
+# late import for faster auto-complete""")
         fprint("from %s import %s" % (mod, func_name))
-        fprint(
-            """
+        fprint("""
 args = parser.parse_args()
 
 # Call the function %s, using command line arguments
 %s(**args.__dict__)
 
-"""
-            % (func_name, func_name)
-        )
+""" % (func_name, func_name))
     # alternative, but I think 0o755 is easier to read
     # import stat
     # os.chmod(filename,stat.S_IRWXU|stat.S_IRGRP|stat.S_IXGRP|stat.S_IROTH|stat.S_IXOTH)
--- a/bin/bout-v5-factory-upgrader.py
+++ b/bin/bout-v5-factory-upgrader.py
@@ -5,7 +5,6 @@ import copy
 import difflib
 import re
 
-
 # Dictionary of factory methods that may need updating
 factories = {
     "Interpolation": {
@@ -62,9 +61,7 @@ def find_factory_calls(factory, source):
         \s*=\s*
         {factory_name}::
         .*{create_method}.*
-        """.format(
-            **factory
-        ),
+        """.format(**factory),
         source,
         re.VERBOSE,
     )
@@ -75,9 +72,7 @@ def find_type_pointers(factory, source):
         r"""
         \b{type_name}\s*\*\s*   # Type name and pointer
         ([\w_]+)\s*;            # Variable name
-        """.format(
-            **factory
-        ),
+        """.format(**factory),
         source,
         re.VERBOSE,
     )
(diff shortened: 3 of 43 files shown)


ZedThree:
Lots of CMake pain here

ZedThree:
I've removed `AUTO_TRACE` completely, and `TRACE` where it was just essentially repeating the function name (or something like `intialising PETSc solver` in `PetscSolver::init`).

That leaves uses like `TRACE("PetscMatrix setting values at ({}, {})", petscRow, petscCol)` which include function arguments or local variables that the stacktrace won't capture. I think they likely have some value for debugging, but quite limited?

Same goes for uses like `TRACE("Creating Outer SOL communicators for Single Null operation")` in `BoutMesh`. I've left them in for now, but I'm not sure how much more information they offer over the exception message + stacktrace.

Here's an example with `CMAKE_BUILD_TYPE=RelWithDebInfo`:

```
Error encountered: Stack trace (most recent call first):
#0 (filtered)
#1 (filtered)
#2 (filtered)
#3 in BoutMesh::createCommunicators()
   at /home/peter/Codes/BOUT++/BOUT-dev/src/mesh/impls/bout/boutmesh.cxx:709:64
     707:       }
     708:       // Unconditional exception for demo purposes
   > 709:       throw BoutException("Single null outer SOL not correct\n");      
                                                                         ^
     710:       MPI_Group_free(&group);
     711:     }
#4 in BoutMesh::load()
   at /home/peter/Codes/BOUT++/BOUT-dev/src/mesh/impls/bout/boutmesh.cxx:575:22
     573:   /// Communicator
     574: 
   > 575:   createCommunicators();
                               ^
     576:   output_debug << "Got communicators" << endl;
#5 in BoutInitialise(int&, char**&)
   at /home/peter/Codes/BOUT++/BOUT-dev/src/bout++.cxx:201:28
     199:   bout::globals::mesh = Mesh::create();
     200:   // Load from sources. Required for Field initialisation
   > 201:   bout::globals::mesh->load();
                                     ^
     202: 
     203:   // time_report options are used in BoutFinalise, i.e. after we
#6 in main
   at /home/peter/Codes/BOUT++/BOUT-dev/examples/elm-pb/elm_pb.cxx:2161:1
    2159: };
    2160: 
  > 2161: BOUTMAIN(ELMpb);
          ^
#7 (filtered)
#8 (filtered)
#9 (filtered)

Single null outer SOL not correct


Additional information:
====== Back trace ======
 -> Creating Outer SOL communicators for Single Null operation on line 687 of '/home/peter/Codes/BOUT++/BOUT-dev/src/mesh/impls/bout/boutmesh.cxx'
```

And here's what it looks like with `CMAKE_BUILD_TYPE=Release` (no debug info):

```
Error encountered: Stack trace (most recent call first):
#0 (filtered)
#1 (filtered)
#2 (filtered)
#3 in BoutMesh::createCommunicators() [clone .cold]
   at /home/peter/Codes/BOUT++/BOUT-dev/build_release/lib/libbout++.so.5.2.0
#4 in BoutMesh::load()
   at /home/peter/Codes/BOUT++/BOUT-dev/build_release/lib/libbout++.so.5.2.0
#5 in BoutInitialise(int&, char**&)
   at /home/peter/Codes/BOUT++/BOUT-dev/build_release/lib/libbout++.so.5.2.0
#6 in main
   at /home/peter/Codes/BOUT++/BOUT-dev/examples/elm-pb/build/elm_pb
#7 (filtered)
#8 (filtered)
#9 (filtered)

Single null outer SOL not correct


Additional information:
====== Back trace ======
 -> Creating Outer SOL communicators for Single Null operation on line 687 of '/home/peter/Codes/BOUT++/BOUT-dev/src/mesh/impls/bout/boutmesh.cxx'
```

Clearly the `TRACE` call is more useful when we don't have debug info, but if we recommend always building with it, it's probably not that useful?

@bendudson Thoughts on completely removing `MsgStack`? Maybe in a separate PR so we can more easily revert it if we change our minds?

---

CUDA tests are failing due to #3233

manual/sphinx/developer_docs/debugging.rst
```diff
@@ -37,11 +37,10 @@ add a ``-v -v`` flag to see ``output_debug`` messages.
...
 
-The second utility BOUT++ has to help debugging is the message stack
-using the `TRACE` (and related `AUTO_TRACE`) macro. These are very
-useful for when a bug only occurs after a long time of running, and/or
-only occasionally. The ``TRACE`` macro can simply be dropped in
-anywhere in the code::
+The second utility BOUT++ has to help debugging is the message stack using the
+`TRACE` macro. This is very useful for when a bug only occurs after a long time
+of running, and/or only occasionally. The ``TRACE`` macro can simply be dropped
+in anywhere in the code::
```
dschwoerer: Should we not recommend `-g`?
Even with `TRACE` still around, most times `-g` is more helpful, imho.

manual/sphinx/developer_docs/debugging.rst
```diff
@@ -95,52 +95,61 @@ the ``fmt`` syntax also used by the loggers::
...
-(``-DBOUT_ENABLE_BACKTRACE=ON``, the default, requires the
-``addr2line`` program to be installed), and debug symbols to be turned
-on (``-DCMAKE_BUILD_TYPE=Debug`` or ``=RelWithDebInfo``) in BOUT++
-_and_ the physics model. If debug symbols are only present in part, the
-backtrace will be missing names for the other part.
-
-The output looks something like this:
+Lastly, BOUT++ can also automatically print a backtrace in the event of a
+crash. This is very useful to include if you ever need to report a bug to the
+developers! The output looks something like this:
```
dschwoerer: Maybe move this up, as I think this is the most helpful information in general?

src/sys/boutexception.cxx
```diff
@@ -51,62 +43,36 @@ BoutException::~BoutException() {
...
-    if (dladdr(ptr, &info) != 0) {
-      // Additionally, check whether this is the default offset for an executable
-      if (info.dli_fbase != reinterpret_cast<void*>(0x400000)) {
-        ptr = reinterpret_cast<void*>(reinterpret_cast<size_t>(ptr)
-                                      - reinterpret_cast<size_t>(info.dli_fbase));
-      }
-    }
+  const auto colours = isatty(stdout_fileno) || isatty(stderr_fileno)
+                           ? formatter::color_mode::always
+                           : formatter::color_mode::none;
```
dschwoerer, line 53: might be nice to have an option to overwrite the default?
So I can force color mode, even if I write to a file, so that I can later on `cat` it and get colors?
Either by ENV var or CLI option?

src/sys/boutexception.cxx
```diff
@@ -51,62 +43,36 @@ BoutException::~BoutException() {
...
-    // If we are compiled as PIE, need to get base pointer of .so and substract
-    Dl_info info;
-    const void* ptr = trace[i];
-    if (dladdr(ptr, &info) != 0) {
-      // Additionally, check whether this is the default offset for an executable
-      if (info.dli_fbase != reinterpret_cast<void*>(0x400000)) {
-        ptr = reinterpret_cast<void*>(reinterpret_cast<size_t>(ptr)
-                                      - reinterpret_cast<size_t>(info.dli_fbase));
-      }
-    }
```
dschwoerer, line 85: I am quite happy this madness goes :-)

src/sys/boutexception.cxx
ZedThree: Yes, that's a good idea. I was thinking of separately improving the colour output -- we have `BOUT_USE_COLOR` compile time option, and the `--color` CLI option, which use the `bout-log-color` script, but we could replace these with [`fmt`'s colour API](https://fmt.dev/12.0/api/#terminal-colors-and-text-styles).

I'll leave all that for a separate PR though

manual/sphinx/developer_docs/debugging.rst
ZedThree: Yes, that's a good point. `TRACE` is now very much for additional info, so I'll try and make sure it's clear that we recommend always using either `Debug` or `RelWithDebInfo` `CMAKE_BUILD_TYPE`, and moving the backtrace section up (your next comment)

src/sys/boutexception.cxx
ZedThree: `reinterpret_cast<void*>(0x400000)` this sends shivers up my spine everytime I see it!

ZedThree:
@dschwoerer Do you have any opinion on completely removing `TRACE` and `MsgStack`? Do you find it useful enough to keep?

ZedThree:
> cpptrace seems not that tricky to build, but I am wondering whether we should allow to disable it? Would it only require to #ifdef part of one file? For example if you do not have internet while building? Should we try to bundle it into our release tars?

I think it's probably simpler if we just always require it. Setting `BOUT_DOWNLOAD_CPPTRACE=off` lets you supply a pre-built version.

Issue #3242, opened 2026-01-08
Parallelising in `z`


ZedThree:
To support parallelisation in `z`, we need to do a few things:

- [ ] check loops over `z` that currently go over `[0, LocalNz)` should be switched to `[zstart, zend]` instead
- [ ] check ranges over `z` that currently go over `[0, LocalNz)` should be switched to `[zstart, zend]` instead
- [ ] convert remaining loops over `x, y, z` to use `BOUT_FOR` where possible
- [ ] guard anything using FFTs on `NZPE==1`
- [ ] switch from using modulo for `z` periodicity to guard cells
  - [ ] add communication in `z` (even for serial-in-z)
  - [ ] default `MZG`
  - [ ] performance tests!
- [ ] domain decomposition in `z`

PR #3244, opened 2026-01-08
Use `zstart/zend` in more loops

ZedThree:
In support of #3242

It's not always obvious when loops that aren't simple triple-loops over the full
domain should use `[zstart, zend]`, but I think all the time and inversion
solvers want to exclude boundaries. Probably derivatives too.

### DIFF
--- a/examples/dalf3/dalf3.cxx
+++ b/examples/dalf3/dalf3.cxx
@@ -71,8 +71,6 @@ private:
 
   BoutReal viscosity, hyper_viscosity;
 
-  bool smooth_separatrix;
-
   FieldGroup comms;
 
   std::unique_ptr<Laplacian> phiSolver{nullptr};  // Laplacian solver in X-Z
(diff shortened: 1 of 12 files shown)


manual/sphinx/developer_docs/data_types.rst
```diff
@@ -255,9 +255,9 @@ parallelise and vectorise. Some tuning of this is possible, see below
 for details. It replaces the C-style triple-nested loop::
 
    Field3D f(0.0);
-   for (int i = mesh->xstart; i < mesh->xend; ++i) {
-     for (int j = mesh->ystart; j < mesh->yend; ++j) {
-       for (int k = 0; k < mesh->LocalNz; ++k) {
+   for (int i = mesh->xstart; i <= mesh->xend; ++i) {
+     for (int j = mesh->ystart; j <= mesh->yend; ++j) {
+       for (int k = mesh->zstart; k <= mesh->zend; ++k) {
```
dschwoerer, line 260: Upsies. Good catch!

I wish we could fix it, so the old code is correct, but I am not sure how to, without breaking everything :sob:

dschwoerer (dismissed):
Thanks :+1:

manual/sphinx/developer_docs/data_types.rst
ZedThree: Luckily this was just the docs! I've not seen any real bugs :)

PR #3245, opened 2026-01-09
Add some `Mesh` methods for handling Z parallelisation 

ZedThree:
Supports #3242 

Note these don't actually _add_ Z parallelisation, just let us handle it correctly in various places

- `Mesh::GlobalZ` for global Z index in `[0, 1]`
- `Mesh::getNZPE/getZProcIndex` for Z processor info
- `bout::fft::checkZSerial` to guard routines using FFT

These also make a lot of code more consistent between the directions

### DIFF
--- a/src/mesh/parallel/shiftedmetricinterp.cxx
+++ b/src/mesh/parallel/shiftedmetricinterp.cxx
@@ -30,7 +30,10 @@
 #include <algorithm>
 
 #include "shiftedmetricinterp.hxx"
+
+#include "bout/boutexception.hxx"
 #include "bout/constants.hxx"
+#include "bout/field3d.hxx"
 #include "bout/parallel_boundary_region.hxx"
 
 ShiftedMetricInterp::ShiftedMetricInterp(Mesh& mesh, CELL_LOC location_in,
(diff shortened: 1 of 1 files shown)


src/invert/laplace/impls/serial_tri/serial_tri.cxx
```diff
@@ -33,14 +33,15 @@
...
-#include <cmath>
-
 #include <bout/output.hxx>
+#include <bout/utils.hxx>
 
 LaplaceSerialTri::LaplaceSerialTri(Options* opt, CELL_LOC loc, Mesh* mesh_in,
                                    Solver* UNUSED(solver))
     : Laplacian(opt, loc, mesh_in), A(0.0), C(1.0), D(1.0) {
+
+  bout::fft::checkZSerial(*localmesh, "`tri` inversion");
```
dschwoerer: Would it be clearer to name this `assertZSerial()`?

src/mesh/impls/bout/boutmesh.cxx
```diff
@@ -3224,6 +3231,26 @@ BoutReal BoutMesh::GlobalY(BoutReal jy) const {
...
+  return static_cast<BoutReal>(getGlobalZIndex(jz)) / static_cast<BoutReal>(MZ);
+}
+
+BoutReal BoutMesh::GlobalZ(BoutReal jz) const {
+
+  // Get global Z index as a BoutReal
+  const BoutReal zglo = getGlobalZIndex(jz);
+
+  if (symmetricGlobalZ) {
+    // With this definition the boundary sits dx/2 away form the first/last inner points
```
dschwoerer:
```suggestion
    // With this definition the boundary sits dz/2 away form the first/last inner points
```

src/mesh/index_derivs.cxx
```diff
@@ -497,6 +498,7 @@ class FFT2ndDerivativeType {
     ASSERT2(bout::utils::is_Field3D_v<T>); // Should never need to call this with Field2D
```
dschwoerer, line 494: Isn't that known at compile time?
So there is no runtime overhead, to have this check always included?

(Same for several of the above)

src/mesh/parallel/shiftedmetricinterp.cxx
```diff
@@ -114,11 +116,10 @@ ShiftedMetricInterp::ShiftedMetricInterp(Mesh& mesh, CELL_LOC location_in,
 
   interp_from_aligned->calcWeights(zt_prime_from);
 
-  int yvalid = mesh.LocalNy - 2 * mesh.ystart;
   // avoid overflow - no stencil need more than 5 points
```
dschwoerer:
```suggestion
  // avoid overflow - no stencil needs more than 5 points
```

dschwoerer (approved):

src/mesh/index_derivs.cxx
ZedThree: Yes, but checking at compile implies that we need to not build some of these for certain types, which might complicate the various macros. At any rate, it's something for a different PR!

PR #3252, opened 2026-01-16
SNES and PETSc solvers call MPI_Abort on exceptions

bendudson:
Catch exceptions and call MPI_Abort rather than trying to continue.

There is no way to recover and synchronise processors unless all processors threw an exception at the same point. If only one processor throws an exception then the others will wait indefinitely on the next MPI communication or collective operation.

Fixes Hermes-3 issue https://github.com/boutproject/hermes-3/issues/448


ZedThree (approved):
Thanks @bendudson!

PR #3257, opened 2026-01-27
switch to `ruff format`

dschwoerer:
See https://github.com/boutproject/hermes-3/pull/486

To ensure BOUT++ and hermes-3 are more aligned


ZedThree:
I think this is no longer needed?

PR #3260, opened 2026-01-29
Add checks to chooseProcessorSplit function in boutmesh.cxx

wearysebas:
Setting NXPE or NYPE would not trigger the checks of number of points / number of specified processes. Also added tests to it and modified previous tests to work.

### DIFF
--- a/src/mesh/impls/bout/boutmesh.cxx
+++ b/src/mesh/impls/bout/boutmesh.cxx
@@ -268,7 +268,11 @@ void BoutMesh::chooseProcessorSplit(Options& options) {
           _f("Number of processors ({:d}) not divisible by NPs in x direction ({:d})\n"),
           NPES, NXPE);
     }
-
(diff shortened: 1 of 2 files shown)


dschwoerer (changes requested):
Thank you for adding checks and providing tests 🎉 

The tests are failing to compile, as the tests re-use names.
Do you mind fixing that?

dschwoerer (dismissed):

mikekryjak:
Hi @wearysebas thanks for doing this. I'm wondering whether you could add some verbose comments to the check to tell the user why a given processor number doesn't work. I came up with error text for each issue here which you can copy:
https://github.com/boutproject/xhermes/blob/5588fbd0ceb2ec98344cd1752f2acf69c25055e6/xhermes/load.py#L689-L720

bendudson (approved):

PR #3261, opened 2026-01-30
Pull out common classes for FFT/DST transforms in some Laplacians

ZedThree:
The Laplacian inversion solvers `LaplaceCyclic`, `LaplacePCR`, `LaplacePCR_THOMAS` all use identical code for transforming using FFTs/DSTs. This commit pulls out that common code to make it easier to maintain.

Also, make the transforms parallel-Z-ready by using the Z guard cells as offsets


src/invert/laplace/common_transform.cxx
```diff
@@ -0,0 +1,381 @@
...
+
+      // Copy into array, transposing so kz is first index
+      for (int kz = 0; kz < nmode; kz++) {
+        result.bcmplx(((iy - ys) * nmode) + kz, ix - xs) = k1d[kz];
+      }
+    }
+
+    // Get elements of the tridiagonal matrix
+    // including boundary conditions
+    BOUT_OMP_PERF(for nowait)
```
dschwoerer, line 68: I think  the `nowait` has no impact?
I would assume the threads anyway have to wait at return?

But then I think it anyway waited at the end of the parallel region, so nothing really changed.

dschwoerer (approved):
Looks good. We might remove the `nowait`, but that is optional ...

src/invert/laplace/common_transform.cxx
ZedThree: I just copied it from the existing implementation :shrug:

PR #3262, opened 2026-01-30
Add unit tests for `ShiftedMetricInterp`

ZedThree:
Because this method should be identical to `ShiftedMetric`, we can just parameterise the existing unit tests over the two types (although the `FieldPerp` overloads are not implemented so we have to skip those). This is in support of the work to parallelise in Z, by getting more tests on bits that that touches.

One important thing to note is that `ZHermiteSpline` (the z-only interpolation) explicitly doesn't do anything in _any_ of the guards. This means that `f.yup(mesh->yend + 1)` does not get filled in. There's [a comment](https://github.com/boutproject/BOUT-dev/blob/b5661c57aeb4e80e9b859b347a8083c252688796/src/mesh/interpolation/interpolation_z.cxx#L30-L31) the guards should be filled in by the boundary conditions, but I'm not actually sure that's true for the parallel slices.

In the second commit, I end up just masking the guards out basically, just to make the test pass.

If I instead comment out the region masking in `ZInterpolation`, and just use `RGN_NOY` instead, all the tests seem to pass, including the integrated tests that use MAST grids.

@dschwoerer @johnomotani Thoughts? Am I missing something why we can't just use `RGN_NOY` for the Z interpolation here?


johnomotani:
Not sure what the right thing to do is. It seems like I got "RGN_NOBNDRY" (`region = localmesh->getRegion3D("RGN_NOBNDRY");`), then tried to remove the boundary points at the y-boundaries, but surely they were not included already? Maybe I have just forgotten what "RGN_NOBNDRY" includes.

I think my expectation when writing this was that if using parallel slices, y boundary conditions would always be applied directly to the parallel slices, but I think that's not necessarily the case (e.g. boundary conditions that are applied by transforming to/from field-aligned)? So sometimes the interpolation should be done in the y-boundaries? My concern would be that in cases where the boundary condition is directly applied to the parallel slices, it is not (?) applied to the Field3D data, so interpolating would potentially overwrite the boundary conditions in the parallel slices with junk?

Thinking out loud:
* When the bc is applied directly to the parallel slices, the boundary condition has to be applied after the parallel slices are calculated.
* If the bc is applied to the Field3D data (e.g. using to/from field aligned method) it needs to be applied before calculating parallel slices (unless these boundary conditions know to fill in the parallel slices?).
Supporting both cases neatly seems tricky!

PR #3293, opened 2026-03-02
Convert `LaplacePetsc` to use `PetscMatrix`

ZedThree:
As well as making this implementation a _lot_ cleaner, this is necessary for the Z parallelisation so that we can use the `GlobalIndexer`, otherwise it's really horrible to work out the domain decomposition (as the grid is Z-first and the processors are X-first, and the matrix may have different number of rows on each processor due to the X boundaries).

I had to remove the `TRACE` macro from the `PetscMatrix` as this gets called in a hot loop (`(nx * nz)^2` times!) and the string formatting really adds up -- yet more reason to remove `MsgStack`?

In debug builds, this has a bit of a weird effect on performance:

`next`:

Timer name | Total time (s) | % of top | Hits    | Mean time/hit (s)
-----------|----------------|----------|---------|------------------
petscsolve | 4.36366336     | 1.00     | 6       | 0.727277226      
petscsetup | 1.50196716     | 0.34     | 6       | 0.250327859    

this PR:

Timer name | Total time (s) | % of top | Hits | Mean time/hit (s)
-----------|----------------|----------|------|------------------
petscsolve | 3.76327226     | 1.00     | 6    | 0.627212044      
petscsetup | 2.25778754     | 0.60     | 6    | 0.376297923   

The solve time goes down, but the setup time goes up.

### DIFF
--- a/src/invert/laplace/impls/petsc/petsc_laplace.cxx
+++ b/src/invert/laplace/impls/petsc/petsc_laplace.cxx
@@ -234,7 +234,9 @@ FieldPerp LaplacePetsc::solve(const FieldPerp& b, const FieldPerp& x0) {
   checkFlags();
 #endif
 
-  const int y = b.getIndex(); // Get the Y index
+  // Set member variable so that we can pass through to shell preconditioner if
+  // required
+  yindex = b.getIndex();
   {
     const Timer timer("petscsetup");
 
@@ -243,9 +245,9 @@ FieldPerp LaplacePetsc::solve(const FieldPerp& b, const FieldPerp& x0) {
 
     // Set the operator matrix
     if (fourth_order) {
-      setFourthOrderMatrix(y, inner_X_neumann, outer_X_neumann);
+      setFourthOrderMatrix(yindex, inner_X_neumann, outer_X_neumann);
     } else {
-      setSecondOrderMatrix(y, inner_X_neumann, outer_X_neumann);
+      setSecondOrderMatrix(yindex, inner_X_neumann, outer_X_neumann);
     }
 
     operator2D.assemble();
@@ -358,7 +360,7 @@ FieldPerp LaplacePetsc::solve(const FieldPerp& b, const FieldPerp& x0) {
   }
 
   auto sol = guess.toField();
-  sol.setIndex(y);
+  sol.setIndex(yindex);
   checkData(sol);
 
   // Return the solution
@@ -651,7 +653,7 @@ void LaplacePetsc::setFourthOrderMatrix(int y, bool inner_X_neumann,
 
 /// Preconditioner function
 int LaplacePetsc::precon(Vec x, Vec y) {
-  FieldPerp xfield(indexer->getMesh(), location, sol.getIndex());
+  FieldPerp xfield(indexer->getMesh(), location, yindex);
   xfield = 0.0;
 
   BOUT_FOR_SERIAL(i, indexer->getRegionAll()) {
--- a/src/invert/laplace/impls/petsc/petsc_laplace.cxx
+++ b/src/invert/laplace/impls/petsc/petsc_laplace.cxx
@@ -47,6 +47,7 @@
 #include <bout/utils.hxx>
 
 #include <cmath>
+#include <memory>
 #include <set>
 #include <vector>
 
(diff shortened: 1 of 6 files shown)


ZedThree:
Hmm, the 3D build is failing with

> [0]PETSC ERROR: Unable to find requested PC type mumps

even though I explicitly check that PETSc was compiled with mumps. I guess it's fine to always use `sor`

src/invert/laplace/impls/petsc/petsc_laplace.cxx
```diff
@@ -986,99 +486,199 @@ void LaplacePetsc::Coeffs(int x, int y, int z, BoutReal& coef1, BoutReal& coef2,
...
+    const auto dx = sliceXZ(coords->dx, y);
+    const auto g11 = sliceXZ(coords->g11, y);
+
+    BOUT_FOR_SERIAL(i, indexer->getRegionOuterX()) {
+      const auto factor = 1. / dx[i] / std::sqrt(g11[i]);
+      operator2D(i, i) = (25.0 / 12.0) * factor;
+      operator2D(i, i.xm(1)) = -4.0 * factor;
+      operator2D(i, i.xm(2)) = 3.0 * factor;
+      operator2D(i, i.xm(3)) = (-4.0 / 3.0) * factor;
+      operator2D(i, i.xm(4)) = (1.0 / 4.0) * factor;
```
dschwoerer, line 590: There should be two points where this is used, but on one of them we should be able to use also `i.xp()` ?
Would that not be better?

src/invert/laplace/impls/petsc/petsc_laplace.cxx
```diff
@@ -986,99 +486,199 @@ void LaplacePetsc::Coeffs(int x, int y, int z, BoutReal& coef1, BoutReal& coef2,
...
+    operator2D(l, l.xp().zmm()) = A3 / (18.0 * dxdz);
+    operator2D(l, l.xp().zm()) = -4.0 * A3 / (9.0 * dxdz);
+    operator2D(l, l.xp()) = (4.0 * A1 / (3.0 * dx2)) + (2.0 * A4 / (3.0 * dx));
+    operator2D(l, l.xp().zp()) = 4.0 * A3 / (9.0 * dxdz);
+    operator2D(l, l.xp().zpp()) = -1.0 * A3 / (18.0 * dxdz);
+    operator2D(l, l.xpp().zmm()) = -1.0 * A3 / (144.0 * dxdz);
+    operator2D(l, l.xpp().zm()) = A3 / (18.0 * dxdz);
+    operator2D(l, l.xpp()) = (-1.0 / 12.0) * ((A1 / dx2) + (A4 / dx));
+    operator2D(l, l.xpp().zp()) = -1.0 * A3 / (18.0 * dxdz);
+    operator2D(l, l.xpp().zpp()) = A3 / (144.0 * dxdz);
```
dschwoerer, line 650: We could pull out the `A3/dxdz` and similar.

I guess MMS tests are checking whether all the factors are correct?

`-1.0 * A3 / (18.0 * dxdz)` could be simplified to 
```
auto const A3fac = A3 / dxdz`
A3fac / -18.0
```
Not sure the compiler will do that, as that might only be allowed with `-Ofast` and similar.

src/invert/laplace/impls/petsc/petsc_laplace.cxx
```diff
@@ -65,6 +72,81 @@ PetscErrorCode laplacePCapply(PC pc, Vec x, Vec y) {
 
   PetscFunctionReturn(laplace->precon(x, y)); // NOLINT
 }
+
+// TODO: handle fourth order
```
dschwoerer: That seems to be done now?

src/invert/laplace/impls/petsc/petsc_laplace.cxx
ZedThree: I haven't touched any of the factors, but yes, likely some of them could be rewritten. I don't think there are MMS tests on anything but the default Laplacian

src/invert/laplace/impls/petsc/petsc_laplace.cxx
ZedThree: Yep, thanks!

dschwoerer (dismissed):

ZedThree:
@bendudson This is now ready

bendudson (approved):

PR #3294, opened 2026-03-02
Added the finalized converter tool from an INGRID to BOUT grid.

wearysebas:
Works for all topologies.


bendudson:
Thanks @wearysebas ! I think this is going into boutdata (https://github.com/boutproject/boutdata/pull/146) so closing this PR.

PR #3296, opened 2026-03-03
Make Field2D more like Field3D

dschwoerer:
Make it easier to write FCI compatible code.


include/bout/field.hxx
```diff
@@ -125,6 +128,12 @@ public:
     swap(first.directions, second.directions);
   }
 
+  virtual void setRegion(size_t UNUSED(regionID)) {}
```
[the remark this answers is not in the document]
ZedThree: Could you switch the `UNUSED` macro to the `[[maybe_unused]]` attribute?

include/bout/field2d.hxx
```diff
@@ -133,8 +133,12 @@ public:
     return *this;
   }
 
-  /// Check if this field has yup and ydown fields
+  /// Dummy functions to increase portability
   bool hasParallelSlices() const { return true; }
+  void calcParallelSlices() const {}
+  void splitParallelSlices() const {}
+  void clearParallelSlices() const {}
+  int numberParallelSlices() const { return 0; }
```
ZedThree: Should these be virtual? Or do they only need to be dummy methods?

include/bout/field3d.hxx
```diff
@@ -510,6 +510,8 @@ public:
 
   std::weak_ptr<Options> getTracking() { return tracking; };
 
+  bool allowCalcParallelSlices{true};
```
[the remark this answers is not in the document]
ZedThree: Could we add a public getter (and setter) instead of making the member variable public?

include/bout/fieldperp.hxx
```diff
@@ -157,6 +157,25 @@ public:
     return *this;
   }
 
+  /// Dummy functions to increase portability
+  bool hasParallelSlices() const { return true; }
+  void calcParallelSlices() const {}
+  void clearParallelSlices() {}
+  int numberParallelSlices() { return 0; }
+
+  FieldPerp& yup(std::vector<FieldPerp>::size_type UNUSED(index) = 0) { return *this; }
```
ZedThree, line 161: `std::size_t` is fine

ZedThree (dismissed):
Nice, thanks @dschwoerer! I guess my main question is could/should the new methods actually be virtual? Then we could just have a default empty implementation on `Field`.

bendudson (dismissed):

include/bout/field3d.hxx
```diff
@@ -511,6 +513,9 @@ public:
 
   std::weak_ptr<Options> getTracking() { return tracking; };
 
+  bool allowCalcParallelSlices() const { return _allowCalcParallelSlices; };
+  void disallowCalcParallelSlices() { _allowCalcParallelSlices = false; };
```
ZedThree, line 517: What do these do? It doesn't look like they're used anywhere.

Also they sound like they should set `_allowCalcParallelSlices` to `true/false`, but the first one is a getter and the second one a setter.

Maybe the first should be `areCalcParallelSlicesAllowed()`?

include/bout/field3d.hxx
```diff
@@ -263,24 +265,24 @@ public:
 
   /// Check if this field has yup and ydown fields
   /// Return reference to yup field
-  Field3D& yup(std::vector<Field3D>::size_type index = 0) {
+  Field3D& yup(size_t index = 0) {
```
ZedThree, line 268: I would've thought we would also need these to be virtual to make things properly generic? Although I guess you're mostly interested in `FieldMetric` and compile-time polymorphism really, and not runtime polymorphism?

We could make these virtual by returning `Field&`, but we probably don't actually want that.

ZedThree (changes requested):

include/bout/field3d.hxx
dschwoerer: Yes, that is the reason. We do not want to return Field, so there is no way to make this virtual.
Compile time polymorphism is right now sufficient, as long as 3D metrics are a compile time choice.

include/bout/field3d.hxx
dschwoerer: > What do these do? It doesn't look like they're used anywhere.

They slipped in from a different PR.

> Also they sound like they should set `_allowCalcParallelSlices` to `true/false`, but the first one is a getter and the second one a setter.
> 
> Maybe the first should be `areCalcParallelSlicesAllowed()`?

Will change it :+1:

include/bout/field3d.hxx
```diff
@@ -511,6 +513,9 @@ public:
 
   std::weak_ptr<Options> getTracking() { return tracking; };
 
+  bool allowCalcParallelSlices() const { return _allowCalcParallelSlices; };
```
dschwoerer:
```suggestion
  bool areCalcParallelSlicesAllowed() const { return _allowCalcParallelSlices; };
```

ZedThree (approved):

PR #3298, opened 2026-03-03
Improvements for hermitespline interpolation

dschwoerer:
Includes X-splitting also for monotonichermitespline and more.


src/mesh/parallel/fci_comm.hxx
```diff
@@ -0,0 +1,310 @@
...
+  GlobalField3DAccessInstance(const GlobalField3DAccess* gfa,
+                              std::vector<BoutReal>&& data)
+      : gfa(gfa), data(std::move(data)){};
+
+private:
+  const GlobalField3DAccess* gfa;
+  std::vector<BoutReal> data;
+};
+
+class GlobalField3DAccess {
```
ZedThree, line 114: This needs a docstring and comments throughout the implementation explaining what it is, what it's for, and how it works please.

src/mesh/parallel/fci_comm.hxx
```diff
@@ -0,0 +1,310 @@
...
+#ifdef _OPENMP
+    ASSERT2(o_ids.size() > static_cast<size_t>(omp_get_thread_num()));
+    o_ids[omp_get_thread_num()].emplace(ind.ind);
+#else
+    ids.emplace(ind.ind);
+#endif
+  }
+
+  void operator[](IndG3D ind) { get(ind); }
+  void setup() {
```
ZedThree: This is large enough it should be in the `.cxx` file really.

Is there anyway this can be part of the constructor? Having two-part constructors like this makes it easier to use wrong.

src/mesh/parallel/fci_comm.hxx
```diff
@@ -0,0 +1,310 @@
...
+  }
+  GlobalField3DAccessInstance communicate(const Field3D& f) {
+    return {this, communicate_data(f)};
+  }
+  std::unique_ptr<GlobalField3DAccessInstance> communicate_asPtr(const Field3D& f) {
+    return std::make_unique<GlobalField3DAccessInstance>(this, communicate_data(f));
+  }
+
+private:
+  void commCommLists() {
```
ZedThree: Should probably also be in .cxx file

src/mesh/parallel/fci_comm.hxx
```diff
@@ -0,0 +1,310 @@
...
+      ASSERT3(ind != MPI_UNDEFINED);
+      ASSERT2(static_cast<size_t>(ind) < toSend.size());
+      ASSERT3(toSendSizes[ind] >= 0);
+      if (toSendSizes[ind] == 0) {
+        continue;
+      }
+      sendBufferSize += toSendSizes[ind];
+      toSend[ind].resize(toSendSizes[ind], -1);
+
+      ret = MPI_Irecv(static_cast<void*>(toSend[ind].data()), toSend[ind].size(), MPI_INT,
```
ZedThree: No need to cast to `void*`, all pointers are implicitly convertable

src/mesh/parallel/fci_comm.hxx
```diff
@@ -0,0 +1,310 @@
...
+  const fci_comm::XYZ2Ind<Ind3D> xyzl;
+  const fci_comm::XYZ2Ind<IndG3D> xyzg;
+
+private:
+  std::vector<std::vector<int>> toGet;
+  std::vector<std::vector<int>> toSend;
+  std::vector<int> getOffsets;
+  int sendBufferSize{0};
+  MPI_Comm comm;
+  std::vector<BoutReal> communicate_data(const Field3D& f) {
```
ZedThree: Also for .cxx file

src/mesh/interpolation/hermite_spline_xz.cxx
```diff
@@ -422,6 +472,24 @@ Field3D XZHermiteSpline::interpolate(const Field3D& f, const std::string& region
     f_interp[iyp] =
         +f_z * h00_z[i] + f_zp1 * h01_z[i] + fz_z * h10_z[i] + fz_zp1 * h11_z[i];
 
+    if constexpr (monotonic) {
+#endif
```
ZedThree: I _really_ don't like constructions like this, interleaving the preprocessor with partial statements, it's a recipe for accidentally breaking builds down the line.

src/mesh/interpolation/hermite_spline_xz.cxx
```diff
@@ -347,16 +391,21 @@ XZHermiteSpline::getWeightsForYApproximation(int i, int j, int k, int yoffset) {
...
+  const auto region2 = y_offset != 0 ? fmt::format("RGN_YPAR_{:+d}", y_offset) : region;
 
-#if USE_NEW_WEIGHTS
-#ifdef HS_USE_PETSC
+  std::unique_ptr<GlobalField3DAccessInstance> gf;
+  if constexpr (monotonic) {
+    gf = gf3daccess->communicate_asPtr(f);
+  }
+
+#if USE_NEW_WEIGHTS and defined(HS_USE_PETSC)
```
ZedThree: I still very much think we should split these out into separate implementations. As it is, verifying that all three variations work requires three separate builds! Having separate implementations means that they can all be tested in the same build

src/mesh/parallel/fci_comm.hxx
```diff
@@ -0,0 +1,310 @@
...
+  const int nz;
+  ind convert(const int x, const int y, const int z) const {
+    return {z + ((y + x * ny) * nz), ny, nz};
+  }
+  ind operator()(const int x, const int y, const int z) const { return convert(x, y, z); }
+  XYZ2Ind(const int nx, const int ny, const int nz) : nx(nx), ny(ny), nz(nz) {}
+};
+} // namespace fci_comm
+
+class GlobalField3DAccessInstance {
```
ZedThree, line 102: Why does this need a separate class?

src/mesh/parallel/fci_comm.hxx
```diff
@@ -0,0 +1,310 @@
...
+        proc = npe - 1;
+      }
+    }
+    const int loc = id - (local * proc);
+#if CHECK > 1
+    if ((loc < 0 or loc > localwith or proc < 0 or proc >= npe)
+        or (periodic and (loc < mg or loc >= local + mg))) {
+      printf("globalToLocal1D failure: %d %d, %d %d, %d %d %s\n", id, idwo, globalwith,
+             npe, proc, loc, periodic ? "periodic" : "non-periodic");
+      ASSERT0(false);
```
ZedThree: This should just throw `BoutException` directly and explain what the problem is. At the minute it's very cryptic

src/mesh/parallel/fci_comm.hxx
```diff
@@ -0,0 +1,310 @@
...
+#include <utility>
+#include <vector>
+class GlobalField3DAccess;
+
+namespace fci_comm {
+struct ProcLocal {
+  int proc;
+  int ind;
+};
+struct globalToLocal1D {
```
ZedThree: Classes should start with a capital:

```suggestion
struct GlobalToLocal1D {
```

Needs a docstring explaining purpose and limitations

src/mesh/parallel/fci_comm.hxx
```diff
@@ -0,0 +1,310 @@
...
+  std::vector<BoutReal> data;
+};
+
+class GlobalField3DAccess {
+public:
+  friend class GlobalField3DAccessInstance;
+  GlobalField3DAccess(Mesh* mesh)
+      : mesh(mesh), g2lx(mesh->xstart, mesh->getNXPE(), mesh->LocalNx, false),
+        g2ly(mesh->ystart, mesh->getNYPE(), mesh->LocalNy, true),
+        g2lz(mesh->zstart, 1, mesh->LocalNz, true),
```
ZedThree:
```suggestion
        g2lz(mesh->zstart, mesh->getNZPE(), mesh->LocalNz, true),
```

src/mesh/parallel/fci_comm.hxx
```diff
@@ -0,0 +1,310 @@
...
+  const GlobalField3DAccess* gfa;
+  std::vector<BoutReal> data;
+};
+
+class GlobalField3DAccess {
+public:
+  friend class GlobalField3DAccessInstance;
+  GlobalField3DAccess(Mesh* mesh)
+      : mesh(mesh), g2lx(mesh->xstart, mesh->getNXPE(), mesh->LocalNx, false),
+        g2ly(mesh->ystart, mesh->getNYPE(), mesh->LocalNy, true),
```
ZedThree: Are the periodicities hardcoded here because this only works with X non-periodic and Y periodic? If so, we should actually check this!

src/mesh/parallel/fci_comm.hxx
```diff
@@ -0,0 +1,310 @@
...
+    o_ids.clear();
+#endif
+    toGet.resize(static_cast<size_t>(g2lx.npe * g2ly.npe * g2lz.npe));
+    for (const auto id : ids) {
+      const IndG3D gind{id, g2ly.globalwith, g2lz.globalwith};
+      const auto pix = g2lx.convert(gind.x());
+      const auto piy = g2ly.convert(gind.y());
+      const auto piz = g2lz.convert(gind.z());
+      ASSERT3(piz.proc == 0);
+      toGet[(piy.proc * g2lx.npe) + pix.proc].push_back(
```
ZedThree: We should also take into account Z processors here, or this will just need to be fixed soon anyway

src/mesh/parallel/fci_comm.hxx
```diff
@@ -0,0 +1,310 @@
...
+      : mesh(mesh), g2lx(mesh->xstart, mesh->getNXPE(), mesh->LocalNx, false),
+        g2ly(mesh->ystart, mesh->getNYPE(), mesh->LocalNy, true),
+        g2lz(mesh->zstart, 1, mesh->LocalNz, true),
+        xyzl(g2lx.localwith, g2ly.localwith, g2lz.localwith),
+        xyzg(g2lx.globalwith, g2ly.globalwith, g2lz.globalwith), comm(BoutComm::get()) {
+#ifdef _OPENMP
+    o_ids.resize(omp_get_max_threads());
+#endif
+  };
+  void get(IndG3D ind) {
```
ZedThree: This needs a better name, it doesn't `get` anything!

ZedThree (commented):
Looks ok, but needs a lot more explanatory comments and docstrings. I can guess what a lot of things do, but I'd rather read a high-level overview first.

Lots of things need renaming too -- again, I can guess, but I'd really rather not. We have to think about new developers, are they going to be able to work out what `xyzl` or `o_ids` means?

The template base class for the monotonic version is a good idea, but I still _really_ don't like squashing together the PETSc, "new weights", and non-PETSc versions. It means three separate builds to test them all, rather than being able to test all of them in a single build. The interleaved preprocessor statements also make it much harder to understand what's going on.

Lastly, it would be good to make this already work for parallelised-Z so we have to do less work when that's ready. There's also already two other mechanisms for local-to-global mapping, `GlobalIndexer` and `GlobalField` (which I'm not sure is even used for anything seriously now), it would be good to understand if they could be used here at all.

src/mesh/parallel/fci_comm.hxx
dschwoerer: What is the order with Z-splitting? Is it still fortran-ordering or something else?

src/mesh/interpolation/hermite_spline_xz.cxx
dschwoerer: Removed this :+1:

src/mesh/parallel/fci_comm.hxx
ZedThree: Yes, unfortunately the processor splitting goes the other way to the grid points: grid points go (z, y, x), processors go (x, y, z).

It's also quite painful if X is non-periodic and the boundaries are included in the global numbering, then `firstX or lastX` processors have different number of points to `not (firstX or lastX)`

src/mesh/parallel/fci_comm.hxx
dschwoerer: It needs to be constant for the whole simulation. I am not sure how to easily check for that.

Y periodic is anyway always a good idea. If you want somewhere boundaries, make the mesh periodic, but set boundaries. If you want something else, that would also fail in other bits of the code.

dschwoerer:
> Looks ok, but needs a lot more explanatory comments and docstrings. I can guess what a lot of things do, but I'd rather read a high-level overview first.

I have added some docs.

> Lots of things need renaming too -- again, I can guess, but I'd really rather not. We have to think about new developers, are they going to be able to work out what `xyzl` or `o_ids` means?

I improved the naming

> The template base class for the monotonic version is a good idea, but I still _really_ don't like squashing together the PETSc, "new weights", and non-PETSc versions. It means three separate builds to test them all, rather than being able to test all of them in a single build. The interleaved preprocessor statements also make it much harder to understand what's going on.

I extended on the templating. Now all versions are templated.

> Lastly, it would be good to make this already work for parallelised-Z so we have to do less work when that's ready. There's also already two other mechanisms for local-to-global mapping, `GlobalIndexer` and `GlobalField` (which I'm not sure is even used for anything seriously now), it would be good to understand if they could be used here at all.

I added an interface to the mesh, to do the mapping. If you think this should be reproduced here, I would need to know how you intend to do it.

All comments addressed, ready for re-review :+1:

bendudson (approved):

PR #3299, opened 2026-03-03
Allow to overwrite MYG, if it is set from the mesh file

dschwoerer:
This allows to create a mesh with 2 guard cells, but run simulations with only one.


bendudson (dismissed):

ZedThree (approved):
Neat, thanks @dschwoerer!

PR #3301, opened 2026-03-04
Run clang-tidy on formatted files

dschwoerer:


ZedThree:
Might be nicer as a separate build that relies on the `clang-format` build, rather than bundling them together?

dschwoerer:
Yes, but I was not sure how to do that. For some reason the CI does not run on the bot commits, otherwise we could just error out if clang-format is finding issues.

If you have a better idea, I am all for it :+1:

dschwoerer:
Replaced by #3315

PR #3302, opened 2026-03-04
Allow more parallel boundary regions

dschwoerer:
Also allow to set boundaries for bndry_par_xin/xout and bndry_par_yup/ydown

### DIFF
--- a/src/mesh/boundary_factory.cxx
+++ b/src/mesh/boundary_factory.cxx
@@ -228,45 +228,81 @@ BoundaryOpBase* BoundaryFactory::createFromOptions(const string& varname,
 
   string prefix("bndry_");
 
-  string side;
+  std::array<string, 5> sides;
+  sides[0] = region->label;
   switch (region->location) {
   case BNDRY_XIN: {
-    side = "xin";
+    sides[1] = "xin";
     break;
   }
   case BNDRY_XOUT: {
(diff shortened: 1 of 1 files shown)


src/mesh/boundary_factory.cxx
```diff
@@ -228,91 +229,107 @@ BoundaryOpBase* BoundaryFactory::createFromOptions(const string& varname,
 
   string prefix("bndry_");
 
-  string side;
+  std::array<string, 5> sides;
```
ZedThree, line 232: Maybe just needs a comment explaining that this is the list of suffixes to try when setting boundaries, from most to least specific?

ZedThree (dismissed):

src/mesh/boundary_factory.cxx
dschwoerer, line 232:
```suggestion
  // suffixes, from most to least specific
  std::array<string, 5> sides;
```

ZedThree (dismissed):

src/mesh/boundary_factory.cxx
ZedThree: Looks like it's lost this comment -- can you just add it back with `[skip ci]` in the message, and I'll merge?

bendudson (approved):

PR #3308, opened 2026-03-05
New Y-Boundaries for unified sheath code

dschwoerer:
This allows to implement unified BCs for yup/ydown as well as FCI / FA.
Taking into acount things like whether the boundary is next to the domain, or between the parallel slices.


dschwoerer:
I rebased this, and applied clang-format to all commits, with respect to origin/next.

I fail to see why clang-format in CI is not happy. I am using `git-clang-format-21.1.8-4.fc43.x86_64`.

And this is the reason why I think having CI fail for missing clang-format is an issue.

dschwoerer:
Some ToDos:
- [x] remove setup, move code to constructor
- [x] move to coordinates, so that users of the API can get an object from coordinates (or mesh?), and do not have to setup there own copy / rely on a global instance

manual/sphinx/user_docs/boundary_options.rst
```diff
@@ -436,6 +436,130 @@ the upper Y boundary of a 2D variable ``var``::
...
+.. list-table:: Members for boundary operation
+   :widths: 15 70
+   :header-rows: 1
+
+   * - Function
+     - Description
+   * - ``pnt.ythis(f)``
+     - Returns the value at the last point in the domain
+   * - ``pnt.ynext(f)``
+     - Returns the value at the first point in the domain
```
mikekryjak: This should read "the first point beyond the domain, i.e. the first (innermost) guard cell"

dschwoerer: Thanks, I have updated the documentation :+1: 

Just to clarify, for FCI `ythis` can also be a guard cell, i.e. an interpolated value, if `MYG>1`, but not a boundary cell.

include/bout/sys/parallel_stencils.hxx
```diff
@@ -0,0 +1,41 @@
+#pragma once
+#include <bout/assert.hxx>
+#include <bout/bout_types.hxx>
+
+namespace parallel_stencil {
```
ZedThree:
```suggestion
namespace bout::parallel_stencil {
```

include/bout/sys/parallel_stencils.hxx
```diff
@@ -0,0 +1,41 @@
...
+// generated by src/mesh/parallel_boundary_stencil.cxx.py
+inline BoutReal pow(BoutReal val, int exp) {
+  // constexpr int expval = exp;
+  // static_assert(expval == 2 or expval == 3, "This pow is only for exponent 2 or 3");
+  if (exp == 2) {
+    return val * val;
+  }
+  ASSERT3(exp == 3);
+  return val * val * val;
+}
```
ZedThree: It doesn't look like we call this with anything other than `2`, and from what I can tell, this isn't actually faster than `std::pow`, so let's just use that?

src/mesh/parallel/fci.cxx
```diff
@@ -354,17 +364,20 @@ FCITransform::FCITransform(Mesh& mesh, const Coordinates::FieldMetric& dy, bool
...
     for (const auto& bndry2 : bndries) {
       if (bndry->dir == bndry2->dir) {
         continue;
       }
-      for (bndry->first(); !bndry->isDone(); bndry->next()) {
-        if (bndry2->contains(*bndry)) {
-          bndry->setValid(0);
+      for (auto pnt : *bndry) {
+        for (auto pnt2 : *bndry2) {
+#warning this could likely be done faster
```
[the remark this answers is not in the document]
ZedThree: I think this is `zip`? Unfortunately, that's C++23, but we could add an implementation if necessary. Either way, we don't want to merge this with the warning!

src/mesh/parallel/fci.cxx
```diff
@@ -344,17 +354,20 @@ FCITransform::FCITransform(Mesh& mesh, const Coordinates::FieldMetric& dy, bool
...
+          if (pnt.ind() == pnt2.ind()) {
+            pnt.setValid(
+                static_cast<signed char>(std::abs((pnt2.offset() - pnt.offset())) - 2));
```
ZedThree: This is not obvious to me, please could you add a comment about what this is doing?

src/mesh/parallel/fci.cxx
```diff
@@ -252,11 +261,12 @@ FCIMap::FCIMap(Mesh& mesh, [[maybe_unused]] const Coordinates::FieldMetric& dy,
     // need at least 2 points in the domain.
     ASSERT2(map_mesh->xend - map_mesh->xstart >= 2);
     auto boundary = (xt_prime[i] < map_mesh->xstart) ? inner_boundary : outer_boundary;
-    boundary->add_point(x, y, z, x + dx, y + (0.5 * offset_),
-                        z + dz, // Intersection point in local index space
-                        0.5,    // Distance to intersection
-                        1       // Default to that there is a point in the other direction
-    );
+    if (!boundary->contains(x, y, z)) {
+      boundary->add_point(x, y, z, x + dx, y + offset - (sgn(offset) * 0.5),
```
ZedThree:
```suggestion
      boundary->add_point(x, y, z, x + dx, y + offset - std::copysign(0.5, offset),
```

src/mesh/parallel/fci.cxx
```diff
@@ -62,6 +62,13 @@
 #include <string>
 #include <string_view>
 
+namespace {
+template <typename T>
+int sgn(T val) {
+  return (T(0) < val) - (val < T(0));
+}
+} // namespace
+
```
ZedThree: This is just [`std::copysign`](https://en.cppreference.com/w/cpp/numeric/math/copysign.html)

```suggestion
```

src/mesh/impls/bout/boutmesh.cxx
```diff
@@ -2928,6 +2933,9 @@ void BoutMesh::addBoundaryRegions() {
 }
 
 RangeIterator BoutMesh::iterateBndryLowerInnerY() const {
+  if (this->isFci()) {
```
ZedThree, line 2937: Do we want these checks only at `CHECK > 0` maybe?

src/invert/laplace/invert_laplace.cxx
```diff
@@ -218,10 +218,10 @@ Field3D Laplacian::solve(const Field3D& b, const Field3D& x0) {
 
   // Setting the start and end range of the y-slices
   int ys = localmesh->ystart, ye = localmesh->yend;
-  if (localmesh->hasBndryLowerY() && include_yguards) {
+  if (include_yguards && localmesh->hasBndryLowerY()) {
```
ZedThree, line 221: Why this change?

manual/sphinx/user_docs/boundary_options.rst
```diff
@@ -436,6 +436,127 @@ the upper Y boundary of a 2D variable ``var``::
...
+
+    class yboundary_example {
+    public:
+      yboundary_example(Options* opt, const Field3D& N, const Field3D& V) :
+      N(N), V(V) {}
+
+      void rhs() {
+        BoutReal totalFlux = 0;
+        mesh->getCoordinates()->getYBoundary()->iter_pnts([&](auto& pnt) {
+          BoutReal flux = pnt.interpolate_sheath_o2(N) * pnt.interpolate_sheath_o2(V);
```
ZedThree: Should this be `return` instead of `BoutReal flux`?

include/bout/yboundary_regions.hxx
```diff
@@ -0,0 +1,153 @@
...
+  void iter_regions(const F& f) {
+    for (auto& region : boundary_regions) {
+      f(*region);
+    }
+    for (auto& region : boundary_regions_par) {
+      f(*region);
+    }
+  }
+  template <class F>
+  void iter_pnts(const F& f) {
```
ZedThree: We can afford those extra two characters :)

```suggestion
  void iter_points(const F& f) {
```

Please can we use `point(s)` everywhere, rather than `pnt`?

include/bout/yboundary_regions.hxx
```diff
@@ -0,0 +1,153 @@
...
+/// It makes it easier to write code for FieldAligned boundaries, but if a bit
+/// care is taken the code also works with FluxCoordinateIndependent code.
+///
+/// An example how to replace old code is given here:
+/// ../../manual/sphinx/user_docs/boundary_options.rst
+
+class YBoundary {
+public:
+  template <class F>
+  void iter_regions(const F& f) {
```
ZedThree: Given that we almost always use `f` for a field, maybe this would be clearer as:


```suggestion
  template <class Func>
  void iter_regions(const Func& func) {
```

include/bout/yboundary_regions.hxx
```diff
@@ -0,0 +1,153 @@
...
+      }
+    });
+  }
+
+  template <class F>
+  void iter(const F& f) {
+    return iter_regions(f);
+  }
+
+  YBoundary(YBndryType type, Options* options_ptr, Mesh* mesh) {
```
ZedThree: Given `mesh` cannot be `nullptr`, why not `const Mesh&`?

include/bout/coordinates.hxx
```diff
@@ -221,9 +226,13 @@ public:
   // solver
   Field2D Laplace_perpXY(const Field2D& A, const Field2D& f);
 
+  friend std::shared_ptr<YBoundary> getYBoundary(Coordinates* coords, YBndryType type);
```
ZedThree, line 229: Why can't this be a member function?

include/bout/boundary_iterator.hxx
```diff
@@ -0,0 +1,218 @@
...
+#include "bout/field3d.hxx"
+#include "bout/mesh.hxx"
+#include "bout/parallel_boundary_region.hxx"
+#include "bout/region.hxx"
+#include "bout/sys/parallel_stencils.hxx"
+#include "bout/sys/range.hxx"
+#include <functional>
+#include <utility>
+
+class BoundaryRegionIter {
```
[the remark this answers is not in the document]
ZedThree: `clang-tidy` can add default implementations of these functions for you

include/bout/boundary_iterator.hxx
```diff
@@ -0,0 +1,218 @@
...
+    if (z == nz()) {
+      z = 0;
+      _next();
+    }
+    return *this;
+  }
+  virtual void _next() = 0;
+  BoundaryRegionIter& operator*() { return *this; }
+
+  void dirichlet_o2(Field3D& f, BoutReal value) const {
```
ZedThree, line 42: Do these need to be implemented as methods, or can they be free functions? Having them as methods makes it more difficult to extend and add new functions.

They also all need docstrings, some of these are very non-obvious!

include/bout/boundary_iterator.hxx
```diff
@@ -0,0 +1,218 @@
...
+  void setAll(Field3D& f, const BoutReal val) const {
+    for (int i = -localmesh->ystart; i <= localmesh->ystart; ++i) {
+      f[ind().yp(by * i).xp(bx * i)] = val;
+    }
+  }
+
+  static int abs_offset() { return 1; }
+
+#if BOUT_USE_METRIC_3D == 0
+  BoutReal& ynext(Field2D& f) const { return f[ind().yp(by).xp(bx)]; }
```
ZedThree: Does this really not compile for 2D metrics?

include/bout/boundary_iterator.hxx
```diff
@@ -0,0 +1,218 @@
...
+  BoundaryRegionIter& operator++() {
+    ASSERT3(z < nz());
+    z++;
+    if (z == nz()) {
+      z = 0;
+      _next();
+    }
+    return *this;
+  }
+  virtual void _next() = 0;
```
ZedThree, line 37: It looks like this should be `protected`?

include/bout/boundary_iterator.hxx
```diff
@@ -0,0 +1,218 @@
...
+
+  void limit_at_least(Field3D& f, BoutReal value) const {
+    ynext(f) = std::max(ynext(f), value);
+  }
+
+  BoutReal& ynext(Field3D& f) const { return f[ind().yp(by).xp(bx)]; }
+  const BoutReal& ynext(const Field3D& f) const { return f[ind().yp(by).xp(bx)]; }
+  BoutReal& yprev(Field3D& f) const { return f[ind().yp(-by).xp(-bx)]; }
+  const BoutReal& yprev(const Field3D& f) const { return f[ind().yp(-by).xp(-bx)]; }
+  BoutReal& ythis(Field3D& f) const { return f[ind()]; }
```
ZedThree: I'm a bit confused by these functions, they start with `y`, but the implementation has `.xp` in there, so I guess these are supposed to be generic for X and Y boundaries?

include/bout/boundary_iterator.hxx
```diff
@@ -0,0 +1,218 @@
...
+  static int abs_offset() { return 1; }
+
+#if BOUT_USE_METRIC_3D == 0
+  BoutReal& ynext(Field2D& f) const { return f[ind().yp(by).xp(bx)]; }
+  const BoutReal& ynext(const Field2D& f) const { return f[ind().yp(by).xp(bx)]; }
+  BoutReal& yprev(Field2D& f) const { return f[ind().yp(-by).xp(-bx)]; }
+  const BoutReal& yprev(const Field2D& f) const { return f[ind().yp(-by).xp(-bx)]; }
+#endif
+
+  const int dir;
```
ZedThree: Please sort out all these clang-tidy comments, basically just add getters:

```cpp
public:
  int dir() { return dir_; }

private:
  int dir_;
```

plus docstrings

include/bout/coordinates.hxx
```diff
@@ -249,17 +258,19 @@ private:
   void checkCovariant();
   // check that contravariant tensors are positive (if expected) and finite (always)
   void checkContravariant();
+
+  mutable std::array<std::shared_ptr<YBoundary>, 3> ybndrys;
```
ZedThree, line 262: Where does the `3` come from? How do we check that it is still the correct value?

include/bout/parallel_boundary_region.hxx
```diff
@@ -115,37 +128,99 @@ public:
...
+  BoundaryRegionParIterBase& operator++() {
+    ++bndry_position;
+    return *this;
+  }
+
+  bool operator!=(const BoundaryRegionParIterBase& rhs) {
+    return bndry_position != rhs.bndry_position;
+  }
+
+#define ITER() for (int i = 0; i < localmesh->ystart - abs_offset(); ++i)
```
ZedThree, line 210: We don't really need a macro for this

include/bout/boundary_iterator.hxx
```diff
@@ -0,0 +1,218 @@
...
+#include "bout/field3d.hxx"
+#include "bout/mesh.hxx"
+#include "bout/parallel_boundary_region.hxx"
+#include "bout/region.hxx"
+#include "bout/sys/parallel_stencils.hxx"
+#include "bout/sys/range.hxx"
+#include <algorithm>
+#include <functional>
+
+class BoundaryRegionIter {
```
ZedThree, line 16: I guess ideally this would be templated on the direction (X or Y)

ZedThree (commented):
I've not finished going through this, but it looks like it's almost there. The thing that would make this really nice is to template the new boundary region stuff so we can handle X/Y boundaries without any runtime switches or checking `bx/by` etc.

There's lots of clang-tidy comments that should be fixed as well.

And more docstrings and explanatory comments please!

include/bout/sys/parallel_stencils.hxx
dschwoerer: I tested it with gcc 15.2.1:
```
g++ example.cxx -std=c++20 -ffast-math -O2 -march=native -isystem include   -Lbuild/src -lbenchmark -lpthread -o mybenchmark && ./mybenchmark 2>&1   | tail -n2
BM_StdPow       0.234 ns        0.234 ns   3000470135
BM_multi        0.444 ns        0.444 ns   1573472940
[dave@raven03: benchmark]👍 g++ example.cxx -std=c++20 -O2 -march=native -isystem include   -Lbuild/src -lbenchmark -lpthread -o mybenchmark && ./mybenchmark 2>&1   | tail -n2
BM_StdPow        20.5 ns         20.5 ns     34629261
BM_multi        0.441 ns        0.441 ns   1600645212
```

So `std::pow` is faster, but only with `-ffast-math`.

src/mesh/impls/bout/boutmesh.cxx
dschwoerer: Probably. I though that does not matter for FCI, but maybe someone still uses FA :-)

src/invert/laplace/invert_laplace.cxx
dschwoerer: `hasBndryLowerY()` throws for FCI, without this change it first throws, and then it does not matter anymore that the result would not be needed.

include/bout/yboundary_regions.hxx
dschwoerer: I just read [CppCoreGuidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#rc-constref) - which suggests to better store a pointer. Does it make sense to pass a reference, but store a pointer?

include/bout/boundary_iterator.hxx
dschwoerer: I think it is only needed for 2D metrics, but it should also compile with 3D metrics.

include/bout/boundary_iterator.hxx
dschwoerer: They are only used for y, but could also be used for x.

include/bout/boundary_iterator.hxx
dschwoerer: Does it need to be templated? It could also just be generic without templating?
Are you worried about performance, or what is your motivation for templating?

include/bout/sys/parallel_stencils.hxx
dschwoerer: Correction: with `-ffast-math` they are the same speed, otherwise `std::pow` is a factor 100 slower.

bendudson (approved):
Thanks @dschwoerer !

PR #3313, opened 2026-03-09
sort .git-blame-ignore-revs

dschwoerer:
This should reduce the number of merge conflicts
if we do not always append to the file


ZedThree:
I'm not sure that sorting the commits will help that much, it will just make it less likely. It's probably better to only add commits to the file when they explicitly re-format large chunks of code.

(I'm going to cancel the tests by the way, as this won't affect anything!)

dschwoerer:
Closed in favor of forcing users running clang-format

PR #3314, opened 2026-03-10
pvode: Don't include bits/basic_string

bendudson:
Not included in Clang distributions. Does not seem to be needed.


dschwoerer (approved):
clang-tidy was apparently a bit eager.
Should we have one CI build with clang?

PR #3315, opened 2026-03-11
Consistent developer tooling; refactor formatting CI

ZedThree:
TL;DR:

CI will now fail on unformatted commits. Instead, run automatic formatting when you make commits:

```console
uv sync --only-dev         # Install consistent versions of clang-format, etc
uv tool run prek install   # Install pre-commit hooks
```

---

`clang-format` and `clang-tidy` are both essential tools for keeping the code clean and catching potential bugs, but the way they are currently run in CI is getting a bit annoying. Namely:

1. The last CI check ends up being the automated commit applying formatting, which always passes, and hides the actual test results
2. The formatting moves things around and the `clang-tidy` comments end up in the wrong places 
3. Adding clang-format commits to `.git-blame-ignore-revs` keeps git blame clean but makes merge conflicts much more likely, requiring more merges to fix them + another round of CI
4. Inconsistent versions of `clang-format` on dev machines and CI means formatting can flip-flop back and forth

This PR tries to fix these issues by:

- making the CI check fail if it has to do any formatting
- adding config so we can all get consistent versions of the tools locally and on CI
- adding pre-commit config so that formatting can run when making commits, meaning code should _always_ be formatted
- no automated commits, and commits always be correctly formatted means not having to add commits to `.git-blame-ignore-revs`

### DIFF
--- a/manual/sphinx/developer_docs/contributing.rst
+++ b/manual/sphinx/developer_docs/contributing.rst
@@ -142,45 +142,6 @@ then they will be able to checkout your branch::
 *Note*: If you have write access to the central BOUT-dev repository, you
 can push your branches there.
 
-Running Tests
-~~~~~~~~~~~~~
-
-We run many tests and checks automatically on GitHub for a variety of build
-configurations. See :ref:`sec-runtestsuite` for how to run them
-locally. Running the full test suite can take some time, but it's a good idea to
-run at least the unit tests locally when you're making changes.
-
-Install Pre-Commit Hooks (Optional but recommended)
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-Along with the automated tests, we also run things like formatters and static
-analysis. For frequent developers, we strongly recommend installing them locally
-and building them into your regular workflow. You can install the majority of
-our developer tools at once using `uv <https://docs.astral.sh/uv/>`_:
-
-.. code-block:: console
-
-    uv sync --only-dev
-
-This will create a new virtual environment (typically under ``.venv``) and
-install all the tools into that. You can then activate the virtual environment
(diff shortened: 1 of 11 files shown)


ZedThree:
I'm going to make a separate PR into this one applying all the formatting rules across everything -- turns out we have a lot of trailing whitespace

dschwoerer (dismissed):
Looks good, thanks 👍

dschwoerer:
> I'm going to make a separate PR into this one applying all the formatting rules across everything -- turns out we have a lot of trailing whitespace

Maybe we should first try to reduce the number of open PRs, to avoid to many conflicts?

ZedThree:
I've expanded the docs on this further to make things clearer about how things work.

I've also switched `clang-tidy-review` to post annotations instead of comments. We'll see how that fairs

ZedThree:
Just a heads up once this gets merged and people start using it: you can pass `git commit --no-verify` to temporarily skip the pre-commit hooks. This might be necessary if switching between branches without this branch merged

manual/sphinx/developer_docs/contributing.rst
```diff
@@ -212,6 +173,78 @@ give you comments to improve the code. You can make additional changes
...
+    cmake --build build --target check-unit-tests
+
+Along with the automated tests, we also run things like formatters and static
+analysis in CI, which may provide further feedback (see :ref:`sec-coding-style`
+for more details). If your PR fails the format check on CI, please install the
+formatters and run them according to the next section. Usually this is just:
+
+.. code-block:: console
+
+    git clang-format next
```
dschwoerer, line 195:
```suggestions
    git clang-format origin/next
```

`next` may be an old, stale local branch. But then, `origin` may also be there own fork, which may also be outdated ...

dschwoerer (approved):

PR #3316, opened 2026-03-11
Options: Convert Tensor<int> to Tensor<BoutReal> and similar

bendudson:
Read `Tensor<int>` from NetCDF files, and allow conversion to `Tensor<BoutReal>` and then Field3D.


ZedThree (changes requested):
Looks good! Please could you add a test :)

ZedThree (approved):
Nice, thanks @bendudson!

PR #3320, opened 2026-03-12
Field3DParallel

dschwoerer:
A field that can be used to compute parallel derivatives. For FCI that means, it has parallel slices, and will preserve them.

Any Field3D can be turned into a Field3DParallel, if it has parallel slices. If not, it will raise a runtime error, as that is not known at compile time.

This can be used to preserve parallel slices, by turning a Field3D into a Field3DParallel:
```
Field3D f = (f1.asField3DParallel() * f2) + f3;
```
`f` will have parallel slices. If f2 or f3 do not have parallel slices, this will throw an error.

Includes #3334


dschwoerer:
This needs https://github.com/boutproject/BOUT-dev/tree/par-metrics which needs https://github.com/boutproject/zoidberg/pull/60

include/bout/field2d.hxx
```diff
@@ -274,6 +274,9 @@ public:
 
   int size() const override { return nx * ny; }
 
+  Field2D& asField3DParallel() { return *this; }
```
bendudson, line 283: This looks potentially very confusing: surely `asField3DParallel` should return `Field3DParallel`, not `Field2D`?

dschwoerer: Indeed, it looks confusing. However, this is needed to write FA / FCI independent code.
Maybe we should rename `asField3DParallel()` to something along the lines of `.allowYDerivatives()`?

bendudson: Ah I see. Perhaps just a comment might be sufficient.

bendudson (dismissed):
Thanks @dschwoerer !

bendudson (approved):

PR #3323, opened 2026-03-12
Communicate `FieldPerp` consistently with other Fields

ZedThree:
Simpler but perhaps less elegant alternative to #3321 

- Add `Field::field_type()` virtual method
- Removes `Mesh::communicate(FieldPerp&)` overload

### DIFF
--- a/src/mesh/mesh.cxx
+++ b/src/mesh/mesh.cxx
@@ -376,38 +376,28 @@ void Mesh::communicate(FieldGroup& g) {
   }
 }
 
-/// This is a bit of a hack for now to get FieldPerp communications
-/// The FieldData class needs to be changed to accomodate FieldPerp objects
-void Mesh::communicate(FieldPerp& f) {
-  comm_handle recv[2];
-
-  int nin = xstart;              // Number of x points in inner guard cell
-  int nout = LocalNx - xend - 1; // Number of x points in outer guard cell
-
-  // Post receives for guard cell regions
-
-  recv[0] = irecvXIn(f[0], nin * LocalNz, 0);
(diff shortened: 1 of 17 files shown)


src/mesh/mesh.cxx
```diff
@@ -376,38 +376,28 @@ void Mesh::communicate(FieldGroup& g) {
...
+  const auto y_length = ylt - yge;
+  const auto z_length = LocalNz;
 
   /// Loop over variables
   for (const auto& var : var_list) {
-    if (var->is3D()) {
-      len += (xlt - xge) * (ylt - yge) * LocalNz * var->elementSize();
-    } else {
-      len += (xlt - xge) * (ylt - yge) * var->elementSize();
+    switch (var->field_type()) {
```
dschwoerer, line 391: We now have a few switches, do we need some check that if we ever add a new type, we will get an error, in case we forgot to add it? Or do we rely on compiler warnings for this?

dschwoerer (dismissed):
Thanks, looks good 👍

dschwoerer:
Now `is3D()` could be deprecated, and also be made non-virtual, as it only needs to check the field_type.

src/mesh/mesh.cxx
ZedThree: We will get a compiler warning for non-exhaustive `switch` statement. We should try and eliminate the last few warnings, then we can error on warnings in CI and catch it immediately

ZedThree:
> Now `is3D()` could be deprecated, and also be made non-virtual, as it only needs to check the field_type.

Good points!

dschwoerer (approved):

PR #3046, opened 2024-12-16
Utility for setting up coordinates for a tokamak system

tomc271:
Extracted duplicated code for creating a Coordinates object for a tokamak system to a new file.

PR #3326, opened 2026-03-13
Add cell volume and areas

dschwoerer:
They should make code more readable, and also potentially faster, as now cell_areas at the cell faces can be requested from the coordinates.

This should make the hermes-3 FA operator much more readable, and at the same time also FCI compatible, as the cell areas can be done correctly once for FCI, and then used transparently in user codes. @mikekryjak 

Todo:
- [x] add some tests
- ~~use in FA operators~~ (needs FA operators from hermes-3)


mikekryjak:
@dschwoerer excellent work, this has been on my wishlist for a very long time. The amount of places we calculate the areas and volumes is considerable, and so is the number of potential mistakes this generates.

I have some questions on what is actually being calculated here. I see there are areas for each side of X, Y and Z. What about parallel areas?

What is Jxz? Is that just J?

include/bout/coordinates.hxx
```diff
@@ -102,6 +102,149 @@ public:
...
+  const FieldMetric& g_22_ylow() const;
+  const FieldMetric& g_22_yhigh() const;
+  FieldMetric& g_22_ylow();
+  FieldMetric& g_22_yhigh();
+  /// get Jxz at the cell faces or cell centre
+  const FieldMetric& Jxz_ylow() const {
+    if (!_jxz_ylow.has_value()) {
+      _compute_Jxz_cell_faces();
+    }
+    return *_jxz_ylow;
```
ZedThree: I think this `clang-tidy` check is unfortunately not clever enough. If `_compute_Jxz_cell_faces()` returns the computed value, this can actually just be


```suggestion
  const FieldMetric& Jxz_ylow() const {
    return _jxz_low.value_or(_compute_Jxz_cell_faces());
```

include/bout/coordinates.hxx
```diff
@@ -102,6 +102,149 @@ public:
   /// Covariant metric tensor
   FieldMetric g_11, g_22, g_33, g_12, g_13, g_23;
 
+  /// get g_22 at the cell faces;
+  const FieldMetric& g_22_ylow() const;
+  const FieldMetric& g_22_yhigh() const;
+  FieldMetric& g_22_ylow();
+  FieldMetric& g_22_yhigh();
```
ZedThree, line 110: Probably don't need these overloads?

ZedThree (commented):
Looks great, thanks @dschwoerer! A couple of suggestions:

include/bout/coordinates.hxx
dschwoerer: We probably want to add the cell length, both in the cell centre, as well as at the faces, i.e. the distance between the cell centres.

That would replace `g_22` then, right now they are still needed.

dschwoerer:
> @dschwoerer excellent work, this has been on my wishlist for a very long time. The amount of places we calculate the areas and volumes is considerable, and so is the number of potential mistakes this generates.
> 
> I have some questions on what is actually being calculated here. I see there are areas for each side of X, Y and Z. What about parallel areas?

What do you mean by "parallel" areas? Parallel to what? Or for the parallel slices?

> What is Jxz? Is that just J?

No, that is the 2D jacobian in the x-z plane. Probably not needed anymore, we should use the areas directly.

include/bout/coordinates.hxx
dschwoerer: Unfortunately `_compute_Jxz_cell_faces()` computes 3 different quantities.
We could split up the functions to do that, but that would add even more boiler plate code.

include/bout/coordinates.hxx
ZedThree: Sorry, I meant the non-`const` ones!

dschwoerer: We need them for normalization. But I am starting to be worried, that that might be a bad idea.

Right now:
```
coords->g_22 *= norm;
coords->g_22_ylow() *= norm;
coords->g_22_yhigh() *= norm;
```
gives the correct thing for FCI, but wrong for FA.

```
coords->g_22_ylow() *= norm;
coords->g_22_yhigh() *= norm;
coords->g_22 *= norm;
```
or
```
coords->g_22 *= norm;
if (coord->g_22.isFci()) {
  coords->g_22_ylow() *= norm;
  coords->g_22_yhigh() *= norm;
}
``` 
works for both
and 
```
coords->g_22 *= norm;
coords->g_22_ylow();
coords->g_22_yhigh();
```
works only for FA.

So we probably want to remove write access to the fields, and let BOUT++ handle all of the normalisation. I think that is part of github.com/boutproject/BOUT-dev/pull/2873 or https://github.com/boutproject/BOUT-dev/pull/3046 - but that is 900 commits behind next and with its 500 comits has probably more merge conflicts that what I can deal with right now :-(

ZedThree: #3046 is into #2873 by the way.

We (@tomc271 and myself) can take care of bringing those PRs up-to-date -- can you review them? Starting with #3046

dschwoerer: I was waiting for the wrappers that make `coords->dy[i]` working again, so not all the code needs to be changed. I thought you still wanted to do that?

mikekryjak:
> > @dschwoerer excellent work, this has been on my wishlist for a very long time. The amount of places we calculate the areas and volumes is considerable, and so is the number of potential mistakes this generates.
> > I have some questions on what is actually being calculated here. I see there are areas for each side of X, Y and Z. What about parallel areas?
> 
> What do you mean by "parallel" areas? Parallel to what? Or for the parallel slices?
> 
> > What is Jxz? Is that just J?
> 
> No, that is the 2D jacobian in the x-z plane. Probably not needed anymore, we should use the areas directly.

I mean the cross-sectional area in the parallel direction. It's probably the most used area in the code due to many BCs being at the target, see e.g.:

https://github.com/boutproject/hermes-3/blob/66469077639e04cfa4a4299ce2cd082ff236a9e8/src/sheath_boundary_simple.cxx#L668-L670

As you noticed, it doesn't have any cross terms which may mean it needs updating for non-orthogonal grids.
I am also wondering how you calculate the angle between the field line and the ylow face in both poloidal and parallel, and whether we can add it to the list along with the areas and volumes. This would be needed for e.g. calculating heat fluxes perpendicular to the target as well as reflection angles for both fluid and kinetic neutral BCs.

totork:
@dschwoerer Does it make sense to add here another attributes like "has_bndry_ylow" and "has_bndry_yhigh", which is a Field3D of bools signaling if the cell has boundaries? These are quite important, because we have to disable the conduction at the sheaths.

dschwoerer:
@totork it would be possible, but I think this should go to #3308 
There you can say:
```
getYBoundary(coords, YBndryType::sheath)->contains<-1>(ind);
```

dschwoerer:
> > > @dschwoerer excellent work, this has been on my wishlist for a very long time. The amount of places we calculate the areas and volumes is considerable, and so is the number of potential mistakes this generates.
> > > I have some questions on what is actually being calculated here. I see there are areas for each side of X, Y and Z. What about parallel areas?
> > 
> > 
> > What do you mean by "parallel" areas? Parallel to what? Or for the parallel slices?
> > > What is Jxz? Is that just J?
> > 
> > 
> > No, that is the 2D jacobian in the x-z plane. Probably not needed anymore, we should use the areas directly.
> 
> I mean the cross-sectional area in the parallel direction. It's probably the most used area in the code due to many BCs being at the target, see e.g.:
> 
> https://github.com/boutproject/hermes-3/blob/66469077639e04cfa4a4299ce2cd082ff236a9e8/src/sheath_boundary_simple.cxx#L668-L670
> 
> As you noticed, it doesn't have any cross terms which may mean it needs updating for non-orthogonal grids.

That would be now `cell_area_yhigh()` and `cell_area_ylow()`.

> I am also wondering how you calculate the angle between the field line and the ylow face in both poloidal and parallel, and whether we can add it to the list along with the areas and volumes. This would be needed for e.g. calculating heat fluxes perpendicular to the target as well as reflection angles for both fluid and kinetic neutral BCs.

PRs welcome, although I think angles between targets and the magnetic field should go to #3308 - and need to be provided by the grid generator.

include/bout/coordinates.hxx
ZedThree: What if the `_compute_*` functions return something like:

```c++
struct _compute_result {
  FieldMetric low;
  FieldMetric high;
};
```

then this could be:

```c++
return _jxz_low.value_or(_compute_Jxz_cell_faces().low);
```

dschwoerer: I thought that would work, but that would cause the compute function to be always be called.

But this should work:
```
  const FieldMetric& Jxz_ylow() const {
    if (!_jxz_ylow.has_value()) {
      return _compute_Jxz_cell_faces().low;
    }
    return *_jxz_ylow;
```

The advantage of the initial version is, that it fails if the compute functions do not set the value. With this version that will always return a valid FieldMetric, but it might be recomputed every time. Imho, I find this worse then having a segfault in my code ...

dschwoerer: I added an `ASSERT2` before each dereference. That way the dereference is not unchecked anymore (for CHECK >= 2). Are you happy with that, @ZedThree?

dschwoerer:
@mikekryjak thought it might be better to put the documention about the meaning of the areas and volumes and there meaning in the hermes-3 repo, as hermes-3 only supports 2 or 3 coordinate systems, so it might be easier to document that.

He wanted to propose a draft, where I would help fill in the FCI specific bits, but that would be independent of this PR.

include/bout/coordinates.hxx
ZedThree: Ah, yes, we need C++23's `.or_else()` for the lazily evaluated version.

Yeah, the `ASSERT2` is fine, it just seems like a waste of `std::optional`'s checking is all! This should probably be a `FIXME` for when we move to C++23.

ZedThree:
Yes, the meanings of the areas in terms of coordinates are specific to the exact coordinate system used, which we can't really know anything about here. Documenting them in Hermes-3 makes sense, maybe with a link to them from here.

ZedThree (approved):
There's two outstanding things that I think should be fixed in the longer-term:

- use `std::optional::or_else` to make the setting cleaner, once we get C++23 (this feature actually only needs GCC 12)
- remove the non-`const&` overloads, which probably requires the coordinates refactor

PR #3330, opened 2026-03-16
Petsc matrix operators for FCI

bendudson:
Reads matrix operators from the mesh file in CSR format, creating PETSc matrices that apply to Field3Ds. Mainly intended to implement FCI parallel operators as an alternative to yup/down communication.

This enables:
1. **Arbitrary parallelisation** : PETSc handles the communication pattern, so there is no restriction on the field-line trajectories and processor boundaries. The mapping between mesh global indices and petsc global indices is handled automatically.
2. **Exact conservation of fluxes**: Using the Support Operator Method to derive the divergence operator from the transpose of the gradient operator (as in e.g. GRILLIX) ensures that the numerical integral of the divergence of a flux cancels to numerical precision.
3. **Better preconditioning**: The LU/MUMPS/STRUMPACK preconditioning used in the `snes` solver needs to know which cells are connected, so that PETSc can perform coloring and calculate the system Jacobian efficiently. Knowing the forward and backward matrices will enable this to be done.

There are a couple of downsides:
1. **Linearity**: The operators are fixed matrices so are linear. Monotonic splines, for example, require nonlinearity.
2. **More communications**: Each matrix-vector operation will perform communications. Perhaps there is a way to batch operations into some kind of composite matrix?


ZedThree:
This looks amazing, thanks @bendudson!

I've not finished going through this yet, but how much is this a replacement for `GlobalIndexer` and/or `PetscMatrix/PetscVector`? Or does it work together with them? It looks like this can only do `Field3D`, whereas `GlobalIndexer` can do the other fields too?

bendudson:
Thanks for going through this @ZedThree ! Sorry there's quite a lot, though most is self-contained.  The implementation is independent of `GlobalIndexer`, `PetscVector` and `PetscMatrix`. Perhaps there is a way to reuse that machinery, but the way these operators are created is quite specific: The mesh generator assigns global indices to cells, including y boundary cells,  and `PetscMapping` is calculating the mapping between the mesh global indices and PETSc's global indices that depend on the number of processors / domain decomposition. 

I've added an example of how these operators are used (https://github.com/boutproject/BOUT-dev/pull/3330/changes#diff-d24773f587abce8257e893cfcfec54002aafaf875fb9cf93f54076b5c60ec7abR16). There is some amount of setup needed to calculate differential operators from the forward / backward maps, that I've put into a helper function that returns a struct of operators (https://github.com/boutproject/BOUT-dev/pull/3330/changes#diff-62e947100b5e999ca6d6d57a6f1281449adc7e2db9e8b7fc5e9273fea887ad88R302). Not sure this is the right API; suggestions welcome!

ZedThree:
> The mesh generator assigns global indices to cells, including y boundary cells, and `PetscMapping` is calculating the mapping between the mesh global indices and PETSc's global indices that depend on the number of processors / domain decomposition.

I had thought that's essentially what `GlobalIndexer` was doing, so possibly I am misunderstanding either that or this! Maybe we can discuss at the next dev meeting?

bendudson:
> > The mesh generator assigns global indices to cells, including y boundary cells, and `PetscMapping` is calculating the mapping between the mesh global indices and PETSc's global indices that depend on the number of processors / domain decomposition.
> 
> I had thought that's essentially what `GlobalIndexer` was doing, so possibly I am misunderstanding either that or this! Maybe we can discuss at the next dev meeting?

The `GlobalIndexer` assigns a global index to each cell but doesn't map this to a mesh global index that is independent of the domain decompositions: Changing the number of processors will change the global indices. The key part of `PetscMapping` is the consistency of cell numbering in the mesh file, row and column indices in the CSR matrices for the operators.

ZedThree:
Ahhh, I see! Thanks!

mikekryjak:
@bendudson "arbitrary parallelisation" is quite a phrase! What exactly will it enable us to do?

dschwoerer:
> @bendudson "arbitrary parallelisation" is quite a phrase! What exactly will it enable us to do?

It is compatible with `x`, `y` and `z` splitting.

dschwoerer:
I noticed we run 56 test in the "static, optimized case", including the test-petsc-operator, while fedora runs 52 + 8 + 1 - and the `test-petsc-operator` is not among them. It does get however build, just not run.

The Fedora CI run has 3 additional tests disabled, according to the configure stage:
```
-- Not building test test-laplace-hypre3d, requirement not met: BOUT_HAS_HYPRE
-- Not building test test-laplacexy2-hypre, requirement not met: BOUT_HAS_HYPRE
-- Not building test test-slepc-solver, requirement not met: BOUT_HAS_SLEPC
```
So on the non-fedora CI should have 64 tests run - so 8 tests are missing.
Taking into account that `test-petsc-operator` is missing on fedora, that makes at least 9 missing tests.

Any idea why some tests are not run @ZedThree ?

ZedThree:
> Any idea why some tests are not run @ZedThree ?

Looks like the Fedora build runs `make -C build check`, but the other builds run `cd build && ctest`.

Here's the diff between what's run on the latest CI:

```diff
diff /tmp/static_optimised /tmp/fedora 
1,2d0
< MMS-bracket
< MMS-derivatives3
4,5d1
< MMS-shiftedmetricinterp
< MMS-upwinding3
11,16d6
< test-boutpp-collect
< test-boutpp-collect-staggered
< test-boutpp-legacy-model
< test-boutpp-mms-ddz
< test-boutpp-print
< test-boutpp-simple-model
37a28
> test-laplace-hypre3d
40a32
> test-laplacexy2-hypre
48a41
> test-slepc-solver
60a54
> test_beuler
```

So we've got the hypre and slepc tests which aren't running due to missing the dependencies, but also beuler, which only requires petsc, so not sure why that's skipped.

I think the others all require Python, which isn't found on the the static optimised build.

include/bout/field3d.hxx
```diff
@@ -372,6 +372,14 @@ public:
...
   };
 
+  /// Convert (x, y, z) to an Ind3D
+  /// Requires mesh size.
+  Ind3D indexAt(int x, int y, int z) const {
+    const int ny = this->getNy();
+    const int nz = this->getNz();
+    return Ind3D{(((x * ny) + y) * nz) + z, ny, nz};
+  }
+
```
dschwoerer, line 382: I have also added this in the past in many places. Is the Field the best place for this? Or should it go to to the mesh?

Would it be worth to have it for all fields?

include/bout/petsc_operators.hxx
```diff
@@ -0,0 +1,1082 @@
...
+  /// @tparam Function Callable with signature
+  ///         <tt>(PetscInt local_row, Ind3D i_ym) -> void</tt>.
+  /// @param func Callback invoked for each locally stored backward boundary cell.
+  template <typename Function>
+  void mapLocalYdown(Function func) const {
+    PetscInt row = evolving_region.size() + xin_region.size() + xout_region.size()
+                   + yup_region.size();
+    BOUT_FOR_SERIAL(i, ydown_region) {
+      func(row, i.ym());
+      ++row;
```
dschwoerer, line 289: More idiomatic C++:
```suggestion
      func(row++, i.ym());
```
If accepted, it also needs to be changed above.

dschwoerer (approved):
Looks good, thanks @bendudson!

include/bout/field3d.hxx
bendudson: Perhaps in Field, or adding to other fields e.g. Field2D::indexAt should return an Ind2D. It's definitely a useful utility; I think Hermes-3 has at least one like it.

PR #3331, opened 2026-03-17
Use pytest for integrated tests

tomc271:


tests/run_integrated_tests.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+PROJECT_ROOT=$(cd "$(dirname "$0")/.." && pwd)
+
+export PYTHONPATH="$PROJECT_ROOT/tools/pylib"
```
ZedThree: If we use `${PROJECT_BINARY_DIR}` here, then CMake's `configure_file` will expand this to the correct value

tests/run_integrated_tests.sh
```diff
@@ -0,0 +1,11 @@
...
+
+# Pre-build the project to prevent concurrent CMake race conditions
+cmake --build "$PROJECT_ROOT"
+
+pytest -m "not serial" --cache-clear -n auto --dist=loadgroup -q "$PROJECT_ROOT/tests/integrated" "$@"
+pytest -m serial "$PROJECT_ROOT/tests/integrated" "$@"
```
ZedThree: We should use `uv` to ensure we've got the test dependencies installed

pyproject.toml
```diff
@@ -4,6 +4,8 @@ requires = []
 # Defined by PEP 517:
 build-backend = "backend"
 backend-path = ["tools/pylib/_boutpp_build/"]
+[tool.pytest.ini_options]
+console_output_style = "plain"
 
 [dependency-groups]
```
ZedThree, line 11: Add `test` group to capture `pytest` and `pytest-xdist` dependencies (+ required versions)

CMakeLists.txt
```diff
@@ -766,7 +766,7 @@ if(BOUT_TESTS)
   endif()
 
   add_custom_target(
-    check-integrated-tests COMMAND ctest -R "test-" --output-on-failure
+    check-integrated-tests COMMAND pytest -k "test-"
```
ZedThree: Use the `run_integrated_tests.sh` script here

.ci_with_cmake.sh
```diff
@@ -13,7 +13,7 @@ fi
 
 cmake --build build --target build-check -j 2
 cd build
-ctest --output-on-failure --timeout 300
+cmake --build . --target check
```
ZedThree, line 16: Should we keep some kind of `timeout`?

tests/integrated/test-boutpp/slicing/test_slicing.py
```diff
@@ -0,0 +1,55 @@
...
+        try:
+            nout = ex(ndat)
+            fout = ex(field)
+            assert (
+                fout.shape == nout.shape
+            ), f"Field3D returned {{ fout.shape }} but numpy {{ nout.shape }}"
+            assert np.all(fout == nout), f"data mismatch, {{ fout == nout }}"
+        except:
+            print("Failed to test", inspect.getsource(ex))
+            raise
```
ZedThree, line 55: Probably don't want `try... except` around the `asserts`

tests/integrated/test-boutpp/slicing/test_slicing.py
```diff
@@ -0,0 +1,55 @@
...
+def test_slicing():
+
+    bc.init("-d test")
+
+    mesh = bc.Mesh.getGlobal()
+    field = bc.Field3D.fromMesh(mesh)
+    ndat = np.random.random(field.shape)
+    field[:] = ndat
+
+    assert np.all(ndat == field[:])
```
ZedThree, line 17: Would be better to use `numpy.testing.assert_allclose`

tests/integrated/test-boutpp/slicing/test_slicing.py
```diff
@@ -0,0 +1,55 @@
...
+    print(field.shape)
+    for ex in examples:
+        print("testing", inspect.getsource(ex))
+        try:
+            nout = ex(ndat)
+            fout = ex(field)
+            assert (
+                fout.shape == nout.shape
+            ), f"Field3D returned {{ fout.shape }} but numpy {{ nout.shape }}"
+            assert np.all(fout == nout), f"data mismatch, {{ fout == nout }}"
```
ZedThree, line 52: Prefer `numpy.testing.assert_allclose` -- that will catch both these `assert`s in fact

tests/integrated/test-command-args/test_command_args.py
```diff
@@ -52,46 +53,46 @@ def testNoArgumentsDefaultDirectory(self):
         self.makeDirAndCopyInput("data")
         launch_safe(self.command, pipe=True, nproc=1, mthread=1)
         self.assertTrue(
-            os.path.exists("data/BOUT.settings"),
+            Path.exists(Path("data/BOUT.settings")),
```
ZedThree, line 55:
```suggestion
            Path("data/BOUT.settings").exists(),
```

I think?

Ditto for instances below

tests/run_integrated_tests.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
```
ZedThree, line 3:
```suggestion
set -e
```

ZedThree:
@dschwoerer This PR now puts a hard requirement of `pytest` (and `pytest-xdist`) for running the tests. The dependencies are in `pyproject.toml` -- where's the best place to install these for the Fedora CI?

dschwoerer:
The best place is to put it here: https://github.com/boutproject/bout-container-base/pull/50

Is there a plan to add a test to ensure the examples can at least be compiled?

dschwoerer:
The builds are not failing, even though they should:
```
.....F..................F..............................
=================================== FAILURES ===================================
```

Also, do you know why for fedora everything but one test is skipped?

ZedThree:
@tomc271 is already on it :)

Fedora skips all but one because it's currently missing the `pytest-xdist` dependency. That one test is separate for some reason (I think the timeout, @tomc271 ?) and doesn't need that dependency

dschwoerer:
It should be there, as soon as the new container image has been build

manual/sphinx/developer_docs/contributing.rst
```diff
@@ -206,7 +206,7 @@ install the majority of our developer tools at once using `uv
 
 .. code-block:: console
 
-    uv sync --only-dev --inexact
+    uv sync --only-dev --group pytest --inexact
```
ZedThree: Not sure this is the correct place for this, probably in the section above on tests?

tests/integrated/test-fieldgroupComm/test_fieldgroupComm.py
```diff
@@ -0,0 +1,60 @@
...
+
+#
+# Run the test, compare results
+#
+
+# Requires: all_tests
+# Requires: netcdf
+# Requires: not metric_3d
+# Cores: 4
+
```
dschwoerer: I doubt the current test system still parses and process them, so the can be removed?

dschwoerer (changes requested):
It seems one of the MMS tests is missing. That certainly needs fixing.

I find it also increadibly hard to review, as in some case the rename from runtest to test_* is not rendered correctly on github. In some cases this seems to be done such that git itself is not aware of the the rename, e.g. `test-fci-mpi/runtest`.

This makes it even harder to figure out what was changed, as you clearly did not write the whole test new.

I have reviewed `test-fci-mpi/runtest` more carefully, and it is doing only part what is in next. It is skipping the iteration over the different test cases:
https://github.com/boutproject/BOUT-dev/blob/next/tests/integrated/test-fci-mpi/runtest#L49

Also, it would only output the errors after the assert:
https://github.com/boutproject/BOUT-dev/blob/7f920ac02550edf26ad613a62dc0d7627229fa7f/tests/integrated/test-fci-mpi/test_fci_mpi.py#L72

So this is never printed.

dschwoerer:
I forgot - the new pytest is much faster, which is nice, but I am wondering this comes from skipping tests.

.github/workflows/tests.yml
```diff
@@ -133,39 +133,48 @@ jobs:
...
-      - name: Install pip packages
+      - name: Setup virtual environment with uv
         run: |
-          python -m pip install --upgrade pip setuptools
-          python -m pip install -r requirements.txt
+          # uv sync automatically creates a .venv and installs the groups
+          uv sync --no-dev --group pytest --inexact
+
+          # Install the core requirements into the same venv
+          uv pip install -r requirements.txt
```
ZedThree, line 175: Orthogonal PR -> this file should be merged into `pyprojects.toml`

tests/integrated/test-boutpp/collect-staggered/test_boutpp_collect_staggered.py
```diff
@@ -0,0 +1,54 @@
...
+    bc.writeDefaultOutputFile(dump)
+
+    fc = bc.create3D("sin(y)", outloc="CENTRE")
+
+    fe = bc.Field3D.fromCollect("f3d_evolve", path="data", info=False)
+    fo = bc.Field3D.fromCollect("f3d_once", path="data", info=False)
+
+    if fe.getLocation() == fc.getLocation():
+        print("The loaded field should not be at CELL_CENTRE")
+        fail = 1
```
ZedThree, line 28: This could be
```suggestion
    assert fe.getLocation() != fc.getLocation(), "The loaded field should not be at CELL_CENTRE"
```

tests/integrated/test-boutpp/collect-staggered/test_boutpp_collect_staggered.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def compare(a, b):
+        diff = a - bc.interp_to(b, a.getLocation())
+        err = bc.max(bc.sqrt(diff * diff))
+        return err
+
+    if compare(fc, f) > 1e-3:
+        print("Something is wrong. Maybe interpolation is broken")
+        fail = 1
```
ZedThree, line 41: These could also be

```suggestion
    assert compare(fc, f) <= 1e-3, "Maybe interpolation is broken"
```

tests/integrated/test-cyclic/test_cyclic.py
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env python3
+
+#
+# Run the test, check it completed successfully
+#
+
+# Requires: netcdf
+# Cores: 4
```
ZedThree: We can delete all of these `# Requires` and `# Cores` comments now too

tests/integrated/test-cyclic/test_cyclic.py
```diff
@@ -0,0 +1,48 @@
...
+                print("PASSED")
+            else:
+                print("FAILED")
+                code = 1
+            assert status, f"Test failed for flag={f}"
+
+    if code == 0:
+        print(" => All cyclic reduction tests passed")
+    else:
+        print(" => Some failed tests")
```
ZedThree, line 45:
```suggestion
```

tests/integrated/test-cyclic/test_cyclic.py
```diff
@@ -0,0 +1,48 @@
...
+                f.write(out)
+
+            r = r + 1
+
+            # Find out if it worked
+            if status:
+                print("PASSED")
+            else:
+                print("FAILED")
+                code = 1
```
ZedThree, line 39:
```suggestion
```

tests/integrated/test-cyclic/test_cyclic.py
```diff
@@ -0,0 +1,48 @@
...
+
+def test_cyclic():
+
+    code = 0  # Return code
+    for nproc in [1, 2, 4]:
+        cmd = "./test_cyclic"
+
+        print("   %d processors...." % (nproc))
+        r = 0
+        for f in flags:
```
ZedThree, line 22: This could be turned into a parameterised test

tests/integrated/test-fci-boundary/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-bout_add_mms_test(
```
ZedThree, line 1: This is surprising! Thanks for catching this!

tests/integrated/test-fci-boundary/test_fci_boundary.py
```diff
@@ -0,0 +1,76 @@
...
+import os
+import pytest
+
+from boututils.run_wrapper import launch_safe
+from boututils.datafile import DataFile
+from boutdata.collect import collect as _collect
+
+from numpy.testing import assert_allclose
+
+if not os.path.exists(os.path.join(os.path.dirname(__file__), "grid.fci.nc")):
```
ZedThree, line 15: Can use `pathlib` here

tests/integrated/test-fci-boundary/test_fci_boundary.py
```diff
@@ -0,0 +1,76 @@
...
+        cmd = "./get_par_bndry"
+        _, out = launch_safe(cmd, nproc=nproc, mthread=mthread, pipe=True)
+
+    for k, v in regions.items():
+        data = collect(f"field_{k}")
+        assert_allclose(data, v)
+    for i in range(-1, 2):
+        name = f"ybndry_{i}"
+        data = collect(name)
+        assert_allclose(bndrys[name], data)
```
ZedThree, line 76: This should be inside the `nprocs` loop, currently it's only testing the last iteration

tests/integrated/test-fci-mpi/test_fci_mpi.py
```diff
@@ -0,0 +1,78 @@
...
+            try:
+                npt.assert_allclose(val, collect(name, **COLLECT_KW))
+            except AssertionError as e:
+                failures.append((nxpe, nype, name, e))
+
+        return failures
+
+    failures = []
+
+    for implementation in ["hermitespline", "monotonichermitespline"]:
```
ZedThree, line 47: Could parameterise the test over `implementation`

tests/integrated/test-fci-mpi/test_fci_mpi.py
```diff
@@ -0,0 +1,78 @@
...
+                failures.extend(failures_)
+
+        success = len(failures) == 0
+
+        assert success, "\nSome tests failed:"
+
+        if not success:
+            for nxpe, nype, name, error in failures:
+                print("----------")
+                print(f"case {nxpe=} {nype=} {name=}\n{error}")
```
ZedThree: Two things:

1. we collect and display all the failures rather than erroring on the first one -- the nicest way to do this would be with a fixture + parameterisation, but that's a fair amount of effort, so let's put a pin in that for now
2. if the assert fires, we leave the function, so won't see the list of failures (actually, we probably don't want to `print` in tests anyway), so this should be moved into the assert message

tests/integrated/test-griddata-yboundary-guards/test_griddata_yboundary_guards.py
```diff
@@ -0,0 +1,138 @@
...
+    )
+
+    s, out = launch_safe(f"./test_griddata -d {datadir}", nproc=nproc, pipe=True)
+
+    with open(f"run.log.{topology}.{n_yguards}.{nproc}", "w") as f:
+        f.write(out)
+
+    testfield = collect("test", path=str(datadir), info=False, yguards=True)
+
+    if topology == "doublenull":
```
ZedThree, line 99: A "cleaner" way of achieving this might be to have all the previous bits as a fixture, then two separate tests for the topologies

tests/integrated/test-griddata/test_griddata.py
```diff
@@ -0,0 +1,38 @@
...
+
+        rwidth = 0.4
+        dr = float(rwidth) / nx
+
+        # Test value of dx
+        npt.assert_allclose(
+            dx, dr * Bpxy * Rxy, atol=1e-7, err_msg="Failed: dx does not match"
+        )
+
+        print("Passed")
```
ZedThree, line 38:
```suggestion
```

tests/integrated/test-gyro/test_gyro.py
```diff
@@ -0,0 +1,66 @@
...
+            v, path=source_data_dir, prefix="benchmark", info=False, xguards=False
+        )
+    return bmk
+
+
+def test_gyro(benchmark_data):
+
+    print("Running Gyro-average inversion test")
+
+    for nproc in [1, 2, 4]:
```
ZedThree, line 36: Could be parameterised

tests/integrated/test-interchange-instability/test_interchange_instability.py
```diff
@@ -0,0 +1,89 @@
...
+        # Calculate RMS value in Z
+        Nirms = np.zeros(nt)
+        for t in np.arange(nt):
+            Nirms[t] = np.sqrt(np.mean(Ni[t, 0, 0, :] ** 2))
+
+        # Get the growth rate
+        ln = np.log(Nirms)
+        return (ln[nt - 5] - 8.0 * ln[nt - 4] + 8.0 * ln[nt - 2] - ln[nt - 1]) / (
+            6.0 * (tarr[nt - 2] - tarr[nt - 4])
+        )
```
ZedThree, line 48: I guess this is really more like a fixture, then we have two separate tests here?

tests/integrated/test-invpar/test_invpar.py
```diff
@@ -0,0 +1,62 @@
...
+            if s == 0:
+                print("PASSED")
+            else:
+                print("FAILED")
+            assert s == 0, f"Test failed for flag={f}"
+
+    if code == 0:
+        print(" => All inversion tests passed")
+    else:
+        print(" => Some failed tests")
```
ZedThree:
```suggestion
```

tests/integrated/test-laplace-hypre3d/test_laplace_hypre3d.py
```diff
@@ -0,0 +1,34 @@
...
+
+        error_max = collect("error_max", path=directory, info=False)
+
+        if error_max > tolerance:
+            print(directory + " failed with maximum error {}".format(error_max))
+            success = False
+        else:
+            print(directory + " passed with maximum error {}".format(error_max))
+
+    assert success
```
ZedThree, line 32:
```suggestion
        assert error_max <= tolerance
```

tests/integrated/test-laplace-hypre3d/test_laplace_hypre3d.py
```diff
@@ -0,0 +1,34 @@
...
+    ("data_circular_core-sol", 1),
+]
+
+tolerance = 1.0e-6
+
+
+def test_laplace_hypre3d():
+
+    success = True
+    for directory, nproc in test_directories:
```
ZedThree, line 19: Could be parameterised

tests/integrated/test-laplace-petsc3d/test_laplace_petsc3d.py
```diff
@@ -0,0 +1,45 @@
...
+]
+
+tolerance = 1.0e-6
+
+
+def test_laplace_petsc3d():
+
+    errors = {}
+
+    for directory, nproc in test_directories:
```
ZedThree, line 21: Could be parameterised

tests/integrated/test-laplace-petsc3d/test_laplace_petsc3d.py
```diff
@@ -0,0 +1,45 @@
...
+            errors[directory] = error_max
+
+    print("**********")
+
+    if errors:
+        print("Some failures:")
+        for name, error in errors.items():
+            print(f"{name}, max error: {error}")
+
+    assert not errors
```
ZedThree, line 43:
```suggestion
        assert status == 0

        error_max = collect("error_max", path=directory, info=False)
        assert error_max <= tolerance
```

tests/integrated/test-laplacexy/test_laplacexy.py
```diff
@@ -52,53 +52,41 @@
...
+    shell(["rm -f data/BOUT.dmp.*.nc"])
 
-            shell("rm data/BOUT.dmp.*.nc > /dev/null 2>&1")
+    if "hypre" in args:
+        s, out = launch(cmd, nproc=nproc, pipe=True, verbose=True)
+        if s == 134:
+            # PETSc did not recognise pctype option, probably means it
+            # was not compiled with hypre, so skip tests that need
+            # hypre to converge
+            pytest.skip("hypre not available as pre-conditioner in PETSc. Skipping...")
```
ZedThree, line 71: Nice!

tests/integrated/test-naulin-laplace/runtest_unsheared
ZedThree, line 1: Not sure this one has been converted?

tests/integrated/test-restarting/test_restarting.py
```diff
@@ -0,0 +1,75 @@
...
+        print("Fail: Field2D field has wrong shape")
+        success = False
+
+    if not np.allclose(f3d_1, f3d_0, atol=tolerance):
+        print("Fail: Field3D values differ")
+        success = False
+
+    if not np.allclose(f2d_1, f2d_0, atol=tolerance):
+        print("Fail: Field2D values differ")
+        success = False
```
ZedThree, line 47: Could be using `npt.assert_allclose` here

tests/integrated/test-restarting/test_restarting.py
```diff
@@ -0,0 +1,75 @@
...
+
+    ###########################################
+    # Test restart
+
+    print("-> Testing restart")
+
+    shell(["rm -f data/BOUT.dmp.0.nc"])
+    s, out = launch_safe("./test_restarting solver:nout=5", nproc=1, pipe=True)
+    s, out = launch_safe("./test_restarting solver:nout=5 restart", nproc=1, pipe=True)
```
ZedThree, line 56: This (and below) should be a separate test function

tests/integrated/test-squash/test_squash.py
```diff
@@ -79,51 +79,55 @@ def verify(f1, f2):
...
-timed_shell_safe("rm -f f2.nc")
-timed_launch_safe("./squash -q -q -q solver:nout=2", nproc=4, mthread=1)
-timed_shell_safe("{} -qdcl 9 data --outputname ../f2.nc".format(bout_squashoutput))
+    print("Parallel test")
+    timed_shell_safe("rm -f f2.nc")
+    timed_launch_safe("./squash -q -q -q solver:nout=2", nproc=4, mthread=1)
+    timed_shell_safe("{} -qdcl 9 data --outputname ../f2.nc".format(bout_squashoutput))
 
-verify("f1.nc", "f2.nc")
+    verify("f1.nc", "f2.nc")
```
ZedThree, line 92: I think each snippet upto a `verify` is really its own test? Then it would be possible to run them all in parallel with `pytest-xdist`
@dschwoerer thoughts?

tests/integrated/test-stopCheck-file/test_stopCheck_file.py
```diff
@@ -0,0 +1,49 @@
...
+    check_values = [True, False]
+    expected_steps = [1, 11]
+    data_dirs = ["data", "dataSecond"]
+    stop_files = ["BOUT.stop", "otherStop.check"]
+
+    print("Running stopCheck test")
+    success = True
+
+    for data_dir, stop_file in zip(data_dirs, stop_files):
+        for check, expected in zip(check_values, expected_steps):
```
ZedThree, line 24: Could be parameterised

tests/integrated/test-naulin-laplace/runtest_multiple_grids
ZedThree, line 1: Not sure this has been converted?

pyproject.toml
```diff
@@ -19,6 +22,10 @@ dev = [
      "sphinx-lint~=1.0",
      "sync-with-uv~=0.5.0",
 ]
+pytest = [
+     "pytest~=8.4",
+     "pytest-xdist~=3.8",
+]
```
ZedThree, line 28: Can we add the other test dependencies here too? `numpy`, `scipy`, `zoidberg`, etc? And perhaps rename:


```suggestion
test = [
     "pytest~=8.4",
     "pytest-xdist~=3.8",
]
```

.ci_with_cmake.sh
ZedThree: This is marked as resolved, but I don't see how? Am I missing something?

tests/integrated/test-yupdown/test_yupdown.py
```diff
@@ -0,0 +1,42 @@
...
+
+    diff = max(abs(ddy - ddy_check))
+
+    print(f"shifttype: {shift_type} Max difference {diff}")
+    if diff < 2e-5:
+        return True
+    return False
+
+
+@pytest.mark.parametrize("shift_type, variable", list(zip(shift_types, variables)))
```
ZedThree: I think this is wrong? This runs the following cases:

- `"shifted"` + `("ddy", "ddy_check")`
- `"shiftedinterp"` + `("ddy2", "ddy_check")`

but we want to run once for each of the `shift_types`, and check all of `variables` for each run

ZedThree (changes requested):
There's a few things that haven't been ported exactly right, and there's still quite a few patterns that I would like to update, but they could be done in a separate PR so we can get the bulk of the changes in sooner. Maybe just leave all the "could be..." for now, and concentrate on fixing the other comments

I noticed some particular patterns that we could improve upon in a second PR:

Could have a helper function to simplify this pattern:
```py
s, out = launch_safe(...)
with open("run.log", "w") as f:
   f.write(out)
```

We have quite a few tests that do something vaguely like this:

```py
# some setup
launch(cmd, ...)

success = True

for v in vars:
   r = collect(v)
   success &= r < tol

assert success
```
where we run an executable, and want to check all the variables before failing, rather than just failing on the first bad one.
This pattern is maybe best transformed into a fixture, and then the test is parameterised over the variables:

```py
@fixture(scope="module")
def fixture():
  # some setup
  launch(cmd, ...)

@parametrize(variables)
def test(variables):
   assert_allclose(r, expected)
```
then we only actually run the code once, but can parallelise over the variables. I think there's a way to parameterise fixtures too, so this could also run once e.g. per `nproc`.

dschwoerer:
I think this should be merged quickly, as it it seems like a great improvement, as it makes the tests faster.

I am however worried that there are regressions. @ZedThree did you compare test by test, to make sure this (still) matches what is in next? It is a bit difficult to get a feeling of the state of this, with so little comments from @tomc271 

I would be willing to review at least part of this, but only if you think this is close to being merged, so the review does not get outdated.

ZedThree:
@dschwoerer Yeah, I went through one by one, comparing the `runtest` to the `test_*.py`. I think I only caught one actual regression (`test_yupdown`). It's probably wise to have a second set of eyes check things too.

@tomc271 Can you comment when you've fixed `test_yupdown`, and we'll merge then. Everything else can wait for a second PR

.ci_with_cmake.sh
tomc271: I think it was this commit: 
https://github.com/boutproject/BOUT-dev/commit/e05bd5183a43950ac3391949c1904873c3e1f166

tests/integrated/test-squash/test_squash.py
dschwoerer: Yes, it should be possible to run in parallel, assuming you have different names, and the `rm` also needs to be modified.
The `f1.nc` needs to be squashed first, for comparison, but after that they should all be independent.

dschwoerer (dismissed):
I did not see any major regressions, there are several tests where the output on failure is worse than it used to, because it already fails on the first failure.

I have pushed a few fixes. That was easier then turning them into comments. Feel free to revert, if they seem / are wrong.

One thing that is worrying me, ist that something like this:
`@pytest.mark.parametrize("nproc", [1, 2, 3, 4])` will cause race conditions - because they run the simulations in the same folder.
I guess that applies to most parametrisation - they will cause race conditions if the simulation is in the same folder.
I am fine with merging it now, if that will be resolved soonish, so that we do not need to check all the tests again. But that means we may not be testing what we want to test, and it might also fail, because the dump files are still written for a different simulation.

tests/integrated/test-cyclic/test_cyclic.py
dschwoerer: I removed all but these:
	modified:   tests/MMS/advection/runtest
	modified:   tests/MMS/diffusion2/runtest
	modified:   tests/MMS/elm-pb/runtest.broken
	modified:   tests/MMS/fieldalign/runtest.broken
	modified:   tests/MMS/hw/runtest
	modified:   tests/MMS/laplace/runtest
	modified:   tests/MMS/spatial/advection/runtest
	modified:   tests/MMS/spatial/d2dx2/runtest
	modified:   tests/MMS/spatial/d2dz2/runtest
	modified:   tests/MMS/spatial/diffusion/runtest
	modified:   tests/MMS/time/runtest
	modified:   tests/MMS/tokamak/runtest.broken
They are not ruff clean, so I cannot commit without fixing.

dschwoerer (dismissed):

dschwoerer (dismissed):

ZedThree:
> One thing that is worrying me, ist that something like this:
> `@pytest.mark.parametrize("nproc", [1, 2, 3, 4])` will cause race conditions - because they run the simulations in the same folder.

There's a fixture `copy_and_cwd_to_unique_tmp_dir` which is used automatically (and implicitly) and takes care of running each test case in a unique tmpdir, so there shouldn't be any race conditions

dschwoerer:
Amazing :heart: 
Thanks @tomc271!

tomc271:
> @tomc271 Can you comment when you've fixed `test_yupdown`, and we'll merge then. Everything else can wait for a second PR

@ZedThree: test_yupdown is now fixed

ZedThree (approved):
Thanks @tomc271 and @dschwoerer!

PR #3334, opened 2026-03-19
Fix unused argument warning

dschwoerer:
This needs https://github.com/boutproject/zoidberg/pull/60 - thus it explicitly checks out that branch. If https://github.com/boutproject/zoidberg/pull/60 gets merged, this needs to be changed here again.


dschwoerer:
This has been mostly merged as part of #3320

It still contains one fix for a warning in 2D metrics, so I keep this open.

ZedThree (approved):

PR #3337, opened 2026-03-23
po: Add Georgian language

EkaterinePapava:


dschwoerer (approved):
Thanks!

locale/ka/libbout.po
```diff
@@ -0,0 +1,1030 @@
...
+
+#: ../src/mesh/impls/bout/boutmesh.cxx:300
+msgid "\t -> Good value\n"
+msgstr "\t -> კარგი მნიშვნელობა\n"
+
+#: ../src/mesh/impls/bout/boutmesh.cxx:171
+#, c++-format
+msgid ""
+"\t -> Leg region jyseps1_1+1 ({:d}) must be a multiple of MYSUB ({:d})\n"
+msgstr ""
```
bendudson, line 55: If a string is empty will the original be used?

bendudson: Ah ok according to Google yes: An empty `msgstr` means untranslated so the original is used. Thanks for this contribution!

bendudson (approved):
Thanks @EkaterinePapava !

bendudson:
Replaced by #3400

PR #3338, opened 2026-03-23
Remove using enum statements for C++17 compatibility

Steven-Roberts:
`using enum` statements require C++20, so I was getting compilation errors with an old compiler.


bendudson:
Hi @Steven-Roberts ! Do you need to use an old compiler? The C++ version is set to C++20 in CMakeLists: https://github.com/boutproject/BOUT-dev/blob/next/CMakeLists.txt#L533

Steven-Roberts:
No, I can use a newer one. The README says C++17 can be used, though
https://github.com/boutproject/BOUT-dev/blob/8ab455275dcb8d036f84edce13197e68810343f6/README.md?plain=1#L65

bendudson:
> No, I can use a newer one. The README says C++17 can be used, though
> 
> https://github.com/boutproject/BOUT-dev/blob/8ab455275dcb8d036f84edce13197e68810343f6/README.md?plain=1#L65

Ah sorry! We should fix the README.

PR #3360, opened 2026-04-24
Add utility for setting up tokamak coordinate system

ZedThree:
Replaces #3046 (rebased onto `next` and refactored)

- Pull out common code across the various examples and tests that use tokamak
  coordinates
- Reads grid quantities, applies normalisations, and calculates the metric
  components in one function call
- Returns a struct with the grid quantities, as some models use them further
- Some rearrangement of the existing code was sometimes necessary as we now need
  the normalisation factors before reading the grid quantities


dschwoerer (approved):
LGTM!

PR #3365, opened 2026-04-24
Field setBoundary and applyBoundary use options

bendudson:
When a boundary value uses options then these were only parsed on processors with boundaries. This uses a dummy boundary region to force all processors to parse options.

This should fix the problem that means unused options checking needs to be disabled in many transport simulations.

### DIFF
--- a/src/field/field3d.cxx
+++ b/src/field/field3d.cxx
@@ -4,7 +4,7 @@
  * Class for 3D X-Y-Z scalar fields
  *
  **************************************************************************
- * Copyright 2010 - 2025 BOUT++ developers
+ * Copyright 2010 - 2026 BOUT++ developers
  *
  * Contact: Ben Dudson, dudson2@llnl.gov
  *
@@ -36,6 +36,7 @@
 #include <cstddef>
 #include <memory>
 #include <optional>
+#include <string>
 #include <utility>
 
 #include "bout/parallel_boundary_op.hxx"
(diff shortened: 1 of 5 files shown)


dschwoerer:
Should that be handled by setting the options as conditionally used?

bendudson:
> Should that be handled by setting the options as conditionally used?

Hi @dschwoerer ! The issue is that some processors have no boundaries so options are not used on those processors. 
The unused options then kill the simulation. If the boundary condition depends on expressions then those expression options should be parsed and recursively marked used. 

One possibility would be to have the unused options code coordinate across processors. This is nontrivial because some processors could use one set of options and another set of processors use a different set of options.

dschwoerer:
> > Should that be handled by setting the options as conditionally used?
> 
> Hi @dschwoerer ! The issue is that some processors have no boundaries so options are not used on those processors. The unused options then kill the simulation. If the boundary condition depends on expressions then those expression options should be parsed and recursively marked used.

We might want to have the conditionally used part to do that recursive anyway, but that is probably a bit more involved.

> One possibility would be to have the unused options code coordinate across processors. This is nontrivial because some processors could use one set of options and another set of processors use a different set of options.

Then the right thing would be to only fail if something is not used anywhere. The current approach feels more like a workaround, then a fix. It would be much nicer to have the state synced to all procs. But that is indeed a bit more complicated, but should be doable. It is only done twice in the simulation, and it should not be a huge amount of data that we have to communicate, even if we just send the whole list that is used on one proc to all other procs.

bendudson:
Hi @dschwoerer Yes I agree that the right solution is to coordinate across processors, and only raise an exception if an option is not used on any processor. I'll close this PR and create an issue.

Issue #3366, opened 2026-04-28
checkForUnusedOptions should gather across MPI ranks


bendudson:
The `checkForUnusedOptions` throws an exception when options are unused. This incorrectly throws when expressions are used to set boundary conditions in multi-processor runs e.g. `bndry_core = dirichlet(core_ne)` the `core_ne` option is used on processors that have a core boundary, but not on other processors.

PR #3365 was an attempt to work around this by adding zero-width boundaries to all processors, but that doesn't work and would be fragile. 

Since `checkForUnusedOptions` is not performance-critical, we can afford to do some kind of reduction over all processors.

ZedThree:
I feel like this has come up before and we had a fix already, but for the life of me I can't remember what it is.

bendudson:
Hey @ZedThree ! There is `setConditionallyUsed` but that doesn't parse the option so
```
[var]
bndry_core = dirichlet(1.0)
```
will work if `bndry_core` is set as conditionally used, but

```
some_value = 1.0
[var]
bndry_core = dirichlet(some_value)
```
The `some_value` option won't be marked used. One possibility would be to modify `setConditionallyUsed` so it recursively marks expressions as used. I think a better solution is to do a reduction over all processors (as @dschwoerer suggested).

bendudson:
Closed by #3367

PR #3367, opened 2026-04-29
Global unused options

bendudson:
Only throw exception on unused option if it's not used on any MPI rank.

Fixes issue #3366


dschwoerer (dismissed):
Thanks @bendudson, that looks good!

It might be sufficient to use `Gatherv` rather then `Allgatherv` - but that is just a nitpick.
It would be sufficient to only check on node0, as that is the only one that writes it out anyway.

src/sys/options.cxx
```diff
@@ -1137,13 +1137,110 @@ void checkForUnusedOptions() {
...
+///   1. MPI_Allgather the per-processor serialised key lists so every rank
+///      knows the full union of locally-unused keys.
+///   2. For each key in that union, MPI_Allreduce with MPI_PROD over a flag
+///      that is 1 if the key is locally unused and 0 if it was used.  A
+///      product of 1 means every rank left it unused.
+///
+/// Keys are serialised as a newline-separated string.  Newlines are safe as a
+/// separator because option keys use ':' as their only structural character.
+std::set<std::string> getGlobalUnusedSet(std::vector<std::string> local_unused_keys) {
+  MPI_Comm comm = BoutComm::get();
```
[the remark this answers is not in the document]
ZedThree: This is supposed to be ignored, but it turns out we've got duplicate `misc-include-cleaner.IgnoreHeaders` keys in `.clang-tidy` :(

src/sys/options.cxx
```diff
@@ -1137,13 +1137,110 @@ void checkForUnusedOptions() {
...
+
+  // Gather the byte-lengths from all processors so we can allocate the
+  // receive buffer and build the displacement array for MPI_Allgatherv.
+  std::vector<int> all_lens(nprocs);
+  MPI_Allgather(&local_len, 1, MPI_INT, all_lens.data(), 1, MPI_INT, comm);
+
+  std::vector<int> displs(nprocs, 0);
+  for (int i = 1; i < nprocs; ++i) {
+    displs[i] = displs[i - 1] + all_lens[i - 1];
+  }
```
ZedThree, line 1178: I think this is `std::partial_sum`?

src/sys/options.cxx
```diff
@@ -1137,13 +1137,110 @@ void checkForUnusedOptions() {
...
+///   2. For each key in that union, MPI_Allreduce with MPI_PROD over a flag
+///      that is 1 if the key is locally unused and 0 if it was used.  A
+///      product of 1 means every rank left it unused.
+///
+/// Keys are serialised as a newline-separated string.  Newlines are safe as a
+/// separator because option keys use ':' as their only structural character.
+std::set<std::string> getGlobalUnusedSet(std::vector<std::string> local_unused_keys) {
+  MPI_Comm comm = BoutComm::get();
+  int nprocs{};
+  MPI_Comm_size(comm, &nprocs);
```
ZedThree: I think we can just use this?
```suggestion
  const int nprocs = BoutComm::size();
```

ZedThree (dismissed):
LGTM. There's a couple of bits that could use existing functions, but very minor

src/sys/options.cxx
bendudson: It's close but `displs` should start with a zero. It might be `exclusive_scan` in C++17 https://en.cppreference.com/cpp/algorithm/exclusive_scan

PR #3372, opened 2026-05-11
Reduce duplication in `Coordinates` constructors

ZedThree:
We had a lot of duplicated complicated code in the `Coordinates` initialisation, this attempts to reduce it. This should make the refactor in #2873 smaller and easier to review as well


src/mesh/coordinates.cxx
```diff
@@ -138,52 +154,56 @@ Field2D interpolateAndExtrapolate(const Field2D& f, CELL_LOC location, bool extr
...
 #endif
-
-  return result;
 }
 
-#if BOUT_USE_METRIC_3D
-Field3D interpolateAndExtrapolate(const Field3D& f_, CELL_LOC location,
-                                  bool extrapolate_x, bool extrapolate_y,
-                                  bool no_extra_interpolate, ParallelTransform* pt_) {
+/// Interpolate a Field2D to a new CELL_LOC with interp_to.
```
dschwoerer:
```suggestion
/// Interpolate a Field to a new CELL_LOC with interp_to.
```

dschwoerer (changes requested):
I would prefer to revert the last two commits, as it requires as reading the fields twice, so it is imho not that much cleaner then the old design.

ZedThree:
It is a little annoying, but note that we don't actually read twice -- at worst, the data will be read on the first `mesh->get` call, and we're just paying for an additional copy. And this is only done for two fields during setup, so this shouldn't really affect performance.

Mostly I am now determined that `Mesh::get` has reached the end of its life and we need to replace it

ZedThree:
@dschwoerer I've reverted those commits

dschwoerer (approved):

PR #3376, opened 2026-05-21
Add remaining topology documentation (images + supported_topologies)

wearysebas:
Some of the topology documentation was lost at some point, this includes all of the missing files.


dschwoerer:
I feel like it would be worth that any of the mentioned restrictions apply only to FA grids.
We should add a few references to FCI, where we have no restrictions in topology, maximum or minimum number of x-points, o-points, chaos as well as minimum number of cores.

PR #3382, opened 2026-05-29
Refactor PETSc preconditioner for CVODE

bendudson:
Refactors the Jacobian coloring preconditioner and makes it available as a CVODE preconditioner.

test2 from scratch. 10 steps of size 1.

No preconditioning:
```
CVODE: nsteps 803, nfevals 967, nniters 966, npevals 0, nliters 4132
    -> Newton iterations per step: 1.202989e+00
    -> Linear iterations per Newton iteration: 4.277433e+00
    -> Preconditioner evaluations per Newton: 0.000000e+00
    -> Last step size: 2.046315e-02, order: 2
    -> Local error fails: 0, nonlinear convergence fails: 64
    -> Stability limit order reductions: 0
1.000e+01        292       1.22e+01     3.8    0.0   32.9    0.7   62.6
Run time : 3 m 38 s
```

User preconditioning:
```
CVODE: nsteps 237, nfevals 280, nniters 279, npevals 913, nliters 669
    -> Newton iterations per step: 1.177215e+00
    -> Linear iterations per Newton iteration: 2.397849e+00
    -> Preconditioner evaluations per Newton: 3.272401e+00
    -> Last step size: 1.171224e-01, order: 3
    -> Local error fails: 0, nonlinear convergence fails: 2
    -> Stability limit order reductions: 0
1.000e+01         69       5.32e+00   -10.9    0.0   31.0    2.4   77.5
Run time : 1 m 14 s
```

petsc preconditioning (ksp_type=preonly, pc_type=hypre):
```
CVODE: nsteps 282, nfevals 326, nniters 325, npevals 787, nliters 479
    -> Newton iterations per step: 1.152482e+00
    -> Linear iterations per Newton iteration: 1.473846e+00
    -> Preconditioner evaluations per Newton: 2.421538e+00
    -> Last step size: 1.976899e-01, order: 4
    -> Local error fails: 0, nonlinear convergence fails: 6
    -> Stability limit order reductions: 0
1.000e+01         16       1.36e+00     2.0    0.0   13.8    7.7   76.6
Run time : 2 m 55 s
```
With PETSc preconditioning the number of linear iterations per Newton step is small, and the timesteps seem to be bigger than other methods. When using STRUMPACK rather than hypre there is an error on the first iteration but it continues and accelerates after:
```
[9]PETSC ERROR: STRUMPACK error: factorization failed
[9]PETSC ERROR: #1 MatLUFactorNumeric_STRUMPACK() at /Users/dudson2/code/petsc/src/mat/impls/aij/mpi/strumpack/strumpack.c:986
[9]PETSC ERROR: #2 MatLUFactorNumeric() at /Users/dudson2/code/petsc/src/mat/interface/matrix.c:3312
[9]PETSC ERROR: #3 PCSetUp_LU() at /Users/dudson2/code/petsc/src/ksp/pc/impls/factor/lu/lu.c:121
[9]PETSC ERROR: #4 PCSetUp() at /Users/dudson2/code/petsc/src/ksp/pc/interface/precon.c:1086
[9]PETSC ERROR: #5 KSPSetUp() at /Users/dudson2/code/petsc/src/ksp/ksp/interface/itfunc.c:427
```
It runs fine with STRUMPACK after the first step, but this factorization failure makes me a bit suspicious.


src/solver/impls/petsc/petsc.cxx
```diff
@@ -559,6 +519,13 @@ int PetscSolver::init() {
     // preconditioner.
 
     if (use_coloring) {
+      Field3D index = globalIndex(0);
+      PetscCall(petsc_preconditioner.createJacobianPattern(
+          index, *options, nlocal, n2Dvars(), n3Dvars(), BoutComm::get()));
+      output_progress.write("Creating Jacobian coloring\n");
+      updateColoring();
+
+#if 0
```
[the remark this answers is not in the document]
dschwoerer: Should this be deleted instead?

src/solver/impls/cvode/cvode.cxx
```diff
@@ -143,6 +165,22 @@ CvodeSolver::CvodeSolver(Options* opts)
   has_constraints = false; // This solver doesn't have constraints
   canReset = true;
 
+  if ((*options)["use_precon"].isSet()) {
+    output_warn << "WARNING: solver:use_precon is deprecated for CVODE and is now "
+                   "ignored. Use solver:cvode_precon_method=none to disable "
+                   "preconditioning.\n";
+  }
```
dschwoerer, line 177: Should this be an error?
We throw also errors for unused options, so I think we should also force the users here to update.

src/solver/impls/cvode/cvode.cxx
```diff
@@ -143,6 +165,22 @@ CvodeSolver::CvodeSolver(Options* opts)
...
+
+#if BOUT_HAS_PETSC
+  // This is a temporary fix to initialise PetscLib early, working
+  // around a bug in Hermes-3 braginskii_conduction that creates a
+  // "petsc:type" subsection.
```
dschwoerer: How hard would it be to fix it in hermes?

src/solver/impls/cvode/cvode.cxx
```diff
@@ -373,42 +454,87 @@ int CvodeSolver::init() {
...
+      }
+#endif // BOUT_HAS_PETSC
+
+    } else if (selected_precon == CvodePreconMethod::bbd) {
+      output_info.write("\tUsing BBD preconditioner\n");
+
+      const int band_width_default = std::accumulate(
+          begin(f3d), end(f3d), 0, [](int a, const VarStr<Field3D>& fvar) {
+            const Mesh* localmesh = fvar.var->getMesh();
+            return a + localmesh->xend - localmesh->xstart + 3;
```
dschwoerer, line 516: That means the ddx must only use one guard cell, and higher order stencils will perform poorly, do to wrong preconditioning?

src/solver/impls/petsc/petsc.cxx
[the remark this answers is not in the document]
dschwoerer: That seems to be the code that has been moved to petsc_preconditioner.cxx, thus I deleted it.

src/solver/impls/cvode/cvode.cxx
bendudson: It's fairly straightforward to fix (I've fixed it in a branch) so hopefully this is a short-lived patch.

src/solver/impls/cvode/cvode.cxx
bendudson: The whole BBD preconditioner is of quite dubious benefit really. It doesn't know anything about spatial stencils so only preconditions coupling between variables that appear next to each other in the 1D state vector. In practice I think that means it's only good for preconditioning coupling between variables in the same cell. I've not yet seen a case where BBD preconditioning improved performance, so it's really here in case someone wants to try it as a last resort.

src/solver/impls/cvode/cvode.cxx
dschwoerer: I would be in favor to patch it in Hermes-3, with the PR that updates the BOUT++ version to include this version. Having a workaround here, we need to remember to revert it.

src/solver/impls/cvode/cvode.cxx
dschwoerer: Would it make sense to remove it the, to keep the code smaller?

Let people use PETSc, if they want a non physics motivated preconditioning?

src/solver/impls/cvode/cvode.cxx
ZedThree: Should also add this to the input upgrader in `boutdata`?

src/solver/impls/cvode/cvode.cxx
```diff
@@ -346,8 +408,27 @@ int CvodeSolver::init() {
...
+    CvodePreconMethod selected_precon = precon_method;
+    if (selected_precon == CvodePreconMethod::Auto) {
+      if (hasPreconditioner()) {
+        selected_precon = CvodePreconMethod::user;
+      } else {
+#if BOUT_HAS_PETSC
+        selected_precon = CvodePreconMethod::petsc;
+#else
+        selected_precon = CvodePreconMethod::bbd;
+#endif
```
ZedThree: We can use `bout::build::has_petsc` to make this a normal `if`:


```suggestion
      } else if (bout::build::has_petsc) {
        selected_precon = CvodePreconMethod::petsc;
      } else {
         selected_precon = CvodePreconMethod::bbd;
```

src/solver/impls/cvode/cvode.cxx
ZedThree: Is this also the reason we do the funny business with interlacing fields?

ZedThree (dismissed):
I can't really comment on the correctness of the preconditioner itself, but I definitely support the reduction in duplicated code!

Just a couple of minor comments

src/solver/impls/cvode/cvode.cxx
bendudson: Done. Fixed in https://github.com/boutproject/hermes-3/pull/585 and this patch is removed.

PR #3385, opened 2026-06-08
Support GPU lazy field arithmetic operators

bendudson:
Expressions involving fields build trees that are merged into kernels. Aims to achieve good GPU performance without needing a lot of user code changes.

This PR implements the template expression infrastructure (`BinaryExpr`) that fuses arithmetic operations on fields.

Next steps will include extending this to include differential operators.

Running elm-pb example on 1 CPU core, the output is identical between `next` and this branch. Oddly this branch seems to be slightly slower than `next`.

`next`:
```
1.000e+00         76       1.66e+00    68.1   20.5    0.1    0.8   10.5
2.000e+00         48       1.05e+00    67.8   20.5    0.1    1.0   10.6
3.000e+00         51       1.11e+00    67.8   20.6    0.1    0.9   10.6
4.000e+00         41       8.89e-01    68.0   20.6    0.1    1.1   10.2
5.000e+00         45       1.02e+00    68.7   20.1    0.1    1.0   10.0
6.000e+00         39       9.55e-01    68.8   19.7    0.2    1.2   10.1
...
3.900e+01         63       1.38e+00    68.1   20.4    0.1    0.7   10.6
4.000e+01         57       1.25e+00    68.0   20.4    0.1    0.8   10.6
Run time : 38 s
```

This branch:
```
1.000e+00         76       2.07e+00    74.7   16.2    0.1    0.6    8.3
2.000e+00         48       1.32e+00    74.6   16.1    0.1    0.8    8.4
3.000e+00         51       1.39e+00    74.6   16.2    0.1    0.7    8.4
4.000e+00         41       1.12e+00    74.7   16.2    0.1    0.9    8.1
5.000e+00         45       1.22e+00    74.8   16.2    0.1    0.8    8.1
6.000e+00         39       1.06e+00    74.7   16.2    0.1    0.9    8.0
...
3.900e+01         63       1.76e+00    74.6   16.2    0.1    0.6    8.4
4.000e+01         57       1.58e+00    74.6   16.1    0.1    0.7    8.5
Run time : 48 s
```


bendudson:
The (slight) performance reduction in elm_pb does not seem to be due to switching from region blocks to index array lookup. I tried an experiment where `BinaryExpr::evaluate` contains:
```
if constexpr (std::is_same_v<ResT, Field3D>) {
      if (const auto* region = bout::detail::getField3DRegion(mesh, regionID)) {
        BOUT_FOR_SERIAL(i, *region) {
          data[i.ind] = operator()(i.ind); // single-pass fusion
        }
        return;
      }
    }
```
so that for Field3D operations it used the previous loop style. This slowed the run further (50s vs 48 vs 38 original). Chatty suggests perhaps (a) Lost auto-vectorization. I remember this was very fragile previously. (b) More inlining pressure and code complexity.(c) Extra integer work on every access: (idx * mul) / div in `Field3D::View::operator(); (d) Simple expressions may dominate the benchmark, outweighing the benefits of merging.

PR #3251, opened 2026-01-14
Snes with constrains

malamast:
Work in progress.  

- Added some functionality to SNES solver to support constraints (solving an differential algebraic equation DAE). I based the implementation from other solvers:  
- Solvers which can include constraints in BOUT++ are:
    IDA, from the SUNDIALS suite
    IMEX-BDF2, which uses PETSc SNES

- Currently, the constraints only work with **equation_form = backward_euler**. We will update the rest of the options later. 
- I added an option to split the pc and apply a different preconditioner for the algebraic equation. 
- I named the splited vectors as "**diff**" and "**alg**" 
- To enable the split, set in the imput file **pc_type = fieldsplit** 

In the input file, one can do something like this and explore different petsc options: 

      [solver]
      nout = 5          # Number of output steps
      output_step = 100.0    # Output timestep, normalised ion cyclotron times [1/Omega_ci]
      
      type = snes                           
      snes_type = newtonls            
      equation_form = backward_euler  
      ksp_type = gmres             
      max_nonlinear_iterations = 26  
      pc_type = fieldsplit            
      lag_jacobian = 12               
      atol = 1e-12                   
      rtol = 1e-5                
      stol = 1e-18
      maxl = 200                   
      diagnose = true
      pid_controller = true
      target_its = 8
      matrix_free_operator = false
      
      [petsc]
      # Preconditioner spliting
      pc_fieldsplit_type = multiplicative # additive,multiplicative,symmetric_multiplicative,schur,gkb
      
      fieldsplit_diff_ksp_type = preonly
      fieldsplit_diff_pc_type = hypre
      fieldsplit_diff_pc_hypre_type = ilu
      fieldsplit_diff_pc_hypre_ilu_type = Block-Jacobi-ILUT
      fieldsplit_diff_pc_hypre_ilu_drop_threshold = 1e-3
      fieldsplit_diff_pc_hypre_ilu_max_nnz_per_row = 70
      fieldsplit_diff_pc_hypre_ilu_local_reordering = true
      fieldsplit_diff_pc_hypre_ilu_tri_solve = true
      
      fieldsplit_alg_ksp_type = preonly
      fieldsplit_alg_pc_type  = hypre
      fieldsplit_alg_pc_hypre_type = boomeramg
      
      fieldsplit_alg_pc_hypre_boomeramg_cycle_type = V
      fieldsplit_alg_pc_hypre_boomeramg_coarsen_type = HMIS
[...]

PR #3386, opened 2026-06-11
Merge SNES constraints into refactor

bendudson:
Merges the changes in #3251 into the refactor-petsc-precon branch.


src/solver/impls/snes/snes.cxx
```diff
@@ -356,7 +357,9 @@ SNESSolver::SNESSolver(Options* opts)
                      .withDefault<bool>(false)),
       asinh_vars((*options)["asinh_vars"]
                      .doc("Apply asinh() to all variables?")
-                     .withDefault<bool>(false)) {}
+                     .withDefault<bool>(false)) {
+  has_constraints = true; ///< This solver can handle constraints
```
ZedThree, line 369: Should this not default to `false`? Then L384 could be removed?

Also, maybe better to set the default in the declaration in the header

src/solver/impls/snes/snes.cxx
```diff
@@ -377,6 +380,34 @@ int SNESSolver::init() {
...
+      have_constraints = true;
+      break;
+    }
+  }
+  for (int i = 0; i < n3Dvars(); i++) {
+    if (f3d[i].constraint) {
+      have_constraints = true;
+      break;
+    }
+  }
```
ZedThree, line 415: We can use the shiny ranges algorithms now!

```suggestion
  using std::ranges::any_of;
  has_constraints = any_of(f2d, &VarStr<Field2D>::constraint)
                    or any_of(f3d, &VarStr<Field3D>::constraint);
```

I think that pointer-to-member function syntax is right, you might need a lambda:

```cpp
  auto f_has_constraint = []<class T>(const VarStr<T>& f) { return f.constraint; };
```

as an intermediate. This has the added advantage of using all the valid bracket types in one expression :)

src/solver/impls/snes/snes.cxx
```diff
@@ -377,6 +380,34 @@ int SNESSolver::init() {
...
+  if (have_constraints) {
+    is_dae.reallocate(nlocal);
+    // Call the Solver function, which sets the array
+    // to one when not a constraint, zero for constraint
+    set_id(std::begin(is_dae));
+
+    if (equation_form != BoutSnesEquationForm::backward_euler) {
+      throw BoutException(
+          "SNES constraints currently require equation_form=backward_euler");
+    }
```
ZedThree: Can we do the check first?

```suggestion
    if (equation_form != BoutSnesEquationForm::backward_euler) {
      throw BoutException(
          "SNES constraints currently require equation_form=backward_euler");
    }

    is_dae.reallocate(nlocal);
    // Call the Solver function, which sets the array
    // to one when not a constraint, zero for constraint
    set_id(std::begin(is_dae));
```

ZedThree (dismissed):
Some minor suggestions

bendudson:
Replaced by #3421

PR #3389, opened 2026-06-12
Fix indexing issue for output monitoring

Steven-Roberts:
Output monitoring functions would get called twice in a row with `iter = 0` (unless `dump_on_restart = false`) and would never reach iteration `NOUT` which was used as a check for cleanup:
https://github.com/boutproject/BOUT-dev/blob/c7020b4fbf94032b4466ffe3b1f99613ad646b00/src/solver/solver.cxx#L931

This PR fixes solver iterations to be 1-indexed.

### DIFF
--- a/include/bout/solver.hxx
+++ b/include/bout/solver.hxx
@@ -451,19 +451,19 @@ protected:
   ///
   /// There are two important things to note about how \p iter is
   /// passed along to each monitor:
-  /// - The solvers all start their iteration numbering from zero, so the
-  ///   initial state is calculated at \p iter = -1
+  /// - The initial state is written at \p iter = 0, and solver output
+  ///   steps are numbered from 1 to NOUT
   /// - Secondly, \p iter is passed along to each monitor *relative to
   ///   that monitor's period*
   ///
   /// In practice, this means that each monitor is called like:
   ///
   ///     monitor->call(solver, simulation_time,
-  ///                   ((iter + 1) / monitor->period) - 1,
+  ///                   iter / monitor->period,
   ///                   NOUT / monitor->period);
   ///
-  /// e.g. for a monitor with period 10, passing \p iter = 9 will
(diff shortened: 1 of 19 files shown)


dschwoerer (commented):
Does this break the output? I think there are somewhere some tricky details, especially with smaller / slower output monitors, that make things a bit non-intuitive ...

dschwoerer:
Why not change the check to `NOUT - 1` instead of changing 19 files?

ZedThree:
> Does this break the output? I think there are somewhere some tricky details, especially with smaller / slower output monitors, that make things a bit non-intuitive ...

I'm also a bit wary about touching this stuff because I recall how fiddly the details are, but we do have pretty good tests on calling multiple monitors with different frequencies (`SolverTest.AddMonitorCheckFrequencies` for example), and that still seems to pass, so I'm in favour of merging this. It does also seem to make the logic easier to understand, even if do end up with some 1-indexing loops!

dschwoerer:
The output, at least for non restarted simulations also seem to match, so might be good.

@Steven-Roberts could you please merge https://github.com/Steven-Roberts/BOUT-dev/pull/1 to fix the formatting?
The CI is failing because it is from a fork, otherwise it would have found those issues.

Steven-Roberts:
Sorry, just getting back to this today.

> Why not change the check to NOUT - 1 instead of changing 19 files?

It comes down to the question of 0 or 1 indexing. I though the 1 indexing (with 0 for startup) seemed more natural, but I'm fine to switch to 0 indexing if that is preferred.

> Does this break the output? I think there are somewhere some tricky details, especially with smaller / slower output monitors, that make things a bit non-intuitive ...

I'll do a bit more testing this afternoon, but I haven't encountered any issue so far.

Steven-Roberts:
After some more experimenting with an example, it seems to be working as expected for me.

ZedThree (approved):
Thanks @Steven-Roberts, this does make the logic easier to follow

PR #3401, opened 2026-06-21
WIP: Lazy stencil expressions for kernel fusion

bendudson:
Operators return BinaryExpr types that perform derivatives 'lazily'. Like the single index operators they capture a CoordinatesAccessor to provide access to metrics etc. The difference is that the user code doesn't need to create field accessors or explicitly write the loop, capture variables etc. All that is done automatically if the compiler is smart/sufficiently aggressively optimizing.

Implemented DDZ_C2 and DDZ_C4 operators for Z derivatives. These appear to inline as expected in the Hasegawa-Wakatani example.

The DDZ(Field3D) operator now returns a lazy BinaryExpr type. It still allows runtime choice of numerical method (C2 / C4) and staggered fields.


include/bout/field.hxx
```diff
@@ -539,53 +539,127 @@ inline BoutReal mean(const BinaryExpr<ResT, L, R, Func>& f, bool allpe = false,
...
+template <typename T>
+struct expression_result {
+  using type = std::decay_t<T>;
+};
 
-  BOUT_FOR(i, result.getRegion(rgn)) { result[i] = ::pow(lhs[i], rhs[i]); }
+template <typename ResT, typename L, typename R, typename Func>
+struct expression_result<BinaryExpr<ResT, L, R, Func>> {
+  using type = ResT;
+};
```
ZedThree: Might be clearer to consistently just use `Result` instead of `ResT`? It took me a minute to decipher that

include/bout/mesh.hxx
```diff
@@ -95,6 +96,94 @@ using comm_handle = void*;
 
 class Mesh {
 public:
+  struct DerivativeDefaults {
+    static constexpr int num_directions = 3;
+    static constexpr int num_staggers = 3;
+    static constexpr int num_deriv_kinds = 5;
```
ZedThree, line 103: Making these `std::size_t` might fix a bunch of the clang-tidy warnings?

include/bout/mesh.hxx
```diff
@@ -95,6 +96,94 @@ using comm_handle = void*;
...
+        for (auto stagger : {STAGGER::None, STAGGER::C2L, STAGGER::L2C}) {
+          for (auto deriv : {DERIV::Standard, DERIV::StandardSecond,
+                             DERIV::StandardFourth, DERIV::Upwind, DERIV::Flux}) {
+            set(direction, deriv, builtinDefaultMethod(deriv), stagger);
+          }
+        }
+      }
+    }
+
+    static DIFF_METHOD builtinDefaultMethod(DERIV deriv) {
```
ZedThree, line 118: I think these can be also be `constexpr`

include/bout/mesh.hxx
```diff
@@ -95,6 +96,94 @@ using comm_handle = void*;
...
+      switch (deriv) {
+      case DERIV::Standard:
+      case DERIV::StandardSecond:
+      case DERIV::StandardFourth:
+        return DIFF_C2;
+      case DERIV::Upwind:
+      case DERIV::Flux:
+        return DIFF_U1;
+      }
+      throw BoutException("Unhandled derivative kind in builtinDefaultMethod");
```
ZedThree, line 128: We could use something like C++23's [`std::unreachable`](https://en.cppreference.com/cpp/utility/unreachable) instead here

include/bout/mesh.hxx
```diff
@@ -95,6 +96,94 @@ using comm_handle = void*;
...
+        return 3;
+      case DERIV::Flux:
+        return 4;
+      }
+      throw BoutException("Unhandled derivative kind in DerivativeDefaults");
+    }
+
+    DIFF_METHOD get(DIRECTION direction, DERIV deriv,
+                    STAGGER stagger = STAGGER::None) const {
+      return values[(directionIndex(direction) * num_staggers + staggerIndex(stagger))
```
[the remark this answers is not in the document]
ZedThree: I think `values.at()` suppresses this warning

include/bout/stencil_expr.hxx
```diff
@@ -0,0 +1,337 @@
...
+  BOUT_HOST_DEVICE BOUT_FORCEINLINE BoutReal apply_c2(int idx, const LView& lhs) const {
+    switch (stagger) {
+    case STAGGER::None:
+      return apply_c2_none(idx, lhs);
+    case STAGGER::C2L:
+      return apply_c2_c2l(idx, lhs);
+    case STAGGER::L2C:
+      return apply_c2_l2c(idx, lhs);
+    }
+    return 0.0;
```
ZedThree, line 496: This could also use an `unreachable()` function

ZedThree (commented):
Is this intended as a replacement for the existing derivatives generation machinery?

bendudson:
> Is this intended as a replacement for the existing derivatives generation machinery?

Hi @ZedThree ! Yes, this would replace the index_derivs machinery. The current kernels are optimized for CPUs but are too small to run efficiently on GPUs. To have any chance of running on GPUs efficiently, kernels need to be merged into larger expressions. The single index operators were one way, but these template expressions hide a lot of the ceremony of creating field accessors, capturing variables etc. The interface stays essentially the same, apart from explicitly converting BinaryExpr to fields in some places.

PR #3403, opened 2026-06-23
WIP: Improve BoutExpr CPU performance

bendudson:
Some small improvements to encourage vectorization in BoutExpr::evaluate for common cases.

### DIFF
--- a/include/bout/build_config.hxx
+++ b/include/bout/build_config.hxx
@@ -70,7 +70,13 @@ constexpr auto use_msgstack = static_cast<bool>(BOUT_USE_MSGSTACK);
 #define BOUT_FORCEINLINE inline
 #endif
 
-#if defined(_MSC_VER)
+#ifdef __has_cpp_attribute
+#if __has_cpp_attribute(assume) >= 202207L
+#define BOUT_ASSUME(condition) [[assume(condition)]]
+#endif
+#endif
+#else
+#ifdef _MSC_VER
 #define BOUT_ASSUME(condition) __assume(condition)
 #elif defined(__clang__)
 #if __has_builtin(__builtin_assume)
--- a/include/bout/build_config.hxx
+++ b/include/bout/build_config.hxx
@@ -70,4 +70,23 @@ constexpr auto use_msgstack = static_cast<bool>(BOUT_USE_MSGSTACK);
 #define BOUT_FORCEINLINE inline
 #endif
 
+#if defined(_MSC_VER)
+#define BOUT_ASSUME(condition) __assume(condition)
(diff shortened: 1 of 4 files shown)


include/bout/build_config.hxx
```diff
@@ -70,4 +70,23 @@ constexpr auto use_msgstack = static_cast<bool>(BOUT_USE_MSGSTACK);
 #define BOUT_FORCEINLINE inline
 #endif
 
+#if defined(_MSC_VER)
```
ZedThree: C++23 has an [`[[assume]]`](https://en.cppreference.com/cpp/language/attributes/assume) we should prefer:

```suggestion
#ifdef __has_cpp_attribute
#if __has_cpp_attribute(assume) >= 202207L
#define BOUT_ASSUME(condition) [[assume(condition)]]
#endif
#endif
#else
#ifdef _MSC_VER
```

include/bout/build_config.hxx
```diff
@@ -70,4 +70,23 @@ constexpr auto use_msgstack = static_cast<bool>(BOUT_USE_MSGSTACK);
...
+#define BOUT_ASSUME(condition) __assume(condition)
+#elif defined(__clang__)
+#if __has_builtin(__builtin_assume)
+#define BOUT_ASSUME(condition) __builtin_assume(condition)
```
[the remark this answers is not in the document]
ZedThree: We probably want to wrap this whole block in `NOLINTBEGIN(cppcoreguidelines-macro-usage)/NOLINTEND(cppcoreguidelines-macro-usage)` comments

src/field/field3d.cxx
```diff
@@ -927,6 +927,19 @@ std::ostream& operator<<(std::ostream& out, const Field3D& value) {
...
+namespace bout::detail {
+
+const Region<Ind3D>& getField3DRegion(const Mesh* mesh, std::optional<size_t> regionID) {
+  ASSERT1(mesh != nullptr);
+
+  if (regionID.has_value()) {
+    return mesh->getRegion(regionID.value());
+  }
+  return mesh->getRegion("RGN_ALL");
+}
```
ZedThree, line 939: Should this just be in `Mesh`?

ZedThree (changes requested):

src/field/field3d.cxx
bendudson: There is an unpleasant dependency loop: We can't call `mesh->getRegion` from `fieldops.hxx` directly, because Mesh is an incomplete type at this point. We therefore need a small helper function to call.

PR #3404, opened 2026-06-23
Use SNESSetLagJacobianPersists to force recalculating Jacobian

cmacmackin:
Previously when something happened that would require the Jacobian to be recalculated (instead of lagged), SNESSetLagJacobian() would be used to set the lag to 1 for the next time-step. However, this would turn off lagging altogether for that time-step. What we actually want to do is use SNESSetLagJacobianPersists() so that the Jacobian is recalculated at the start of the time-step and then is lagged as usual. This commit changes to doing that.

Note that, in practice, the lag wasn't actually being forced to 1 in most cases, because previously there was a line that would set it back to the value from the configurations. As far as I can tell this line shouldn't have been there.

### DIFF
--- a/src/solver/impls/snes/snes.cxx
+++ b/src/solver/impls/snes/snes.cxx
@@ -881,7 +881,7 @@ int SNESSolver::init() {
   return 0;
 }
 
-PetscErrorCode SNESSolver::rescale(int& saved_jacobian_lag) {
+PetscErrorCode SNESSolver::rescale() {
   // Individual variable scaling
   // Note: If variables are rescaled then the Jacobian columns
   //       need to be scaled or recalculated
@@ -929,12 +929,8 @@ PetscErrorCode SNESSolver::rescale(int& saved_jacobian_lag) {
   }
 
   // Force recalculation of the Jacobian
-  SNESGetLagJacobian(snes, &saved_jacobian_lag);
-  // FIXME: This isn't actually what we want to happen. The lag should
-  // stay as before, we just want to make sure the Jacobian gets
-  // re-evaluated at the start of the solve. We need to use
-  // SNESSetJacobianPersist for that.
-  SNESSetLagJacobian(snes, 1);
+  PetscCall(SNESSetLagJacobianPersists(snes, PETSC_FALSE));
+
   return PETSC_SUCCESS;
 }
 
@@ -957,9 +953,8 @@ int SNESSolver::run() {
     PetscCall(VecRestoreArray(snes_x, &xdata));
   }
 
-  int saved_jacobian_lag = 0;
   if (scale_vars) {
-    PetscCall(rescale(saved_jacobian_lag));
+    PetscCall(rescale());
   }
 
   // Initialise residuals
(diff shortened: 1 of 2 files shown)


src/solver/impls/snes/snes.cxx
```diff
@@ -1162,13 +1161,6 @@ int SNESSolver::run() {
         continue; // Try again
       }
 
-      if (saved_jacobian_lag != 0) {
```
bendudson, line 1165: This should restore the previous Jacobian lag on successful SNES convergence. I think this should restore the JacobianPersists setting to its input option. It looks like `rescale()` will set JacobianPersists to False, but where would it be restored to True (if that's what was in the settings)?

cmacmackin: If rescale determines that JacobianPersists should be false then that applies for the current time-step and it will be reset on the next one.

bendudson: Where does the reset happen? This point (around line 1165) is where the Jacobian lag was being reset: Execution reaches this point only after a successful SNES convergence. If this is removed then I don't see where JacobianPersists is reset.

cmacmackin: The sequence is:

1. [Start the iteration](https://github.com/boutproject/BOUT-dev/pull/3404/changes#diff-556d8075945153b052f4ffe414fbaddc7ecb90f81bae46ef32ac900526c240d2R980)
2. If not persist_jac_lag_off, [reset SNESSetLagJacobianPersists](https://github.com/boutproject/BOUT-dev/pull/3404/changes#diff-556d8075945153b052f4ffe414fbaddc7ecb90f81bae46ef32ac900526c240d2R998). Otherwise leave SNESSetLagJacobianPersists to whatever it was set to in the last iteration and [set persist_jac_lag_off to false](https://github.com/boutproject/BOUT-dev/pull/3404/changes#diff-556d8075945153b052f4ffe414fbaddc7ecb90f81bae46ef32ac900526c240d2R995). 
3. [Rescale, if necessary](https://github.com/boutproject/BOUT-dev/pull/3404/changes#diff-556d8075945153b052f4ffe414fbaddc7ecb90f81bae46ef32ac900526c240d2R1004). If a rescale is performed, [set SNESSetLagJacobianPersists to false](https://github.com/boutproject/BOUT-dev/pull/3404/changes#diff-556d8075945153b052f4ffe414fbaddc7ecb90f81bae46ef32ac900526c240d2R932), but don't change persist_jac_lag_off.
4. Try to perform the timestep. If something happens that will require the length of the timestep to be changed for the next iteration, set SNESSetLagJacobianPersists to false *and* persist_jac_lag_off to true.
5. Repeat.

So, to answer your questions, the value of JacobianPersists is reset to the input value at step 2, or [line 998](https://github.com/boutproject/BOUT-dev/pull/3404/changes#diff-556d8075945153b052f4ffe414fbaddc7ecb90f81bae46ef32ac900526c240d2R998).

cmacmackin:
@bendudson Are you happy with my explanation of how the control flow works now?

src/solver/impls/snes/snes.cxx
bendudson: Thanks @cmacmackin yes that makes sense.

bendudson (approved):

PR #3415, opened 2026-06-30
Capture git diff when building

### LINKED ISSUE #2217: Save git diff?
We already save the git-hash, but in case the code has been edited somewhere it would be nice to also compile the git-diff into the code so we can save it into run output, helping reproducibility. I've done this for STORM, but think I need to hand over to a build-system guru to add it to BOUT++.

One way to save the git hash is to have a Python script, e.g.
```
#!/usr/bin/env python3

from pathlib import Path
from string import ascii_uppercase

from boututils.run_wrapper import shell

scriptdir = Path(__file__).parent

_, diff = shell("git diff", pipe=True)

# Find a string to use for the delimiter
delimiter = "_BOUT_DIFF_"
for letter in ascii_uppercase:
    if ")" + delimiter not in diff:
        delimiter_success = True
        break
    delimiter = "_BOUT_DIFF_" + letter + "_"
if ")" + delimiter in diff:
    # print a return code to stdout so we can capture and test it in the makefile
    print(11)
    raise ValueError(
        "save_git_version.py failed to find a delimiter that is not in the git diff"
    )

with open("bout_git_hash.hxx", "w") as f:
    f.write("constexpr auto bout_git_diff = R\"" + delimiter + "(")
    f.write(diff)
    f.write(")" + delimiter + "\";\n")

# print a return code to stdout so we can capture and test it in the makefile
print(0)
```
although probably it would be better to append to `include/bout/build_defines.hxx` or something. This script only has dependencies on Python standard library (I think) plus `boututils` (I guess the `boututils` import needs fixing to make sure it's always imported from the local submodule).

Not sure whether it's best to do this at configure-time or compile-time. I'd favour compile-time if possible - I think we always rebuild `bout++.cxx` (?) so if we have a header that's only imported there, we could re-generate that header each time `make` is run. What I've got for STORM is some lines in the `makefile` like
```
save_git_version_status := $(shell ./save_git_version.py)
ifeq ($(save_git_version_status), 11)
  $(error "save_git_version.py failed to find a delimiter that is not in the git diff")
endif
ifneq ($(save_git_version_status), 0)
  $(error "save_git_version.py failed with an unrecognised error $(save_git_version_status)")
endif
```

bendudson:
Stores code changes from the current commit into the output dump files. This is to help in reproducing results when code was not committed before building.

- Generates a git diff on build, so the diff used when compiling is stored, not the diff when configuring. Handles out-of-source builds and cases where the source is not under git e.g. release tarballs.

- Stores the diff as a string in a compilation unit git_metadata.cxx

- Saves to output, along with flags indicating whether the diff was available, and whether the build had uncommitted changes.

Not sure how to test this in CI, so I just verified manually that changes in source get stored in the output files (`conduction` example). GPT-5.4 wrote most of the code; I reviewed and modified slightly.

Fixes #2217

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -389,6 +389,7 @@ set(BOUT_SOURCES
     ./src/sys/timer.cxx
     ./src/sys/type_name.cxx
     ./src/sys/utils.cxx
+    ./include/bout/git_metadata.hxx
     ${CMAKE_CURRENT_BINARY_DIR}/include/bout/revision.hxx
     ${CMAKE_CURRENT_BINARY_DIR}/include/bout/version.hxx
 )
@@ -520,8 +521,26 @@ add_custom_command(
   MAIN_DEPENDENCY "${CMAKE_CURRENT_LIST_DIR}/cmake/GenerateDateTimeFile.cmake"
 )
 
-add_library(bout++ ${BOUT_SOURCES} ${CMAKE_CURRENT_BINARY_DIR}/bout++-time.cxx)
+set(BOUT_GIT_METADATA_SOURCE ${CMAKE_CURRENT_BINARY_DIR}/git_metadata.cxx)
+set_source_files_properties(
+  ${BOUT_GIT_METADATA_SOURCE} PROPERTIES GENERATED TRUE
+)
+add_custom_target(
+  bout++-git-metadata
+  COMMAND
+    ${CMAKE_COMMAND} -DOUTPUT_FILE=${BOUT_GIT_METADATA_SOURCE}
+    -DSOURCE_DIR=${PROJECT_SOURCE_DIR} -DGIT_EXECUTABLE=${GIT_EXECUTABLE} -P
+    "${CMAKE_CURRENT_LIST_DIR}/cmake/GenerateGitMetadata.cmake"
+  BYPRODUCTS ${BOUT_GIT_METADATA_SOURCE}
+  COMMENT "Generating git metadata source"
+)
+
+add_library(
+  bout++ ${BOUT_SOURCES} ${CMAKE_CURRENT_BINARY_DIR}/bout++-time.cxx
+         ${BOUT_GIT_METADATA_SOURCE}
+)
 add_library(bout++::bout++ ALIAS bout++)
+add_dependencies(bout++ bout++-git-metadata)
 target_link_libraries(bout++ PUBLIC MPI::MPI_CXX)
 target_include_directories(
   bout++
@@ -566,6 +585,7 @@ set(BOUT_VERSION_MINOR ${PROJECT_VERSION_MINOR})
 set(BOUT_VERSION_PATCH ${PROJECT_VERSION_PATCH})
 
 # Get the git commit
+find_package(Git QUIET)
 include(GetGitRevisionDescription)
 get_git_head_revision(GIT_REFSPEC BOUT_REVISION)
 if(BOUT_REVISION STREQUAL "GITDIR-NOTFOUND")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -512,17 +512,16 @@ else()
 endif()
 set(BOUT_USE_PYTHON ${BOUT_ENABLE_PYTHON})
 
-set(BOUT_GIT_METADATA_SOURCE ${CMAKE_CURRENT_BINARY_DIR}/git_metadata.cxx)
 set_source_files_properties(
-  ${BOUT_GIT_METADATA_SOURCE} PROPERTIES GENERATED TRUE
+  ${CMAKE_CURRENT_BINARY_DIR}/git_metadata.cxx PROPERTIES GENERATED TRUE
 )
 add_custom_target(
   bout++-git-metadata
   COMMAND
-    ${CMAKE_COMMAND} -DOUTPUT_FILE=${BOUT_GIT_METADATA_SOURCE}
+    ${CMAKE_COMMAND} -DOUTPUT_FILE=${CMAKE_CURRENT_BINARY_DIR}/git_metadata.cxx
     -DSOURCE_DIR=${PROJECT_SOURCE_DIR} -DGIT_EXECUTABLE=${GIT_EXECUTABLE} -P
     "${CMAKE_CURRENT_LIST_DIR}/cmake/GenerateGitMetadata.cmake"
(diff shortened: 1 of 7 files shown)


cmake/GenerateGitMetadata.cmake
```diff
@@ -0,0 +1,126 @@
...
+    else()
+      execute_process(
+        COMMAND "${GIT_EXECUTABLE}" diff --no-ext-diff --binary --cached --
+        WORKING_DIRECTORY "${SOURCE_DIR}"
+        RESULT_VARIABLE git_cached_diff_result
+        OUTPUT_VARIABLE git_cached_diff
+        ERROR_QUIET
+      )
+      if(NOT git_cached_diff_result EQUAL 0)
+        set(git_cached_diff "")
```
dschwoerer:
```suggestion
        set(git_cached_diff "git diff unavailable")
```
We could store also that the git diff is not available in the string.
If it is used programmatically, it would fail to apply the diff, which would be fine, I think. If it is used by humans, it is easy enough to see that this is not actually a diff.

cmake/GenerateGitMetadata.cmake
```diff
@@ -0,0 +1,126 @@
...
+  "auto git_metadata_available() -> bool { return git_metadata_available_value; }\n"
+)
+file(APPEND "${output_tmp}"
+     "auto git_dirty() -> bool { return git_dirty_value; }\n"
+)
+file(APPEND "${output_tmp}"
+     "auto git_diff() -> std::string_view { return git_diff_value; }\n"
+)
+file(APPEND "${output_tmp}" "} // namespace version\n")
+file(APPEND "${output_tmp}" "} // namespace bout\n")
```
dschwoerer: For other cases we have a file, that we configure. I think that is easier to read and maintain.

cmake/GenerateGitMetadata.cmake
```diff
@@ -0,0 +1,126 @@
...
+
+execute_process(
+  COMMAND "${CMAKE_COMMAND}" -E copy_if_different "${output_tmp}"
+          "${OUTPUT_FILE}"
+)
+execute_process(COMMAND "${CMAKE_COMMAND}" -E rm -f "${output_tmp}")
```
dschwoerer, line 106: I think cmake should be able to move?

CMakeLists.txt
```diff
@@ -520,8 +521,26 @@ add_custom_command(
   MAIN_DEPENDENCY "${CMAKE_CURRENT_LIST_DIR}/cmake/GenerateDateTimeFile.cmake"
 )
 
-add_library(bout++ ${BOUT_SOURCES} ${CMAKE_CURRENT_BINARY_DIR}/bout++-time.cxx)
+set(BOUT_GIT_METADATA_SOURCE ${CMAKE_CURRENT_BINARY_DIR}/git_metadata.cxx)
+set_source_files_properties(
+  ${BOUT_GIT_METADATA_SOURCE} PROPERTIES GENERATED TRUE
+)
+add_custom_target(
```
dschwoerer, line 518: I would prefer `add_custom_command` - like we use for the timestamp above.

Although, it makes me wonder - could / should they be in the same file? Having to compile just one file should be faster, and both contain dynamic, generated content. They both get modified at the same time.

CMakeLists.txt
```diff
@@ -520,8 +521,26 @@ add_custom_command(
...
+    ${CMAKE_COMMAND} -DOUTPUT_FILE=${BOUT_GIT_METADATA_SOURCE}
+    -DSOURCE_DIR=${PROJECT_SOURCE_DIR} -DGIT_EXECUTABLE=${GIT_EXECUTABLE} -P
+    "${CMAKE_CURRENT_LIST_DIR}/cmake/GenerateGitMetadata.cmake"
+  BYPRODUCTS ${BOUT_GIT_METADATA_SOURCE}
+  COMMENT "Generating git metadata source"
+)
+
+add_library(
+  bout++ ${BOUT_SOURCES} ${CMAKE_CURRENT_BINARY_DIR}/bout++-time.cxx
+         ${BOUT_GIT_METADATA_SOURCE}
```
dschwoerer: I think we can just write the name here (and above), that makes it easier to read.

dschwoerer (commented):
If we want to test this in CI, it would not be that hard. We already have some post `ctest` tests: https://github.com/boutproject/BOUT-dev/blob/next/.ci_with_cmake.sh#L18

We could it there. Modify a source file, adding some white space to a file, and then build a test executable, and check the output. Certainly not a nice implementation, and I am not sure it is worth to have the extra time spent in CI.

cmake/GenerateGitMetadata.cmake
bendudson: According to my parrot, the reason for copy and delete is copy_if_different only updates git_metadata.cxx when contents actually changed. A move/rename would replace the file every build, even if the diff is unchanged. That would retrigger recompilation of the translation unit every build.

CMakeLists.txt
bendudson: Parrot says:

"A custom command with OUTPUT git_metadata.cxx would only rerun when:
the output is missing, or listed dependencies are newer.
But the thing we care about is “working tree state changed”, which includes unstaged edits, staged edits, possibly files outside BOUT_SOURCES. Those changes do not map neatly onto a safe DEPENDS list. A custom target with no outputs is always out of date, so it forces the metadata refresh each build, which is the simple reliable choice here."

(I am now just a mouthpiece for the parrot oracle :) )

cmake/GenerateGitMetadata.cmake
bendudson: Now replaced with a `cmake/git_metadata.cxx.in` file that is configured.

cmake/GenerateGitMetadata.cmake
```diff
@@ -0,0 +1,106 @@
...
+  message(FATAL_ERROR "OUTPUT_FILE must be set")
+endif()
+if(NOT DEFINED SOURCE_DIR)
+  message(FATAL_ERROR "SOURCE_DIR must be set")
+endif()
+
+if(NOT DEFINED GIT_EXECUTABLE OR GIT_EXECUTABLE STREQUAL "")
+  find_program(GIT_EXECUTABLE git)
+endif()
+
```
dschwoerer, line 27: Do we really want to check every time whether git is installed?

```suggestion
```

dschwoerer (approved):

PR #3416, opened 2026-07-01
Cleanup YBoundary

dschwoerer:
Follow up on #3308 

* Removes duplicate implementation for FCI and FA. They where part of the initial design. The new design uses a unified CRTP implementation, that FCI, FA and X-direction share.
* Adds units tests
* Switches yboundary to use unified implementation.


ZedThree (approved):
Thanks @dschwoerer! I really like that we're finally getting rid of some raw pointers in `Mesh`! I feel like we probably have more `dynamic_cast`s than strictly necessary, but I haven't looked into it too hard

PR #3418, opened 2026-07-06
Next version for next will be 6.0.0

dschwoerer:
Or will it even be 6.0.0, @ZedThree ?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,7 +31,7 @@ endif()
 # Set the version here, strip any extra tags to use in `project`
 # We try to use git to get a full description, inspired by setuptools_scm
 set(_bout_previous_version "5.2.0")
-set(_bout_next_version "5.2.1")
+set(_bout_next_version "5.3.0")
 execute_process(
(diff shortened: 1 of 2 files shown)


ZedThree:
we're breaking lots of things, I think, so 6.0.0

dschwoerer:
Ok, updated

dschwoerer:
Slightly worrying that there is a test failure:
```
/home/runner/work/BOUT-dev/BOUT-dev/tests/unit/include/bout/test_single_index_ops.cxx:185: Failure
Errors while running CTest
Value of: IsFieldEqual(difops, indexops, "RGN_NOBNDRY")
  Actual: false (Field3D(1, 1, 3; [45]) == -5.0373076593904145; Expected: -5.0373076593904162)
Expected: true
```

That seems to be a rounding difference, but I would assume that even with the template expression we should not introduce any race conditions or similar, that might be introducing random errors? Is that maybe hardware dependent?

bendudson (approved):

PR #3420, opened 2026-07-08
Legacy boundary conditions

malamast:
Added a legacy implementation of 1st order Dirichlet and Neumann BCs.  
    
  - This is what we used to do in Bout++ version 3.
  - The default BC remain unchanged. 
  - Higher order extrapolation can be unstable sometimes when using C4 so I added back the 1st order BC. 
  - You can use them in the input file by setting e.g. **bndry_all = dirichlet_o1**.
  - Implementation for staggered grids is not done.


dschwoerer:
>     * Implementation for staggered grids is not done.

Could you add an assert?

Also, please fix the clang-tidy and clang-format issues.

malamast:
> > ```
> > * Implementation for staggered grids is not done.
> > ```
> 
> Could you add an assert?
> 
> Also, please fix the clang-tidy and clang-format issues.

I have this line here: 
https://github.com/boutproject/BOUT-dev/blob/a96044748c6464b5fc297460476a306c7719905d/src/mesh/boundary_standard.cxx#L154 
 throw BoutException("neumann_o1 BC is not implementated for staggered grids.");
That throws an Exception if this BC is used with staggered grids. Does this BoutException  work as an error message or a an assert statement?  

Also, I used: git clang-format origin/next for the formatting. Is there another command I sue use for clang-tidy?

dschwoerer:
> I have this line here:
> 
> https://github.com/boutproject/BOUT-dev/blob/a96044748c6464b5fc297460476a306c7719905d/src/mesh/boundary_standard.cxx#L154

That is sufficient. However, it would be better to check whether the fields location is at CELL_CENTRE.
Running with staggered grids is not an issue for this operator, as long as it is only used for fields at CELL_CENTRE.

> Also, I used: git clang-format origin/next for the formatting. Is there another command I sue use for clang-tidy?

Formatting is now fine. For clang-tidy you see the comments that it posts. Some of them would require a refactoring, which is probably not in scope for this PR, like this:

> warning: returning a newly created resource of type 'BoundaryOp *' or 'gsl::owner<>' from a function whose return type is not 'gsl::owner<>' [[cppcoreguidelines-owning-memory](https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/owning-memory.html)]

dschwoerer (approved):
This looks decent. The only question is whether we want to add more stuff to `boundary_standard`

PR #3421, opened 2026-07-08
Snes with constrains updated

malamast:
This is an extension of an older PR #3251 that was later merged in #3386 
~Probably we **will need to merge it with** #3386 before it can be merged to next~

- With this PR the SNES solver now  supports constraints with the other EquationForms.
- Supports solution of algebraic differential equations (e.g. for phi) also for the other equation forms like rearranged_backward_euler and the pseudo_transient method.
- This method worked well for 2D problems for the inversion of the Laplacian to get the phi. 
- The BC grid points are not part of the solution space so no need to apply BC to the ddt(phi) as in the example.


src/solver/impls/snes/snes.cxx
```diff
@@ -1690,6 +1835,11 @@ BoutReal SNESSolver::pid(BoutReal timestep, int nl_its, BoutReal max_dt) {
   // clamp growth factor to avoid huge changes
   BoutReal fac = std::clamp(facP * facI * facD, 0.2, 5.0);
 
+  // Add slow growth when convergence is good and stable
+  if (nl_its <= target_its && nl_its_prev <= target_its) {
+    fac *= 1.1; // or 1.05 for more conservative growth
+  }
+
```
cmacmackin: I feel like there should be more discussion before adopting this. We are aware that timestep setting tends to be too conservative. So far we've been dealing with it by setting `target_its` to be much larger than it should be, [as seen here](https://github.com/mikekryjak/hermes-perftest/blob/master/recipes/SNES-MUMPS-1.txt). That certainly isn't a good long-term solution, but I think there were some other ideas for how this could be handled. @mikekryjak What are your thoughts?

src/solver/impls/snes/snes.cxx
```diff
@@ -385,6 +387,29 @@ int SNESSolver::init() {
   output_info.write("\t3d fields = {:d}, 2d fields = {:d} neq={:d}, local_N={:d}\n",
                     n3Dvars(), n2Dvars(), neq, nlocal);
 
+  // Check if there are any constraints
+  have_constraints = false;
```
cmacmackin, line 513: Just give this a default value of `false`, unless you expect that the `init()` method might be called more than once.

```suggestion
```

src/solver/impls/snes/snes.hxx
```diff
@@ -223,6 +223,13 @@ private:
   int nlocal; ///< Number of variables on local processor
   int neq;    ///< Number of variables in total
 
+  bool have_constraints;  ///< Are there any constraint variables?
```
cmacmackin, line 251:
```suggestion
  bool have_constraints = false;  ///< Are there any constraint variables?
```

src/solver/impls/snes/snes.cxx
cmacmackin, line 513: I find `have_constraints` is too similar a name to `has_constraints` (I was confused by this during my review). Could you choose another name, such as `algebraic_constraints`?

src/solver/impls/snes/snes.cxx
```diff
@@ -449,6 +474,34 @@ int SNESSolver::init() {
...
+        alg_idx.push_back(gi);
+      }
+    }
+
+    PetscCall(ISCreateGeneral(BoutComm::get(), diff_idx.size(), diff_idx.data(),
+                              PETSC_COPY_VALUES, &is_diff));
+    PetscCall(ISCreateGeneral(BoutComm::get(), alg_idx.size(), alg_idx.data(),
+                              PETSC_COPY_VALUES, &is_alg));
+
+    have_is_maps = true;
```
cmacmackin: What is the purpose of this variable? As far as I can tell, if `have_constraints` is true then this will always also be true. Is there any situation when the value of `have_is_maps` and `have_constraints` could differ?

src/solver/impls/snes/snes.cxx
```diff
@@ -580,6 +633,24 @@ int SNESSolver::init() {
...
+
+    // Give PETSc the index sets
+    PetscCall(PCFieldSplitSetIS(pc, "diff", is_diff));
+    PetscCall(PCFieldSplitSetIS(pc, "alg", is_alg));
+
+    // Let the user configure from options (recommended)
+    // Example options you can set in input file:
+    //   -pc_fieldsplit_type additive
+    //   -fieldsplit_alg_pc_type hypre -fieldsplit_alg_pc_hypre_type boomeramg
+    //   -fieldsplit_diff_pc_type ilu
```
cmacmackin: It would be better to put these notes in documentation, rather than as a comment in the code.

src/solver/impls/snes/snes.cxx
```diff
@@ -1510,26 +1581,100 @@ PetscErrorCode SNESSolver::snes_function(Vec x, Vec f, bool linear) {
...
   case BoutSnesEquationForm::rearranged_backward_euler: {
     // Rearranged Backward Euler
-    // f = (x0 - x)/Δt + f
-    // First calculate x - x0 to minimise floating point issues
-    VecWAXPY(delta_x, -1.0, x0, x); // delta_x = x - x0
-    VecAXPY(f, -1. / dt, delta_x);  // f <- f - delta_x / dt
+    // F = (x0 - x)/Δt + f
+    // Algebraic:     F = G(x)  (already stored in f by rhs_function)
+
+    if (!have_constraints) {
```
cmacmackin: There's lots of repetition in this part of the code. I've drafted an alternative approach which should simplify things, but I don't have any problems to test it on. I'll post the diff in a separate comment.

cmacmackin (changes requested):
There are some changes required and some things I think need further discussion. The biggest problem, though, is that there is not test for any of this. You should add an integration test.

cmacmackin:
Here is a proposed refactor to avoid repeating yourself so much in `SnesSovler::snes_function`. You will need to check it actually works though.

```diff
diff --git a/src/solver/impls/snes/snes.cxx b/src/solver/impls/snes/snes.cxx
index bdb32ab04..657000ca1 100644
--- a/src/solver/impls/snes/snes.cxx
+++ b/src/solver/impls/snes/snes.cxx
@@ -1567,9 +1567,27 @@ PetscErrorCode SNESSolver::rhs_function(Vec x, Vec f, bool linear) {
   return PETSC_SUCCESS;
 }
 
+template <typename Func, typename... Args>
+PetscErrorCode performCalculation(Func operation, IS indices, Args... args) {
+  if (indices == nullptr) {
+    PetscCall(operation(args...));
+  } else {
+    constexpr std::size_t N = sizeof...(Args);
+    std::array<Vec, N> vectors{args...};
+    std::array<Vec, N> subvectors;
+    for (std::size_t i = 0; i < N; ++i) {
+      PetscCall(VecGetSubVector(vectors[i], indices, &(subvectors[i])));
+    }
+    PetscCall(std::apply(operation, subvectors));
+    for (std::size_t i = 0; i < N; ++i) {
+      PetscCall(VecRestoreSubVector(vectors[i], indices, &(subvectors[i])));
+    }
+  }
+  return PETSC_SUCCESS;
+}
+
 // Result in f depends on equation_form
 PetscErrorCode SNESSolver::snes_function(Vec x, Vec f, bool linear) {
-
   // Call the RHS function
   if (rhs_function(x, f, linear) != PETSC_SUCCESS) {
     // Tell SNES that the input was out of domain
@@ -1583,34 +1601,13 @@ PetscErrorCode SNESSolver::snes_function(Vec x, Vec f, bool linear) {
     // Rearranged Backward Euler
     // F = (x0 - x)/Δt + f
     // Algebraic:     F = G(x)  (already stored in f by rhs_function)
-
-    if (!have_constraints) {
-
-      // First calculate x - x0 to minimise floating point issues
-      VecWAXPY(delta_x, -1.0, x0, x); // delta_x = x - x0
-      VecAXPY(f, -1.0 / dt, delta_x); // f <- f - delta_x / dt
-
-    } else {
-
-      ASSERT2(have_is_maps);
-      // Some constraints
-
-      Vec x_diff, x0_diff, delta_x_diff, f_diff;
-      PetscCall(VecGetSubVector(x, is_diff, &x_diff));
-      PetscCall(VecGetSubVector(x0, is_diff, &x0_diff));
-      PetscCall(VecGetSubVector(delta_x, is_diff, &delta_x_diff));
-      PetscCall(VecGetSubVector(f, is_diff, &f_diff));
-
-      PetscCall(VecWAXPY(delta_x_diff, -1.0, x0_diff,
-                         x_diff)); // delta_x_diff = x_diff - x0_diff
-      PetscCall(
-          VecAXPY(f_diff, -1.0 / dt, delta_x_diff)); // f_diff <- f_diff - delta_x / dt
-
-      PetscCall(VecRestoreSubVector(x, is_diff, &x_diff));
-      PetscCall(VecRestoreSubVector(x0, is_diff, &x0_diff));
-      PetscCall(VecRestoreSubVector(delta_x, is_diff, &delta_x_diff));
-      PetscCall(VecRestoreSubVector(f, is_diff, &f_diff));
-    }
+    PetscCall(performCalculation(
+        [this](Vec x, Vec x0, Vec delta_x, Vec f) -> PetscErrorCode {
+          PetscCall(VecWAXPY(delta_x, -1.0, x0, x)); // delta_x = x - x0
+          PetscCall(VecAXPY(f, -1.0 / dt, delta_x)); // f <- f - delta_x / dt
+          return PETSC_SUCCESS;
+        },
+        is_diff, x, x0, delta_x, f));
     break;
   }
   case BoutSnesEquationForm::pseudo_transient: {
@@ -1618,34 +1615,15 @@ PetscErrorCode SNESSolver::snes_function(Vec x, Vec f, bool linear) {
     // except that Δt is a vector
     // F = (x0 - x)/Δt + f
     // Algebraic:     F = G(x)  (already stored in f by rhs_function)
-
-    if (!have_constraints) {
-
-      VecWAXPY(delta_x, -1.0, x0, x);
-      VecPointwiseDivide(delta_x, delta_x, dt_vec); // delta_x /= dt
-      VecAXPY(f, -1.0, delta_x);                    // f <- f - delta_x
-
-    } else {
-      ASSERT2(have_is_maps);
-
-      Vec x_diff, x0_diff, delta_x_diff, f_diff, dt_vec_diff;
-      PetscCall(VecGetSubVector(x, is_diff, &x_diff));
-      PetscCall(VecGetSubVector(x0, is_diff, &x0_diff));
-      PetscCall(VecGetSubVector(delta_x, is_diff, &delta_x_diff));
-      PetscCall(VecGetSubVector(f, is_diff, &f_diff));
-      PetscCall(VecGetSubVector(dt_vec, is_diff, &dt_vec_diff));
-
-      PetscCall(VecWAXPY(delta_x_diff, -1.0, x0_diff, x_diff));
-      PetscCall(
-          VecPointwiseDivide(delta_x_diff, delta_x_diff, dt_vec_diff)); // delta_x /= dt
-      PetscCall(VecAXPY(f_diff, -1.0, delta_x_diff)); // f <- f - delta_x
-
-      PetscCall(VecRestoreSubVector(delta_x, is_diff, &delta_x_diff));
-      PetscCall(VecRestoreSubVector(x, is_diff, &x_diff));
-      PetscCall(VecRestoreSubVector(x0, is_diff, &x0_diff));
-      PetscCall(VecRestoreSubVector(f, is_diff, &f_diff));
-      PetscCall(VecRestoreSubVector(dt_vec, is_diff, &dt_vec_diff));
-    }
+    PetscCall(performCalculation(
+
+        [](Vec x, Vec x0, Vec delta_x, Vec f, Vec dt_vec) -> PetscErrorCode {
+          PetscCall(VecWAXPY(delta_x, -1.0, x0, x));
+          PetscCall(VecPointwiseDivide(delta_x, delta_x, dt_vec)); // delta_x /= dt
+          PetscCall(VecAXPY(f, -1.0, delta_x));                    // f <- f - delta_x
+          return PETSC_SUCCESS;
+        },
+        is_diff, x, x0, delta_x, f, dt_vec));
     break;
   }
   case BoutSnesEquationForm::backward_euler: {
@@ -1653,28 +1631,13 @@ PetscErrorCode SNESSolver::snes_function(Vec x, Vec f, bool linear) {
     // Differential:  F = x - x0 - dt*f
     // Algebraic:     F = G(x)  (already stored in f by rhs_function)
 
-    if (!have_constraints) {
-
-      VecAYPX(f, -dt, x);   // f <- x - Δt*f
-      VecAXPY(f, -1.0, x0); // f <- f - x0
-
-    } else {
-
-      ASSERT2(have_is_maps);
-      // Some constraints
-
-      Vec x_diff, x0_diff, f_diff;
-      PetscCall(VecGetSubVector(x, is_diff, &x_diff));
-      PetscCall(VecGetSubVector(x0, is_diff, &x0_diff));
-      PetscCall(VecGetSubVector(f, is_diff, &f_diff));
-
-      PetscCall(VecAYPX(f_diff, -dt, x_diff));   // f_diff <- x_diff - dt*f_diff
-      PetscCall(VecAXPY(f_diff, -1.0, x0_diff)); // f_diff <- f_diff - x0_diff
-
-      PetscCall(VecRestoreSubVector(x, is_diff, &x_diff));
-      PetscCall(VecRestoreSubVector(x0, is_diff, &x0_diff));
-      PetscCall(VecRestoreSubVector(f, is_diff, &f_diff));
-    }
+    PetscCall(performCalculation(
+        [this](Vec x, Vec x0, Vec f) -> PetscErrorCode {
+          PetscCall(VecAYPX(f, -dt, x));   // f <- x - Δt*f
+          PetscCall(VecAXPY(f, -1.0, x0)); // f <- f - x0
+          return PETSC_SUCCESS;
+        },
+        is_diff, x, x0, f));
     break;
   }
   case BoutSnesEquationForm::direct_newton: {
```

bendudson:
Thanks for your review and suggestions @cmacmackin ! I think I have addressed them. I also merged the changes in #3386 into this PR. I think it's ready to go.

bendudson (dismissed):

tests/integrated/test-snes-constraints/test_snes_constraints.py
```diff
@@ -0,0 +1,67 @@
...
+        and "Run finished at" in result.stdout
+        and "internal_Finalize" in result.stderr
+        and "OFI poll failed" in result.stderr
+    )
+
+
+def run_command(command: str) -> str:
+    result = subprocess.run(
+        command, shell=True, capture_output=True, text=True, timeout=600, check=False
+    )
```
cmacmackin, line 29: Shouldn't we use `boututils.run_wrapper.launch_safe` here?

src/solver/impls/snes/snes.cxx
cmacmackin: Is there any reason this has to be called `have_constraints`? I still think it is too easily confused with the `has_constraints` member variables. I actually like the name of that variable even less (I think it would be better to call it something like `supports_constraints`) but it's not worth breaking the API to change it.

tests/integrated/test-snes-constraints/test_snes_constraints.cxx
```diff
@@ -0,0 +1,33 @@
+#include <bout/bout_types.hxx>
+#include <bout/field3d.hxx>
+#include <bout/physicsmodel.hxx>
+#include <bout/unused.hxx>
+
+namespace {
+constexpr BoutReal constraint_factor = 0.5;
+}
+
+class TestSnesConstraints : public PhysicsModel {
```
cmacmackin, line 10: Could you add a description of the system of equations being solved and the expected solution?

tests/integrated/test-snes-constraints/test_snes_constraints.cxx
```diff
@@ -0,0 +1,33 @@
...
+
+    u = 1.0;
+    phi = constraint_factor;
+
+    return 0;
+  }
+
+  int rhs(BoutReal UNUSED(time)) override {
+    ddt(u) = -u + phi;
+    ddt(phi) = phi - constraint_factor * u;
```
cmacmackin, line 27: Am I correct in thinking that the result of this algebraic equation is being stored in `ddt(phi)`? I'm guessing that's because it's the only way to get information to the solver?

This is (unavoidably) confusing and makes it look like it's just a differential equation after all. I'd suggest adding a comment to explain.

tests/integrated/test-snes-constraints/test_snes_constraints.cxx
```diff
@@ -0,0 +1,33 @@
+#include <bout/bout_types.hxx>
+#include <bout/field3d.hxx>
+#include <bout/physicsmodel.hxx>
+#include <bout/unused.hxx>
+
+namespace {
+constexpr BoutReal constraint_factor = 0.5;
```
cmacmackin, line 7: Why is this in an anonymous namespace rather than just a `static constexpr` member variable of `TestSnesConstraints`?

tests/integrated/test-snes-constraints/test_snes_constraints.cxx
bendudson: Good point. I've added a `residual()` alias of `ddt()`

bendudson:
Thanks @cmacmackin ! I've renamed `have_constraints` to `has_constraint_variables` in SNES and IMEXBDF2 solvers. The `has_constraints` variable is protected (not public API) so I've renamed it to `supports_constraints`.

cmacmackin:
Looks good, except that I see you didn't add a `residual()` method to `FieldPerp`. Is there a reason for that?

bendudson:
Hi @cmacmackin ! `FieldPerp` doesn't have a `ddt()` function so I didn't add `residual()`

cmacmackin:
> Hi @cmacmackin ! `FieldPerp` doesn't have a `ddt()` function so I didn't add `residual()`

Ah, that would be a good reason!

cmacmackin (approved):

PR #3423, opened 2026-07-08
Add non const g_22_ylow/yhigh

dschwoerer:
Fix needed for https://github.com/boutproject/hermes-3/pull/551

I am wondering whether it might be easier / cleaner to provide instead `normaliseFCI(BoutReal rho_s)` instead of the two non-const functions.


dschwoerer:
Replaced by #3463

PR #3424, opened 2026-07-08
Add API to normalise metric for FCI

dschwoerer:
Alternative to #3423 

It might be nice to do something similar for FA, but I think there are more variants around.


ZedThree:
The coordinates refactor is almost ready which will change how this needs to be implemented, so let's hold off momentarily

dschwoerer:
Replaced by #3463

PR #3429, opened 2026-07-09
Remove \\ at end of comments

dschwoerer:
This was partially broken by clang-format.
It also make the compiler unhappy, that warns about that.

I think you did that, @ZedThree , is that needed or can it be reverted?


ZedThree:
It's for the LaTeX rendering of the equations in the docstring -- not super important, and I'd rather break that than have a warning. Might need something silly like `\\\\`

dschwoerer:
I think this should work now - the `\\` are still there, and as long as it is not at the EOL, the compiler should also be happy about it.
Clang-format should also not move them around - in order to align the last `\`

@ZedThree Installing check-github-workflows does not seem to work with python3.15 :-(

ZedThree (approved):

PR #3428, opened 2026-07-09
Fixes and unit tests for Field3DParallel

dschwoerer:
The templated expression broke a significant bit of the Field3DParallel implementation.

This forces to to not use the templates for Field3DParallel. It might be nice to fix them instead, but that takes more effort.

It also adds unit tests to avoid future regressions.

PR #3430, opened 2026-07-10
Template expressions with Field3DParallel

bendudson:
Enables expressions with `Field3DParallel` to be merged using templates.
- If a `BinaryExpr` is assigned to a `Field3D` then no calculations are performed on the parallel slices and the result will not have parallel slices.
- If a `BinaryExpr` is assigned to a `Field3DParallel` then if `isFci()` the parallel slices will be calculated. If `CHECK >= 2` then the assignment will assert that it has slices.
- scalars "broadcast" across slices.
- The `pow` and `floor` functions are now lazy.
- Eager arithmetic operators removed from `generated_fieldops`. In-place operators and `FieldPerp` operators are still generated (for now).

Unit tests taken from #3428 are passing.

ToDo:
- [ ] Check that the `BinaryExpr` changes don't affect performance on GPUs, especially CUDA builds. `yup()` and `ydown()` members and Field3D pointers have been added.
- [ ] Quantify performance change on CPU (optimized) builds.


include/bout/fieldops.hxx
```diff
@@ -381,6 +383,43 @@ struct BinaryExpr {
...
+      return lhs.numberParallelSlices();
+    } else {
+      ASSERT2(lhs.numberParallelSlices() == rhs.numberParallelSlices());
+      return lhs.numberParallelSlices();
+    }
+  }
+  auto yup(int slice = 0) const {
+    return BinaryExpr<ResT, L, R, Func>{lhs.yup(slice), rhs.yup(slice), f,
+                                        mesh,           location,       directions,
+                                        regionID,       indices,        yindex};
```
dschwoerer: We should probably compute / lookup the regionID for the overlap of the lhs and rhs regionID's. I assume regionID is the regionID of this.

src/field/gen_fieldops.py
```diff
@@ -365,6 +380,9 @@ def returnType(f1, f2):
                 "out": out,
                 "lhs": lhs,
                 "rhs": rhs,
+                "emit_free_binary": emit_free_binary_wrapper(out),
+                "emit_update_inplace": emit_update_inplace(lhs, out),
+                "emit_member_op_equals": emit_member_operator_equals(lhs, out),
```
dschwoerer, line 385: It would probably be nicer to have matching var and function names.

dschwoerer:
> Enables expressions with `Field3DParallel` to be merged using templates.
> 
>     * If a `BinaryExpr` is assigned to a `Field3D` then no calculations are performed on the parallel slices and the result will not have parallel slices.

I am not a user of this, but do not some derivatives use the parallel slices, even without FCI?

>     * If a `BinaryExpr` is assigned to a `Field3DParallel` then if `isFci()` the parallel slices will be calculated. If `CHECK >= 2` then the assignment will assert that it has slices.

This is probably a much cleaner design than what I had initially :+1: 

>     * Eager arithmetic operators removed from `generated_fieldops`. In-place operators and `FieldPerp` operators are still generated (for now).

It is great to not have functions that should never be used :-)

> Unit tests taken from #3428 are passing.
> 
> ToDo:
> 
>     * [ ]  Check that the `BinaryExpr` changes don't affect performance on GPUs, especially CUDA builds. `yup()` and `ydown()` members and Field3D pointers have been added.

We probably could get rid of the Field3D pointers, if we would ensure that the parallel slices have consecutive memory allocated.
I.e. `yup(0)[Ind3D(0)]` points to `BoutReal ptr1`, and `yup(1)[Ind3D(0)]` points to `ptr1 + nx * ny * nz`. That would need a change to the allocation routine, but would be doable if it helps with optimisation.

dschwoerer (changes requested):
I think the regionID needs fixing, rest should be good.

Thanks a lot, @bendudson

totork:
@bendudson @dschwoerer I am sorry for pressing here a little bit, but is it possible to merge this quite soonish? Having to communicate most arguments of operators makes the code quite bloated and introduces inconsistencies at many places.

dschwoerer:
@totork I fixed the regionID, I think it should be good. There is some cleaning up we could do, but that can probably wait. Lets merge this, and work on this later if we have time.

dschwoerer (dismissed):

dschwoerer (approved):
Thanks @bendudson!

PR #3431, opened 2026-07-10
Fix bad PETSc cleanup

ZedThree:
`PetscVector`, `PetscMatrix`, and `LaplacePetsc` were not cleaning up their owned PETSc objects correctly, which can lead to some `MPI_Comm_free` errors on the XZ-communicator. I don't really understand exactly what the issue is, I think it's something to do with PETSc being over-eager to clean up (or not cleaning up enough?).


ZedThree:
Tobias confirmed this is working for him as well

cmacmackin (approved):

PR #3433, opened 2026-07-10
cpp20 concepts

tomc271:


include/bout/sys/type_name.hxx
```diff
@@ -8,8 +8,10 @@
 #include "bout/bout_types.hxx"
 
 #include <string>
+#include <type_traits> // Required for std::is_same_v
```
dschwoerer: I think the comment can go. clang-tidy will keep it up to date.
```suggestion
#include <type_traits>
```

CMakeLists.txt
```diff
@@ -388,9 +387,7 @@ set(BOUT_SOURCES
     ./src/sys/range.cxx
     ./src/sys/slepclib.cxx
     ./src/sys/timer.cxx
-    ./src/sys/type_name.cxx
-    ./src/sys/utils.cxx
-    ./include/bout/git_metadata.hxx
+        ./src/sys/utils.cxx
```
dschwoerer: Has the formatting gone wrong here?

```suggestion
    ./src/sys/utils.cxx
```

dschwoerer (commented):
Looks good, just some minor cleanups would be nice.

include/bout/where.hxx
```diff
@@ -31,14 +31,18 @@
...
 /// @param[in] test   The value which determines which input to use
 /// @param[in] gt0    Uses this value if test > 0.0
 /// @param[in] le0    Uses this value if test <= 0.0
-template <class T, class U, class V,
-          class ResultType = typename bout::utils::EnableIfField<T, U, V>>
+
+// Overload 1: Three fields
+template <typename T, typename U, typename V,
+          typename ResultType = std::common_type_t<T, U, V>>
+requires IsField<T>&& IsField<U>&& IsField<V>
```
ZedThree: @tomc271 Is there a reason we can't do:
```suggestion
template <IsField T, IsField U, IsField V,
          typename ResultType = std::common_type_t<T, U, V>>
```

Is it because `T` etc in this case are concepts, and `common_type_t` takes templates?

include/bout/sys/type_name.hxx
```diff
@@ -23,44 +25,39 @@ namespace utils {
 
 template <typename T>
 std::string typeName() {
-  return typeid(T).name();
+  if constexpr (std::is_same_v<T, bool>)
+    return "bool";
+  else if constexpr (std::is_same_v<T, int>)
+    return "int";
```
ZedThree: Could you run clang-tidy/clang-format over this? We should always have braces with `if`, and not have `else` following `return`

ZedThree (changes requested):
A couple of little things to tidy up, otherwise looks great!

ZedThree:
You might need to bump `Standard:        c++20` in `.clang-format` btw, otherwise you'll get some weird formatting, I've just discovered!

tomc271:
Replaced by #3501

PR #3435, opened 2026-07-10
cpp20 if with initializer

tomc271:


ZedThree (approved):
CI failures are due to PR coming from a fork

PR #3439, opened 2026-07-16
Hypre solver eliminate boundaries (Redux)

bendudson:
Aims to import hypre convergence by eliminating boundary rows.

Previously merged in #3082 and reverted in #3136 because it failed round-trip tests.


bendudson:
Replaced by #3460

PR #3441, opened 2026-07-18
NVCC fixes

bendudson:
The NVidia C++ compiler is quite picky compared to LLVM and GCC. 

These fixes enable compilation with nvcc V12.9.41 on Perlmutter.

Ready for merging.

### DIFF
--- a/include/bout/assert.hxx
+++ b/include/bout/assert.hxx
@@ -19,7 +19,7 @@
 
 #include "bout/boutexception.hxx"
 
-#include <string_view>
+#include <string_view>  // IWYU pragma: export
 
 #ifndef CHECK
 #define CHECKLEVEL 0
--- a/include/bout/assert.hxx
+++ b/include/bout/assert.hxx
@@ -19,6 +19,8 @@
 
 #include "bout/boutexception.hxx"
 
+#include <string_view>
+
 #ifndef CHECK
 #define CHECKLEVEL 0
 #else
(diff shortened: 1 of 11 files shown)


include/bout/assert.hxx
```diff
@@ -19,48 +19,54 @@
 
 #include "bout/boutexception.hxx"
 
+#include <string_view>
```
ZedThree: clang-tidy is complaining because `std::string_view` is only used in a macro, so add this `include-what-you-use` pragma to silence the warning:
```suggestion
#include <string_view>  // IWYU pragma: export
```

include/bout/region.hxx
```diff
@@ -254,6 +254,7 @@ struct SpecificInd {
     case (DIRECTION::Z):
       return zp(dd);
     }
+    return *this;
```
ZedThree: Hmm, would be nicer if we could just use some `unreachable` thing here, as this is not really correct (and we check we've exhausted all enum values). C++23 has `std::unreachable` -- I think I added a macro to hermes that could be used, although I'm not sure if I tested it with nvcc. Worth using that here too?

src/sys/options.cxx
```diff
@@ -772,27 +770,90 @@ struct ConvertContainer<C<Scalar>> {
...
     throw BoutException(error_message);
   }
 
 private:
   std::string error_message;
   Container similar_to;
 };
+
+/// Visitor to convert an int, BoutReal or Array to the appropriate Array type.
+template <class Scalar, class Backing>
```
ZedThree, line 795: This is a bit disappointing we have to do this. Is the issue something to do with not picking up the second template argument with `C<scalar>`?

Can we write it explicitly, like:

```cpp
template <template <class, class> class C, class Scalar>
struct ConvertContainer<C<Scalar, ArrayData<Scalar>>> {
```

include/bout/fieldops.hxx
```diff
@@ -7,6 +7,7 @@
 #include "bout/bout_types.hxx"
 #include "bout/build_config.hxx"
 #include "bout/build_defines.hxx"
+#include "bout/build_config.hxx"
```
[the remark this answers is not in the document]
ZedThree: easy fix :)

ZedThree (changes requested):
Looks good, we can fix the clang-tidy warnings though

include/bout/region.hxx
bendudson: I've added a `BOUT_UNREACHABLE` macro to `build_config.hxx`

src/sys/options.cxx
bendudson: I think the issue is that `Array` has two template arguments but `Matrix` and `Tensor` only take one. GCC and LLVM happily use the default `Backing` template parameter but NVCC doesn't.

ZedThree: Ah, gotcha. That is unfortunate.

ZedThree (approved):

PR #3445, opened 2026-07-28
Ensure option names updated to match path

cmacmackin:
Previously, using the move-assignment operator for `Options` objects would keep the same full name as the moved object. This was problematic because it meant that the name of the newly-assigned `Options` object could differ from the path used to access that object. E.g.,

```c++
Options opts;
// Add various things in options
opts["a"]["b"] = opts["c"]["d"].copy();
fmt::print("{}\n", opts["a"]["b"].str()); // Output: c:d
```

This posed particular problems for the [permissions system](https://hermes3.readthedocs.io/en/latest/developer.html#permissions) in Hermes-3, as it meant that the name being checked against permissions could be incorrect.

In this PR I've modified the move-assignment operator to ensure that the assigned-to `Options` object (and all children) will have names consistent with the path to access them.

### DIFF
--- a/include/bout/options.hxx
+++ b/include/bout/options.hxx
@@ -934,6 +934,9 @@ private:
   bool similar(T lhs, T rhs) const {
     return lhs == rhs;
   }
+
+  /// Replaces the start of the name of an Option and any children
+  void recursively_update_names(size_t len, const std::string& new_prefix);
 };
 
 // Specialised assign methods for types stored in ValueType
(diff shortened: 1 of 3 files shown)


ZedThree (approved):

cmacmackin:
Thanks for the speedy turnaround on this, but unfortunately I just realised there is a corner-case I missed. I'll open another PR for that.

PR #3446, opened 2026-07-29
Cover corner-cases when updating option names

cmacmackin:
Following #3445, I realised that there were some corner cases in updating option names that I hadn't accounted for. This is related to when one or both of the operands have no name. In that case the path wasn't always updated correctly. This PR adds additional checks and logic to ensure everything works as it should.

### DIFF
--- a/src/sys/options.cxx
+++ b/src/sys/options.cxx
@@ -262,10 +262,21 @@ Options& Options::operator=(Options&& other) noexcept {
   children = std::move(other.children);
   value_used = other.value_used;
 
+  size_t len = other.full_name.size();
+  std::string new_prefix;
+  if (len == 0) {
+    new_prefix = full_name.empty() ? full_name : full_name + ":";
+  } else {
+    new_prefix = full_name;
+    // If new_prefix is empty but the old name is not then we also
+    // need to remove a colon when updating names
+    if (full_name.empty()) {
+      len += 1;
+    }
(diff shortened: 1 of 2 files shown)


src/sys/options.cxx
```diff
@@ -262,10 +262,21 @@ Options& Options::operator=(Options&& other) noexcept {
...
+  if (len == 0) {
+    new_prefix = full_name.empty() ? full_name : full_name + ":";
+  } else {
+    new_prefix = full_name;
+    // If new_prefix is empty but the old name is not then we also
+    // need to remove a colon when updating names
+    if (full_name.empty()) {
+      len += 1;
+    }
+  }
```
ZedThree: I found the nested conditionals a bit difficult to get my head around; flattening them makes it clear there's only two special cases to consider:

```suggestion
  std::string new_prefix = full_name;
  if (len == 0 and not full_name.empty()) {
    // Append section delimiter to new name
    new_prefix += ":";
  } else if (len != 0 and full_name.empty()) {
    // Remove extraneous section delimiter from old name
    len += 1;
  }
```

There's also a way to make both `new_prefix` and `len` `const`, but it gets a little ugly

ZedThree (changes requested):
One small clarification that I think helps (not tested)

ZedThree (approved):

PR #3447, opened 2026-07-30
Fix typos in `LaplacePetsc` after conversion

### LINKED ISSUE #3442: PETSC Laplacian Boundary condition
Hi all! 

I am currently fighting with some simulations that include laplace inversions with PETSC, specifically with frequent inversion fails when I want to run with the inner flag 2 (no local gradient) and the outer flag 16 (setting from value). I tried to look a little bit through the code and found the following code 

https://github.com/boutproject/BOUT-dev/blob/9eb94c7c9d9539c1e4e60ec84377a48d5ab50b22/src/invert/laplace/impls/petsc/petsc_laplace.cxx#L331-L341

Does this work as intended? I have to look a little bit more through the code to understand if there is a reason for this, but I found it counter-intuitive that i check the inner boundary when setting the outer boundary

ZedThree:
Fixes #3442


bshanahan (approved):

ZedThree:
I think this also impresses on us the need for much more comprehensive Laplacian solver tests :|

PR #3452, opened 2026-08-10
Add Conduction_Methods to Div_par_K_Grad_par_mod

dschwoerer:
Copied from Hermes-3: https://github.com/boutproject/hermes-3/pull/578

This makes it a runtime choice, using a templated function. It has currently has no options based control mechanism, so it needs to be selected by the API user, requiring changes in Hermes-3.

### DIFF
--- a/src/mesh/difops.cxx
+++ b/src/mesh/difops.cxx
@@ -377,7 +377,7 @@ Field3D Div_par_K_Grad_par(const Field3D& kY, const Field3D& f, CELL_LOC outloc)
 }
 
 namespace bout {
-enum ConductionMethod : uint8_t { Original, ProductJK, Harmonic };
+enum class ConductionMethod : uint8_t { Original, ProductJK, Harmonic };
 }
 
 namespace {
--- a/src/mesh/difops.cxx
+++ b/src/mesh/difops.cxx
@@ -374,8 +374,13 @@ Field3D Div_par_K_Grad_par(const Field3D& kY, const Field3D& f, CELL_LOC outloc)
          + Div_par(kY, outloc) * Grad_par(f, outloc);
 }
 
-Field3D Div_par_K_Grad_par_mod(const Field3D& Kin, const Field3D& fin, Field3D& flow_ylow,
-                               bool bndry_flux) {
+namespace bout {
+enum ConductionMethod : uint8_t { Original, ProductJK, Harmonic };
+}
+
+template <bout::ConductionMethod conduction_method>
+Field3D Div_par_K_Grad_par_mod_impl(const Field3D& Kin, const Field3D& fin,
+                                    Field3D& flow_ylow, bool bndry_flux) {
   ASSERT2(Kin.getLocation() == fin.getLocation());
 
   const Mesh* mesh = Kin.getMesh();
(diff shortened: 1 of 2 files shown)


src/mesh/difops.cxx
```diff
@@ -475,7 +534,25 @@ Field3D Div_par_K_Grad_par_mod(const Field3D& Kin, const Field3D& fin, Field3D&
 
   return result;
 }
+} // namespace
 
+Field3D Div_par_K_Grad_par_mod(const Field3D& Kin, const Field3D& fin, Field3D& flow_ylow,
+                               bool bndry_flux, const std::string& method) {
```
bendudson: I think I would prefer an enum runtime argument here, rather than passing a string and having to match strings each time. `Options` can do the conversion of input strings to BOUT++ enums. Remembering to omit the `i` in "orginal" is one reason :)

dschwoerer: I have switched to enums and copied your documentation :+1:

src/mesh/difops.cxx
```diff
@@ -475,7 +529,26 @@ Field3D Div_par_K_Grad_par_mod(const Field3D& Kin, const Field3D& fin, Field3D&
...
+    return Div_par_K_Grad_par_mod_impl<bout::ConductionMethod::ProductJK>(
+        Kin, fin, flow_ylow, bndry_flux);
+  }
+  if (method == bout::ConductionMethod::Harmonic) {
+    return Div_par_K_Grad_par_mod_impl<bout::ConductionMethod::Harmonic>(
+        Kin, fin, flow_ylow, bndry_flux);
+  }
+  throw BoutException(
+      "Unknown method `{}` - choose from `Original`, `ProductJK` or `Harmonic`.",
+      static_cast<int>(method));
```
ZedThree: We should be able to convert this to:

```suggestion
  using enum bout::ConductionMethod;

  switch(method) {
  case Original:
    return Div_par_K_Grad_par_mod_impl<Original>(Kin, fin, flow_ylow, bndry_flux);
  case ProductJK:
    return Div_par_K_Grad_par_mod_impl<ProductJK>(Kin, fin, flow_ylow, bndry_flux);
  case Harmonic:
    return Div_par_K_Grad_par_mod_impl<Harmonic>(Kin, fin, flow_ylow, bndry_flux);
  }
  BOUT_UNREACHABLE();
```

C++ doesn't have proper exhaustive enums, so we need `BOUT_UNREACHABLE()` to tell the compiler there are no other options

ZedThree (changes requested):
Nice, I like this idiom for giving us runtime control over compile-time options!

I've not checked the suggestion, so you might need to test it first

src/mesh/difops.cxx
dschwoerer: I think this is wrong though. If you call this like:
```
Div_par_K_Grad_par_mod(K, f, ylow, false, static_cast<bout::ConductionMethod>(10));
```
you get undefined behaviour.
This cannot happen with an exception.

ZedThree: Don't call it like that :)

We should only ever be calling this with either:
a) a `bout::ConductionMethod` variant explicitly, or
b) a value from `Options` which will have already been checked to be a valid variant

dschwoerer: I have not plans on doing that :-)

I still think the throw is cleaner. If you feel strongly about it, I can change it, but I would rather have the throw.

Also, that gives a somewhat helpful error, if a new method is added, but only to the enum and the templated function, not the runtime wrapper.

ZedThree: We actually get a compiler warning in that case (`-Wswitch`), which we can/should make a compile-time error.

I've just checked, either way should be fine, it's the presence of a `default` case that suppresses the compiler warning (so we should make sure to never use that with switches over enums!)

dschwoerer: Ok, then lets go with an exception. That also handles the cases of a user doing bad things.

ZedThree (approved):

src/mesh/difops.cxx
ZedThree: I do think that for switches over enums, we should handle bad enum values when we get them from the user (via `Options`), and then we can use `unreachable` everywhere. But I will try to enforce that consistently in another PR

dschwoerer: > b) a value from `Options` which will have already been checked to be a valid variant

@ZedThree 
How do I get an enum from Options?
Does it need to be a BOUT_ENUM?
I grepped for `.withDefault<>` but did not find any examples of that with enums ...

ZedThree: Yeah, the `BOUT_ENUM` macro adds an `Options::as<enumname>` specialisation

PR #3454, opened 2026-08-11
Update for python3.15 compartibilty

dschwoerer:
### Purpose

It allows to run prek with python3.15

### Change Summary

prek tools where updated

### Validation

This is run by CI and the tools have (hopefully) there on CI.

### AI Assistance

None

### Documentation

None, only minor.

### Review Notes

See here for arguments about not tracking uv.lock:
https://stackoverflow.com/a/79983524

See also https://github.com/boutproject/hermes-3/pull/631


ZedThree:
What's the actual issue here? Do the versions just need bumping?

dschwoerer:
There are two issues:
For python3.15 it is sufficient to just bump versions - PyO3 does otherwise not support python3.15 without extra compile time flags.

The other is, that uv.lock does change, depending on OS, python version and maybe others. I have one install where the uv.lock is always changed by installing prek packages, which is a bit of an annoyance. I heard from others too, that they have the issue. Its just annoying, but I thought, if it does not work as expected, removing it from git might be better.

ZedThree:
But removing this: `uv tool run sync-with-uv --check --diff` don't we now lose the ability to check that the prek versions are in sync with those in the `pyproject.toml`?

dschwoerer:
Maybe. In which case it is probably better to have a non matching uv.lock every now and then?

I will make a new PR to only update the dependencies.

ZedThree:
I mean, it's not that I particularly care about having `uv.lock` or not, but that I really want only one place to specify the tooling dependencies and their versions. We apparently do have to repeat the versions in the prek config file, so as long as we can run _something_ in CI that checks that they're in sync with `pyproject.toml`, I don't care

dschwoerer:
I restored the `uv.lock` - if it is needed to ensure pyproject.toml and the pre-commit hook stay aligned, than that is fine.

ZedThree:
Please can you update the corresponding versions in `pyproject.toml`? `uv.lock` should be generated from `pyproject.toml`, and then prek config checked against that

dschwoerer:
Is there a tool for that?
I was following https://bout-dev.readthedocs.io/en/latest/developer_docs/contributing.html#formatting-and-linters

ZedThree:
Yes, update `pyproject.toml` -> `uv sync --only-dev [--inexact]` to install dependencies and update `uv.lock` -> `uv tool run sync-with-uv` to update prek config

dschwoerer:
But the first step, update `pyproject.toml` - is this to be done manually?

dschwoerer:
I have updated it manually. If there is a better way, I am curious about that!

ZedThree (approved):

ZedThree:
It looks like uv might be getting a way to update `pyproject.toml` files, but it's not here just yet

PR #3460, opened 2026-08-14
Hypre solver eliminate boundaries (New API)

bendudson:
Replaces #3439, using a new API that doesn't require sharing boundary information between matrices, rhs and solution vectors. Avoids hidden mutations that could be hard to reason about.

Includes additional tests and some changes to the boundary row elimination code.

Written almost entirely by Codex/gpt-5.4 with review and iterations.


include/bout/hypre_interface.hxx
```diff
@@ -327,6 +549,8 @@ class HypreMatrix {
   std::vector<HYPRE_BigInt>* I;
   std::vector<std::vector<HYPRE_BigInt>>* J;
   std::vector<std::vector<HYPRE_Complex>>* V;
+  bool elimBE{false};
```
ZedThree: Is this something like `eliminate_boundaries`?

include/bout/hypre_interface.hxx
```diff
@@ -60,6 +70,139 @@ int checkHypreError(int error) {
...
+  HYPRE_Int* bjnum_array;
+  HYPRE_Int* bdep_array;
+  HYPRE_Complex* bii_array;
+  HYPRE_Complex* bij_array;
+  HYPRE_Int na;
+  HYPRE_Int* aoffset_array;
+  HYPRE_Int* aknum_array;
+  HYPRE_Complex* aki_array;
+  std::vector<HYPRE_Int> reduction_order;
+  std::vector<HYPRE_Int> expansion_order;
```
ZedThree, line 151: What's the reason some of these are raw pointers, and some are `std::vector`? Is this a device/host distinction?

include/bout/hypre_interface.hxx
```diff
@@ -697,8 +926,33 @@ public:
...
-    checkHypreError(
-        HYPRE_IJMatrixSetValues(*hypre_matrix, num_rows, num_cols, rawI, cols, vals));
+
+    // Eliminate boundary condition equations in hypre SetValues input arguments
+    if (elimBE) {
+      HYPRE_Int* bi_array;
+      HYPRE_Int* row_indexes;
+      // There must be an easier way to get nb
+      int nb = 0;
+      BOUT_FOR_SERIAL(i, index_converter->getRegionBndry()) { nb++; }
```
ZedThree: I think this is just:

```suggestion
      const int nb = index_converter->getRegionBndry().size();
```

src/sys/hypre_interface.cxx
```diff
@@ -0,0 +1,261 @@
...
+
+      aknums.push_back(aknum);
+      akis.push_back(aki);
+    }
+    aoffsets[bnum + 1] = static_cast<HYPRE_Int>(aknums.size());
+  }
+
+  na = static_cast<HYPRE_Int>(aknums.size());
+  HypreMalloc(aoffset_array, sizeof(HYPRE_Int) * (nb + 1));
+  std::copy(aoffsets.begin(), aoffsets.end(), aoffset_array);
```
ZedThree, line 151: Worth running [`clang-tidy -checks=modernize-use-ranges`](https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-ranges.html) on this file to see which algorithms can be upgraded :)

src/sys/hypre_interface.cxx
```diff
@@ -0,0 +1,261 @@
...
+    boundary_number_for_row.emplace(bi_array[bnum], bnum);
+  }
+
+  std::vector<std::vector<BoundaryOccurrence>> occurrences_by_boundary(nb);
+  for (HYPRE_Int rownum = 0; rownum < nrows; rownum++) {
+    const HYPRE_Int row_start = row_indexes[rownum];
+    for (HYPRE_Int m = 0; m < ncols[rownum]; m++) {
+      const auto boundary_position = boundary_number_for_row.find(cols[row_start + m]);
+      if (boundary_position != boundary_number_for_row.end()) {
+        occurrences_by_boundary[boundary_position->second].push_back({rownum, m});
```
ZedThree, line 64: I'm not sure what this is supposed to be doing, but should `occurrences_by_boundary` just be `std::set`?

I think this block could use a comment explaining what it's doing

src/sys/hypre_interface.cxx
```diff
@@ -0,0 +1,261 @@
...
+    }
+    return static_cast<HYPRE_Int>(std::distance(rows, row_position));
+  };
+
+  // Create the row_indexes array
+  HypreMalloc(row_indexes, sizeof(HYPRE_Int) * nrows);
+  row_indexes[0] = 0;
+  for (HYPRE_Int i = 1; i < nrows; i++) {
+    row_indexes[i] = row_indexes[i - 1] + ncols[i - 1];
+  }
```
ZedThree, line 34: I think this is just [`std::partial_sum`](https://cppreference.com/cpp/algorithm/partial_sum)? Sadly, there's no `std::ranges::partial_sum`, even in C++23 for some reason.

Probably I'm understanding what's going on here, but doesn't this entirely skip the information in the last element of `ncols`? The last value of `i` in this loop is `nrows - 1`, so the last access of `ncols` is at `nrows - 2`.

Ah, no, I see, the loop below at L61 uses `row_indexes[rownum]` as the start element, and `ncols[rownum]` as the end.

src/sys/hypre_interface.cxx
```diff
@@ -0,0 +1,261 @@
...
+#include <unordered_map>
+
+namespace bout {
+
+BoundaryElimination::BoundaryElimination(HYPRE_Int nrows, HYPRE_Int* ncols,
+                                         HYPRE_BigInt* rows, HYPRE_Int** row_indexes_ptr,
+                                         HYPRE_BigInt* cols, HYPRE_Complex* values,
+                                         HYPRE_Int nb, HYPRE_Int* bi_array)
+    : nb(nb) {
+  HYPRE_Int* row_indexes;
```
ZedThree, line 18: I think this constructor should be executed on the host, so we can use `vector` here?

```suggestion
  std::vector<HYPRE_Int> row_indexes;
  row_indexes.reserve(nrows);
```

src/sys/hypre_interface.cxx
```diff
@@ -0,0 +1,261 @@
...
+
+#include <numeric>
+#include <unordered_map>
+
+namespace bout {
+
+BoundaryElimination::BoundaryElimination(HYPRE_Int nrows, HYPRE_Int* ncols,
+                                         HYPRE_BigInt* rows, HYPRE_Int** row_indexes_ptr,
+                                         HYPRE_BigInt* cols, HYPRE_Complex* values,
+                                         HYPRE_Int nb, HYPRE_Int* bi_array)
```
ZedThree, line 16: I think most of these should at least be `const*`, but I don't think there's any need for pointers here? Most of them seem to be copied elementwise?

ZedThree (changes requested):
I can't judge the accuracy of the numerics, but I have a few comments on the code.

include/bout/hypre_interface.hxx
bendudson: Mostly yes. `std::vector` can't be used on host and arrays that are passed into Hypre should use HypreMalloc.

include/bout/hypre_interface.hxx
bendudson: It should be. I think this is eliminate Boundary Equations.

src/sys/hypre_interface.cxx
bendudson: The lifetime of this is somewhat complicated: `row_indexes` is allocated but then assigned to the "out" parameter `row_indexes_ptr`. Very C-like behavior. This could perhaps be tidied up in future.

include/bout/hypre_interface.hxx
ZedThree: Upsetting

ZedThree (approved):

PR #3461, opened 2026-08-17
Use pre-computed cell area and volume for Div_par_K_Grad_par

totork:
The current parallel diffusion operator does not use the stored cell distances, surfaces and volumes. This should adjust the operator to use them.


dschwoerer (approved):
Looks good 👍

bendudson (approved):
Thanks, @totork !

PR #3463, opened 2026-08-17
Add API to normalise metrics

dschwoerer:
With the encapsulation of the metrics, it got a bit more annoying to normalise the metrics.

Doing it in BOUT++ simplifies user code, but still allowing to do it in different ways, if you want something else.


ZedThree:
Is this different from [`TokamakCoordinates`](https://github.com/boutproject/BOUT-dev/blob/next/include/bout/tokamak_coordinates.hxx)?

Could this also be a free function, rather than a member of `Coordinates`?

dschwoerer:
> Is this different from [`TokamakCoordinates`](https://github.com/boutproject/BOUT-dev/blob/next/include/bout/tokamak_coordinates.hxx)?

Yes, this is from the Hermes-3 branch that uses the metrics provided, and just normalises them.
The `TokamakCoordinates` should probably be used to replace https://github.com/boutproject/hermes-3/blob/master/src/recalculate_metric.cxx

> Could this also be a free function, rather than a member of `Coordinates`?

No, because calling the setter on the metric tensor does recompute J and B. This is very wrong for FCI, and even for non-FCI not wanted. We want to just rescale those quantities.

ZedThree:
Ah, that's because we forgot to add an overload to `setMetricTensor()` to also take J and B. If you add that, then these could be free functions (or a method on `TokamakCoordinates`?), and we can avoid tying `Coordinates` to tokamaks.

src/mesh/coordinates.cxx
```diff
@@ -1277,6 +1278,81 @@ void Coordinates::communicateMetricTensor() {
...
+  // No we compute g_22_* - they must not be cleared. If they get
+  // cleared, they will be recomputed, but not normalised!
+  if (Bxy().isFci()) {
+    g_22_ylow();
+    g_22_yhigh();
+    ASSERT2(_g_22_ylow.has_value());
+    (*_g_22_ylow) /= rhoSQ;
+    ASSERT2(_g_22_yhigh.has_value());
+    (*_g_22_yhigh) /= rhoSQ;
+  }
```
dschwoerer: @ZedThree But then we need to expose an API for this?
I really think we should not do that.

ZedThree: What if we move them to `MetricTensor`, and then `TokamakCoordinates` can also take care of normalising them?

dschwoerer: We could, but that surely should be FCICoordinates?

Are we fine with MetricTensor having a pointer to the mesh, so that it can load from the grid?

ZedThree: I don't follow, sorry. `g_22_ylow/high` are computed for both FCI and non-FCI, and this normalisation is a particular tokamak one rather than FCI specific.

We could do that, or we could just load it along with the rest of the metric if it exists?

dschwoerer: > I don't follow, sorry. `g_22_ylow/high` are computed for both FCI and non-FCI,

Somewhat. For FCI, zoidberg does the computation using the actually magnetic field line length. On the BOUT++ side, it just gets loaded. For non-FCI we just interpolate.

> and this normalisation is a particular tokamak one rather than FCI specific.

It needs to be done for the FCI, as we load it, rather then interpolate. And I think I also put it correctly in the FCI branch.

> We could do that, or we could just load it along with the rest of the metric if it exists?

So you want to not do the lazy loading of that anymore? Should we then always throw for FCI if it is missing from the grid, even if the g_22_ylow is not used?

dschwoerer: If we move this to the metric and keep it lazy, the metric now needs to now about the parallel transform.

I am wondering whether all the extra work is worth it? Should we maybe just accept that it is probably easier to have the normalisation done by the coordinates, and have an argument to select what kind of normalisation you want? Then it could not be done by a derived class without issues, but we could still offer the user a choice beyond the FCI and Tokamak geometries?

src/mesh/tokamak_coordinates.cxx
```diff
@@ -75,4 +75,21 @@ TokamakCoordinates set_tokamak_coordinates(Mesh& mesh, BoutReal Lbar, BoutReal B
 
   return {Rxy, Zxy, Bpxy, Btxy, Bxy, hthe, I, I_unnormalised};
 }
+
+MetricNormaliser TokamakOrFCIMetricNormaliser(const Mesh* mesh, BoutReal Bnorm,
+                                              BoutReal rho_s0) {
+  if (mesh->isFci()) {
+    return {.g{SQ(rho_s0)}, .J{rho_s0 * rho_s0 * rho_s0}, .Bxy{Bnorm}};
```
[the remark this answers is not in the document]
dschwoerer: clang-tidy cannot decide. Either https://github.com/boutproject/BOUT-dev/pull/3463/commits/ae9b5e78ed21a8d028c89c48368c47e862bbcfa0 is really redundant, or it is not. This comment seems like https://github.com/boutproject/BOUT-dev/pull/3463/commits/ae9b5e78ed21a8d028c89c48368c47e862bbcfa0 is not redundant and should be reversed.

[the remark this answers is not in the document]
ZedThree: It seems this one is for compatibility with some 16 year old gcc behaviour (or bug, not sure).

One fix is to initialise the members with `= std::nullopt`. This is the same thing as using `{}` or no initial value, but seems to not trigger the redundant initialisation warning. I guess also it's pretty explicit about what's happening!

ZedThree:
Thanks @dschwoerer, I think this looks really clean!

src/mesh/tokamak_coordinates.cxx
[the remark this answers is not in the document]
dschwoerer: Ah, thanks. Updated :+1:

ZedThree (approved):

PR #3466, opened 2026-08-21
Add optional dissipation for Div_par_mod in Fci

totork:
The current Div_par_mod operator does not have any dissipation added in Fci. This PR adds a Rusanov flux. It is optional, so that it can be turned off in physics models for e.g. testing/MMS purposes.


dschwoerer (commented):
Should we add an assert that this flag does not get activated for FA? Otherwise users might be confused why the flag does nothing. It would be better to let them know that it does not work for FCI, as I think input files will (at least currently) not be used a lot to do FCI / FA comparisons.

Issue #3462, opened 2026-08-17
Parallel diffusion into boundaries


totork:
Right now, the parallel diffusion operator for Fci (including https://github.com/boutproject/BOUT-dev/pull/3461) does not differentiate between a cell with a boundary and without. Many physics models like Hermes-3 include the power flux into the sheath as a separate sink in the cell. As they also extrapolate variables like density or temperature at the sheath, this causes a finite gradient into the boundary and thus an additional gradient-driven flux. It thus double-counts the sheath. There is the need to disable the gradient-driven flux in the operator itself. This can be achieved by:

- [ ] Adding a member to coordinates that includes the information, whether a cell has a sheath boundary at the yhigh/ylow position. @dschwoerer Would the easiest way here to just follow the path of if it is done for e.g. coord->cell_volume() or is there a smarter way to do this?
- [ ] Adjusting the Div_par_K_grad_par_mod operator to set the gradient to zero, if there is a sheath present.

dschwoerer:
The first point is already done:
https://github.com/boutproject/BOUT-dev/blob/next/include/bout/yboundary_regions.hxx#L107

You can choose between template argument, runtime argument and function name to choose between up and down direction.

So we "just" need to change all the parallel derivative functions to respect the no-flux flag.

PR #3468, opened 2026-08-21
Add information about which cells have sheath boundaries

totork:
Currently, there is no straight forward way in Fci to determine if a cell has adjacent parallel sheath boundaries inside operators. This prevents e.g. turning of parallel diffusion into the boundaries, as explained in 

https://github.com/boutproject/BOUT-dev/issues/3462

The goal of this PR is to add this information as a member of coordinates. This is then easily accessible inside operators and can be used to adjust the calculation if needed. This is also a first try, I am not sure if there is an easier to way do this.

### DIFF
--- a/src/mesh/coordinates.cxx
+++ b/src/mesh/coordinates.cxx
@@ -1237,7 +1237,7 @@ void Coordinates::_determine_cell_sheath() const {
           if (pnt.dir() > 0) {
             sheath_yhigh[i] = 1;
           } else {
-            heath_ylow[i] = 1;
+            sheath_ylow[i] = 1;
           }
         }
       }
--- a/src/mesh/coordinates.cxx
+++ b/src/mesh/coordinates.cxx
@@ -1032,6 +1032,8 @@ void Coordinates::invalidateCellGeometryCaches() {
   _cell_area_zlow.reset();
   _cell_area_zhigh.reset();
   _cell_volume.reset();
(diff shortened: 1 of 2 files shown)


src/mesh/coordinates.cxx
```diff
@@ -1221,6 +1223,30 @@ void Coordinates::_compute_cell_volume() const {
...
+          } else {
+            sheath_ylow[i] = 1;
+          }
+        }
+      });
+      _cell_sheath_yhigh.emplace(sheath_yhigh);
+      _cell_sheath_ylow.emplace(sheath_ylow);
+    }
+  }
+}
```
dschwoerer, line 1248: I do not think we should duplicate the code from `YBoundary`.

Also, this prevents the user from configuring which y boundary actually is a sheath boundary.

I think it would be nicer if we would store a YBoundary object in coordinates, set it up properly, i.e. with an Options object passed in, and then have an API to get a shared_ptr to that YBoundary object.

Then we also do not need to duplicate the YBoundary API.

Also, I doubt we need to clear the cache in any case, I would not be aware that we allow to change any of the things in a mesh, which would make a recomputation of it necessary.

totork: I see, makes sense. So I can close this PR? Would you implement this? I am not sure what you mean with your suggestion.

dschwoerer:
Ah, it is already there, just the docs need fixing: #3469

PR #3469, opened 2026-08-21
Fix docs to changed signature of `getYBoundary`

dschwoerer:
`getYBoundary` is a free function and takes the coordinates as an argument.


ZedThree:
I'm not sure I understand why it is a free function? It accesses private members of `Coordinates` and is declared a `friend` -- why not just make it a member?

dschwoerer:
I do not remember why I did change it. But I think the docs should match the API at least ...

One advantage of a free function: It can be inlined, but that is probably not sufficient motivation to make it a free function, but maybe enough to keep it as is?

dschwoerer:
Replaced by #3474

PR #3471, opened 2026-08-21
Use C++20 "spaceship" operator

ZedThree:
C++20 can now synthesise all the relational operators from `operator<=>` (and `operator==` if `<=>` is not `= default`)


bendudson (approved):
Nice bit of tidying. Thanks!

PR #3474, opened 2026-08-24
Refactor boundary region iterators

ZedThree:
I've tried to improve the implementation of the new boundary region iterators. The main changes:

- convert the boundary condition methods to free functions
  - in the future, I want to convert all the actual boundary condition functions to use these as kernels, and having them as free functions should make that easier to extend to new kernels
- move the iterator classes into the the region classes
  - this helps clarify the relationship between the classes, and also simplifies the XY version, with the inner class being able to reuse the template parameter of the outer class
- use templates and concepts to reduce code duplication and improve error messages
  - compiler error messages aren't necessarily any shorter, but they can be a lot more precise when using concepts

Also fixes a couple of bugs due to missing functions in the derived classes


include/bout/boundary_region_iter.hxx
```diff
@@ -752,24 +752,21 @@ void iter_boundary(const Bndry& bndry, const Func& func) {
...
-begin(const bout::boundary::BoundaryRegionXY<isX>& reg) {
-  return bout::boundary::BoundaryRegionIterXY<isX>(&reg, true);
+inline auto begin(const BoundaryRegionXY<isX>& reg) {
+  return BoundaryRegionIterXY<isX>(&reg, true);
 }
 template <bool isX>
-inline bout::boundary::BoundaryRegionIterXY<isX>
-end(const bout::boundary::BoundaryRegionXY<isX>& reg) {
-  return bout::boundary::BoundaryRegionIterXY<isX>(&reg, false);
+inline auto end(const BoundaryRegionXY<isX>& reg) {
```
dschwoerer: `inline` can go, for a template?

src/mesh/coordinates.cxx
```diff
@@ -1221,8 +1221,13 @@ void Coordinates::_compute_cell_volume() const {
   }
 }
 
-std::shared_ptr<YBoundary> Coordinates::makeYBoundary(YBndryType type) const {
-  return std::make_shared<YBoundary>(type, localoptions, *localmesh);
+const bout::boundary::YBoundary& Coordinates::getYBoundary(YBndryType type) const {
+  using bout::boundary::YBoundary;
```
dschwoerer, line 1227: It is only used once in this function - should it just be spelled out where it is used?

include/bout/boundary_region_iter.hxx
```diff
@@ -199,11 +199,103 @@ private:
...
-/// Uses `BoundaryRegionIterFCI` as its iterator.
-///
 /// This can't use the legacy iteration methods (`first()`, `next()`, and so on)
 class BoundaryRegionFCI : public BoundaryRegionBase {
 public:
+  /// Iterator over a `BoundaryRegionFCI`
+  class Iterator : public BoundaryRegionIterBase<Iterator> {
+  private:
+    // TODO(dave) make non-const?
+    const BoundaryRegionFCI* region;
```
dschwoerer: If it is part of BoundaryRegionFCI then it should be non-const
```suggestion
    BoundaryRegionFCI* region;
```
It still must not call any non-const method, as that could be bad.

include/bout/boundary_region_iter.hxx
```diff
@@ -256,44 +262,31 @@ public:
...
-      return pos <=> rhs.pos;
+      return index.length;
     }
 
-    bool operator==(Iterator lhs) const {
-      ASSERT3(region == lhs.region);
-      return pos == lhs.pos;
-    }
+    auto operator<=>(const Iterator& rhs) const { return _ind() <=> rhs._ind(); }
+    bool operator==(Iterator rhs) const { return _ind() == rhs._ind(); }
```
dschwoerer: I think this is wrong. It implies that if the `ind()` is the same, the iterators are at the same point. But a point can have several boundaries.

include/bout/boundary_region_iter.hxx
```diff
@@ -256,44 +262,31 @@ public:
...
 
-    Iterator& operator++() {
-      ++pos;
-      return *this;
-    }
-
-    Iterator& operator*() { return *this; }
+    friend auto operator<=>(const Iterator& lhs, Ind3D ind) { return lhs._ind() <=> ind; }
+    friend bool operator==(Iterator lhs, Ind3D ind) { return lhs._ind() == ind; }
```
dschwoerer: Why do we need to compare to `Ind3D`? If we want to do this, I would rather we are explicit about it?

include/bout/boundary_region_iter.hxx
```diff
@@ -2,562 +2,420 @@
...
+  class Iterator : public BoundaryRegionIterBase<Iterator> {
+  private:
+    parallel_boundary_region::Indices index;
+    Mesh* localmesh_m;
+    int _dir;
+
+  public:
+    Iterator() = delete;
+    Iterator(parallel_boundary_region::Indices index, const BoundaryRegionFCI& rgn)
+        : index(index), localmesh_m(rgn.localmesh), _dir(rgn._dir) {}
```
dschwoerer: But that now works on a copy of index?
The point of setValid is that it needs to write through to the non-iterator. This avoids the cast, but breaks `setValid()` - it makes it essentially a no-op.

It seems there is no test that `setValid()` actually works.

dschwoerer (changes requested):
I think this hugely helps cleaning up the iterator!

Also moving operators into free functions makes it much more extendable, that is a huge plus.

There are a few things where I think it changes behaviour, and more tests would probably be good.

include/bout/boundary_region_iter.hxx
ZedThree: Ah, the name `Iterator` is now misleading, and we should change it to `Point` or something. This is no longer an iterator but actually a single point on the boundary, and we store a `std::vector` of these -- the iterator is now actually a `std::vector::iterator`. So when we call `setValid()` we modify the point itself.

I'll fix the names

src/mesh/coordinates.cxx
ZedThree: I think this was to prevent some horrible formatting :)

include/bout/boundary_region_iter.hxx
ZedThree: This is needed for the call to `std::lower_bound` in `contains`, which makes use of the elements being orderable and sorted. Previously, it was using the operators in `parallel_boundary_region::Indices`, but this PR changes the element type.

We can replace this with:

```cpp
    return std::ranges::find(bndry_points, ind,
                             [](const auto& point) { return point._ind(); })
           != std::end(bndry_points);
```

(and with C++23, this can just be `std::ranges::contains`).

The downside is this is a linear search (that is, $\mathcal{O}(N)$ comparisons), while `lower_bound` allows a binary search (at most $\log_2(N)$ comparisons). With C++20 we can now use `std::ranges` to apply a projection and avoid the need for the comparison operators between `Iterator` and `Ind3D`:

```cpp
    return std::ranges::binary_search(bndry_points, ind, {},
                                      [](const auto& point) { return point._ind(); });
```

include/bout/boundary_region_iter.hxx
ZedThree: I'm not sure there's another sensible choice? I guess we could store a `BoundaryParType` in there too?

Are we ever likely to compare points in separate boundaries?

dschwoerer: No, I meant a point can have a boundary in up and down direction. I think that is an implementation detail that right now they are in different vectors.

ZedThree: Oh right, so we just also need to compare `dir` too then, you mean?

include/bout/boundary_region_iter.hxx
```diff
@@ -183,16 +183,16 @@ private:
...
 class BoundaryRegionFCI : public BoundaryRegionBase {
 public:
-  /// Iterator over a `BoundaryRegionFCI`
-  class Iterator : public BoundaryRegionIterBase<Iterator> {
+  /// A single point in a `BoundaryRegionFCI`
+  class Point : public BoundaryRegionIterBase<Point> {
   private:
     parallel_boundary_region::Indices index;
     Mesh* localmesh_m;
     int _dir;
```
dschwoerer: If we duplicate `_dir` and `Mesh` for each point, should we not just move that to `parallel_boundary_region::Indices`?
Why do we need both?

The advantage of not duplicating that, is that the structure needs less memory, and thus improving performance. That was the reason to initially not duplicate it.

We could at least add `_dir` as signed char without extra cost to Indices.

We could even add 5 chars, without extra cost, so maybe we can get the bits from localmesh that we need, instead of a pointer to the mesh?

tests/unit/include/bout/test_boundary_iter.cxx
```diff
@@ -0,0 +1,35 @@
...
+  BoundaryRegionFCI bndry{"test", BndryLoc::yup, +1, bout::globals::mesh};
+
+  bndry.add_point(1, 0, 3, 0.0, 0.0, 0.0, 0.0, 0, 0);
+  bndry.add_point(1, 1, 3, 0.0, 0.0, 0.0, 0.0, 0, 0);
+  bndry.add_point(1, 2, 3, 0.0, 0.0, 0.0, 0.0, 0, 0);
+
+  ASSERT_TRUE(bndry.contains(1, 2, 3));
+  ASSERT_FALSE(bndry.contains(1, 3, 3));
+  ASSERT_FALSE(bndry.contains(3, 2, 1));
+}
```
dschwoerer, line 22:
```suggestion
TEST_F(BoundaryIteratorTest, Contains) {
  BoundaryRegionFCI bndry{"test", BndryLoc::yup, +1, bout::globals::mesh};

  bndry.add_point(1, 0, 3, 0.0, 0.0, 0.0, 0.0, 0, 0);
  bndry.add_point(1, 1, 3, 0.0, 0.0, 0.0, 0.0, 0, 0);
  bndry.add_point(1, 2, 3, 0.0, 0.0, 0.0, 0.0, 0, 0);

  ASSERT_TRUE(bndry.contains(1, 2, 3));
  ASSERT_FALSE(bndry.contains(1, 3, 3));
  ASSERT_FALSE(bndry.contains(3, 2, 1));
}

TEST_F(BoundaryIteratorTest, ContainsUnsorted) {
  BoundaryRegionFCI bndry{"test", BndryLoc::yup, +1, bout::globals::mesh};

  bndry.add_point(1, 0, 3, 0.0, 0.0, 0.0, 0.0, 0, 0);
  bndry.add_point(1, 1, 3, 0.0, 0.0, 0.0, 0.0, 0, 0);
  bndry.add_point(1, 2, 3, 0.0, 0.0, 0.0, 0.0, 0, 0);
  bndry.add_point(0, 1, 0, 0.0, 0.0, 0.0, 0.0, 0, 0);
  bndry.add_point(0, 1, 1, 0.0, 0.0, 0.0, 0.0, 0, 0);
  bndry.add_point(0, 1, 2, 0.0, 0.0, 0.0, 0.0, 0, 0);
  bndry.add_point(0, 1, 3, 0.0, 0.0, 0.0, 0.0, 0, 0);
  bndry.add_point(0, 0, 0, 0.0, 0.0, 0.0, 0.0, 0, 0);
  bndry.add_point(0, 0, 1, 0.0, 0.0, 0.0, 0.0, 0, 0);
  bndry.add_point(0, 0, 2, 0.0, 0.0, 0.0, 0.0, 0, 0);
  bndry.add_point(0, 0, 3, 0.0, 0.0, 0.0, 0.0, 0, 0);

  ASSERT_TRUE(bndry.contains(1, 2, 3));
  ASSERT_FALSE(bndry.contains(1, 3, 3));
  ASSERT_FALSE(bndry.contains(3, 2, 1));
}
```

Maybe also an unsorted test, to make sure we are sorting?

dschwoerer (dismissed):
I am worried about performance. That will only be measurable for large boundary regions though, if they do not fit in cache easily.

But then, if we really want to improve performance we should switch to a struct of vectors, rather then keep a vector of structs ...

include/bout/boundary_region_iter.hxx
ZedThree: We need the `Mesh*` to create a `generator::Context` in `BoundaryOpPar::getValue`.

But yes, `Point` and `Indices` could be merged -- but also so could all of this and `BoundaryRegionPar`. I will see how easy that merge is, otherwise I will put it off for now or it will be difficult to know where to stop!

ZedThree:
I'm not concerned about performance -- boundaries should only be a small part of the work, both in terms of number of points, and also computation. If your `rhs` does less work than applying a dirichlet BC, you're probably not worried about performance.

I also don't think this should change the cache behaviour anyway. Previously, we had:

```cpp
Ind3D _ind() const { return region->bndry_points[pos].index; }
```

Every time we called `_ind()`, we would have to fetch a whole cache line around `bndry_points[pos]` (i.e. an `Indices`). Using `Indices` directly will give exactly the same behaviour.

Anyway, we will be dominated by cache misses from loading the `Field` values, and seeing as we don't (yet) have boundaries in Z means we're likely to have a cache miss at every single boundary point (at least in X, possibly not in Y, depending on the size of the grid).

I guess you're also making a point about padding. In this PR, we now have:

```cpp
  std::println("size of Point: {}", sizeof (bout::boundary::BoundaryRegionFCI::Point));
  std::println("size of Indices: {}", sizeof (bout::parallel_boundary_region::Indices));
```

```
size of Point: 80
size of Indices: 56
```

So you're correct, we can get some reduction in size from just using `Indices` directly.

Putting a `Mesh* mesh` and an `int dir` into `Indices` it inflates to 64 bytes. We can fit 512 points into 32 kiB of L1 cache, which seems like plenty!

include/bout/boundary_region_iter.hxx
```diff
@@ -1,563 +1,442 @@
...
+    signed char valid_m;
+    /// How many points we are away from the boundary
+    signed char offset_m;
+    unsigned char abs_offset_m;
+
+    Mesh* localmesh_m;
+    /// Direction to the boundary. Positive values are in the sense of the
+    /// increasing coordinate
+    int dir_m;
+
```
dschwoerer:
```suggestion
    Ind3D index;

    Mesh* localmesh_m;

    /// Intersection with boundary in index space
    RealPoint intersection;
    /// Distance to intersection
    BoutReal length_m;

    /// Direction to the boundary. Positive values are in the sense of the
    /// increasing coordinate
    int dir_m;
    /// How many points we can go in the opposite direction
    signed char valid_m;
    /// How many points we are away from the boundary
    signed char offset_m;
    unsigned char abs_offset_m;
```
I think we should be able to move things around for better alignment?

ZedThree: I'm happy to do this, but I don't think we'll see any real benefit!

dschwoerer (approved):

PR #3476, opened 2026-08-27
Add unit tests for normalisation

dschwoerer:


tests/unit/mesh/test_tokamak_coordinates.cxx
```diff
@@ -0,0 +1,46 @@
...
+#include "fake_mesh_fixture.hxx"
+#include "test_extras.hxx"
+
+using bout::globals::mesh;
+
+using TokamakCoordinatesTestFCI = FakeMeshFixtureFCI;
+
+class TokamakCoordinatesTest : public FakeMeshFixture {
+public:
+  using FieldMetric = Coordinates::FieldMetric;
```
ZedThree: Is this used?

```suggestion
```

src/mesh/coordinates.cxx
```diff
@@ -1048,7 +1048,11 @@ void Coordinates::invalidateMetricCaches() {
   g_values_cache.reset();
   Grad2_par2_DDY_invSgCache.clear();
   invSgCache.reset();
-  jacobian_cache.reset();
+  if (jacobian_cache != nullptr) {
+    if (not jacobian_cache->isFci()) {
+      jacobian_cache.reset();
```
ZedThree: Please could you add a comment explaining why we don't want to reset the jacobian in this case? I guess this is something to do with not wanting to recalculate the slices? But if we're calling this, then we've changed the metric, so how do we ensure the jacobian is in sync with the metric?

Also we can collapse the `if` statements into one :)

dschwoerer: > Please could you add a comment explaining why we don't want to reset the jacobian in this case? I guess this is something to do with not wanting to recalculate the slices?

Exactly. We could do this, but it is rather bothersome.

> But if we're calling this, then we've changed the metric, so how do we ensure the jacobian is in sync with the metric?

We hope the user knows what they are doing. We also have no check that co and contra variant tensor are aligned. We could add a `CHECK > 2` check to ensure J is somewhat consistent (and maybe co and contravariant tensors, too)

ZedThree:
Oh, is the actual problem just the ordering of calls?

https://github.com/boutproject/BOUT-dev/blob/61ca9f191276dfd0b7f11fb893a4e0dfc6db6af4/src/mesh/coordinates.cxx#L1284-L1286

https://github.com/boutproject/BOUT-dev/blob/61ca9f191276dfd0b7f11fb893a4e0dfc6db6af4/src/mesh/coordinates.cxx#L1314-L1337

Should we just move the call to `invalidateMetricCaches()` to _before_ calling `setJ`?

bendudson (dismissed):

bendudson (approved):

PR #3477, opened 2026-08-27
Some improvements for FCI

dschwoerer:
Summary:

* By taking a Field3DParallel, if we want to take derivatives, user code mostly just works for FCI
* `.asField3DParallel()` is mostly not needed any more
* Introduce `bout::FieldMetricParallel`
```
#if BOUT_USE_METRIC_3D
Field3DParallel sqrtB = sqrt(Bxy);
#else
Field2D sqrtB = sqrt(Bxy);
#endif
```
can be replaced by
```
bout::FieldMetricParallel sqrtB = sqrt(Bxy);
```

### DIFF
--- a/include/bout/difops.hxx
+++ b/include/bout/difops.hxx
@@ -77,7 +77,7 @@ Field3D Grad_par(const Field3DParallel& var, CELL_LOC outloc = CELL_DEFAULT,
  * Combines the parallel and perpendicular calculation to include
  * grid-points at the corners.
  */
-Field3D Grad_parP(const Field3D& apar, const Field3D& f);
+Field3D Grad_parP(const Field3D& apar, const Field3DParallel& f);
 
 /*!
  * vpar times parallel derivative along unperturbed B-field (upwinding)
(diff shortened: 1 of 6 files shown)


ZedThree (dismissed):

ZedThree:
Looks good! Before merging this, please could you make a note of these types in the docs? I can see we've got a mention of `Field3DParallel` and `Coordinates::FieldMetric` (which should now be `s/Coordinates/bout/`) in `field_expressions.rst` -- maybe something there about `FieldMetricParallel` too?

bendudson (approved):
Thanks @dschwoerer !
The change from `Coordinates::FieldMetric` to `bout::FieldMetric` is probably good, but will require downstream changes.

dschwoerer:
> The change from `Coordinates::FieldMetric` to `bout::FieldMetric` is probably good, but will require downstream changes.

 `Coordinates::FieldMetric` still works, and can be used. `bout::FieldMetric` is however the new and preferred way, so we have time to adopt. Not sure we will drop the old way, certainly no rush to port things for now.

PR #3478, opened 2026-08-27
Add DAGP_fv for FCI

dschwoerer:
Add the DAGP_fv operator needed for https://github.com/boutproject/hermes-3/pull/643


include/bout/fv_ops.hxx
```diff
@@ -185,5 +187,37 @@ Field3D Div_par_fvv_heating(const Field3D& f_in, const Field3D& v_in,
 template <typename CellEdges = MC>
 Field3D Div_a_Grad_perp_limit(const Field3D& a, const Field3D& g, const Field3D& f);
 
+/// Div ( a g Grad_perp(f) )
+///
+/// This version uses pre-computed coefficient. It can also be used
```
ZedThree: I think there's some text missing?

It could definitely do with some more docs, I guess that `g` is precomputed here given that the call operator only takes `a` and `f`?

Certainly we need something about how to create one

include/bout/fv_ops.hxx
```diff
@@ -185,5 +187,37 @@ Field3D Div_par_fvv_heating(const Field3D& f_in, const Field3D& v_in,
...
+class dagp_fv {
```
ZedThree: This is in the `FV` namespace, right? So we can drop the `_fv` (unless it doesn't stand for finite volume here?)

I'm also not a big fan of the name -- all our other operators are spelt out (well, most of them. I guess `Delp2` is not great!) which helps the readability of the physics models.

So can we go for:


```suggestion
class Div_a_g_Grad_perp {
```

include/bout/fv_ops.hxx
```diff
@@ -185,5 +187,37 @@ Field3D Div_par_fvv_heating(const Field3D& f_in, const Field3D& v_in,
...
+  Field3D operator()(const Field3D& a, const Field3D& f, Field3D& low_xlow,
+                     Field3D& flow_zlow, bool upwinding);
+  Field3D operator()(const Field3D& a, const Field3D& f, bool upwinding);
+  dagp_fv(Mesh& mesh);
+  dagp_fv& operator*=(BoutReal fac);
+  dagp_fv& operator/=(BoutReal fac);
+
+private:
+  template <bool extra, bool upwinding>
+  Field3D operator()(const Field3D& a, const Field3D& f, Field3D* low_xlow,
```
ZedThree, line 218: `const`?

include/bout/fv_ops.hxx
```diff
@@ -185,5 +187,37 @@ Field3D Div_par_fvv_heating(const Field3D& f_in, const Field3D& v_in,
...
+  Field3D fac_ZX;
+  Field3D fac_ZZ;
+  Field3D volume;
+
+  bool isNormalised{false};
+
+  template <bool upwinding>
+  BoutReal xflux(const Field3D& a, const Field3D& f, const Ind3D& i);
+  template <bool upwinding>
+  BoutReal zflux(const Field3D& a, const Field3D& f, const Ind3D& i);
```
ZedThree: `const`?

src/mesh/fv_ops.cxx
```diff
@@ -670,4 +675,124 @@ template Field3D Div_par_fvv_heating<WENO3>(const Field3D& f_in, const Field3D&
...
+  }
+  result.setRegion("RGN_NOBNDRY");
+  if constexpr (extra) {
+    flow_xlow->setRegion("RGN_NOBNDRY");
+    flow_zlow->setRegion("RGN_NOBNDRY");
+    *flow_xlow /= volume;
+    *flow_zlow /= volume;
+  }
+  result /= volume;
+  return result;
```
ZedThree:
```suggestion
  return result / volume;
```

include/bout/fv_ops.hxx
```diff
@@ -185,5 +187,37 @@ Field3D Div_par_fvv_heating(const Field3D& f_in, const Field3D& v_in,
...
 Field3D Div_a_Grad_perp_limit(const Field3D& a, const Field3D& g, const Field3D& f);
 
+/// Div ( a g Grad_perp(f) )
+///
+/// This version uses pre-computed coefficient. It can also be used
+class dagp_fv {
+public:
+  Field3D operator()(const Field3D& a, const Field3D& f, Field3D& low_xlow,
+                     Field3D& flow_zlow, bool upwinding);
+  Field3D operator()(const Field3D& a, const Field3D& f, bool upwinding);
```
ZedThree: Should these be `const`?

`low_xlow` and `flow_zlow` need some docs -- not sure that `flow` is a great name either, I think it's `f_low` rather than `flow`?

src/mesh/fv_ops.cxx
```diff
@@ -670,4 +675,124 @@ template Field3D Div_par_fvv_heating<WENO3>(const Field3D& f_in, const Field3D&
...
+  const auto dx = 0.5 * (f[i.xp()] - f[i.xm()] + f[izp.xp()] - f[izp.xm()]);
+  return -((fac_ZX[i] * dx) + (fac_ZZ[i] * dz)) * av;
+}
+dagp_fv& dagp_fv::operator*=(BoutReal fac) {
+  ASSERT2(not isNormalised);
+  isNormalised = true;
+  volume /= fac * fac;
+  return *this;
+}
+dagp_fv& dagp_fv::operator/=(BoutReal fac) { return operator*=(1 / fac); }
```
ZedThree: Is this used anywhere? Or is it anticipated to be used in user code?

src/mesh/fv_ops.cxx
```diff
@@ -670,4 +675,124 @@ template Field3D Div_par_fvv_heating<WENO3>(const Field3D& f_in, const Field3D&
                                             Field3D& flow_ylow, bool fixflux);
 template Field3D Div_a_Grad_perp_limit<WENO3>(const Field3D& a, const Field3D& g,
                                               const Field3D& f);
+
+namespace {
+std::map<Mesh*, std::weak_ptr<dagp_fv>> dagp_fv_cache;
+}
+
+std::shared_ptr<dagp_fv> getDagp_fv(Mesh* mesh, BoutReal rho_s0) {
```
ZedThree: If this were instead a `static` method on the class (and named something like `create`?), could we inline `operator*=` here and drop it? Or is it also needed elsewhere?

ZedThree (changes requested):
Looks good, a few comments though. Also:

- please expand the docs
- needs tests -- should be pretty easy to add to `fv_mms`?
- please can we change the name? :)

include/bout/fv_ops.hxx
dschwoerer: They are the flow in x direction and the flow in z direction. Do you have a better name?

include/bout/fv_ops.hxx
dschwoerer: I copy pasted and did not notice the `g` - that should go!

It needs the coefficients in the grid. Zoidberg does add them.

That also means most of the testing is actually testing of zoidberg, as that does most of the computation, BOUT++ just applies the weights.

dschwoerer:
> Looks good, a few comments though. Also:
> 
>     * please expand the docs

I added some, if you want more, please let me know what.

>     * needs tests -- should be pretty easy to add to `fv_mms`?

No, it needs zoidberg to create the coefficients. Zoidberg has MMS tests, so a unit test should be fine?

>     * please can we change the name? :)

Sure, do you have a good suggestion?

bshanahan:
> > ```
> > * please can we change the name? :)
> > ```
> 
> Sure, do you have a good suggestion?

`Div_a_Grad_perp_FV_FCI` ?

bendudson (approved):

PR #3480, opened 2026-08-27
Mesh topology

wearysebas:
Created the MeshTopology function that identifies the topology from the branch cuts at the beginning of the run. Works with Single Null, Connected Double-Null, Unconnected Double-Null, and closed field lines.


ZedThree:
I think you need to rebase off a more recent `next` or something -- there's way too many unrelated changes!

PR #3481, opened 2026-08-27
Expression `unit_integral` to normalize sources

bendudson:
In expressions `unit_integral(expr)` will normalise the given `expr` so that the volume integral over the domain is 1. This is intended to make specifying sources easier, removing the need to manually scale sources to the required power or fuelling rate. This is a common confusion and errors when running source-driven cases.

Some care will still be needed when using this: The integral will use the normalized metrics so the input expression needs to normalise the volume; If a fraction of the torus is being simulated then that needs to be corrected for. 

An alternative would be to implement in Hermes-3 more custom inputs e.g.
```
full_torus_power = 1e6 # Heating power for a full torus i.e. a 1/4 torus simulation would receive 1/4 of this power
power_shape = is_y_periodic * exp(-x^2) # The shape of the power deposition profile.
```
The `unit_integral()` function is intended as a more general solution.


tests/unit/field/test_field_factory.cxx
```diff
@@ -29,6 +30,43 @@
...
+BoutReal volumeIntegral(const Field3D& field) {
+  const auto* coords = field.getCoordinates();
+  if (coords == nullptr) {
+    throw BoutException("Field has no coordinates");
+  }
+
+  BoutReal local = 0.0;
+  BOUT_FOR(i, field.getRegion("RGN_NOBNDRY")) {
+    local +=
+        field[i] * coords->J()[i] * coords->dx()[i] * coords->dy()[i] * coords->dz()[i];
```
ZedThree, line 43:
```suggestion
        field[i] * coords->cell_volume[i];
```

tests/unit/field/test_field_factory.cxx
```diff
@@ -29,6 +30,43 @@
...
+  auto* coords = localmesh->getCoordinates(location);
+  if (coords == nullptr) {
+    throw BoutException("Mesh has no coordinates");
+  }
+
+  BoutReal local = 0.0;
+  Field3D ones{localmesh};
+  ones.allocate();
+  BOUT_FOR(i, ones.getRegion("RGN_NOBNDRY")) {
+    local += coords->J()[i] * coords->dx()[i] * coords->dy()[i] * coords->dz()[i];
```
ZedThree, line 61:
```suggestion
    local += coords->cell_volume[i];
```

tests/unit/field/test_field_factory.cxx
```diff
@@ -29,6 +30,43 @@
 using namespace bout::globals;
 using bout::generator::Context;
 
+namespace {
+BoutReal volumeIntegral(const Field3D& field) {
```
ZedThree, line 34: I'm surprised that we don't already have a `sum(Field3D)` given that we have `mean` -- if we add that, then these two functions are just `sum(field * coords->cell_volume())` and `sum(coords->cell_volume())`.

We could then use this in the generator implementation?

src/field/fieldgenerators.cxx
```diff
@@ -47,6 +48,69 @@ BoutReal FieldHeaviside::generate(const Context& ctx) {
...
+
+  cached_mesh = localmesh;
+  cached_time = ctx.t();
+  cached_location = ctx.location();
+  cache_valid = true;
+}
+
+BoutReal FieldUnitIntegral::generate(const Context& ctx) {
+  if (!cacheMatches(ctx)) {
+    std::lock_guard<std::mutex> guard(cache_mutex);
```
ZedThree, line 104: Is this AI by any chance? We don't have usual threads, we use OpenMP, so this should be `BOUT_OMP(single)` instead?

ZedThree (changes requested):
Looks good, a few comments

PR #3482, opened 2026-08-28
Normalisation fixes

dschwoerer:
Both commits are required to have the 2D-production test for Hermes-3. I tested locally only the 2nd patch, and in CI the first patch only. Both failed on there own.

### DIFF
--- a/src/mesh/tokamak_coordinates.cxx
+++ b/src/mesh/tokamak_coordinates.cxx
@@ -81,14 +81,18 @@ MetricNormaliser TokamakOrFCIMetricNormaliser(const Mesh* mesh, BoutReal Bnorm,
   if (mesh->isFci()) {
     return {.g{SQ(rho_s0)}, .J{rho_s0 * rho_s0 * rho_s0}, .Bxy{Bnorm}};
   }
-  return {.g11{1 / SQ(Bnorm * rho_s0)},
+  return {.g11{SQ(Bnorm * rho_s0)},
+          .g11_mul{true},
           .g22{SQ(rho_s0)},
           .g33{SQ(rho_s0)},
-          .g12{1 / Bnorm},
-          .g13{1 / Bnorm},
+          .g12{Bnorm},
+          .g12_mul{true},
+          .g13{Bnorm},
+          .g13_mul{true},
           .g23{SQ(rho_s0)},
           .dx{rho_s0 * rho_s0 * Bnorm},
-          .J{rho_s0 / Bnorm},
+          .J{Bnorm / rho_s0},
+          .J_mul{true},
           .Bxy{Bnorm}};
 }
 
--- a/include/bout/coordinates.hxx
+++ b/include/bout/coordinates.hxx
@@ -564,17 +564,26 @@ std::string parallelSliceFieldName(std::string_view field, int offset);
 /// tensor, which is multiplied by the normalisation factor.
 struct MetricNormaliser {
   std::optional<BoutReal> g = std::nullopt;
+  bool g_mul = false;
   std::optional<BoutReal> g11 = std::nullopt;
+  bool g11_mul = false;
   std::optional<BoutReal> g22 = std::nullopt;
+  bool g22_mul = false;
   std::optional<BoutReal> g33 = std::nullopt;
+  bool g33_mul = false;
   std::optional<BoutReal> g12 = std::nullopt;
+  bool g12_mul = false;
   std::optional<BoutReal> g13 = std::nullopt;
+  bool g13_mul = false;
   std::optional<BoutReal> g23 = std::nullopt;
+  bool g23_mul = false;
   std::optional<BoutReal> dx = std::nullopt;
   std::optional<BoutReal> dy = std::nullopt;
   std::optional<BoutReal> dz = std::nullopt;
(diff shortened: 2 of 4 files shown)


src/mesh/tokamak_coordinates.cxx
```diff
@@ -81,14 +81,18 @@ MetricNormaliser TokamakOrFCIMetricNormaliser(const Mesh* mesh, BoutReal Bnorm,
   if (mesh->isFci()) {
     return {.g{SQ(rho_s0)}, .J{rho_s0 * rho_s0 * rho_s0}, .Bxy{Bnorm}};
   }
-  return {.g11{1 / SQ(Bnorm * rho_s0)},
+  return {.g11{SQ(Bnorm * rho_s0)},
+          .g11_mul{true},
```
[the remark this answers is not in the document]
dschwoerer: Not sure what clang-tidy is trying to do here, but g++ 15.2.0 is not happy with the suggested change.

src/mesh/tokamak_coordinates.cxx
```diff
@@ -81,14 +81,18 @@ MetricNormaliser TokamakOrFCIMetricNormaliser(const Mesh* mesh, BoutReal Bnorm,
...
           .g22{SQ(rho_s0)},
           .g33{SQ(rho_s0)},
-          .g12{1 / Bnorm},
-          .g13{1 / Bnorm},
+          .g12{Bnorm},
+          .g12_mul{true},
```
ZedThree, line 89: I don't understand this at all -- why is this not equivalent? Is it just loss of precision?

dschwoerer: Yes, and that seems to cause the failures in Hermes-3. If you think we should not do this, and just update the expected results, then we can do that, too. But I thought it might be good if we do not change results without a somewhat good reason, and the `_mul` additions do not seem like a huge addition of code.

dschwoerer:
For the record, without these changes, Hermes-3 fails with:
```
      Start 14: 2D-production
14/20 Test #14: 2D-production ..............................***Failed   12.69 sec

libgomp: Invalid value for environment variable OMP_NUM_THREADS: 
2D-Production test: download took 0.00 seconds
2D-Production test: zip extraction took 1.77 seconds
2D-Production test: simulation took 2.08 seconds
2D-Production test: loading results took 6.95 seconds
2D-production test: Failed in 11.24 seconds
2D-production test failures:

inner_lower_target [ddt(Pe)]:

Not equal to tolerance rtol=1e-05, atol=1e-08

Mismatched elements: 1 / 16 (6.25%)
Mismatch at index:
 [8]: 0.03679871893512578 (ACTUAL), 0.036799153418756995 (DESIRED)
Max absolute difference among violations: 4.34483631e-07
Max relative difference among violations: 1.1806892e-05
 ACTUAL: array([-0.050165, -0.067434, -0.095253, -0.140382, -0.18905 , -0.238565,
       -0.344202, -0.562915,  0.036799, -0.091465, -0.586319, -0.396325,
       -0.224284, -0.064263,  0.009454,  0.049401])
 DESIRED: array([-0.050165, -0.067434, -0.095253, -0.140382, -0.18905 , -0.238565,
       -0.344202, -0.562915,  0.036799, -0.091465, -0.586319, -0.396325,
       -0.224284, -0.064263,  0.009454,  0.049401])

sol_ring [ddt(Pe)]:

Not equal to tolerance rtol=1e-05, atol=1e-08

Mismatched elements: 1 / 50 (2%)
Mismatch at index:
 [0]: 0.03679871893512578 (ACTUAL), 0.036799153418756995 (DESIRED)
Max absolute difference among violations: 4.34483631e-07
Max relative difference among violations: 1.1806892e-05
 ACTUAL: array([ 0.036799, -1.645953, -1.76924 , -1.600019, -1.298552, -0.991483,
       -0.854056, -0.782628, -0.775318, -0.861349, -0.864964, -0.775591,
       -0.790239, -0.855372, -0.999383, -1.299618, -1.591131, -1.75628 ,...
 DESIRED: array([ 0.036799, -1.645953, -1.76924 , -1.600019, -1.298552, -0.991483,
       -0.854056, -0.782628, -0.775318, -0.861349, -0.864964, -0.775591,
       -0.790239, -0.855372, -0.999383, -1.299618, -1.591131, -1.75628 ,...

```

So it seems the rounding errors in doubles, cause differences in 1e-5 range.

dschwoerer:
There are rounding differences between `B/Bnorm` and `B * (1/Bnorm)` - they are numerically not the same. It seems such differences cause enough of a differences to cause larger differences in the end, i.e. the 2D-production test is very sensitive to such grid details

The PR address this by normalising just the same way that Hermes-3 used to normalise the metrics

ZedThree:
Hmm, if the test is really that sensitive, probably it needs its tolerances loosening instead.

The difference between `B / Bnorm` and `B * (1. / Bnorm)` should be around 1-2 [ULP](https://en.wikipedia.org/wiki/Unit_in_the_last_place), because there are only two floating point operations. That means a relative difference of about 1e-16. Obviously that's going to get compounded over time, but we shouldn't really be testing against differences in the grid of that level as it will tie us quite strictly to the exact implementation.

PR #3483, opened 2026-08-28
Honor bndry_flux for FCI

### LINKED ISSUE #3462: Parallel diffusion into boundaries

dschwoerer:
Resolves #3462

### DIFF
--- a/src/mesh/difops.cxx
+++ b/src/mesh/difops.cxx
@@ -44,6 +44,7 @@
 #include <bout/solver.hxx>
 #include <bout/unused.hxx>
 #include <bout/utils.hxx>
+#include <bout/yboundary_regions.hxx>
 
 #include <cmath>
 #include <limits>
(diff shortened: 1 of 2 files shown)


src/mesh/difops.cxx
```diff
@@ -386,23 +387,32 @@ Field3D Div_par_K_Grad_par_mod_impl(const Field3DParallel& Kin,
     Field3D result{zeroFrom(fin)};
     flow_ylow = zeroFrom(fin);
 
+    const auto yboundary = coord->getYBoundary();
```
totork, line 390: How is the handling here when getYBoundary is not set? Does it throw an error or generate it then?

totork (approved):

bendudson (approved):

PR #3484, opened 2026-08-28
Add missing non-const `BoundaryRegionIterBase::prev` overload

ZedThree:
Missing non-const ref overload meant it was not possible to assign to `point.prev(f)`

@dschwoerer Please can this go into the Hermes-3 BOUT++ update PR?

### DIFF
--- a/include/bout/yboundary_regions.hxx
+++ b/include/bout/yboundary_regions.hxx
@@ -11,8 +11,6 @@
 #include <memory>
 #include <vector>
 
-class Options;
-
 namespace bout {
 namespace boundary {
 /// This class allows to simplify iterating over y-boundaries.
@@ -24,7 +22,7 @@ namespace boundary {
 /// ../../manual/sphinx/user_docs/boundary_options.rst
 class YBoundary {
 public:
-  YBoundary(YBndryType type, Options* options_ptr, const Mesh& mesh);
+  YBoundary(const Mesh& mesh, bool lower_y, bool upper_y);
 
   /// Iterate over the boundary.
   /// This function takes a lamda / templated function, that applies the boundary on the given point.
--- a/src/mesh/coordinates.cxx
+++ b/src/mesh/coordinates.cxx
@@ -1223,13 +1223,45 @@ void Coordinates::_compute_cell_volume() const {
   }
 }
 
-const bout::boundary::YBoundary& Coordinates::getYBoundary(YBndryType type) const {
-  using bout::boundary::YBoundary;
-  const auto itype = static_cast<int>(type);
-  if (ybndrys.at(itype) == nullptr) {
-    ybndrys.at(itype) = std::make_shared<YBoundary>(type, localoptions, *localmesh);
+const bout::boundary::YBoundary& Coordinates::getYBoundary(Options* options) const {
+  Options* opts = (options == nullptr) ? localoptions : options;
+
+  const bool lower_y =
+      localmesh->isFci()
+          ? (*opts)["outer_x"].doc("Boundary on outer x?").withDefault<bool>(true)
+          : (*opts)["lower_y"].doc("Boundary on lower y?").withDefault<bool>(true);
+  const bool upper_y =
+      localmesh->isFci()
+          ? (*opts)["inner_x"].doc("Boundary on inner x?").withDefault<bool>(false)
+          : (*opts)["upper_y"].doc("Boundary on upper y?").withDefault<bool>(true);
+
+  return getYBoundary(lower_y, upper_y);
+}
+
+namespace {
+std::size_t boundary_bools_to_index(bool lower_y, bool upper_y) {
+  if (lower_y and upper_y) {
+    return 2;
   }
-  return *ybndrys.at(itype);
+  if (upper_y) {
+    return 1;
+  }
+  if (lower_y) {
+    return 0;
+  }
+  return 3;
+}
+} // namespace
+
+const bout::boundary::YBoundary& Coordinates::getYBoundary(bool lower_y,
+                                                           bool upper_y) const {
+  const auto index = boundary_bools_to_index(lower_y, upper_y);
+  if (ybndrys.at(index) == nullptr) {
+    ybndrys.at(index) =
+        std::make_shared<bout::boundary::YBoundary>(*localmesh, lower_y, upper_y);
+  }
+
+  return *ybndrys.at(index);
 }
 
 void Coordinates::setBxy(FieldMetric Bxy, const bool communicate) {
--- a/src/mesh/yboundary_regions.cxx
+++ b/src/mesh/yboundary_regions.cxx
@@ -1,52 +1,22 @@
 #include "bout/yboundary_regions.hxx"
 
-#include "bout/boundary_region.hxx"
 #include "bout/boundary_region_iter.hxx"
 #include "bout/field_data.hxx"
-#include "bout/options.hxx"
+#include "bout/mesh.hxx"
 
 #include <memory>
 
 namespace bout::boundary {
-YBoundary::YBoundary(YBndryType type, Options* options_ptr, const Mesh& mesh)
+YBoundary::YBoundary(const Mesh& mesh, bool lower_y, bool upper_y)
     : _contains_low(&mesh, false), _contains_high(&mesh, false) {
-  bool lower_y = true;
-  bool upper_y = true;
-  bool outer_x = true;
-  bool inner_x = false;
-  if (options_ptr != nullptr) {
-    auto& options = *options_ptr;
-    if (!mesh.isFci()) {
-      lower_y = options["lower_y"].doc("Boundary on lower y?").withDefault<bool>(lower_y);
-      upper_y = options["upper_y"].doc("Boundary on upper y?").withDefault<bool>(upper_y);
-    } else {
-      outer_x = options["outer_x"].doc("Boundary on outer x?").withDefault<bool>(outer_x);
-      inner_x = options["inner_x"].doc("Boundary on inner x?").withDefault<bool>(inner_x);
-    }
-  }
-  switch (type) {
-  case YBndryType::sheath:
-    break;
-  case YBndryType::not_sheath:
-    lower_y = !lower_y;
-    upper_y = !upper_y;
-    outer_x = !outer_x;
-    inner_x = !inner_x;
-    break;
-  case YBndryType::all:
-    lower_y = true;
-    upper_y = true;
-    outer_x = true;
-    inner_x = true;
-  }
 
   if (mesh.isFci()) {
-    if (outer_x) {
+    if (lower_y) {
       for (auto& bndry : mesh.getBoundariesPar(BoundaryParType::xout)) {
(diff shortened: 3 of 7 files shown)


ZedThree:
Actually there's at least one other function missing and some functions duplicated between the base and derived classes.

dschwoerer:
> Missing non-const ref overload meant it was not possible to assign to `point.prev(f)`

It might be beneficial to expose the `<check>` template parameter for this one.

> @dschwoerer Please can this go into the Hermes-3 BOUT++ update PR?

I think it would be best if they first go into next, together with #3482

ZedThree:
I'm still not convinced `check` is needed at all, it's probably better to have separate `valid_next` and `valid_prev` variables to check in the separate directions, or to only check the point is valid when looking into the boundary.

There's also another big pain point in the design: `getYBoundary()` only takes `YBndryType` which is insufficient to control which boundaries are actually required at the point of calling. Additionally, `YBndryType` has variants `sheath/not_sheath/all` which also doesn't give sufficient control.

Instead I've refactored it to have two overloads: one that takes an `Options` and one that takes two `bool`s. This allows us to create them as before, but then retrieving them doesn't require a lookup in `Options`.

We still only need to cache three `YBoundary` values for lower/upper/both, rather than sheath/not sheath/all, but no enum is exposed -- only the two bools are required.

src/mesh/yboundary_regions.cxx
```diff
@@ -1,52 +1,22 @@
...
-    lower_y = true;
-    upper_y = true;
-    outer_x = true;
-    inner_x = true;
-  }
 
   if (mesh.isFci()) {
-    if (outer_x) {
+    if (lower_y) {
       for (auto& bndry : mesh.getBoundariesPar(BoundaryParType::xout)) {
```
dschwoerer: So `lower_y` means xout?
Why do you think this is a good idea?

dschwoerer:
> I'm still not convinced `check` is needed at all, it's probably better to have separate `valid_next` and `valid_prev` variables to check in the separate directions, or to only check the point is valid when looking into the boundary.

I think the implementation of the sheath BC will tell. I am fine either way.

> There's also another big pain point in the design: `getYBoundary()` only takes `YBndryType` which is insufficient to control which boundaries are actually required at the point of calling. Additionally, `YBndryType` has variants `sheath/not_sheath/all` which also doesn't give sufficient control.

I do not understand. Why is it not sufficient? What control do you need? E.g. in https://github.com/boutproject/BOUT-dev/pull/3483 I want to special handle sheath BCs. How do I do that with the new interface?

> Instead I've refactored it to have two overloads: one that takes an `Options` and one that takes two `bool`s. This allows us to create them as before, but then retrieving them doesn't require a lookup in `Options`.

But that is super confusing. Sometimes the API means yup/ydown. Sometimes core/SOL.
Does that not lead to lots of if/else in calling code? And it needs to parse the right bit of the options tree to figure out how the simulation is configured? I tried that for Hermes-3 initially, and it was leading to duplication in the config, or each component needed to parse over and over again the global options tree.

> We still only need to cache three `YBoundary` values for lower/upper/both, rather than sheath/not sheath/all, but no enum is exposed -- only the two bools are required.

But figuring out which values to pass is tricky (outside of unit tests).

ZedThree:
> Why is it not sufficient? What control do you need? 

Just so everyone is on the same page, the current behaviour works like this:

- `Coordinates` is created, taking an `Options` from `Mesh` -- this will be the `[mesh]` section, and it's not possible to control this
  - FA meshes will read `lower/upper_y` from this section
  - FCI meshes will read `inner/outer_x` instead
- `Coordinates::getYBoundary` only takes `YBndryType`, which can be either: `sheath`, `not_sheath`, or `all`
  - `all` always uses both boundaries
  - `sheath` takes the input values as given
  - `not_sheath` inverts the input values -- setting `lower_y = true` in the input file and calling `coords->getYBoundary(YBndryType::not_sheath)` uses `lower_y = false` instead

So unless the `YBndryType` is an input value, _all_ calling sites will be tied to the one input section, and it won't be possible to independently control whether or not an individual boundary is enabled at a particular side. Even if a component makes the boundary type an input value, the value to use will be dependent on the values in `[mesh]`.

Concretely, with this input file:

```conf
[mesh]
lower_y = true
upper_y = true
```

it is not possible to turn off a single boundary condition at just one of the boundaries:

| `YBndryType` value | Resulting `lower_y` | Resulting `upper_y` |
|--|--|--|
| `sheath` | `true` | `true` |
| `not_sheath` | `false` | `false` |
| `all` | `true` | `true` |

Even in the cases where this works, `sheath`/`not_sheath` are not super descriptive about whether or not a boundary is enabled in a given direction. They're also not relevant for all physics models used in BOUT++ either.

---

The behaviour in this PR works instead as follows:

- The object using the `YBoundary` is created, taking an appropriate `Options`
  - this _could_ be `[mesh]`, or it could be a different section
  - FA meshes will read `lower/upper_y` from this section
  - FCI meshes will read `inner/outer_x` instead, as before
- The object stores either:
  - the input values of `lower/upper_y` during creation, and calls `Coords::getYBoundary()` with the stored values at runtime
  - just stores the `YBoundary` and uses it directly

Now, it will be possible to set up boundary conditions that can be toggled independently, while still falling back to a global input value.

For example, one could have:

```conf
[mesh]
lower_y = true
upper_y = true

[A]
lower_y = false
# use upper_y from [mesh]

[B]
# use !mesh:lower_y and !mesh:upper_y
# behaves like `not_sheath` previously
```

So this PR gives us the most control while still a set of boundary conditions to be toggled together.

---

> But that is super confusing. Sometimes the API means yup/ydown. Sometimes core/SOL.

I mean, I agree, but this retains the existing behaviour. How do you think it should behave?

> each component needed to parse over and over again the global options tree.

I'm not quite sure what you mean here. Parsing the options tree should be done during setup, and not during runtime (that is, not during `PhysicsModel::rhs()` or `Component::transform()`), so any duplicated parsing should be very cheap, plus the parsing is cached, so there should be any performance cost here.

I think the correct behaviour for any physics model is, in general, that a given component or numerical object (Laplacian solver, time solver, etc) should read from a specific individual section, with a fallback to a more generic section.

So if a set of components should be linked together by default somehow, I would expect to see something like:

```cpp
Options& opt = options == nullptr ? Options::root()["mesh"] : *options;
```

or

```cpp
lower_y = options["lower_y"].with_default(
   Options::root()["mesh"]["lower_y"].with_default(false)
);
```

The first one being a bit more idiomatic BOUT++.

> But figuring out which values to pass is tricky (outside of unit tests).

Ah, I think you're saying that we read in either `lower/upper_y` or `inner/outer_x` but don't actually return or store which values actually got read anywhere. We could store them in the `YBoundary` and add getters. Or we can also just store the `YBoundary` itself

src/mesh/coordinates.cxx
```diff
@@ -1223,13 +1223,45 @@ void Coordinates::_compute_cell_volume() const {
...
+  if (lower_y) {
+    return 0;
   }
-  return *ybndrys.at(itype);
+  return 3;
+}
+} // namespace
+
+const bout::boundary::YBoundary& Coordinates::getYBoundary(bool lower_y,
+                                                           bool upper_y) const {
```
dschwoerer, line 1257: I think this should be private.
The API naming is confusing, as it seems to talk about `lower_y` and `upper_y`, but that also might be `inner_x` / `outer_x`

include/bout/yboundary_regions.hxx
```diff
@@ -24,7 +22,7 @@ namespace boundary {
 /// ../../manual/sphinx/user_docs/boundary_options.rst
 class YBoundary {
 public:
-  YBoundary(YBndryType type, Options* options_ptr, const Mesh& mesh);
+  YBoundary(const Mesh& mesh, bool lower_y, bool upper_y);
```
dschwoerer: I think this should be private.
The API naming is confusing, as it seems to talk about `lower_y` and `upper_y`, but that also might be `inner_x` / `outer_x`

src/mesh/yboundary_regions.cxx
```diff
@@ -1,52 +1,22 @@
...
 #include "bout/boundary_region_iter.hxx"
 #include "bout/field_data.hxx"
-#include "bout/options.hxx"
+#include "bout/mesh.hxx"
 
 #include <memory>
 
 namespace bout::boundary {
-YBoundary::YBoundary(YBndryType type, Options* options_ptr, const Mesh& mesh)
+YBoundary::YBoundary(const Mesh& mesh, bool lower_y, bool upper_y)
```
dschwoerer: I think this should be private.
The API naming is confusing, as it seems to talk about `lower_y` and `upper_y`, but that also might be `inner_x` / `outer_x`

tests/unit/include/bout/test_yboundary.cxx
```diff
@@ -21,70 +21,70 @@ using bout::globals::mesh;
 TEST_F(YBTest, dirichlet_o2_rgn) {
   dynamic_cast<FakeMesh*>(mesh)->createBoundaries();
   Field3D test = 1.0;
-  YBoundary sheath(YBndryType::all, nullptr, *mesh);
+  YBoundary sheath(*mesh, true, true);
```
dschwoerer, line 24: I think this API should not be used (outside of unit tests, maybe)

dschwoerer (commented):
Ah, thanks, I had indeed missed that the API with options was your preferred way of doing things now 👍 
I do agree that that is a cleaner implementation, but only if the interfaces with the two bools are not used by user code. That should be an implementation detail.

It would help if you would update the documentation as well, I think that still refers to the `YBndryType`?

We should make sure that the `Options` API is used, and not the one passing in bools, as they are confusing to use.

src/mesh/coordinates.cxx
ZedThree: How about just `lower/upper`? Eventually, I'm hoping that something like this will be generic across all three dimensions, and so it will be good to have generic terminology

dschwoerer: I do not think that would work. lower_y / upper_y will always be in the SOL, at least in the common topologies I have seen. For FCI it is just different.

ZedThree: I'm not sure what you mean. X and Y can have boundaries at either end, and Y may have more boundaries but these are still classified as either "lower" or "upper". FCI has parallel boundaries described as some combination of `xin/xout` and `fwd/bwd`; `YBoundary` operates over `xin/xout`.

Are you saying there are boundaries which could not be generically described as either `lower` or `upper`?

dschwoerer: No, they probably can be described as lower and upper, but we do not map it to lower and upper.
We could, but I think for good reasons, we choose to map it to inner/outer for FCI. We could map it to upper / lower (fwd / bwd in FCI speak) for FCI, but I think that would not lead to a usable interface. I think we should just take an options pointer / nullptr and use that.

But I am not sure any of that is helpful. So maybe let's try to focus on a use case.
How would you expect a that API to be used, outside of unit tests?

Issue #3300, opened 2026-03-03
Allow mesh file to be read from data_dir


oparry-ukaea:
I'd like to be able to specify a mesh file name in BOUT.inp and have it read from the directory passed with `-d`.

mikekryjak:
I've always wondered why we don't do this already... is there a reason for it?
I'd also like to be able to read it from a directory defined in the input file.

bendudson:
I agree this would make more sense. The reason for the current system is that it's much more common to change the BOUT.inp than the mesh file: A set of simulations will all share the same mesh file, which can be in the root directory, with separate runs or restarted runs in subdirectories. I would however be happy to switch to the mesh file path being relative to the BOUT.inp file, so BOUT.inp could contain
```
[mesh]
file = "../some_mesh.nc"
```
to use subdirectories for separate runs. 

It would be a breaking change, but probably not very difficult for people to adapt to.

ZedThree:
One way to not make it a breaking change would be to first search (relative to?) the same directory as the input file, and then if not found, try the current working directory (or the other way round). That should probably not break things? Might be confusing?

See also #3208

oparry-ukaea:
That makes sense to me. Searching cwd first would be guaranteed not to break anybody's current setup, presumably?

PR #3486, opened 2026-09-03
MeshFactory::create read mesh from data directory

bendudson:
Tries to read the specified mesh from data directory, falling back to the previous behavior of reading relative to current directory. If both `data_dir/file` and `./file` exist then an exception will be thrown rather than picking one. If the mesh is given as an absolute path then it's unambiguous.

This does change behavior but any ambiguous situations will result in an error rather than silently doing something strange.

Includes updates to docs and some unit tests.

Uses C++20 <filesystem> to manipulate paths and test existence.

Fixes issue #3300

### DIFF
--- a/manual/sphinx/user_docs/bout_options.rst
+++ b/manual/sphinx/user_docs/bout_options.rst
@@ -486,12 +486,9 @@ can be set with the global option ``grid``, or ``mesh:file``:
     [mesh]
     file = "cbm18_8_y064_x260.nc"
 
-Relative grid paths are first resolved relative to ``datadir``. If that
-file does not exist, BOUT++ falls back to resolving the path relative to
-the current working directory. If matching files exist in both places,
+Relative grid paths are resolved relative to ``datadir`` and the current working directory. If matching files exist in both places,
 BOUT++ throws an error rather than guessing which one to use. To avoid
-ambiguity entirely, use an explicit absolute path when the grid file is
-not inside ``datadir``.
+ambiguity, an explicit absolute path for the grid file can be used.
 
 
 Communications
--- a/manual/sphinx/user_docs/bout_options.rst
+++ b/manual/sphinx/user_docs/bout_options.rst
@@ -476,17 +476,22 @@ minimises ``abs(sqrt(NPES * (nx - 4) / ny) - NXPE)``).
 
 If you need to specify complex input values, e.g. numerical values
 from experiment, you may want to use a grid file. The grid file to use
-is specified relative to the root directory where the simulation is
-run (i.e. running “``ls ./data/BOUT.inp``” gives the options
-file). You can use the global option ``grid``, or ``mesh:file``:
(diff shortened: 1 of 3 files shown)


ZedThree:
Looks good! Please could you add something to the docs about this behaviour?

bendudson:
Thanks @ZedThree ! Good point. I modified the behavior a bit and have added docs and unit tests.

manual/sphinx/user_docs/bout_options.rst
```diff
@@ -476,17 +476,22 @@ minimises ``abs(sqrt(NPES * (nx - 4) / ny) - NXPE)``).
...
+    grid = "cbm18_8_y064_x260.nc"
 
     # Alternatively:
     [mesh]
-    file = "data/cbm18_8_y064_x260.nc"
+    file = "cbm18_8_y064_x260.nc"
+
+Relative grid paths are first resolved relative to ``datadir``. If that
+file does not exist, BOUT++ falls back to resolving the path relative to
+the current working directory. If matching files exist in both places,
```
dschwoerer:
```suggestion
Relative grid paths are resolved relative to ``datadir`` and the current working directory. If matching files exist in both places,
```
The order does not matter, if it must be unique.

manual/sphinx/user_docs/bout_options.rst
```diff
@@ -476,17 +476,22 @@ minimises ``abs(sqrt(NPES * (nx - 4) / ny) - NXPE)``).
...
+BOUT++ throws an error rather than guessing which one to use. To avoid
+ambiguity entirely, use an explicit absolute path when the grid file is
+not inside ``datadir``.
```
dschwoerer: That bit also reads a bit weird.
Maybe:
```suggestion
BOUT++ throws an error rather than guessing which one to use. To avoid
ambiguity, an explicit absolute path for the grid file can be used.
```

src/mesh/mesh.cxx
```diff
@@ -67,14 +70,37 @@ MeshFactory::ReturnType MeshFactory::create(const std::string& type, Options* op
...
     }
-    output << "\nGetting grid data from file " << grid_name << "\n";
 
-    // Create a grid file, using specified format if given
-    const auto grid_ext =
-        (*options)["format"].withDefault(Options::root()["format"].withDefault(""));
+    // Resolve mesh files relative to datadir first, then the current directory.
+    const auto datadir = fs::path(Options::root()["datadir"].withDefault("data"));
+    const auto grid_path = fs::path(grid_name);
+    auto full_path = datadir / grid_path;
```
dschwoerer: Should we make sure that `grid_path` is not absolute?

Right now it would fail if I use `$HOME/grid.nc` but also have a grid in `data/$HOME/grid.nc`. Not that is really likely ...

dschwoerer (dismissed):
Looks good, documentation could be improved.

Thanks @bendudson

dschwoerer (approved):

PR #3487, opened 2026-09-04
ADIOS2: Fix restart read/write

bendudson:
Restarting from an ADIOS2 `BOUT.restart.bp` dataset and then writing to it would produce an unusable restart dataset. Attempting to restart a second time would either segfault or fail because variables had the wrong shape.

Quoth GPT:

      The main issue was that a restart read was preserving ADIOS’s
      internal __xarray_dimensions__ metadata in the Options tree,
      then writing it back out in a lossy form on the next restart
      write. On the second restart that made ADIOS interpret generic
      arrays as mesh fields with the wrong rank. I also tightened the
      ADIOS stream lifecycle so finished streams retire their IO
      objects cleanly instead of leaving stale cached state around.

Includes a new unit test that reproduced the original error and now passes.

### DIFF
--- a/include/bout/adios_object.hxx
+++ b/include/bout/adios_object.hxx
@@ -55,6 +55,8 @@ public:
 
   ~ADIOSStream();
 
+  bool isRetired() const { return retired_; }
+
   template <class T>
   adios2::Variable<T> GetValueVariable(const std::string& varname) {
     auto v = io.InquireVariable<T>(varname);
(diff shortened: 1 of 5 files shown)


Steven-Roberts:
Thanks for catching this @bendudson. I will review tomorrow.

Steven-Roberts (approved):

PR #3489, opened 2026-09-07
Fix CI pipeline failures when UMPIRE is enabled

jonycastagna:
Resolved compilation errors and test failures that occurred specifically when building with the UMPIRE configuration turned ON. This will be useful for GPU offloading optimization

### DIFF
--- a/src/mesh/parallel/shiftedmetric.cxx
+++ b/src/mesh/parallel/shiftedmetric.cxx
@@ -489,6 +489,26 @@ void ShiftedMetric::calcParallelSlices(Field3D& f) {
   f.splitParallelSlices();
 
 #if BOUT_HAS_CUDA
+  const bool cuda_fft_supported =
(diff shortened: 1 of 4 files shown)


src/mesh/parallel/shiftedmetric.cxx
```diff
@@ -489,6 +489,26 @@ void ShiftedMetric::calcParallelSlices(Field3D& f) {
   f.splitParallelSlices();
 
 #if BOUT_HAS_CUDA
+  const bool cuda_fft_supported =
+      mesh.LocalNz == 16 || mesh.LocalNz == 64 || mesh.LocalNz == 128
```
dschwoerer: Why not 32?

jonycastagna: My bad, just missed it! Thanks

PR #3490, opened 2026-09-11
Added the Mesh_topology enum class.

wearysebas:
The getMeshTopology function identifies the current topology and the variable now stores the value so that it can be used anywhere. 

To be expanded with the Snowflake and X-Point topology.

### DIFF
--- a/src/mesh/impls/bout/boutmesh.cxx
+++ b/src/mesh/impls/bout/boutmesh.cxx
@@ -27,8 +27,10 @@
 #include "boutmesh.hxx"
 
 #include <bout/assert.hxx>
+#include <bout/assert.hxx>
 #include <bout/boundary_region.hxx>
 #include <bout/boundary_region_iter.hxx>
+#include <bout/bout_types.hxx>
 #include <bout/bout_types.hxx>
 #include <bout/boutcomm.hxx>
 #include <bout/boutexception.hxx>
@@ -41,10 +43,19 @@
 #include <bout/field_data.hxx>
 #include <bout/fieldgroup.hxx>
 #include <bout/globals.hxx>
+#include <bout/field2d.hxx>
+#include <bout/field3d.hxx>
+#include <bout/field_data.hxx>
+#include <bout/fieldgroup.hxx>
+#include <bout/globals.hxx>
 #include <bout/griddata.hxx>
 #include <bout/msg_stack.hxx>
 #include <bout/options.hxx>
 #include <bout/output.hxx>
+#include <bout/region.hxx>
+#include <bout/sys/gettext.hxx>
+#include <bout/sys/range.hxx>
+#include <bout/parallel_boundary_region.hxx>
 #include <bout/region.hxx>
 #include <bout/sys/gettext.hxx>
 #include <bout/sys/range.hxx>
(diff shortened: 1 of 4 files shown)


src/mesh/impls/bout/boutmesh.cxx
```diff
@@ -173,12 +173,47 @@ void BoutMesh::setXDecompositionIndices(const XDecompositionIndices& indices) {
   ixseps2 = indices.ixseps2;
 }
 
+MeshTopology BoutMesh::getMeshTopology(int jyseps1_1_, int jyseps2_1_, int jyseps1_2_,
+                                       int jyseps2_2_, int ny_inner_, int ixseps1_,
+                                       int ixseps2_) { //Returns MeshTopology that is type enum
```
ZedThree: No need for this comment

```suggestion
                                       int ixseps2_) {
```

src/mesh/impls/bout/boutmesh.cxx
```diff
@@ -173,12 +173,47 @@ void BoutMesh::setXDecompositionIndices(const XDecompositionIndices& indices) {
...
+
+  if (jyseps1_1 < 0 and jyseps2_2 >= ny - 1) {
+    return MeshTopology::closed_field_line; //write whole thing
+  } else if (jyseps2_1 == jyseps1_2) {
+    return MeshTopology::single_null;
+  } else if (ixseps1 == ixseps2) {
+    return MeshTopology::connected_double_null;
+  } else {
+    return MeshTopology::unconnected_double_null;
+  }
```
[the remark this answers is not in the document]
ZedThree: Clang-tidy should be able to apply this fix from your editor

src/mesh/impls/bout/boutmesh.cxx
```diff
@@ -173,12 +173,47 @@ void BoutMesh::setXDecompositionIndices(const XDecompositionIndices& indices) {
...
 CheckMeshResult checkBoutMeshYDecomposition(int num_y_processors, int ny,
                                             int num_y_guards, int jyseps1_1,
                                             int jyseps2_1, int jyseps1_2, int jyseps2_2,
                                             int ny_inner) {
 
+  // Preserve legacy behaviour (Single / Double Null)
+  return checkBoutMeshYDecomposition(num_y_processors, ny, num_y_guards, jyseps1_1,
+                                     jyseps2_1, jyseps1_2, jyseps2_2, ny_inner,
+                                     MeshTopology::unconnected_double_null);
+}
```
ZedThree: A simpler way of achieving this is to just give `mesh_topology` a default argument of `MeshTopology::unconnected_double_null`, and then we don't need a second overload


But more to the point, this seems a bit confusing. Does this function check that `mesh_toplogy` is correct for these parameters? Or does it change which checks to do? In either case, it doesn't seem like a single value for `mesh_toplogy` can be correct for all cases -- so maybe it makes more sense to just drop this version and always require `mesh_toplogy`?

src/mesh/impls/bout/boutmesh.cxx
```diff
@@ -228,8 +263,9 @@ CheckMeshResult checkBoutMeshYDecomposition(int num_y_processors, int ny,
                      "be a multiple of MYSUB ({:d})\n"),
                   jyseps1_2, ny_inner, jyseps1_2 - ny_inner + 1, num_local_y_points)};
     }
-  } else {
-    // Single Null
+  } else if ((mesh_topology == MeshTopology::single_null)
+             || (mesh_topology == MeshTopology::closed_field_line)) {
+    // Single Null or connected Double Null
```
ZedThree: This comment is either unnecessary or wrong!

```suggestion
```

PR #3496, opened 2026-09-16
Support AMD GPU via HIP backend

jonycastagna:
Added HIP directive for offloading on AMD GPU via RAJA and binary expression backend


tests/MMS/diffusion/diffusion.cxx
```diff
@@ -10,12 +10,15 @@ class Diffusion : public PhysicsModel {
 protected:
   int init(bool UNUSED(restarting)) override;
   int rhs(BoutReal t) override;
+
+private:
+  // Keep Umpire-backed field storage owned by the model so that it is
+  // released before BoutFinalise() tears down BOUT++'s array storage.
+  Field3D N;
```
ZedThree, line 17: Oh, this is fine for this PR, but we should just move everything inside the model

tests/MMS/advection/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
 add_executable(MMS-advection advection.cxx)
+bout_target_enable_hip(MMS-advection)
```
ZedThree, line 2: What's this for? Or rather, why only for this test?

src/mesh/parallel/shiftedmetric.cxx
```diff
@@ -429,6 +431,16 @@ static void shiftZ_block_fft(const int Nz, const BoutReal** in, BoutReal** out,
...
         <<<grid, block, 0, stream>>>(in, out, phs, nbatches, nblocks);
+  } else if (Nz == 32) {
+    constexpr int FFTS_PER_BLOCK = 8;
+    constexpr int THREADS_PER_FFT = 32;
+
+    dim3 block(THREADS_PER_FFT, FFTS_PER_BLOCK);
+    dim3 grid((total_ffts + FFTS_PER_BLOCK - 1) / FFTS_PER_BLOCK);
+
+    fft_block_cooperative<32, FFTS_PER_BLOCK>
+        <<<grid, block, 0, stream>>>(in, out, phs, nbatches, nblocks);
```
ZedThree, line 442: Again, fine for this PR, but this whole section is a little suspicious -- looks like we should be able to reduce the duplication here

cmake/SetupCompilers.cmake
```diff
@@ -32,6 +32,35 @@ set(BOUT_USE_OPENMP ${BOUT_ENABLE_OPENMP})
...
+  endif()
+  # Keep compiler detection and compilation on the same ROCm installation.
+  # Otherwise a Spack view can be mistaken for an implicit include directory.
+  if(CMAKE_HIP_COMPILER_ROCM_ROOT)
+    string(APPEND CMAKE_HIP_FLAGS_INIT
+           " --rocm-path=${CMAKE_HIP_COMPILER_ROCM_ROOT}"
+    )
+  endif()
+  enable_language(HIP)
+  set(CMAKE_HIP_STANDARD 20)
```
ZedThree, line 59: Does this need to stay in-sync with the C++ standard?

cmake/SetupBOUTThirdParty.cmake
```diff
@@ -53,12 +66,6 @@ set(BOUT_HAS_UMPIRE ${BOUT_ENABLE_UMPIRE})
 option(BOUT_ENABLE_RAJA "Enable RAJA" OFF)
 if(BOUT_ENABLE_RAJA)
   find_package(RAJA REQUIRED)
-  message(STATUS "RAJA_CONFIG:" ${RAJA_CONFIG})
-  string(FIND ${RAJA_CONFIG} "raja" loc)
-  math(EXPR value "${loc} + 5" OUTPUT_FORMAT DECIMAL)
-  string(SUBSTRING ${RAJA_CONFIG} 0 ${value} RAJA_PATH)
-  message(STATUS "RAJA_PATH" ${RAJA_PATH})
-  target_include_directories(bout++ PUBLIC ${RAJA_PATH}/include)
```
ZedThree, line 61: Do we not need these bits any more? Or have they moved, and I've missed them?

ZedThree (approved):
Looks good, I've got some questions but they might not be blockers

tests/MMS/diffusion/diffusion.cxx
jonycastagna: Agree

tests/MMS/advection/CMakeLists.txt
jonycastagna: It’s explicit there because MMS-advection uses add_executable() directly, bypassing the usual test helper. Other MMS tests use bout_add_mms_test() I think...

jonycastagna:
Thanks Peter/All, I need to check one thing one the option N=32 and should be ready then
Jony

Sent from Outlook for iOS<https://aka.ms/o0ukef>
________________________________

src/mesh/parallel/shiftedmetric.cxx
jonycastagna: Yes, we could template the call. Maybe I can do in a second round for a cleaner progress track?

cmake/SetupCompilers.cmake
jonycastagna: It would be better to be in sync. Would it help to have across the full code a variable like set(CMAKE_CXX_STANDARD 20) that we can reuse?

cmake/SetupBOUTThirdParty.cmake
jonycastagna: Those lines tried to infer RAJA’s installation directory from the config file path, which assumes the installation directory is named "raja". However, an earlier occurrence in the path could produce the wrong result, that's why I removed. This is part of the headache I was try to explain about the installation of RAJA (in a very cumbersome way... sorry!). Maybe we need something more robust here

jonycastagna:
Hi @ZedThree, I just added some comments above and merged latest from "next" but seems some checks were not succesful. Do you know what I am doing wrong? Thanks!

ZedThree:
The failing test is unrelated

jonycastagna:
Ok, thanks! I see now is all good

cmake/SetupBOUTThirdParty.cmake
ZedThree: It does look pretty hacky -- if RAJA is built with CMake, we really shouldn't need anything like this, but I don't have much experience installing it in different configurations/environments.

Probably best to remove it and see who screams..

cmake/SetupCompilers.cmake
ZedThree: Yes, we can use something like `set_property(TARGET ${LIB_NAME} PROPERTY CXX_STANDARD ${BOUT_CXX_STANDARD})`.

We probably want to avoid `set(CMAKE_CXX/HIP_STANDARD 20)` as this will set it globally

ZedThree:
@jonycastagna Is this ready to merge? You mentioned some other work -- is that a second PR?

jonycastagna:
> @jonycastagna Is this ready to merge? You mentioned some other work -- is that a second PR?

Yes, it is. The other work is related to Laplacian inversion, so better to keep separate

